PR #2, opened 2017-02-08
Now compatible with Philipp's AnnData class

falexwolf:


scanpy/ann_data.py
```diff
@@ -91,21 +92,20 @@ def _check_dimensions(data, smp, var):
...
-        u"""
-        Annotated Matrix
+    def __init__(self, ddata=None, X=None, smp=None, var=None, vis=None, **meta):
+        """
+        Annotated Data
 
-        Represents a sample × variable matrix (e.g. cell × gene) with the
-        possibility to store an arbitrary number of annotations for both
-        samples and variables.
+        Stores a data matrix X of dimensions nr_samples x nr_variables,
```
flying-sheep, line 99: why did you remove the “×” symbols? isn’t it clearer that way? i always need a moment to process that some “x” isn’t actually an x, but an ASCII approximation to “×”

scanpy/ann_data.py
```diff
@@ -164,27 +202,37 @@ def __init__(self, X, smp=None, var=None, vis=None, **meta):
...
+                        self.smp[k] = v
+                elif key == 'col':
+                    for k, v in ddata['col'].items():
+                        self.var[k] = v
+                elif key == 'var':
+                    for k, v in ddata['var'].items():
+                        self.var[k] = v
+                    
+    def smp_keys(self):
+        return list(self.smp.dtype.names)[1:]
```
flying-sheep, line 223: `smp_names` aren’t really guarenteed to be first, right?

either this should be changed to `return set(self.smp.dtype.names) - {SMP_NAMES}` or we should make sure to reorder the columns appropriately when setting them

i also think those should be properties instead of functions and maybe named “smp_columns”. “keys” is dictionary terminology and we have a structured array.

scanpy/ann_data.py
```diff
@@ -164,27 +202,37 @@ def __init__(self, X, smp=None, var=None, vis=None, **meta):
...
+
+    def var_keys(self):
+        return list(self.var.dtype.names)[1:]
+
+    def to_dict(self):
+        smp = {k: self.smp[k] for k in self.smp_keys() if not k=='smp_names'}
+        var = {k: self.var[k] for k in self.var_keys() if not k=='var_names'}
```
flying-sheep, line 230: please use the constants, it’s less error prone.

also, smp_names/var_names are already excluded in that method.

scanpy/ann_data.py
falexwolf: Because of compatibility with Python 2, which threw an error. I think that could have been fixed by specifying the encoding. But then, everything should be done the same way. I can restore it when doing some final polishing on the docstrings.

scanpy/ann_data.py
falexwolf: yes, that's a hack. but I think the np.recarray is ordered, right?

both your suggestions are fine. i just wanted it to get to work. I'm happy if you go over it again. maybe not right now, as everything is tested. but we can sit together maybe tomorrow, if you want.

scanpy/ann_data.py
flying-sheep: yeah: `# coding: utf-8` and an `u` in front fo the docstring.

> But then, everything should be done the same way

i don’t understand what you mean by that. everything *is* done the same way. the default encoding in python 2 is ASCII, which is a subset of utf-8, which means that having or not having the UTF-8 encoding declaration is irrelevant as long as there is no error immediately when loading the module.

scanpy/ann_data.py
falexwolf: yes, you're completely right. the whole thing needs to be redone. i haven't slept a minute for 29 hours. ;)

flying-sheep:
great stuff 😄

scanpy/ann_data.py
falexwolf: ok! I'll restore it then.

falexwolf:
no, not great, but it work's and one should now be much better settled for the future with AnnData.

for example, the gene plots and different subgroups work. if you have a good suggestion for a default color map for continuous and categorial columns in smp, I'm very happy to adapt it. :)

https://github.com/falexwolf/collab_alex/blob/master/scanpy/examples/maehr17.md

or here directly in the main readme

https://github.com/theislab/scanpy#moignard15

scanpy/ann_data.py
flying-sheep: it’s ordered, but doing `AnnData(X, dict(a=..., smp_names=...))` is valid and will* result in smp_names being the second column.


*on python ≤3.5 randomly, on python 3.6 always

scanpy/ann_data.py
flying-sheep: why would you do that, go to sleep!

PR #18, opened 2017-05-05
switch filter_genes_dispersion to recarray

### LINKED ISSUE #17: don’t return dicts
e.g `pp.filter_genes_dispersion` returns a dict with known keys. if you know the keys, this means you should return an object where you can use attribute access instead. usually the pythonic thing to do in this case is to return a namedtuple.

in this case though, we have

```py
>>> {k: v.shape for k, v in filter_result.items()}
{'dispersions': (173351,),
 'dispersions_norm': (173351,),
 'gene_filter': (173351,),
 'means': (173351,)}
```

which is a perfect case for a recarray.

flying-sheep:
fixes #17


falexwolf:
Thank's for the suggestion! I will get back to you!

PR #20, opened 2017-05-18
Build directly from Cython

flying-sheep:
I think I got it 😉


falexwolf:
Hey Phil! I'm still a bit hesitant to adopt this option. As discussed before, I'm mainly planning on integrating c++ extensions. Then doing everything via Cython seems overhead. Even the latest Cython distributions recommends **not** doing it. See [here](http://cython.readthedocs.io/en/latest/src/reference/compilation.html) and search for *distributing cython modules*. They provide a solution very similar to the one fom stackoverflow that we have adopted right now.

PR #23, opened 2017-06-08
Fix reference links

jhung0:


falexwolf:
Thanks! :smile: 

FYI: I plan to remove the references in the future, though, as they are quite cumbersome to maintain. In the future, I plan to just reference papers using `[First Author *et al.*](https://doi.org/...)` instead of `[First Author *et al.*](#ref_firstauthor)`.

PR #37, opened 2017-08-29
moved docs {generated→api}

flying-sheep:
i think that’s a better URL


falexwolf:
It is! "generated/" is just what numpy, scipy and seaborn used. But your suggestion makes much more sense!

falexwolf:
Go ahead and merge this! :smile:

PR #38, opened 2017-08-30
automated setup.py

flying-sheep:
this makes it possible to use `pip install` without installing numpy.

it also includes automation for cython again, as currently the `python setup.py build_ext` command will never use cython, even if available.

once the .pyx is changed and `build_ext` is executed, this now refreshes the `.c` file.


falexwolf:
Phil, thanks for this! I'm slowly finding time again to deal with these things. I looked through it and it's a very nice solution. I'll test it these days and merge it into master. :)

PR #60, opened 2018-01-03
Allow passing non-string groups in rank_genes_groups

flying-sheep:
I also re-activated tests


falexwolf:
Great, thank you! I wasn't completely happy with how Tobias wrote these tests on pickled files - they never actually passed on travis, it's much better that now, they pass! :smile: :smile:

flying-sheep:
Yeah 😃

PR #67, opened 2018-01-12
Fix module path of download_progress and create dir correctly.

gokceneraslan:
`sc.datasets.paul15_raw()` fails with `attempted relative import beyond top-level package` error due to the wrong module path in `sc.utils.check_presence_download `. 

Simply run `sc.datasets.paul15_raw()` or `sc.datasets.paul15()` to reproduce.


falexwolf:
Thank you very much, Gökcen!

PR #76, opened 2018-02-06
Scoring functions

dawe:
I believe everything is in place. We should have now a generic `add_score` which scores cells according to expression of gene lists. That is wrapped twice in `cell_cycle_score`.

### DIFF
--- a/scanpy/__main__.py
+++ b/scanpy/__main__.py
@@ -29,6 +29,7 @@ dtools = odict([
     ('dpt', 'perform Diffusion Pseudotime analysis'),
     ('louvain', 'cluster cells using the louvain algorithm'),
     ('rank_genes_groups', ' test groups for differential expression'),
+    ('score', ' calculates cell scores according to given genes'),
     ('sim', 'simulate stochastic gene expression models'),
 ])
 
--- a/scanpy/api/tl.py
+++ b/scanpy/api/tl.py
@@ -9,3 +9,5 @@ from ..tools.dpt import dpt
 from ..tools.louvain import louvain
 from ..tools.sim import sim
 from ..tools.top_genes import correlation_matrix, ROC_AUC_analysis
+
+from ..tools.score import add_score, cell_cycle_score
--- a/scanpy/tests/score.py
+++ b/scanpy/tests/score.py
@@ -1,0 +1,17 @@
+import numpy as np
+import scanpy.api as sc
+from anndata import AnnData
+
+def test_add_score():
+    adata = AnnData(np.random.randint(0, 1000, 100000).reshape((100, 1000)))
+    gene_names = np.array([''.join([chr(x) for x in np.random.randint(65, 90, 6)]) for n in range(2000)])
+    adata.var_names = gene_names[:1000]
+    sc.pp.normalize_per_cell(adata, counts_per_cell_after =1e4)
+    sc.pp.log1p(adata)
+    some_genes = np.concatenate([np.unique(gene_names[np.random.randint(0, 1000, 10)]), np.unique(gene_names[np.random.randint(1000, 2000, 3)])])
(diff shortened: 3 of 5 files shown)


falexwolf:
Hi Davide!

Thank you very much for this! Sorry that I tend to be late these days, have two 6 week old baby twins to take care of...

I'm happy to merge this and I'll add you to the author list! I hope it is ok if I rename the module and the top-level function to `score_gene_lists`  and the second top-level function to `score_cell_cylce_genes`? Simply `score` is a bit generic... there might be many other scores in the future and then people will get confused. It's also good if both start with `score` so that auto-lookup gives you directly these suggestions? 

Also, do you have a notebook with an example? It would be cool to see this at work. You could push this to a new subdirectory in `scanpy_usage`: https://github.com/theislab/scanpy_usage. I just sent you a collaborator invitation. From the example, we can then mayb design a test that goes a bit more into detail. Would be cool to benchmark with Seurat, for example.

Also, one could think about providing a default list of genes, right? In particular for the cell cycle, it would be nice to directly call the function with default parameters - one can then still add user-specified lists. Do you want to provide such a list? I also sent you collaborator invitation for scanpy - maybe only temporarily if we get too many people at some point - so that you can quickly add this, if you like.

Cheers,
  Alex

falexwolf:
Finally: it would be great to add a bit more explanation on how the comparison with the random gene set works exactly. I can also do this...

falexwolf:
I've now also added this to the [documentation](https://scanpy.readthedocs.io/en/latest/api/index.html#tools-tl) so that we see how "it looks".

Do you think the naming convention is good like that? Or shall we rather make it `score_gene_sets` and `score_genes_cell_cycle` [more symmetry and more rigour in naming?]. Now we can easily change this without having to worry about backwards compat... ;)

dawe:
@falexwolf I'm glad you merged. A couple of things

- ok for the name change. I see you have a different naming and syntax style, I'll try to stick to that (e.g. no space around operators)
- I do not have any notebook ready, I'll prepare one
- As for the gene list, I used the Regev lab one, I'm adding that into the notebook
- I will try to explain how the comparison works. I actually had to figure it out myself, as reverse engineering `R` code is far beyond human capabilities :-)
- you don't have to apologize for being late. I'm father of two and I haven't worked after 21:00 since my first was born 7 years ago

falexwolf:
@dawe 
- the coding style is the [official python coding style](https://www.python.org/dev/peps/pep-0008/), we should all stick to that. in particular: [white spaces](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements) around operators but **not** around optional keyword arguments
- thank you for a notebook!
- gene list: why not add it as an attribute of your module? or make a class `GeneLists` with a few gene lists in your model? of course, these will not be comprehensive, but might provide a good starting point
- in a few instances, I had to go through the reverse R engineering myself - in particular, if it comes to benchmarking code to floating point precision, it's really a hassle to dig out all the hidden different conventions... but I think it really pays off in the sense that it provides a lot more confidence in code and methods if several tools provide the same result on basic things - even across languages

dawe:
@falexwolf 

Too bad I have never read the official coding style, I guess that happens when you are a self-taught programmer :-) Not to mention I still have difficulties with git. You have been warned :-D

dawe:
@falexwolf I've added a silly notebook in the scanpy_usage repo showing cell cycle regression, pretty much following most important bits of [Seurat's vignette](http://satijalab.org/seurat/cell_cycle_vignette.html#assign-cell-cycle-scores). Note that this will work only after the scoring function code will be fixed. As mentioned in #82 , when my functions were merged and names changed, not all calls were renamed.

falexwolf:
Oh, thank you! I'll go through this tonight!

I will polish it, check it and once again rethink the naming conventions, as mentioned just a few minutes ago.

Issue #73, opened 2018-01-30
Plotting difference in png file vs. Jupyter output


jorvis:
I assume I'm missing something here, but when I try a simple example of plotting a gene dispersion I get two plots 'normalized' and 'not normalized' version in Jupyter, but when I use the save argument to sc.pl.filter_genes_dispersion() I get an image with only one of these.  Screenshot attached.   Just in case, I tried also passing the multi_panel argument but that caused an error. 

Also, is it no possible to specify the path where the files should be stored when using the save arguments to the plotting methods?  I want to point to a directory where it should place them, but it seems "./figures/" is hard-coded and you can only modify the end of that.

Thanks

The attached screenshot shows the dual image within Jupyter, but only the single plot which appears in the PNG file exported.

![screenshot from 2018-01-30 12-34-56](https://user-images.githubusercontent.com/330899/35584410-945d7bb6-05ba-11e8-89fc-14f615a9c6a6.png)

falexwolf:
The figures directory can be set using - I will add this to the documentation...
```
sc.settings.figdir = 'whateverpathyoulike'
```
The default is './figures/'.

You only get to save the non-normalized version as this is what Cell Ranger and Seurat allow you to visualize. Do you need the other version? I can easily add this...

jorvis:
Great info about the figdir, thank you.

I think it would be good to be able to access/save both plots, yes.  I want this sort of pipeline script to be able to generate the same output for inspection as if the user were running the commands within Jupyter, and both display there.  I don't think it's a critical thing though.

falexwolf:
How images are saved has also improved in Scanpy 1.0 and produces less white space etc.. In addition, you can set `sc.settings.autosave = True` and this will automatically save all images instead of showing them, see [here](https://github.com/theislab/scanpy_usage/blob/3ba68ab52ee4e5db39da13f449b575402f404df4/170522_visualizing_one_million_cells/cluster.py#L7).

PR #80, opened 2018-02-08
Now writes both normalized and standard image.  I didn't change the b…

jorvis:
…asename of the standard one so existing tool expectations wouldn't break.  Addresses scanpy issue #73

### DIFF
--- a/scanpy/plotting/preprocessing.py
+++ b/scanpy/plotting/preprocessing.py
@@ -42,11 +42,8 @@ def filter_genes_dispersion(result, log=False, save=None, show=None):
         pl.legend()
         pl.xlabel(('$log_{10}$ ' if False else '') + 'mean expression of gene')
 
-        if id == 0:
-            pl.ylabel(('$log_{10}$ ' if False else '') + 'dispersion of gene'
-                  + (' (normalized)'))
-            utils.savefig_or_show('filter_genes_dispersion_normalized', show=show, save=save)
-        else:
-            pl.ylabel(('$log_{10}$ ' if False else '') + 'dispersion of gene'
-                  + (' (not normalized)'))
-            utils.savefig_or_show('filter_genes_dispersion', show=show, save=save)
+        file_name = 'filter_genes_dispersion{}'.format('_normalized' if id == 0 else '')
+        pl.ylabel(('$log_{10}$ ' if False else '') + 'dispersion of gene'
+                  + (' (normalized)' if id == 0 else ' (not normalized)'))
+        utils.savefig_or_show(file_name, show=show, save=save)
+        
--- a/scanpy/plotting/preprocessing.py
+++ b/scanpy/plotting/preprocessing.py
@@ -41,6 +41,12 @@ def filter_genes_dispersion(result, log=False, save=None, show=None):
             pl.ylim(y_min, 1.05*np.max(dispersions))
         pl.legend()
         pl.xlabel(('$log_{10}$ ' if False else '') + 'mean expression of gene')
-        pl.ylabel(('$log_{10}$ ' if False else '') + 'dispersion of gene'
-                  + (' (normalized)' if id == 0 else ' (not normalized)'))
-    utils.savefig_or_show('filter_genes_dispersion', show=show, save=save)
+
+        if id == 0:
+            pl.ylabel(('$log_{10}$ ' if False else '') + 'dispersion of gene'
+                  + (' (normalized)'))
+            utils.savefig_or_show('filter_genes_dispersion_normalized', show=show, save=save)
+        else:
+            pl.ylabel(('$log_{10}$ ' if False else '') + 'dispersion of gene'
+                  + (' (not normalized)'))
+            utils.savefig_or_show('filter_genes_dispersion', show=show, save=save)
--- a/scanpy/preprocessing/simple.py
+++ b/scanpy/preprocessing/simple.py
@@ -136,24 +136,11 @@ def filter_genes(data, min_cells=None, min_counts=None, max_counts=None,
 
     See :func:`~scanpy.api.pp.filter_cells`.
     """
-    multi_option_error = False
-    
-    if min_cells is not None:
-        if min_counts is not None or max_cells is not None or max_counts is not None:
-            multi_option_error = True
-    if min_counts is not None:
-        if min_cells is not None or max_cells is not None or max_counts is not None:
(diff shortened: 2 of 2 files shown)


scanpy/plotting/preprocessing.py
```diff
@@ -41,6 +41,12 @@ def filter_genes_dispersion(result, log=False, save=None, show=None):
...
-    utils.savefig_or_show('filter_genes_dispersion', show=show, save=save)
+
+        if id == 0:
+            pl.ylabel(('$log_{10}$ ' if False else '') + 'dispersion of gene'
+                  + (' (normalized)'))
+            utils.savefig_or_show('filter_genes_dispersion_normalized', show=show, save=save)
+        else:
+            pl.ylabel(('$log_{10}$ ' if False else '') + 'dispersion of gene'
+                  + (' (not normalized)'))
+            utils.savefig_or_show('filter_genes_dispersion', show=show, save=save)
```
flying-sheep: Could you please eliminate repetition here? The only difference between the if branches is `'filter_genes_dispersion_normalized'` vs `'filter_genes_dispersion'`.

You can simply do

```py
file_name = 'filter_genes_dispersion{}'.format('_normalized' if id == 0 else '')
```

and use the variable in the block without repeating the whole thing. Also makes for a smaller diff.

scanpy/preprocessing/simple.py
```diff
@@ -122,38 +123,75 @@ def filter_cells(data, min_counts=None, min_genes=None, max_counts=None,
...
+            multi_option_error = True
+    if max_cells is not None:
+        if min_cells is not None or min_counts is not None or max_counts is not None:
+            multi_option_error = True
+    if max_counts is not None:
+        if min_cells is not None or min_counts is not None or max_cells is not None:
+            multi_option_error = True
+    if multi_option_error:
+        raise ValueError("Only provide one of the optional arguments (`min_counts`," +
+                         "`min_cells`, `max_counts`, `max_cells`) per call.")
```
flying-sheep: This is confusing and super long. I wouldn’t see at a glance what it really does, but it looks like it’s just doing this:

```py
n_given_options = sum(option is not None for option in [min_cells, min_counts, max_cells, max_counts])
if n_given_options > 1:
    raise ValueError(...)
```

Is that correct?

We should also maybe add a check if they all are `None`, like `if n_given_options == 0`

flying-sheep (changes requested):
Thanks for your contribution!

There’s a few things that can be made shorter and clearer, could you please do this?

scanpy/preprocessing/simple.py
jorvis: That is better, yes.  I originally started doing the conditional check like in the [filter_cells() method](https://github.com/theislab/scanpy/blob/master/scanpy/preprocessing/simple.py#L87), since it had all the options, but realized that checking would fail if someone passed some options together, such as min_counts and max_genes.  I will admit to often not doing the (quite pythonic) inline loop/conditional as you have done, but in this case it reads well.

jorvis: Please also remember to see my note that this isn't actually working still.  Not the option parsing, but my attempt to make it filter properly fails with a "ValueError: Length of values does not match length of index".

flying-sheep (approved):

flying-sheep:
Thank you! So you say it doesn’t work, but I see a green checkmark. Would you mind adding a test that exposes the error?

jorvis:
No problem - is there a standard input dataset you use for testing?  Otherwise I can just use one I have on-hand.

flying-sheep:
We have a few built in: https://github.com/theislab/scanpy/blob/master/scanpy/datasets/builtin.py

But AFAIK there’s no scRNA-Seq ones that aren’t dynamically downloaded.

I think for tests we should add a small built-in one, and make sure it doesn’t end up in the binary wheels when building.

gokceneraslan:
`paul15` is downloaded automatically, very practical.

gokceneraslan:
Ah, I didn't know others are also automatically downloaded, very nice.

paul15 uses `sc.utils.check_presence_download()`, while others use `check_datafile_present_and_download()` via `readwrite.read()`, though. Is this intentional?

flying-sheep:
> `paul15` is downloaded automatically, very practical.

Yeah, it’s really cool for interactive use, but not for automated testing / continuous integration I think.

jorvis:
Here's [test code](https://gist.github.com/jorvis/da877d89fd159b2fb7dfba26705f7ceb) and my output is:

```pytb
Initial shape: 737280x28002
After min_genes: 5128x28002
After max_genes: 1431x28002
  File "/tmp/test_cell_and_gene_filter.py", line 22, in <module>
    sc.pp.filter_genes(adata, min_cells=3)
  File "/home/jorvis/git/scanpy/scanpy/preprocessing/simple.py", line 152, in filter_genes
    adata.var['n_cells'] = number
  File "/usr/local/lib/python3.6/dist-packages/pandas/core/frame.py", line 2519, in __setitem__
    self._set_item(key, value)
  File "/usr/local/lib/python3.6/dist-packages/pandas/core/frame.py", line 2585, in _set_item
    value = self._sanitize_column(key, value)
  File "/usr/local/lib/python3.6/dist-packages/pandas/core/frame.py", line 2760, in _sanitize_column
    value = _sanitize_index(value, self.index, copy=False)
  File "/usr/local/lib/python3.6/dist-packages/pandas/core/series.py", line 3121, in _sanitize_index
    raise ValueError('Length of values does not match length of ' 'index')
ValueError: Length of values does not match length of index
```

Note that this same error displays on both of the following lines:

```python
sc.pp.filter_genes(adata, min_cells=3)
sc.pp.filter_genes(adata, max_cells=1000)
```

falexwolf:
Sorry for taking so long to merge this. I got a bit confused about all the dicussions here... Anyways. This seems to be fine, now!

Thank you, Joshua & Philipp!

jorvis:
Hmmm, you got it to successfully run?   The last thing I had posted was failures in testing (see above)

falexwolf:
Oh, damn... It wasn't clear to me that you were actually saying that this is broke... Sorry, I went over a few pull requests too quickly.

Anyways, so the master branch was broke for a day. It's resolved in https://github.com/theislab/scanpy/commit/96890730972162aa531c3289b38ad728a7585c85

The next pull request goes smoother... :wink:

PR #82, opened 2018-02-09
Fixes erroneous function call

dawe:
The `add_score` function comes from pull request and has been renamed in `score_gene_list` when was merged

### DIFF
--- a/scanpy/tools/score_gene_lists.py
+++ b/scanpy/tools/score_gene_lists.py
@@ -133,10 +133,10 @@ def score_cell_cycle_genes(
 
     ctrl_size = min(len(s_genes), len(g2m_genes))
     # add s-score
-    add_score(adata, gene_list=s_genes, score_name='S_score', ctrl_size=ctrl_size)
+    score_gene_list(adata, gene_list=s_genes, score_name='S_score', ctrl_size=ctrl_size)
 
     # add g2m-score
-    add_score(adata, gene_list=g2m_genes, score_name='G2M_score', ctrl_size=ctrl_size)
+    score_gene_list(adata, gene_list=g2m_genes, score_name='G2M_score', ctrl_size=ctrl_size)
 
     scores = adata.obs[['S_score', 'G2M_score']]
 


falexwolf:
Dear Davide, again sorry for the late response...

I'll, of course, merge this and you can directly push such small things on the master... 

In the meanwhile, I've made another release (see notes on https://scanpy.readthedocs.io).

It would be cool to further work on the gene scoring with the small amendments (default gene list class, a small notebook with a use case) to get this in a new release.

Also, in this context. I know think that `score_genes` instead of `score_gene_lists` would be better. The `_lists` does not contain actual information and is self-understood. It would also fit nicely with other functions (`cluster_genes`, e.g.).

Until all of this is fixed, I removed this stuff from the API docs...

Have a good start into the week!
  Alex

dawe:
Hi @falexwolf , 
I'd like to directly push such small things but either I do not have permissions or I'm not proficient enough in git :-/
That said, following another thread about gene lists, I'm quite skeptical about the implementation of gene list class, or at least I believe it is out of scope. How would you manage, for example, gene identifiers? The list used for cell cycle is for humans and include HGCN symbols, but Entrez ID or ENSEMBL ID may be used, and since there may be confusion in ID translation, I'd leave the definition/translation part to the final user.
As for the notebook, I pushed a notebook for cell cycle in scanpy_usage repo, it could be modified/prettified/whateverfied.
cheers

d

falexwolf:
Dear Davide,
sorry, I guess this is my bad. I was sure I added you as a collaborator already more than a week ago,  but it still didn't list you. So, I hope, that in the meanwhile, you received a list.
OK, I agree with you regarding the gene list class.
Cheers,
  Alex

falexwolf:
I'm right now editing the notebook and making some small changes to your code again.

falexwolf:
Hi @dawe!

OK, I'm done. I think now everything makes a nice impression and is consistent with the other naming conventions etc. in Scanpy
https://scanpy.readthedocs.io/en/latest/api/scanpy.api.tl.score_genes.html
https://scanpy.readthedocs.io/en/latest/api/scanpy.api.tl.score_genes_cell_cycle.html
Both link to the notebook in the "Examples" section
https://nbviewer.jupyter.org/github/theislab/scanpy_usage/blob/master/180209_cell_cycle/cell_cycle.ipynb

I also added you to the author list (already wanted to do this before)...

Sorry that it took so long... the kids...

Thank you very much! :smile:
  Alex

raph06:
Hi, 

I just stumbled upon the same error, maybe due to the installation method.
Anyway, I got it fixed but since my package was outdated and it's been a while since I used python maybe the next point as already been solved:

-typo issue in the notebook examples (phase instead of Phase)
at In [8] when you call the pca.scatter function with color 'phase'

> ValueError: "phase" is invalid! specify valid sample annotation, one of ['n_genes', 'percent_mito', 'n_counts', 'dropouts', 'complexity', 'S_score', 'G2M_score', 'Phase', 'X_diffmap0', 'louvain_groups']


Very nice package and notebooks BTW,
Raphaël

falexwolf:
Hi Raphaël!

Thanks!

What do you mean with "I just stumbled upon the same error, maybe due to the installation method." - which error?

Regarding the typo: Hm, are you running Scanpy 0.4.4; if you run an early version, this was 'Phase' with a captical 'P'; since 0.4.3+7 it's 'phase'; like all the other annotations.

Cheers,
  Alex

raph06:
Hi, 

I indeed run Scanpy 0.4.3, that's why I got stucked.
The score_gene_list function was named add_score, and the 'phase' needed to be capitalized..

Raphaël

PR #83, opened 2018-02-09
Show score name in logging

dawe:


falexwolf:
Just push things like this on the master branch. :wink: And, consider using `'a string with a variable: {}'.format(variable)` for formatting strings. :smile:

PR #87, opened 2018-02-15
PCA function returns eigenvalues and stores it in adata.uns

LucasESBS:
PCA function now returns and store eigenvalues of each PCs computed during PCA in _adata.uns['pca_eigenvalues']_


falexwolf:
Thank you! I'll rename the result according to the scikit-learn convention, though. If they call it `explained_variance_`, we should also call it that way - we can add in the docs, that it equals the eigenvalues.

falexwolf:
OK, done: https://github.com/theislab/scanpy/commit/4124d8b1a72f4e1fd7ce9313353d8d08723bbb7a

PR #91, opened 2018-02-21
Clarify ambiguous help string of recompute argument in pca

gokceneraslan:


falexwolf:
Thank you, Gokcen!

All these recompute arguments are, of course, not a good thing - they arose when I first drafted Scanpy thinking that everything should be concise single-word command-line arguments... As mentioned on our Mattermost channel, I want to get away from it; there is one convenience preprocessing function `neighbors` and one class `Neighbors` with more control that should be called as a step in the preprocessing; all other tools will then delegate graph parameters to this...

PR #92, opened 2018-02-22
Add UMAP visualization support

gokceneraslan:
I added UMAP support for visualization \o/ Here is how MNIST "single cells" look like:

image

image

I'm not so familiar with code sytle and variable naming etc. yet, and I haven't fully tested things like additional umap kwargs ~and 3d visuazliation~ etc. but let's keep PR here and resolve things along the way.

### DIFF
--- a/docs/references.rst
+++ b/docs/references.rst
@@ -3,7 +3,7 @@
 
        <style type="text/css">
        .label {
-       }    
+       }
        </style>
 
 References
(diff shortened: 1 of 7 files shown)


gokceneraslan:
OK, this was easier than I thought:

image

docs/references.rst
```diff
@@ -3,7 +3,7 @@
 
        <style type="text/css">
        .label {
-       }
```
gokceneraslan, line 6: Sorry about removing whitespace. I just couldn't resist 🙈

gokceneraslan:
More examples:

image

image

falexwolf:
Nice, nice, nice! :smile: Great!

docs/references.rst
falexwolf: sure! there should be no whitespace here. maybe some relict from when this was all markdown and three whitespaces meant a line break

terooatt:
gokceneraslan 

Thank you very much for the example you have provided with umap utilization. It looks pretty straight forward to use. If I had initially combined different samples and generate a umap plot do you know how you could visualize one sample while keeping the overall structure of the umap pot?

PR #93, opened 2018-02-22
Fixes for sc.datasets.moignard15

gokceneraslan:
pandas is required by moignard15


gokceneraslan:
Also `groups_names` was undefined in the same function, I changed it to `groups_order` but I'm not sure if it's the correct variable.

falexwolf:
grrr, sorry, it's not been used for some time it seems... thank you!

PR #100, opened 2018-03-06
Downsample branch

LuckyMD:
Hey!

Here's the downsample function I wrote to downsample count matrices. Now the function is also loaded via the api.

Best,

Malte


falexwolf:
thank you very much! one thing that we could consider is renaming this to "downsample_counts"; for some people, "downsampling observations" is an alias to "subsampling observations" and for these the function name is not descriptive enough. what do you think? can I make this change?

LuckyMD:
Sounds good to me.

LuckyMD:
Also, you may want to change the error that is shown if a non-AnnData object is passed to the function. I put it as TypeError, because I didn't find how you have previously called that error.

PR #104, opened 2018-03-14
correct definition of dispersion

ckmah:


falexwolf:
Thank you! :smile: There is also taking the logarithm of the means involved and I'll add this for clarifcation, but of course you're right... there's no logarithm of the variance taken.

PR #114, opened 2018-04-01
Make diffusion maps compatible with new Neighbors class

gokceneraslan:
sc.tl.diffmap is not using new neighbors attributes.


falexwolf:
Oh man, how did that happen? To myself: Is there really no separate call to diffmaps anywhere in the notebooks?... Thank you so much!

gokceneraslan:
No problem. I think increasing the test coverage should be prioritised to make scanpy more robust.

falexwolf:
Yes... you're of course right. For now, I added the call of diffmaps to a notebook... When graph abstraction is finished, I'll think about systematically adding a lot more tests...

falexwolf:
PS: Your fix is also out in 1.0.1.

PR #116, opened 2018-04-05
Update PyPI badge

flying-sheep:
The new one is smaller, because it doesn’t say “package”. It’s also blue which shows it’s not a build status

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,9 +1,9 @@
-|Docs| |PyPI| |Build Status|
+|PyPI| |Docs| |Build Status|
 
+.. |PyPI| image:: https://img.shields.io/pypi/v/scanpy.svg
+    :target: https://pypi.org/project/scanpy
 .. |Docs| image:: https://readthedocs.org/projects/scanpy/badge/?version=latest
    :target: https://scanpy.readthedocs.io
-.. |PyPI| image:: https://badge.fury.io/py/scanpy.svg
-    :target: https://pypi.python.org/pypi/scanpy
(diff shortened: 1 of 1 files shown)


gokceneraslan:
Switching README.rst to markdown would also look nice on new pypi interface e.g. https://pypi.org/project/markdown-description-example/

flying-sheep:
The rendering being broken is a consequence of using HTML to make the readme look nicer on GitHub, and that won’t improve by using markdown. (if they filter HTML-in-rst, they’ll also filter HTML-in-md, right?)

Also, what advantage would the switch have?

PS: I think for a small, mostly-html-anyway README, md is a good choice, but it’s crap for technical documentation.

gokceneraslan:
Yeah, actually I don't get why even the header doesn't get rendered properly. I thought it's an rst issue, but you're right maybe md also doesn't make a difference.

falexwolf:
Thank you for the note Gökcen!

Yes, all this is due to the html snippets... indeed, on GitHub, html in Markdown renders properly and respects, e.g., `width` and would enable the same layout as on http://scanpy.rtfd.io, where the readme is [reused](https://raw.githubusercontent.com/theislab/scanpy/master/docs/index.rst). I don't know about PyPI. In any case, it would be ugly to transform between rst and md. On the other hand, the important thing is that the narrative is reused and the image could be embedded separately using a markdown version for GitHub and PyPi and the present html version for the docs... I'll do this when there is some more time. :wink:

PR #118, opened 2018-04-06
savefig with bbox_inches = 'tight'

iandriver:
Currently output figures are often cut off at the edges. Adding bbox_inches = 'tight' to savefig no longer cuts off figure text at edges. As far as I've tested it this shouldn't cause any other problems and simply saves figures with proper padding.


falexwolf:
Thank you! I hadn't actually noticed this problem. Did you call [`settings.set_figure_params()`](https://github.com/theislab/scanpy/blob/c3c2df8457322529e2e6a787cb86874f9ceb6ae1/scanpy/settings.py#L154-L182) [or `plotting.set_rcParams_scanpy()`]? But `bbox_inches='tight'` should indeed do no harm. So happy to merge! :smile:

PR #119, opened 2018-04-09
Added intersphinx, easier types in docs

flying-sheep:
As example for the change I [simplified scanpy.preprocessing.simple.filter_cells](https://github.com/theislab/scanpy/pull/119/commits/ae85520fcd16abbb1fb9748cf0b03fb35ce858b6#diff-1aa47c128676c77be1123acc601efd9eL19). Also I fixed a few docs problems.

The new format is now custom, so if you dislike the grey headers, that’s easy.

Before | After
----------|----------
before | after

### DIFF
--- a/docs/basic_usage.rst
+++ b/docs/basic_usage.rst
@@ -13,40 +13,40 @@ in ``sc.tl``, e.g.::
 
     sc.tl.louvain(adata, **tool_params)  # cluster cells using Louvain clustering
 
-where ``adata`` is an :class:`~scanpy.api.AnnData` object. Each of these calls adds annotation to an expression matrix *X*, which stores *n_obs* observations of *n_vars* gene expression variables. For each tool, there is at least one associated plotting function in ``sc.pl``, which retrieves and plots the added annotation::
+where ``adata`` is an :class:`~anndata.AnnData` object. Each of these calls adds annotation to an expression matrix *X*, which stores *n_obs* observations of *n_vars* gene expression variables. For each tool, there is at least one associated plotting function in ``sc.pl``, which retrieves and plots the added annotation::
 
     sc.pl.louvain(adata, **plotting_params)
 
 If you pass ``show=False``, a `matplotlib.Axes <https://matplotlib.org/api/axes_api.html>`_ instance is returned and you have all of matplotlib's detailed configuration possibilities.
 
-To facilitate writing memory-efficient pipelines, by default, Scanpy tools operate *inplace* on ``adata`` and return ``None`` - this also allows to easily transition to `out-of-memory pipelines <http://falexwolf.de/blog/171223_AnnData_indexing_views_HDF5-backing/>`_. If you want to return a copy of the :class:`~scanpy.api.AnnData` object and leave the passed ``adata`` unchanged, pass ``copy=True``.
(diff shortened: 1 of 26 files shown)


falexwolf:
Cool, very interesting! :smile: Greatest advantage at first sight for me: `scanpy.api.AnnData` is now `anndata.AnnData`. Also, you don't seem to have to mingle around with autodoc anymore, which seems a good thing... The simpler docstrings are also nice... but it's going to be a lot of work to rewrite all the docstrings... also, there might be some danger of introducing bugs as one needs to rewrite the function headers. Hm, ... I'm a bit hesitant to just do this right away... Let's discuss! :smile:

flying-sheep:
> Greatest advantage at first sight for me: scanpy.api.AnnData is now anndata.AnnData

to be fair, this was a simple consequence of adding intersphinx and would have been possible without the rest

> Also, you don't seem to have to mingle around with autodoc anymore, which seems a good thing...

yes, but by now i added another, less invasive hack to get the parameter doc style the way you want them. it would be nice to have numpydoc-style parameter rendering as a separate extension or sphinx option.

:skull_and_crossbones: the hack is also not finished, as in its current form, it’ll break docstrings with indentation (code blocks, lists, …). optimally the hack would be rewritten as a sphinx extension that can be loaded after the others. (it’s only a hack because it piggypacks on another extension just to ensure it runs last)

> it's going to be a lot of work to rewrite all the docstrings...

We don’t lose anything if we do it gradually: Unconverted Docstrings just render as they do now.

> there might be some danger of introducing bugs as one needs to rewrite the function headers

i don’t think it’s possible to get bugs this way: we’re just adding type annotations, we don’t change the defaults or the order or anything.

falexwolf:
Hey! I just wanted to find you in your office... We should discuss and look at this in person... Non-working indentation, for instance, would be a serious problem... I'd suggest waiting a little bit longer before you write more hacks... in principle, I was satisfied with the docs except for referencing `scanpy.api.AnnData`... I still don't see the big advantage of using type annotation outside of the docstrings... As they are right now, they look very good when rendered as html and they look good as plain text when invoked within a Jupyter notebook... Of course, you usually have the better arguments on such questions in the end, but I'm not fully convinced at this stage. So, let's discuss in person. :smile:

PS: Evidently, scanpy's style for docstrings simply imitates numpy, pandas, seaborn, scikit-learn. I'm not sure whether one should break these conventions... Also, you imagine how much work it was - many iterations over the past 12 months - to get all the type annotations, the "optional" keyword and the default value into the docstrings... as I'm busy like crazy with other stuff, I'm again hesitant to make such big changes... :wink:

flying-sheep:
> Hey! I just wanted to find you in your office... We should discuss and look at this in person...

Sure! I’m at my gradma’s place right now, but I’ll be back tomorrow or thursday

> Non-working indentation, for instance, would be a serious problem...

Yaya, that’s a very temporary hack because I just wanted a working version, nothing that stays.

> in principle, I was satisfied with the docs except for referencing scanpy.api.AnnData... I still don't see the big advantage of using type annotation outside of the docstrings...

The big one is that many of the bugs we had in the past and that we’ll have in the future can be prevented if your IDE/editor tells you “you can’t pass that thing here, wrong type.”

And by using type annotations in the code, it’s impossible to have them wrong (to forget a comma or so) and break the type format. I’ve seen quite some commits by you just fixing such a thing.

 > Evidently, scanpy's style for docstrings simply imitates numpy, pandas, seaborn, scikit-learn. I'm not sure whether one should break these conventions...

We don’t break them. Typeless parameter annotations are still in numpy style 😉

falexwolf:
I guess this here is now no longer needed...

flying-sheep:
Right.

PR #125, opened 2018-04-17
Added MNN_correct support

chriscainx:
Hi, I wrapped a R function from scran (mnnCorrect) in the preprocessing module. It's a new algorithm for batch effect removal ([Batch effects in single-cell RNA-sequencing data are corrected by matching mutual nearest neighbors](https://www.nature.com/articles/nbt.4091)) and has not been implemented in Python, so I thought it would be useful. (Extremely useful for myself 😁 )

### DIFF
--- a/requires.txt
+++ b/requires.txt
@@ -11,3 +11,4 @@ networkx
 natsort
 joblib
 numba
+rpy2
--- a/scanpy/preprocessing/simple.py
+++ b/scanpy/preprocessing/simple.py
@@ -9,12 +9,79 @@ import warnings
 from scipy.sparse import issparse
 from sklearn.utils import sparsefuncs
 from pandas.api.types import is_categorical_dtype
+from rpy2.robjects.packages import importr
+from rpy2.robjects import numpy2ri
 from anndata import AnnData
 from .. import settings as sett
 from .. import logging as logg
 
 N_PCS = 50  # default number of PCs
 
+def mnn_concatenate(*adatas, geneset=None, k=20, sigma=1, n_jobs = None, **kwargs):
+    """Merge AnnData objects and correct batch effects using the MNN method.
+
+    Batch effect correction by matching mutual nearest neighbors [Haghverdi18]_ 
+    has been implemented as a function 'mnnCorrect' in the R package 
+    `scran <https://bioconductor.org/packages/release/bioc/html/scran.html>`_ 
+    This function provides a wrapper to use the mnnCorrect function when 
+    concatenating Anndata objects by using the Python-R interface `rpy2 
+    <https://pypi.org/project/rpy2/>`_.
+
+    Parameters
+    ----------
+    adatas : :class:`~scanpy.api.AnnData`
+        AnnData matrices to concatenate with. Each dataset should generally be
+        log-transformed, e.g., log-counts. Datasets should have the same number
+        of genes, or at lease have all the genes in geneset.
+    geneset : `list`, optional (default: `None`)
+        A list specifying the genes with which distances between cells are 
+        calculated in mnnCorrect, typically the highly variable genes.
+        All genes are used if no geneset provided. See the `scran manual 
+        <https://bioconductor.org/packages/release/bioc/html/scran.html>`_ for
+        details.
+    k : `int`, ptional (default: 20)
+        See the `scran manual <https://bioconductor.org/packages/release/bioc/html/scran.html>`_ 
+        for details.
+    sigma : `int`, ptional (default: 20)
+        See the `scran manual <https://bioconductor.org/packages/release/bioc/html/scran.html>`_ 
+        for details.
+    n_jobs : `int` or `None` (default: `sc.settings.n_jobs`)
+        Number of jobs.
+    kwargs :
+        Keyword arguments passed to Anndata.concatenate
+
+    Returns
+    -------
+    An AnnData object with MNN corrected data matrix X.
+
+    Example
+    -------
+    >>> adata1
+    AnnData object with n_obs × n_vars = 223 × 33694 
+        obs: 'n_genes', 'percent_mito', 'n_counts', 'Sample', 'Donor', 'Tissue'
+        var: 'gene_ids', 'n_cells'
+    >>> adata2
+    AnnData object with n_obs × n_vars = 1457 × 33694 
+        obs: 'n_genes', 'percent_mito', 'n_counts', 'Sample', 'Donor', 'Tissue'
+        var: 'gene_ids', 'n_cells'
+    >>> adata3 = sc.pp.mnnconcatenate(adata2, adata1, geneset = hvgs)
+    """
+    adata = AnnData.concatenate(*adatas, **kwargs)
+    if geneset is None:
+        datamats = tuple([adata.X.T for adata in adatas])
+    else:
+        datamats = tuple([adata[:,geneset].X.T for adata in adatas])
+    n_jobs = settings.n_jobs if n_jobs is None else n_jobs
+    numpy2ri.activate()
+    rbase      = importr('base')
+    rscran     = importr('scran')
+    bpparam    = importr('BiocParallel').MulticoreParam(workers = n_jobs) if n_jobs > 1 else importr('BiocParallel').SerialParam()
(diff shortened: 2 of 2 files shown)


chriscainx:
The build environment doesn't have R installed so the checks failed....

gokceneraslan:
One of the aims of scanpy is to be self-contained and easy-to-install for users and also to be easy to maintain by the developers. Heavy dependencies like louvain and python-igraph are already troublesome, expecting users to have rpy2 + proper R installation + Bioconductor + scran would risk smooth user experience and easy maintainability.

I was wondering whether it makes sense to have a community-maintained `scanpy-contrib`  or `scanpy-extensions` repository (and python package) similar to https://github.com/keras-team/keras-contrib ? There are also couple of things I have in mind like `sc.pl.netsne(adata, anotheradata)` for embedding unseen samples via parametric tSNE, or `sc.tl.simlr` and `sc.pl.simlr` for [SIMLR](https://github.com/BatzoglouLabSU/SIMLR) via RPy2 bridge... 

These are popular requests for Scanpy and people expect the same convenient API and an easy integration with AnnData objects. However, they will probably not be included in the mainstream Scanpy because of the reasons I mentioned above.

What do you think @falexwolf and @flying-sheep ?

falexwolf:
Completely agree, Gökcen!

How I just thought about dealing with this in the past couple of minutes: could we not make a submodule *rtools*? We could show the contained wrapper functions on an extra page of the API. All of the dependencies of this would be optional. In effect, this would be a very shallow wrapper that is only interesting for people who already have a working R installation etc. and use Scanpy along with R packages. As there are quite many of these people, this is definitely meaningful.

The code would still look proper. Implementing tests for these wrappers is maybe not so important as these are only shallow interfaces. It would be easier to have this in the main scanpy repository than setting up a `scanpy-contrib`: I imagine less people will like to contribute and take the burden of maintaining another repository. PS: `anndata` is a different story. That's something that is meant to be so basic that it doesn't need a lot of maintenance an contributions.

What do you think?

chriscainx:
I completely agree that including the R/scran requirements will be troublesome and harms user experience. The reason I used a R-py interface is that there's no decent MNN correct on python yet, and scran's implementation is already fast and efficient enough, and I think this is meant to be an optional feature that provides a handy fix for those in need. Personally I would prefer if you guys create a submodule _rtools_, and put wrappers inside. This is going to be awesome to use and easy to maintain.

gokceneraslan:
> Completely agree, Gökcen!
> 
> How I just thought about dealing with this in the past couple of minutes: could we not make a submodule rtools? We could show the contained wrapper functions on an extra page of the API. All of the dependencies of this would be optional. In effect, this would be a very shallow wrapper that is only interesting for people who already have a working R installation etc. and use Scanpy along with R packages. As there are quite many of these people, this is definitely meaningful.
> 

That'd make things a lot easier for many people (including myself 😃), I agree. However

1) There are (and will be) so many R packages about single cell, so once we open the door, there might be so many requests about these packages so that it'd be difficult to decide what to include and what not to include. The decision might be a bit arbitrary. This is why I suggested a contrib repo, which will have everything users request (as soon as there is someone who is willing to maintain it), in a `use at your own risk` way...

2) There might be several bug reports about rpy2 itself or thin wrappers or R installation or R packages themselves. I was wondering whether this might introduce more maintenance burden, although supported packages will be limited.

> The code would still look proper. Implementing tests for these wrappers is maybe not so important as these are only shallow interfaces. It would be easier to have this in the main scanpy repository than setting up a scanpy-contrib: I imagine less people will like to contribute and take the burden of maintaining another repository. PS: anndata is a different story. That's something that is meant to be so basic that it doesn't need a lot of maintenance an contributions.
> 
> What do you think?

Alternatively, we can just prepare jupyter notebooks with some Python 3 and some R cells in it (which is super easy via rpy2 magics anyway) for some R packages/functions like mnn or SIMLR and put those in scanpy_usage as a reference for the community. For example:

image

Another question is other single cell Python packages like magic, ZIFA or DCA, for example. There will hopefully be more in the future. A contrib repo might include these, as well i.e. `sc.tl.magic`

falexwolf:
I still don't see 100% why the submodule `rtools` would be so much worse than `scanpy-contrib`. The submodule would be separate from the rest of the package and we could write something on top of its  API overview page like: interfaces for R tools, address the maintainers of these tools for help... We need to keep some structure so that things remain clean, but yeah, additions will always be somewhat arbitrary. If someone suggests a "meaningful addition", we will accept it, if someone suggests something that does not seem to be of good quality, we will reject it. But this is not a problem only for `rtools` but for wrappers of python packages as well... see, for instance, https://github.com/theislab/scanpy/pull/126.

So, I would set up the `rtools` submodule to save us the work of maintaining a different repo and the user the work of installing a `scanpy-contrib` package and figuring out which namespaces to use so that notebooks don't get completely messed up.



So, if you don't mind, I'd set up the "rtools" submodule...

falexwolf:
Regarding the other packages: of course, we will also interface those as optional dependencies... But I'd do it from the original Scanpy repo. To me, the whole problem is simply about keeping a clean structure and throwing clear error messages if optional dependencies are not installed.

falexwolf:
Regarding this pull request: I'd merge it, put the `rpy2` dependency in the interface and remove it from the requirements. and put the interface into `rtools`.

falexwolf:
I merged it and moved it into `rtools`. There is a reexport so that you should be able to call it using `sc.rtools.mnn_concatenate` where `sc` is `scanpy.api`.

Could you test whether this works?

falexwolf:
The above is meant for @chriscainx :smile:

falexwolf:
btw: have you by any chance also used the seurat cca implementation? i'm tending to like this a bit better...

chriscainx:
Thanks @falexwolf !! I will test if the feature works. Of course I've tested seurat cca, but it seems to work better when integrating data from different sequencing platforms. As for my own data, several batches generated by 10x, output if the MNN method looks more pleasing...
unknown

Plus, I haven't looked into the maths of CCA, but I have for MNN and feel more comfortable using it.

Actually, @gokceneraslan 's comments do make sense to me, and I've spent quite some time working on a native implementation of MNN correct on python. Now it's nearly complete and features more complete multicore support than the scran implementation.
screen shot 2018-04-25 at 20 25 17
I built it to be fully compatible with anndata and scanpy. Now it already runs much faster than the scran version, and I'm planning to add more speedups, eg Cython and CUDA. I'm thinking of creating a full toolbox for scanpy, like scran for scater/sce, in python. Perhaps we could work together? 😄

I'm currently writing docstrings and will pack and upload the code to a repository shortly.

falexwolf:
Cool! I'm of course happy to work together! :smile: And, of course, a Python implementation is way cooler than an R one. :wink:

The last addition to Scanpy was an interface for [pypairs](https://github.com/theislab/scanpy/blob/master/scanpy/tools/pypairs.py). It's still not in the docs but will be very soon, I'm right now missing a link to a short example, there.

I'm guess you're planning to put your own package on PyPI, right? If yes, than it's not so crucial: for Scanpy, I changed from cython to numba. Many people currently do that, if you try numba, you'll realize why. So, I'd recommend to use numba if you need to speedup something.

chriscainx:
You are right I'm putting it on PyPI, and guess what, I'm already using numba on it😄

I finished some docs and uploaded it to my repo [mnnpy](https://github.com/chriscainx/mnnpy). There is still much to optimize, according to my tests, the 'adjust variance' step takes most time, and the best solution may still be rewriting it in C, as scran did, although somehow they made it single-threaded. In the current version I used jit and multiprocess.

I'm afraid we'll have to remove the 'mnn_concatenate' from /rtools though 😂

falexwolf:
Cool! Very happy to get another pull request for an interface to `mnnpy`. :smile:

Regarding writing it in C: I disagree, numba-boosted Python code is much nicer for these type of "relatively simple" algorithms... 

Regarding `rtools`. I'll remove it from the `api` but leave it in scanpy as an example for how one could wrap other r packages... No user will notice that...

gokceneraslan:
`mnnpy` looks really great! Thanks a lot for the implementation!

PR #131, opened 2018-04-26
Added wrapper for mnnpy

chriscainx:
Added wrapper for mnnpy in pp


falexwolf:
Thank you very much! Looks awesome! I'll put it into a separate file, though.

PR #134, opened 2018-04-27
Exclude nans from normalized dispersion vector

### LINKED ISSUE #115: filter_genes_dispersion(adata, n_top_genes=x) retrieves fewer genes if zero-expression genes exist
I was trying to get top 1000 variable genes in 1.3M dataset, but every time I ended up with zero genes after the `sc.pp.filter_genes_dispersion(adata, n_top_genes=1000)` call.

The reason is that `sc.pp.filter_genes_dispersion(adata, n_top_genes=x)` actually returns `x - num_zero_expression_genes` genes instead of x, where num_zero_expression_genes represents number of genes without any expression. 

Here is a small reproducible example:

image

It's easy to fix with a prior `sc.pp.filter_genes(adata, min_counts=1)` call, but I think filter_genes_dispersion should retrieve n_top_genes, regardless of presence of zero expression genes.

gokceneraslan:
`nan`s accumulate in the beginning of `dispersion_norm` vector and lead to an overestimated dispersion cutoff, if there are zero-expression genes.

Fixes #115.

### DIFF
--- a/scanpy/preprocessing/simple.py
+++ b/scanpy/preprocessing/simple.py
@@ -349,6 +349,7 @@ def filter_genes_dispersion(data,
         raise ValueError('`flavor` needs to be "seurat" or "cell_ranger"')
     dispersion_norm = df['dispersion_norm'].values.astype('float32')
     if n_top_genes is not None:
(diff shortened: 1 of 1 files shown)


falexwolf:
Thank you!

PR #136, opened 2018-05-01
Add PHATE integration

scottgigante:
Hi! I've added two possible APIs for PHATE: I imagine you'll only want to include one of them. 

The first, `sc.tl.PHATE`, is an object-oriented interface that matches `phate.PHATE`, which looks like an `sklearn` estimator class. The second is a functional interface which is equivalent to running `phate.PHATE().fit_transform(data)`. The second is more simple, but less powerful as any changes in parameters will require the entire operator to be recomputed, where the object-oriented interface allows partial recomputation.

Please let me know which you prefer (or if you would like to include both!) If we only include the functional version I would probably rename it to `phate` rather than `run_phate`.

### DIFF
--- a/docs/references.rst
+++ b/docs/references.rst
@@ -91,6 +91,10 @@ References
    *Visualizing data using t-SNE*,
    `JMLR <http://www.jmlr.org/papers/v9/vandermaaten08a.html>`_.
 
+.. [Moon17] Moon *et al.* (2017),
+   *PHATE: A Dimensionality Reduction Method for Visualizing Trajectory Structures in High-Dimensional Biological Data*,
+   `BioRxiv <http://biorxiv.org/content/early/2017/03/24/120378>`_.
+
 .. [Satija15] Satija *et al.* (2015),
    *Spatial reconstruction of single-cell gene expression data*,
    `Nature Biotechnology <https://doi.org/10.1038/nbt.3192>`_.
--- a/requires.txt
+++ b/requires.txt
@@ -11,4 +11,3 @@ networkx
 natsort
 joblib
 numba
-phate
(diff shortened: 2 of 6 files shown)


falexwolf:
Hi Scott!

Sorry that I didn't see this before writing the following email this moring:

> Sorry about forgetting to mention this in the first email below (just thought I'd send you the link to the most recent wrapper). The tSNE tool is a wrapper, too, and also has an example for a visualization function
> https://github.com/theislab/scanpy/blob/master/scanpy/tools/tsne.py
> https://github.com/theislab/scanpy/blob/master/scanpy/plotting/tools/__init__.py#L426-L518
> You'll find the corresponding pair in the docs. If your wrapper parallelizes this, this will make it very easy for Scanpy users to use out PHATE.

Scanpy's toplevel API does - on purpose - not feature objects, but convenience functions. If people want to do more complicated stuff, they should directly use the PHATE package, in this case.

So, it would be nice if you'd only provide the convenience function - which is already very similar to what I pasted about tSNE.

Also, please remove phate from the requirements... and raise a corresponding ImportError in your convenience function. Scanpy has quite some optional dependencies and Phate will be one of them.

Thank you!

falexwolf:
PS: You don't need a test for this... it would require installing phate on travis and this would take time... Also, the interface is trivial. You should simply link to your package within the docs to redirect people for bugs and more info.

scottgigante:
@falexwolf thanks very much for the extra information! I've copied the t-SNE API so it should fit much more cleanly into the scanpy interface. Let me know if there's anything else you need me to change. Thanks again!

falexwolf:
Merged. :wink:

falexwolf:
Nice! You also provided an example in the doc strings...!

PR #139, opened 2018-05-03
Add PHATE to API docs

scottgigante:
Sorry for the noise - I didn't realise until after the last one merged that I had to add PHATE manually to the API docs.

Thanks!

### DIFF
--- a/scanpy/api/__init__.py
+++ b/scanpy/api/__init__.py
@@ -62,6 +62,7 @@ Tools: TL
    tl.pca
    tl.tsne
    tl.umap
+   tl.phate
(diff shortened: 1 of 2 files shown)


falexwolf:
No worries, I would have done this for the next release... 

Btw: cool that you made your actual `fit_transform` accept `AnnData` objects... Would not have been necessary, but nice! :wink:

scottgigante:
Thanks, and you're welcome! I figure if a scanpy user wants to take the plunge and use the OO interface, they shouldn't have to change the way they interact with their AnnData. Thanks again!

PR #141, opened 2018-05-03
Programmatic mitochondrial gene symbols retrieval

fbrundu:
I added a method for programmatic retrieval of mitochondrial gene symbols through BioMart (instead of using a regular expression, this may be less error-prone and constantly up-to-date).
Let me know if you are interested in merging it, and if the code style is acceptable for this library.
I was unsure on how to test it, in case do you have any suggestions?

Thanks,
Francesco


falexwolf:
Nice! Thank you! That's certainly very meaningful. It should just go somewhere else, not in `preprocessing.simple`... Let me think about where to put these queries... We'll have more of this sort of thing in the future! I'll certainly merge this and move it around... However, we'll not make bioservices a hard requirement.

fbrundu:
I updated the commits, making bioservices optional.

fbrundu:
Hopefully last update on this PR. What I did:
- I noticed a regression on the method `rank_genes_groups_violin`, therefore I reverted back the code to the original one and I added an additional method `genes_groups_violin` which should be used if we want to pass the list of genes directly to the violin plot. The code is just a POC, but maybe it can be integrated
- Within the same method `rank_genes_groups_violin`, I found a bug: the ax variable was overwritten for each group (I don't know if it gave you error before). In my case, all the plots were merged into a single figure, every one on top of the previous ones
- Additionally, the parameters `gene_symbols` and `computed_distribution` were not defined within the method `rank_genes_groups_violin`. I added a default parameter (`None`) for `gene_symbols`, since it was defined in the docstring. With `computed_distribution` I didn't know what you wanted to do so I temporarily commented the line that used it

falexwolf:
Sorry for taking so long with this, I'm merging this now... I'll add a few more commits on the master branch subsequently.

Thank you again!

PR #147, opened 2018-05-04
Fixed Pandas .loc notation and added gene_names parameter to violin

fbrundu:
With this PR, I tried to fix issue https://github.com/theislab/scanpy/issues/140. Needs to be tested still.

Additionally, I found it useful to provide to scanpy directly the gene names when plotting violins. Use case: you have a different algorithm that is not implemented in scanpy for differential gene expression, and you want to plot its results with scanpy.

Feel free to reject if there's a better way to do it currently in scanpy. I didn't find it though.


falexwolf:
Makes sense! Thank you!

PR #157, opened 2018-05-19
SPRING v2 export

swolock:
I updated the export function `sc.export_to.spring_project` to work with our new version of SPRING (currently at https://github.com/allonkleinlab/spring_dev).


falexwolf:
Awesome! Thank you! I'm merging this and will play around with it...

PR #161, opened 2018-05-25
Test on 3.5

flying-sheep:
I expect unrelated test fails; see #162


flying-sheep:
Only the tests that fail on master also fail here, so this is fine and can be merged

PR #164, opened 2018-05-29
regress_out multiprocessing

fidelram:
This PR adds multiprocessing to the pp.regress_out function. 

Here some benchmarks:

```python
import numpy as np
import pandas as pd
import scanpy.api as sc
from anndata import AnnData
from scipy.sparse import random

# create a matrix with 20.000 cells with 3000 genes
adata = AnnData(random(20000, 3000, density=0.6, format='csr'))
```
**Benchmark using ordinal variables**
```python
# create a categorical column and run regress out using 
# the categorical column
adata.obs['batch'] = pd.Categorical(np.random.randint(1, 4, size=adata.X.shape[0]))
%timeit res = sc.pp.regress_out(adata, keys='batch', n_jobs=20, copy=True)
```
> 8.44 s ± 292 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```python
# import previous version of the function (which I saved in the file simple_old.py)
from scanpy.preprocessing.simple_old import regress_out_old
%timeit res_old = regress_out_old(adata, keys='batch', n_jobs=1, copy=True)
```
> 1min 5s ± 4.45 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

**Compare that the previous and the new output are the same**
```python
np.array_equal(res.X, res_old.X)
```
> True

**Benchmark using ordinal variables**
```python
adata.obs['percent_mito'] = np.random.rand(adata.X.shape[0])
adata.obs['n_counts'] = adata.X.sum(axis=1)

%timeit res2 = sc.pp.regress_out(adata, keys=['n_counts', 'percent_mito'], n_jobs=32, copy=True)
```
> 4.99 s ± 501 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```python
%timeit res2_old = regress_out_old(adata, keys=['n_counts', 'percent_mito'], n_jobs=1, copy=True)
```
> 41.2 s ± 7.79 s per loop (mean ± std. dev. of 7 runs, 1 loop each)

```python
np.array_equal(res2.X, res2_old.X)
```
> True

### DIFF
--- a/scanpy/preprocessing/simple.py
+++ b/scanpy/preprocessing/simple.py
@@ -703,29 +703,32 @@ def regress_out(adata, keys, n_jobs=None, copy=False):
         The annotated data matrix.
     keys : `str` or list of `str`
         Keys for observation annotation on which to regress on.
-    n_jobs : `int` or `None`, optional (default: `None`)
-        Number of jobs for parallel computation. Currently has no effect.
+    n_jobs : `int` or `None`, optional. If None is given, then the n_jobs seting is used (default: `None`)
+        Number of jobs for parallel computation.
     copy : `bool`, optional (default: `False`)
         If an :class:`~scanpy.api.AnnData` is passed, determines whether a copy
         is returned.
 
     Returns
     -------
-    Depening on `copy` returns or updates `adata` with the corrected data matrix.
+    Depending on `copy` returns or updates `adata` with the corrected data matrix.
     """
     logg.info('regressing out', keys, r=True)
     if issparse(adata.X):
         logg.info('    sparse input is densified and may '
                   'lead to high memory use')
     adata = adata.copy() if copy else adata
-    if isinstance(keys, str): keys = [keys]
+    if isinstance(keys, str):
+        keys = [keys]
+
     if issparse(adata.X):
         adata.X = adata.X.toarray()
-    if n_jobs is not None:
-        logg.warn('Parallelization is currently broke, will be restored soon. Running on 1 core.')
+
     n_jobs = sett.n_jobs if n_jobs is None else n_jobs
+
     # regress on a single categorical variable
     sanitize_anndata(adata)
+    variable_is_categorical = False
     if keys[0] in adata.obs_keys() and is_categorical_dtype(adata.obs[keys[0]]):
         if len(keys) > 1:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


scanpy/preprocessing/simple.py
```diff
@@ -740,52 +743,78 @@ def regress_out(adata, keys, n_jobs=None, copy=False):
...
-    def _regress_out_chunk(chunk, responses, regressors):
-        chunk_array = np.zeros((responses.shape[0], chunk.size),
-                               dtype=responses.dtype)
-        for i, col_index in enumerate(chunk):
-            chunk_array[:, i] = _regress_out(col_index, responses, regressors)
-        return chunk_array
-
-    for chunk in chunks:
-        # why did this break after migrating to dataframes?
-        # result_lst = Parallel(n_jobs=n_jobs)(
```
falexwolf, line 780: Hi @fidelram! Thank you very much for the PR! Absolutely awesome! Also with the benchmarks... Just one question: Does your solution not have the memory issues that I encountered here? I'm just a bit worried that things will crash when you pass, say a list of 20 genes as regressors. If you're confident that this is fine, I'm happy to merge this immediately.

fidelram: In the current implementation, each processor receives a) a copy of the adata.X matrix containing **only** the columns to regress and b) the columns used as independent variables. 

a) is efficient because only the portion needed is passed
b) is copied multiple times as jobs. If b is big this could take some memory but not much. The size of b is the number of cells x num. of independent variables. Even if the num of independent variables is 20, this should be too big. Let me run some tests to check this out.

fidelram: I forgot to mention that if a categorical variable is used, in which case a matrix of size equal to adata.X is generated, 'b' (from the above comment) is also efficient because only the part of the matrix needed is passed to each processor.

falexwolf: So you think that
```
result_lst = [_regress_out(	
                     col_index, adata.X, regressors) for col_index in chunk]
```
caused problems because I used the same reference to `adata.X` and `regressors` in several chunks/jobs? It definitely didn't generate additional memory and as these references are only used for reading, I thought it would be ok.

Anyways, I believe that you're implementation will probably not suffer from the problems I encountered with the previous solution when regressors contained, say, more than 5 to 10 variables.

Have you run the checks? Otherwise I'm happy to merge. It's really cool that you made a PR for this...

fidelram:
Thanks for merging. I did some test and the memory usage was not very high with 1 or 20 regressors. Nevertheless, I tried more memory efficient methods but the gain was minimal.

PR #169, opened 2018-06-08
added highest expressed genes QC

fidelram:
This PR adds the option to make an image like the following:

```python
sc.pl.highest_expr_genes(adata, n_top=40)
```

This plot is similar to the one produced by `scater` function `plotQC` and is useful to identify highly expressed genes in a sample.

To keep the code tidy I added the new plot on  `scanpy/plotting/qc.py` I imagine that other QC plots can be added in the future. Possible future improvements can plot multiple panels by splitting the data using batch for example.

Additionally, this PR:

 * Changes the *grey* dot color of `_scatter_obs` to *ligh_grey*. This results in a better contrast of colors. E.g.:
```python
sc.pl.umap(bdata, color='batch', groups=['PBMC'])
```

 * Added option to select the number of panels for `rank_genes_groups`. Without this option, if there are too many louvain groups, then the image is too wide. With the new parameter, it is easy to select how many panels per row should be plotted.

### DIFF
--- a/scanpy/plotting/qc.py
+++ b/scanpy/plotting/qc.py
@@ -1,0 +1,57 @@
+import seaborn as sns
+from matplotlib import pyplot as plt
+import pandas as pd
+from . import utils
+from ..preprocessing.simple import normalize_per_cell
+
+
+def highest_expr_genes(adata, n_top=30, save=None, show=None, ax=None, **kwargs):
+    """
+    Computes, for each gene, the fraction of counts assigned
+    to that gene within a cell. The n_top genes with the highest
+    mean fraction over all cells are plotted as boxplots.
+
+    This plot is similar to the `scater` package function
+    `plotQC(type = "highest-expression")`.
+    See (https://bioconductor.org/packages/devel/bioc/vignettes/scater/inst/doc/vignette-qc.html)
+
+    Parameters
+    ----------
+    adata : :class:`~scanpy.api.AnnData`
+        Annotated data matrix.
+    n_top : int, optional (default:30)
+        Number of top
+    save : `bool` or `str`, optional (default: `None`)
+        If `True` or a `str`, save the figure. A string is appended to the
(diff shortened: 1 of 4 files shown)


falexwolf:
Awesome! :smile:

scanpy/plotting/qc.py
```diff
@@ -0,0 +1,57 @@
...
+    **kwargs : keyword arguments
+        Are passed to `seaborn.boxplot`.
+
+    Returns
+    -------
+    A `matplotlib.Axes` object
+
+    """
+    # compute the percentage of each gene per cell
+    dat = normalize_per_cell(adata, counts_per_cell_after=100, copy=True)
```
fidelram, line 40: @falexwolf I am using `normalize_per_cell` to get the percentage of counts per gene in each cell. As a second thought this may problematic in case the `normalize_per_cell` is updated in the future. Maybe it is better to compute the values here directly. What do you think?

falexwolf: Yes, I noted that. I don't see a problem, there is a test in place for that, I'll add a remark to the test.

falexwolf: Here is the test and the remark: https://github.com/theislab/scanpy/blob/f48476a089cea14260f44c44286dd77b6307b039/scanpy/tests/preprocessing.py#L20-L21

falexwolf: PS: I added you to the author list 🙂

flying-sheep:
Another possible extension: Allow to select different distribution plots (for all of our distribution plots):

- box plot
- [violin plot](https://seaborn.pydata.org/examples/simple_violinplots.html)
- [swarm plot](https://seaborn.pydata.org/examples/scatterplot_categorical.html)
- [joyplot](https://seaborn.pydata.org/examples/kde_joyplot.html)

scanpy/plotting/qc.py
fidelram: Thanks for adding me to the author list! Such a honor!

PR #176, opened 2018-06-14
Add igraph and louvain to requirements

sidneymbell:


gokceneraslan:
Hi Sidney,

Thanks for the pull request. igraph and louvain are kind of heavy dependencies (e.g. takes long time to compile them and they're not easily available via PyPI for all platforms etc.), this is why they are excluded from requirements file. It's written in the [installation document](https://scanpy.readthedocs.io/en/latest/installation.html) that these need to be installed manually.

Also, there should be proper error messages stating that these must be installed separately when their functionality is needed for a function in scanpy and cannot be found. Did you get any other error regarding these packages?

sidneymbell:
Totally understandable, thanks!

flying-sheep:
What we could do instead is to add a [extras_require section](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) to the `setup()` call in setup.py

then people could do `pip install scanpy[louvain]` or `pip install scanpy[all]` to get the whole thing.

falexwolf:
I agree with Phil, but it's not a priority right now. The installation of both igraph and louvain has to be done only once... these packages don't evolve much. So I think it's OK for people to have this little inconvenience as it's only once in the beginning.

flying-sheep:
it’s a one-line-fix: #180

PR #180, opened 2018-06-20
Added louvain extra to setup call

flying-sheep:


falexwolf:
Would you also update the installation docs to show people how to call this?

flying-sheep:
Of course. In fact, I was already on it while you wrote that comment 😉

PR #183, opened 2018-06-28
Fix sc.write() for files without extensions

gokceneraslan:
`sc.write('filewithoutextension', adata)` is broken right now. I noticed it via a related report in DCA (https://github.com/theislab/dca/issues/13). Filenames without extension should fallback to final else and return False, but now IndexError exception is raised since `ext[-1]` is invalid for empty list.


falexwolf:
Thanks, this should do it, yes!

PR #184, opened 2018-06-29
fixed position of labels in pl.heatmap

fidelram:
In some cases the ordering of the labels of the `groupby` categories was not correct. This is fixed now. Also I did some small layout changes to increase the colorbars width.


falexwolf:
Thank you!

PR #186, opened 2018-07-01
Add DCA integration

### LINKED ISSUE #142: DCA integration
I'd like to integrate DCA with scanpy and I need to ask some questions regarding that. 

There are two use cases for DCA so far. One is denoising as a preprocessing step and using denoised matrix in downstream analyses, especially in the ones that require original feature space such as differential expression. This seems to be a better fit for the `preprocessing` package i.e. `sc.pp.dca(adata)` which overwrites `adata.X` with the denoised matrix. Users can use `copy=True` to keep adata.X of original matrix unchanged.

The other use case is the low dimensional representations of cells produced by the encoder. This can also be used for the downstream analyses that do not require original feature space e.g. clustering, and visualization. This is more like `sc.tl.dca(adata)` which stores the new representation into `adata.obsm['X_dca']`. Optionally it can store dropout probabilities (pi matrix) and the dispersion, as well.

However, having both sc.tl.dca and sc.pp.dca might be confusing for the users. Any suggestions?

gokceneraslan:
Added DCA integration, as discussed in #142 . For denoising, uses can call:

`sc.pp.dca(adata)`

which replaces adata.X inplace. For latent representations:

`sc.pp.dca(adata, mode='latent')`

can be called, which adds 'X_dca' to adata.obsm.

Fixes #142 . This integration uses new DCA API (>= 0.2.1). All DCA API arguments are exposed and usable in scanpy integration.

### DIFF
--- a/docs/references.rst
+++ b/docs/references.rst
@@ -22,8 +22,12 @@ References
    `PNAS <https://doi.org/10.1038/nmeth.3971>`_.
 
 .. [Csardi06] Csardi *et al.* (2006),
-   *The igraph software package for complex network researc*,
+   *The igraph software package for complex network research*,
    `InterJournal Complex Systems <http://igraph.org>`_.
+
+.. [Eraslan18] Eraslan and Simon *et al.* (2018),
+   *Single cell RNA-seq denoising using a deep count autoencoder*,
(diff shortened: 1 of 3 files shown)


falexwolf:
Nice! :smile:

PR #191, opened 2018-07-06
Speed up log1p

jamestwebber:
We noticed some performance issues with `scanpy.api.pp.log1p` and I think these changes will make it faster and more memory efficient. The `out` argument of `np.log1p` can be used to modify data in-place, conserving memory (this does require special-casing for sparse matrices, unfortunately). Because it's a ufunc, it's quite fast and efficient.

With those changes I removed the `chunked` and `chunk_size` arguments for this function, because I don't think they would actually help performance. In fact I found that using `chunked=True` was extremely slow, possibly because of having so many function calls. The only advantage I can imagine there is if you used parallelization on the chunks, but until that's implemented I think the option should be removed.

### DIFF
--- a/scanpy/preprocessing/simple.py
+++ b/scanpy/preprocessing/simple.py
@@ -404,7 +404,7 @@ def filter_genes_fano_deprecated(X, Ecutoff, Vcutoff):
     return gene_subset
 
 
-def log1p(data, copy=False, chunked=False, chunk_size=None):
+def log1p(data, copy=False):
     """Logarithmize the data matrix.
 
     Computes `X = log(X + 1)`, where `log` denotes the natural logarithm.
@@ -422,19 +422,20 @@ def log1p(data, copy=False, chunked=False, chunk_size=None):
     -------
     Returns or updates `data`, depending on `copy`.
     """
+    if copy:
+        data = data.copy()
+    
     if isinstance(data, AnnData):
-        adata = data.copy() if copy else data
-        if chunked:
-            for chunk, start, end in adata.chunked_X(chunk_size):
-                adata.X[start:end] = log1p(chunk)
+        if issparse(data.X):
+            np.log1p(data.X.data, out=data.X.data)
         else:
-            adata.X = log1p(data.X)
-        return adata if copy else None
-    X = data  # proceed with data matrix
-    if not issparse(X):
-        return np.log1p(X)
+            np.log1p(data.X, out=data.X)
+    elif issparse(data):
+        np.log1p(data.data, out=data.data)
     else:
-        return X.log1p()
+        np.log1p(data, out=data)
+
+    return data if copy else None
 
 
 def pca(data, n_comps=None, zero_center=True, svd_solver='auto', random_state=0,
--- a/scanpy/preprocessing/simple.py
+++ b/scanpy/preprocessing/simple.py
@@ -404,7 +404,7 @@ def filter_genes_fano_deprecated(X, Ecutoff, Vcutoff):
     return gene_subset
 
 
-def log1p(data, copy=False):
+def log1p(data, copy=False, chunked=False, chunk_size=None):
     """Logarithmize the data matrix.
 
     Computes `X = log(X + 1)`, where `log` denotes the natural logarithm.
@@ -424,16 +424,23 @@ def log1p(data, copy=False):
     """
     if copy:
         data = data.copy()
-    
-    if isinstance(data, AnnData):
-        if issparse(data.X):
-            np.log1p(data.X.data, out=data.X.data)
+
+    def _log1p(X):
+        if issparse(X):
+            np.log1p(X.data, out=X.data)
         else:
-            np.log1p(data.X, out=data.X)
-    elif issparse(data):
-        np.log1p(data.data, out=data.data)
+            np.log1p(X, out=X)
+
+        return X
+
+    if isinstance(data, AnnData):
+        if chunked:
+            for chunk, start, end in data.chunked_X(chunk_size):
+                 data.X[start:end] = _log1p(chunk)
+        else:
+            _log1p(data.X)
     else:
-        np.log1p(data, out=data)
+        _log1p(data)
 
     return data if copy else None
 


jamestwebber:
As a side-note: there isn't much harm in always returning the resulting object (rather than `None` when `copy=False`). It doesn't use memory (just another reference) and it allows for a more functional style of writing a processing pipeline.

falexwolf:
Hi James!

Thank you for the remark! And you're right... several repetitions of the following are consistent:
image

I'm aware of numpy's inplace functionality, and we've discussed it even for log1p... I don't know why we missed using it... Also, I wouldn't have thought that the speedup would be so dramatic,  but of course, already for better memory efficiency we should have done it.

I'll go through the rest of the toolkit and see whether there is another such striking omission...

Regarding the two other things you changed:
- `chunked` and `chunk_size` are in particular important when running an `AnnData` object in `backed` mode, when it's so large that it doesn't fit into memory. To date, this only works for the two functions that were the bottleneck for very large data (`pp.log1p` and `pp.pca`), where it already gives remarkable memory use reduction in `memory` mode. Of course, this is considerably slower than feeding in the full data matrix. We'll use AnnData's chunked functionality  in other tools, soon. We're also using it when working with tensorflow. At some point, when you open an AnnData in `backed` mode, the whole pipeline will run through by processing chunks and the user won't have to do a single change to his or her code. By that, code that has been written for data that fits into memory will automatically scale to many millions of observations. Also, there will be global settings that allow to manually determine whether the whole pipeline should run on chunks but still load the basic data matrix into memory, something we've found useful in several occasions.
- not returning `None` when modifying a reference inplace: the very first draft of Scanpy was written this way. then @flying-sheep remarked, that it shouldn't and I agreed with him right away:  if you return the changed object, you'll allow two different variable names for the same reference. This is a dangerous source for bugs - this was one of the few instances where I produced more bugs than in C++, where one would always write inplace functions (taking pointers or references) that return `void`. In addition, returning `None` directly tells the user that the typical code for writing pipelines does not have to be redundant: `function(adata)` instead of `adata = function(adata)`. Finally: all of Scanpy is consistently written using these principles and it would cause a lot of trouble both changing it in a simple function and changing it everywhere. Why do you think that _it allows for a more functional style of writing a processing pipeline_?

Hence, I'm sorry that I tend to not merge your pull request as is. Either you restore everything else that was there and solely add the inplace `np.log1p` or I'd do that. :smile:

Have a good Sunday!
  Alex

jamestwebber:
No problem, I'll change it to your preferred style. I don't think it's a problem to add the chunking but I'll need to test it for sparse matrices.

Just to clarify, what I meant by "more functional style" is something like this:

```
processed_data = raw_data.log1p().normalize(options...).some_other_method(options...)
```

That is, it allows a [functional programming](https://en.wikipedia.org/wiki/Functional_programming) style. Similar to libraries like `scikit-learn` (e.g. `fit()` returns `self` so you can immediately call another method) or `keras` (see the [functional API guide](https://keras.io/getting-started/functional-api-guide/).

But as you say, that might be a dramatic change in coding style for your library. I find it can lead to simpler code but that's a personal preference. The above examples are notable because they allow both functional and declarative styles of coding, depending on the user.

jamestwebber:
I modified the code to reintroduce `chunked` but I confess I haven't tested it because I've never been able to get it to work. I think there might be other issues with that functionality...

falexwolf:
@jamestwebber Thank you for reintroducing the functionality! :smile:

@Koncopd, could you check whether it still works as you're the one who introduced it in the first place?

Functional programming (scikit-learn): Yes, I assumed you were referring to functional programming. Also, I completely agree that it makes sense that a class method like `.fit(X)` should return `self` in order to be able to further operate on the modified object. But I wouldn't call that _functional programming_ and for a non-class-method, I'd say this is a completely different case. I think we're interpreting the Wikipedia article and all the tutorials on functional vs. object-oriented a bit differently, which is not problematic, of course. 🙂

Functional programming (Scanpy): When thinking about Scanpy's structure in the first place, I thought I'd write it in a largely functional way - where I mean that the central operations are based on functions on some fixed data container, not on objects that contain both data and functionality for the central operations - for two reasons: (1) the data structure was clear from the beginning, it was not terribly complicated but sufficiently complicated to write some non-trival infrastruture (`anndata.AnnData`) (2) the needed functionality of the toolkit was not clear and is not entirely clear to date. Also, a lot of different functionality is needed and putting all of it into one object would require to define a generic god class, which wouldn't be very transparent.

Functional programming (Keras): I'd be very astonished if `processed_sequences = TimeDistributed(model)(input_sequences)` modified `input_sequences` - this is taken from the tutorial you linked and summarizes the central operation. Hence, I think that the Keras functional API doesn't use `inplace` functions, as most APIs. Then, of course, there should be a return value. But maybe I'm mistaken. :wink:

jamestwebber:
I think you're right that Keras doesn't do in-place modifications (it's just building a computation graph there so there is nothing to modify). And the scikit-learn example is a bit different because it's a class method (although it would also make sense for `log1p` to be a class method, given that it only needs to exist for `AnnData` objects).

Your rationale for the Scanpy structure makes a lot of sense for the reasons you mentioned, but to me functional programming means immutable objects and functions that return their result with no side effects. Of course, memory constraints come into play here and immutability isn't viable, but the "return your result" part is still possible.

Personally I don't mind the possibility of multiple references to one object and so I prefer the option of either assigning the result or chaining functions together. In such a library it's easy to switch between an in-place or copying workflow, to inspect intermediate output if desired.

This behavior is what `numpy.log1p` itself is doing here, for that matter–with an `out` argument it still returns the array.

Anyway, obviously it's your call on how to design Scanpy, but I thought I'd give my perspective. Looking forward to see where you go with it.

falexwolf:
Thank you for all your thoughts! That's very interesting and helpful!

> although it would also make sense for log1p to be a class method, given that it only needs to exist for AnnData objects

Yes! I also think so. But then the question is which function makes into AnnData and which doesn't. Right now we only put functionality that is related to bookkeeping of the data into AnnData. Everything else remains out of it, even it's something as simple as `log1p`... but that's just a safeguard towards cluttering the object... I agree that it would be more convenient to have some of this in `AnnData`. I guess numpy went a similar way: not all of numpy's functions are available as `np.ndarray`'s class methods.

> In such a library it's easy to switch between an in-place or copying workflow, to inspect intermediate output if desired.

Interesting! I never thought of this.

> This behavior is what numpy.log1p itself is doing here, for that matter–with an out argument it still returns the array.

Yes! I think that's a good solution. The `out` argument is very verbose and allows setting a second name for the reference to the modified object, which is returned in addition. I thought about making `inplace` the default for Scanpy's function or not for a long time and finally decided for the unorthodox choice of making it the default - having in mind that AnnData's will become pretty large and at some point backed on disk (which hugely limits the possibilities of how you can write pipelines). Then the `out` rationale doesn't work anymore, as, by default, there simply is no second reference around...

Again, thank you for your perspective.

And, I'll merge this as soon as having figured out the `chunked` issue. Should be tomorrow or so...

Koncopd:
Seems it works properly
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/log1p_test.ipynb

falexwolf:
Thank you @Koncopd! Thank you @jamestwebber! I'm merging this.

falexwolf:
@Koncopd I made an actual test from your notebook :smile:: https://github.com/theislab/scanpy/commit/8d4ec6376c5b338456ced0bc051683052f15aa37

PR #193, opened 2018-07-09
Add MAGIC API

scottgigante:
MAGIC API modelled off the DCA API. A couple of comments:

- I changed `threads` to `n_jobs` to match `sc.tl` API
- I use `sc.settings.verbosity` rather than `False` for `verbose` default
- I updated the PHATE API while I was here
- I'm open to suggestions on how to handle `name_list`.

Re: `name_list`: MAGIC accepts both gene names and column indices. I've set it up currently to return to adata.obsm['X_magic'] if genes are specified as the number of columns is reduced from the original data, and to adata.X if all genes are returned. However, adata.obsm['X_magic'] now has no information about the order of the genes, unless I return it from MAGIC as a `pandas.DataFrame`, or as an AnnData object in its own right. What do you think?

P.S. MAGIC handling `genes` on AnnData input is currently on the dev branch, but will be merged to master once we settle on an API.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -52,6 +52,7 @@ For visual quality control, also see :func:`~scanpy.api.pl.highest_expr_gens` an
    :toctree: .
 
    pp.dca
+   pp.magic
 
 **Neighbors**
 
--- a/docs/references.rst
+++ b/docs/references.rst
@@ -152,6 +152,10 @@ References
    *Multicore t-SNE*,
    `GitHub <https://github.com/DmitryUlyanov/Multicore-TSNE>`__.
 
+.. [vanDijk18] Van Dijk D *et al.* (2018),
+   *Recovering Gene Interactions from Single-Cell Data Using Data Diffusion*,
+   `Cell <https://www.cell.com/cell/abstract/S0092-8674(18)30724-4>`__.
+
 .. [Weinreb17] Weinreb *et al.* (2016),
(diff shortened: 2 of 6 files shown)


falexwolf:
Looks good! Thanks! :smile:

falexwolf:
> However, adata.obsm['X_magic'] now has no information about the order of the genes, unless I return it from MAGIC as a pandas.DataFrame, or as an AnnData object in its own right. What do you think?

Just read this again. Right. This is not a good solution. Let met think about.

scottgigante:
One option is to not allow using to request a gene subset with MAGIC from Scanpy, but I don't particularly like that as the returned data is forcibly dense and could be extremely large. Alternatively, I could return a new AnnData object (force copy=True) with var_names corresponding to the genes that were returned.

falexwolf:
I'd say that the second option is very viable for the task of imputation, which is a bit different than all other tools - instead of "adding" annotation, one replaces the whole data matrix.

So, I'd say, if you do _not_ want to impute the whole data matrix as is, one should return a new AnnData for the wished gene selection.

Now the only problem is that people might get confused because magic doesn't do an inplace operation in this case.

So I'd leave `copy=None`, the default behavior being the inplace replacement of the full data matrix. If one passes a gene list, the passed AnnData should stay unchanged but a new imputed "AnnData" is returned. People can then directly use this for further tasks...

Sounds like a good solution to me.

PR #197, opened 2018-07-13
Add missing import

gokceneraslan:
`sc.pl.diffmap(adata)` gives `NameError: name 'pl' is not defined` due to missing pyplot import.


gokceneraslan:
Line 85: `if not settings.autosave and show: pl.show()`

falexwolf:
thanks!

PR #199, opened 2018-07-13
Add a 'dotplot' visualization to scanpy

fidelram:
This PR adds a visualization of genes, where, for each category (eg. cluster), each gene is represented as a circle whose diameter is proportional to the fraction of cells expressing the gene, and the color represent the mean expression of the gene. I saw this type of visualization on a talk by Dr. Hemant Suryawanshi from Rockefeller University but I couldn't find  a citation.

image

Furthermore, some improvements are introduced to the `pl.heatmap` and `pl.violin`. In particular, now is possible to swap the axis in multi_panel plot of `pl.violin` and to change the figure size in both types of plots. 

I updated an example here: https://gist.github.com/fidelram/2289b7a8d6da055fb058ac9a79ed485c


fidelram:
Anyone familiar with the travis error?

gokceneraslan:
Due to a misconfiguration in Travis setup, all tests are now running only with Python 3.7 now and there is a mysterious HDF error somewhat related to Python 3.7 and pytables.
Python version is fixed in https://github.com/theislab/scanpy/pull/201, so until we have Python 3.7 tests, we are good.

falexwolf:
Awesome, Fidel!

PR #201, opened 2018-07-15
Fix PYTHON_VERSION in travis installer

gokceneraslan:
PYTHON_VERSION variable is empty, so we actually pass `python=` in `conda create` so Travis always tests scanpy with latest Python in Conda distribution. Therefore Python 3.5 is actually never tested. Furthermore, conda switched to python 3.7, so now all test are run on Python 3.7. This is also the reason of weird HDF error message we get in tests.


falexwolf:
Thank you very much Gökcen! :smile:

Issue #204, opened 2018-07-16
Testing plotting functions


falexwolf:
Hi @fidelram !

I just reintroduced the simple multi-panel violin feature that is used right in the beginning of the [introductory tutorial](https://nbviewer.jupyter.org/github/theislab/scanpy_usage/blob/master/170505_seurat/seurat.ipynb): https://github.com/theislab/scanpy/commit/2a869d6060ba0de12feaf08a809bdf745d39ef10

For now, I simply hope this didn't break anything in your https://github.com/theislab/scanpy/pull/199.

For future work on the plotting part of Scanpy: I need to think about adding tests for this... In case that you have an idea for doing this in the best way, I'm happy to ready about it... As a simple solution, if you don't mind, I'd add the calls from your [gist](https://gist.github.com/fidelram/2289b7a8d6da055fb058ac9a79ed485c) to the introductory clustering notebook, so that I make sure that I don't break anything.

Best,
  Alex

fidelram:
I will check. Meanwhile, I realized that some errors were introduced in the latest plotting functions, thus I started working in a list of tests to avoid those problems in the future.

PR #207, opened 2018-07-16
 adding initial tests for plotting

fidelram:
and fixed an error with plotting functions that I think I caused when I merged with master.

I will set this PR as work in progress as I will be adding more tests.


fidelram:
@falexwolf regarding #204  the image that didn't work for you is this? I will address the conflict once this is clear because they are related.

image

scanpy/tests/test_plotting.py
```diff
@@ -0,0 +1,64 @@
...
+    sc.pl.dotplot(adata, adata.var_names, 'cell_type', use_raw=False)
+    pl.savefig(outfile.name, dpi=80, bbox_inches='tight')
+    res = compare_images(ROOT + '/master_dotplot.png', outfile.name, tolerance)
+
+    assert res is None, res
+
+    os.remove(outfile.name)
+
+
+def test_dotplot():
```
gokceneraslan: Same name as previous test.

fidelram: Thanks! Probably, I should go home and take some rest.

falexwolf:
Yes, the correct plot. I re-enabled the feature with https://github.com/theislab/scanpy/commit/2a869d6060ba0de12feaf08a809bdf745d39ef10, but I figure you'll have a more elegant solution. 🙂

falexwolf:
Yes, this is the correct plot. I re-enabled the feature with https://github.com/theislab/scanpy/commit/2a869d6060ba0de12feaf08a809bdf745d39ef10. 🙂

Your tests look very good! Seems like a great solution. :smile:

How about, instead of committing extra image files to the test repo, using the image files that we use for the docs? See here https://github.com/theislab/scanpy/tree/master/docs/api. It would be great to test whether what one sees in the docs is actually still the same as the current version produces.

The conflicts are solely due to the fact that I renamed `kwargs` to `kwds` as this fits better with the  "parameters" (instead of "args") convention, and to the fix of the plot above.

falexwolf:
Tell me when you're finished and I'll resolve the conflicts. No problem about that.

fidelram:
I will take a look later to see how we can integrate better the visualizations, the tests and the documentation. I will put back `kwds` also. 

Have you consider adding another dataset to the repository? This will be good for showing examples.

falexwolf:
I don't feel well adding a large dataset to the repository. Adding something subsampled and rather low-dimensional would be fine.

Alternatively, you could also just add a dataset that is automatically downloaded: as [here](https://github.com/theislab/scanpy/blob/7646c947f632ea7b09fea783e32a017136cfed24/scanpy/datasets/__init__.py#L104-L106) or [here](https://github.com/theislab/scanpy/blob/7646c947f632ea7b09fea783e32a017136cfed24/scanpy/datasets/__init__.py#L142-L144). As both travis and readthedocs will cache this, it should be a viable solution that avoids bloating the repository with several MB of data.

fidelram:
@falexwolf I like the idea of downloading the data on the fly!

For an idea of a small dataset that is already filtered and normalized, do you know if  the pbmc data from 10x Genomics has some restrictions? ... nevermind, they have a Creative Commons licence.

fidelram:
I think we can merge this for now. I will add other tests I think next week.

There is something I don't understand: For the docs, I see that an image is presented together with the API documentation. E.g. [api.pl.dotplot](https://scanpy.readthedocs.io/en/latest/api/scanpy.api.pl.dotplot.html) But I don't see how is this image is referred in the function description. How does this work?

falexwolf:
@flying-sheep could you explain it?

falexwolf:
Sorry, but I think there still is a range-related bug in the tests:
```
E           matplotlib.testing.exceptions.ImageComparisonFailure: Image sizes do not match expected size: (376, 439, 3) actual size (376, 440, 3)
```

fidelram:
my local tests pass. Which means that I will need to check the version of matplotlib. As soon as I can I will take a closer look.

fidelram:
Test are passing now! 

@flying-sheep can you tell me more about the images that appear in the automatic API documentation?

flying-sheep:
Sure! I made it as easy as possible, by adding it directly to the [autosummary template](https://github.com/theislab/scanpy/blob/181240602f2face3c49532f40b2fad4f300e3685/docs/_templates/autosummary/base.rst#L3). `api_image` is defined [in `scanpy/docs/conf.py` here](https://github.com/theislab/scanpy/blob/181240602f2face3c49532f40b2fad4f300e3685/docs/conf.py#L213-L218), and simply inserts a `..image` directive if a file exists at the right place.

Just add your (.png) image to [`scanpy/docs/api`](https://github.com/theislab/scanpy/tree/master/docs/api) using the qualified name of the function, e.g. `scanpy/docs/api/scanpy.api.pl.dotplot.png`

Maybe we should set up [Contribution guidelines](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) where people can read this.

falexwolf:
OK, merged this into master on the command line after fixing `plotting/anndata.py`, where some changes would have otherwise been reversed...

Thank you very much, Fidel! Really cool!

fidelram:
I hope that soon I can add some figures into the documentation and add further test.

PR #216, opened 2018-07-26
Update draw_graph.py

maximilianh:
mistake in arg doc


falexwolf:
I'd tend to deprecate the `key_ext` parameter. I think I never really used it in the end and I'm having a hard time imagining a canonical use for it. If you want to rename coordinates, you can always do that manually by operating on the AnnData object after calling the tool.

maximilianh:
Ah ok, fair enough, nevermind then.

PR #217, opened 2018-07-26
Update umap.py

maximilianh:
added keyname to umap function, like graph_draw()

### DIFF
--- a/scanpy/tools/umap.py
+++ b/scanpy/tools/umap.py
@@ -12,6 +12,7 @@ def umap(
         gamma=1.0,
         negative_sample_rate=5,
         init_pos='spectral',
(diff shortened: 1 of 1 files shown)


falexwolf:
Hm, I'm not super happy with that additional flexibility. People can break the visualization function `pl.umap` in that case. Why would you need it?

maximilianh:
I found the same parameter in the draw graph function and used it to make
sure paga doesn’t overwrite the original coords. Given that you don’t like
the original parameter, I agree, let’s just remove it everywhere. Never
mind! Guess I felt like doing pull requests today...

falexwolf:
Good! :smile:

Issue #218, opened 2018-07-26
How to specify color for each group with pl.tsne()


fbrundu:
Hi all,
I want to plot a tSNE plot with categorical groups.
I have already a palette that I need to use to maintain consistency across figures.
I tried to use the method `pl.tsne()` as the following:

```
sc.pl.tsne(adata_p, color='signature', size=15, palette=pal, groups=['Group1', 'Group2', 'Group3'])
```

Originally I set up a dictionary to use as a palette (using Seaborn), in particular it would be:

```
{'Group1': (0.06988755691850207, 0.2682780410742496, 0.405798717591302), 'Group2': (1.0, 0.4749019607843138, 0.4647058823529413), 'Group3': (0.8049999999999999, 0.8049999999999999, 0.8049999999999999)}
```

However, scanpy does not support a dictionary as argument for `palette` so I only used the values. But unexpectedly, the colors are assigned to each group not in alphabetic order (based on the name of the group). I can I assign directly the colors to each group, or make scanpy use the colors I want?

I tried to look into `adata.uns['signature']` but it contains only the list of the colors so I don't know exactly how to define it in the right order.

In particular, I noticed that the palette is in the right order, while the groups are not sorted in alphabetic order.

Thanks,
Francesco

wflynny:
Won't this automatically lexicographically sort, among other things, the order of plot legend labels for all scatter plots now?  That is probably not desirable as default behavior;  the order of the categories should be defined at the data object level, and not altered during plotting.

There should instead maybe be a utility function to properly assigns a palette to your labels prior to plotting, then the actual plotting functionality can stay the same.

fbrundu:
@wflynny actually I discovered that the behavior is a little more subtle: they are ordered based on the categories order.
Usually this is lexicographically. However, if you produce categories through [pandas.cut()](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.cut.html) they are ordered following the original bins - not lexicographically. I think that this original behavior is the more natural, so I think I'll close this issue and the linked PR.

Thanks,
Francesco

falexwolf:
Dear @fbrundu,

very sorry for the late response. I took a couple of days off with the family.

Everything is dictated by the order in `.obs['mycategorical'].cat.categories`. If you're starting off from a string `.obs['mystring']` annotation, then this will default to `natsorted` categories. If you pandas `.reorder_categories` then this will be reflected, too.

[...]

PR #219, opened 2018-07-26
Categories sorted for palette colors

fbrundu:
I think this should solve #218 , assuming what one assigns colors to label sorted in alphabetical order.


fbrundu:
Sorry for the PR, was not prop tested.

PR #221, opened 2018-07-26
 Sorted groups when plotting categorical data 

### LINKED ISSUE #218: How to specify color for each group with pl.tsne()

fbrundu:
Should fix #218. I have tested it, both with all groups and with some restricted groups.

### DIFF
--- a/scanpy/plotting/anndata.py
+++ b/scanpy/plotting/anndata.py
@@ -388,7 +388,7 @@ def _scatter_obs(
         mask_remaining = np.ones(Y.shape[0], dtype=bool)
         centroids = {}
         if groups is None:
-            for iname, name in enumerate(adata.obs[key].cat.categories):
+            for iname, name in enumerate(adata.obs[key].cat.categories.sort_values()):
                 if name not in settings.categories_to_ignore:
                     mask = scatter_group(axs[ikey], key, iname,
                                          adata, Y, projection, size=size, alpha=alpha)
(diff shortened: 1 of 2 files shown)


fbrundu:
I decided to close this PR since the current behavior is more correct in my opinion. See #218 for details.

PR #228, opened 2018-08-07
new ranked genes plotting functions

fidelram:
This PR introduces a family of plots to validate gene markers obtained using `scanpy.api.tl.rank_genes_groups`.

The plots work similarly as `scanpy.api.pl.rank_genes_groups`:
 
image

image

image

image

Those functions are wrappers around  `scanpy.api.pl.heatmap`, `scanpy.api.pl.stacked_violin`, `scanpy.api.pl.dotplot` and `scanpy.api.pl.matrixplot`

 `heatmap` and `dotplot` were modified to allow the 'brackets' on top of the images. This functionality can be used directly from those plots. E.g.:

image

The `pl.stacked_violin` plot was before part of  `pl.violin` but I thought that the code is cleaner by separating the two plots.

Also, this PR adds the new plot `scanpy.api.pl.matrixplot` that plots the average gene expression per category. E.g.:

image

This PR also updates several test for the plotting options and adds new ones. 

I tried to update the documentation to reflect the changes.

Also a new dataset called `pbmc68k_reduced` was added. This dataset is used for tests and for the example notebook [here](https://gist.github.com/fidelram/2289b7a8d6da055fb058ac9a79ed485c)

This dataset contains around 700 cells and 200 genes from the original 68k 10x genomics dataset and is saved as a small anndata object. It contains cell type annotations, UMAP coordinates and rank_gene_groups. The dataset can be accessed as
 ```python
```


gokceneraslan:
This looks super cool! Alternatively, one can assign colors to cell types (both rows and columns) and display a legend for cell type - color matching similar to `pheatmap` in R. E.g. 
image

But I think this brackets are more readable as it eliminates the need for color matching.

fidelram:
@gokceneraslan I like the idea, that would save quite some space. However, as a partial color blind person, I tend to avoid legends based on color because I can not map them reliably back to the figure. But otherwise, I think it is easy to adapt the current code to add such feature.

fidelram:
I don't know why the tests related to violin plots fail. I assume that it has to do with different libraries that produce slightly different shapes. But without access to the images generated during the testing would be difficult to say.

falexwolf:
Thank you so much, Fidel!

PR #234, opened 2018-08-10
Return to image directive

flying-sheep:
This is the only way the README renders on PyPI.

depends on github/markup#1222, which in turn depends on jch/html-pipeline#302

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,16 +13,17 @@
 Scanpy – Single-Cell Analysis in Python
 =======================================
 
-.. raw:: html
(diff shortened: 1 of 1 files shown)


falexwolf:
Cool!

flying-sheep:
Right now the image doesn’t float on GitHub, since the two mentioned PRs/issues aren’t merged/fixed.

If you’re OK with that we can leave it like that; If not we could separate the README from what’s used in the documentation and PyPI

PR #236, opened 2018-08-20
Layers

Koncopd:


falexwolf:
Thanks!

LuckyMD:
Are these layers for the velocyto implementation? I.e. spliced and unspliced count data layers. And is this then also usable for different layers of data processing?

Or am I missing the point entirely here?

falexwolf:
This is just being built up, but yes, the layers have the same constraints as in loom files and are meant to enable the transformation of loom files to AnnData objects without loss of information. Here's the stub of the docs: https://anndata.readthedocs.io/en/latest/anndata.AnnData.layers.html. Here's the very first draft https://github.com/theislab/scvelo of our stochastic RNA velocity package... There will be more information on this very soon. :smile:

LuckyMD:
I'm not that familiar with loom files, but it definitely sounds good ;).

So I guess every tool and plotting function would need a layers argument to determine which layer it should work on? At least if layers is implemented in a sufficiently general way to be used for data processing layers as well.

Any idea on the timeline for this? Then I could adapt the tutorial to work with this.

falexwolf:
In principle, everything works already. @VolkerBergen and @Koncopd have all the info on this.

PR #244, opened 2018-08-27
Simpler scatter functions

fidelram:
I tried to collect in one file the code used for plotting functions that use matplotlib scatter like `sc.pl.tsne`, `sc.pl.pca` and `sc.pl.umap` and others.

Also, I tried to annotate the code and improve the readability. 

Currently, the code is on a separate file called `scatter.py` and not integrated into the API as this facilitates comparison with previous code. 

Besides readability the proposed code can:
 * Plot a large number of plots in multiple columms (instead of a long row of plots)
 * Pass arguments directly to `matplotlib.pyplot.scatter` like vmax and vmin to adjust the color scale. When plotting multiple plots, this is useful to have a consistent range of values)

See cells 15 and 15 in this example:  https://gist.github.com/fidelram/8b43f786e7519bcfb7ffc0d5ccdbb0fe 
If the admins would like to merge these changes I can replaced the previous functions.

An example on how to use the code:

```python
import scanpy.plotting.tools.scatter as spl
spl.tsne(adata, color='louvain')
```

image

Further examples [here](https://gist.github.com/fidelram/8b43f786e7519bcfb7ffc0d5ccdbb0fe)


flying-sheep:
Could you please add images to [scanpy/docs/api](https://github.com/theislab/scanpy/tree/master/docs/api)?

fidelram:
@flying-sheep Sure, I can add the images as well. But first, I want to know if @falexwolf agrees to continue working on these changes.

falexwolf:
This looks good!

I'd also say that we should move towards a more transparent code for the scatter plots. It's a result of 1.5 years of subsequently adding features. No good initial plan about that.

The only problems I see:
- there might be a few bugs in this, so I'd like to wait until after the 1.3 release. 
- some people might want exactly the same appearance of the plots as before. I'd say that this should be possible as you just used the existing code snippets. Currently, however, there are slight differences regarding defaults of spaces etc. This is also why the tests fail.

What do you think?

fidelram:
@falexwolf Thanks for the support. 

I agree that this is not an urgent changes that can quietly be tested. Have you consider having a 'develop' branch were we can put all code like this? 

As you point out some differences are seen with respect to the shape of the plot when multiple panels are plot. This is mostly due to some code to add space for the colorbar and legends that can overlap nearby figures. Nevertheless, I can further adjust this to get plots that are more similar to the actual ones.

I think that the tests are failing because there is a clash between the module and a method called `scatter`. Once the code is cleaned this should go away. 

If you don't mind I would slowly start  removing redundant code and adding further tests. So, lets keep this PR open.

falexwolf:
Apologies for again, the late response @fidel! I married and moved to the US with twin babies last week. And in between, I spilled something over my laptop...

Yes, please go ahead and remove redundant code and add further tests. We'll merge this PR eventually.

And yes, we can think about a `develop` branch starting from 1.3. What do you say, @flying-sheep?

falexwolf:
How about we first merge all of this in a `plotting` branch on theislab/scanpy? As soon as you feel like it, Fidel.

falexwolf:
PS: I updated the [seurat-based tutorial](https://github.com/theislab/scanpy_usage/blob/master/170505_seurat/seurat.ipynb) and added a few of your plotting functions and a link to your gist. Feel free to update it further!

fidelram:
@falexwolf Can you take a look at the updated gist: https://gist.github.com/fidelram/8b43f786e7519bcfb7ffc0d5ccdbb0fe

Most of the previous and new plots are quite similar.

For diffmap I see different results but I suspect that there is a bug in the previous code.

Also, I don't have any example with arrows. Do you have any?

fidelram:
@falexwolf Do you know a more efficient way to get the value of a single column given the gene name. Currently, I am using:

```
adata[:, 'gene_name'].X
```

This is easy but not optimal. Any idea?

I will start updating the test once we are happy we the new results.

falexwolf:
Why is it not optimal, can you be more specific and maybe provide an example? It's supposed to be the canonical way. It should be as fast as it possibly gets. If you have sparse data, it could be worthwhile transforming `.tocsc()` before. I don't know whether I'd want to make this automatic, but one could think about it.

If there's something to improve here, this would be a top priority for me and I'd immediately fix it.

scanpy/plotting/tools/scatterplots.py
```diff
@@ -191,6 +191,10 @@ def plot_scatter(adata,
     else:
         args_3d = {}
 
+    if adata.raw is None and use_raw is True:
```
falexwolf: Since a while, the convention in all the plotting functions is that `use_raw=None` is the default (you might remember that I asked you for your opinion a few weeks back). Hence, you can throw an error if someone sets `use_raw=True` and `.raw is None`.

falexwolf (commented):
I clicked on the wrong button, wanted to comment not review. ;)

scanpy/plotting/tools/scatterplots.py
fidelram: I modify this to adjust to the convention.

fidelram:
@flying-sheep I thought that by doing `adata[:, 'gene_name'].X` the AnnData object does all sorts of checks which are redundant if I only want to access a single column on the data matrix. But if you say this is fine I will not change it.

falexwolf:
The checks shouldn't take any time. But you're right; you're also slicing the `.var` and the `.varm` annotations if you make this call and we could check whether this perceivably slows down things (only for very large data, I guess). If it does, it would be very simple to add an accessor `.slice_X` that enables convenient slicing of the data matrix. That's a bit ugly but would vanish in the plotting function. What do you think?

fidelram:
The tsne test is giving me a headache. There are some small difference even setting a `random_state`. I will remove the test.

falexwolf:
> The tsne test is giving me a headache. There are some small difference even setting a `random_state`. I will remove the test.

I completely get this... the exact UMAP and tSNE plots are simply not _exactly_ reproducible, just very similar... fortunately, clustering (even though that's also a greedy algorithm) and everything else are exactly reproducible. I also removed the tests for UMAP: https://github.com/theislab/scanpy/blob/1df151f678c50b9f85f5d65e7a47d061e4e6784b/scanpy/tests/notebooks/pbmc3k.py#L88-L91 :wink:

.travis_install.sh
```diff
@@ -27,5 +27,5 @@ source activate testenv
 
 pip install -r requires.txt
 conda install -c conda-forge python-igraph louvain --yes
-pip install docutils
+pip install docutils phate
```
falexwolf: Phate is an optional dependency and should stay an optional dependency. It's not (yet) one of the canonical visualization methods in the field and not in Scanpy. It's in Scanpy so that people can try it out.

You don't have to provide a test for it... It's tested in it's own package. There's just a small wrapper in Scanpy.

fidelram:
@falexwolf The UMAP test works because the umap coordinates are saved along the pbmc datataset that I am using. In contrast, the tsne coordinates need to be computed. Thus I suggest to leave the UMAP test.

falexwolf:
Sure, makes sense! Keep it in this case. 🙂

fidelram:
If you want, we can merge now.

falexwolf:
I'll work a little bit with this branch for a couple of days to test it out myself, I might also push little changes to it. I'm super happy to merge after these tests. :smile:

fidelram:
No problem. Maybe something we should consider comes from my attempt to use
plotly with the scatter functions output (probably for bokeh is similar).
Plotly has a function to convert a matplotlib fig object to plotly.
However, for this to work the figure object (the one returned by
pyplot.figure()) has to be returned. Currently, only the axes object are
returned. Thus, we should consider returning the fig object instead of the
axis or add this separately not to break any other code.

falexwolf:
All of this is really nice! :smile:

Regarding your comment on plotly: returning a fig instead of an ax would follow a different convention as in seaborn; which I'd try to mimic as closely as possible. But for the scatter plots, which are the only ones that would profit a lot from interactive exploration, one could think about breaking this convention.

Regarding bugs for now: One bug I could quickly fix myself, two further bugs that I've stumbled across in the past few hours and couldn't fix right away: If I'm calling
```
sc.pl.umap(adata, color=['celltypes', 'prediction'], palette=[sc.pl.palettes.vega_20, sc.pl.palettes.default_64])
```
the color maps are ignored. And if I'm calling
```
sc.pl.umap(adata, color=['louvain', 'Gata1'], legend_loc='on data'
)
```
the first plot has a different shape than the second.

fidelram:
I will get that fixed soon. I don't know why the palette assignment is not
working as I am using the previous code for that.

fidelram:
@falexwolf I took the opportunity to add a change that I wanted with respect to the palette which is the ability to set a palette based on a matplotlib colormap. For example using `palette='tab20'`:

image

or using `palette='Set3'`

image

I didn't want to modify the previous code that handles setting the palette to avoid breaking other code, but if we have some tests for other functions that use that functionality I could try to update the original methods.

falexwolf:
Really awesome changes! Now things are nicely tiled along the same grid independent of whether categorical or continuous annotation is plotted. :smile:

image

I added a docstring for `panels_per_row`; maybe one should simply call it `ncols` as in `matplotlib.GridSpec`? Essentially, in scanpy, sklearn and many other packages all things that are integer numebers are called `nsomething` or `n_something`.

I'd merge immediately, things seem to work perfectly now. Just one tiny cosmetic thing; for these scatter plots, don't you think it would be nice to have them be a perfect square? As there is no meaningful scales on x and y axis? It gave me a bit of a headache when I first wrote it (couldn't make it work with GridSpec, hence all the mess that you encountered)... You're new clean code is definitely more important than this cosmetic thing, but if you have a quick solution, that's the last thing I can think of... 
image

falexwolf:
PS: The last image shows the new scatter compared to the old scatter.

falexwolf:
Argghh... I guess it's really also totally fine if things are no perfect squares anymore, really just change it if you feel it's easy...

fidelram:
It is not much trouble to get something closer to a more square aspect, but the issue is whether the rcParams['figure.figsize'] is respected or not. To make room for the colormap on the right, the plot area is shrinked a bit but the figsize continues to be a square. 

In my view, the solutions are:
  * respect the rcParams['figure.figsize'] but make the colormap thinner, thus aiming towards a more squared image.
 * enlarge the width of the figure or reduce the height to make it more square.

I will experiment a bit to see what is better.

As for the panels_per_row, I think is better to use some standard naming. I will change that.

falexwolf:
I'd definitely say that `rcParams['figure.figsize']` should be respected. Nonetheless, the width of a panel could be `1.2 * rcParams['figure.figsize']` or something like this so that the actual drawing region becomes a square if someone passes a square `figsize` (e.g. `(4, 4)`). This is how I wrote it initially. But as I wanted this to hold _exactly_,  I couldn't use `matplotlib.GridSpec` and then I produced all this mess...

So, having something that resembles a square a bit more, would be pretty, I think; but one should not get into writing messy custom code again... :wink:

fidelram:
I noticed that matplotlib v. 3 is being installed in travis. This may be the reason why some tests not related to the changes are now failing. I am updating my matplotlib version to update de tests.

fidelram:
I don't know what is happening with the new matplotlib but I had to increase the tolerance for the image comparison in order for the tests to pass. Locally I noticed small differences in the margins and axis labels. Also I noticed that running a single test is different than running several tests at once. This probably has to do with some internal matplotlib parameters that are modified. 

At least the tests are passing now.

scanpy/plotting/tools/scatterplots.py
```diff
@@ -179,10 +179,8 @@ def plot_scatter(adata,
 
     sanitize_anndata(adata)
     if color_map is not None:
-        print("instead of color_map use cmap")
```
falexwolf: We would usually do this using `logg.warn` but the convention of having `size` as a parameter is also present in several of matplotlib's and seaborn's functions. I regard a completely obscure parameter name such as `s` not as good coding style. These days, we want to be verbose. So, let's not discourage people from using clear parameter names.

ivirshup:
@fidelram, really excited to try this out! I've got a couple questions about this PR:

* On that grid of plots @falexwolf posted, it looks like there a shared set of colorbars for two genes. Is this a coincidence (i.e. both genes happen to be expressed on a  similar scale), is the colorscale being generated on the range of both genes, or is something else going on?
* When making a set of plots on the same coordinates (different genes on the same UMAP coordinates), have you found any way to reduce computational load? I'd like to think there are repeated computations (like layout) some memoization could speed up, but haven't figured out how.

fidelram:
@ivirshup:
 * If I am looking at the right example, the two genes  have a different scale. There are 6 plots, but only two genes being plotted repeatedly. This is the same behaviour as before. However, now you can pass `vmin` and `vmax` to have the same scale in all plots that are quantitative.
  * The point coordinates (e.g. tSNE or UMAP) are already precomputed, what changes in each plot is the order in which to plot the points because by default `sort_order=True`. The matplotlib scatter function takes care of the layout, but I don't see how this can be further optimized. Do you have any idea? Plotting large numbers of cells does not seem particularly slow for me.

scanpy/plotting/tools/scatterplots.py
fidelram: I am glad that you revised this as I wanted to ask you about those options but I forgot. I agree that `s` is quite misleading. Usually, matplotlib functions have a short and a long name, but in this case only the short is present. Let's remove the warning and use `size` as in seaborn,  this needs to be added to the documentation as well. For `color_map` I don't know. Whatever is more consistent with other scanpy uses should be best.

falexwolf:
Everything is in a very good shape, now. I'm merging this into master. Any further improvements can take off in a new branch from there.

Thank you so much, @fidelram!

ivirshup:
@fidelram 

* Cool. I think I had gotten confused about some of the labelling on there which I'm going to blame jet lag for 😊
* Using the `inline` backend with ~300k cells, it takes about 5 seconds per plot for me. Since I'm trying to improve the experience of sitting with a biologist figuring out labels for clusters, this is a little slow once you get to 5 or more plots – especially when you just want to update one. From `%prun`, it looks like about half of `matplotlib`'s plotting time is spent figuring out where to put points, and the extents of the plot, so I figured that could be a good target for optimization. I've looked into copying the plot after layout, but before coloring, but I'm not sure how feasible that is.

fidelram:
@ivirshup For what you want we need to look into bokeh or plotly as they are optimized to render thousands of points quickly. I think that matplotlib is not going to be a solution here. Last week I played a bit with Dash and I found that is quite easy to set up an app to quickly explore gene expression. Maybe this is something that we can further develop. Currently, it uses matplotlib but I also tried it with plotly with decent results. Here is a very crude but functional demo using human lung airway data from *Plasschaert et a. Nature. 2018. “A Single-Cell Atlas of the Airway Epithelium Reveals the CFTR-Rich Pulmonary Ionocyte.”*  https://doi.org/10.1038/s41586-018-0394-6.: https://demo-scexplorer.herokuapp.com/

PR #245, opened 2018-08-28
Allow choice of metric for finding neighbors when `knn=False`

### LINKED ISSUE #241: Setting `knn=False` in `sc.pp.neighbors` forces use of euclidean distances.
If you use `knn=False` to calculate neighbors, any choice of metric is ignored. [Here's the offending line of code.](https://github.com/theislab/scanpy/blob/6c1daba7448be72de84dec16a038fcaeda1636ad/scanpy/neighbors/__init__.py#L706)

A quick example:

```python
import scanpy.api as sc
import numpy as np

adata = sc.datasets.krumsiek11()
adata.obs_names_make_unique(); sc.pp.pca(adata) # To get rid of warnings
adata_eucl  = sc.pp.neighbors(adata, knn=False, method="gauss", copy=True)
adata_spear = sc.pp.neighbors(adata, knn=False, method="gauss", copy=True, metric="correlation")

assert np.all(adata_eucl.uns["neighbors"]["connectivities"] == adata_spear.uns["neighbors"]["connectivities"])
```

Additionally, I suspect this should throw an error:

```python
sc.pp.neighbors(adata, method="gauss", knn=False, metric="not a real metric")
```

ivirshup:
Fixes #241.

I've allowed the choice of metric for finding nearest neighbors when `knn=False`. I've added a small test to make sure it works, but would open to adding more. The fix I've made killed a few code paths, so I've removed the dead code.

I also reorganized the test cases a bit (split one monolithic test into separate tests), as having more granular feedback helped with a little debugging.

### DIFF
--- a/scanpy/neighbors/__init__.py
+++ b/scanpy/neighbors/__init__.py
@@ -101,82 +101,6 @@ def neighbors(
     return adata if copy else None
 
 
-def compute_euclidean_distances_using_matrix_mult(X, Y):
-    """Compute euclidean distance matrix for data arrays X and Y.
-
-    Parameters
-    ----------
-    X : np.ndarray
-        Data array (rows store observations, columns store variables).
-    Y : np.ndarray
-        Data array (rows store observations, columns store variables).
-
-    Returns
-    -------
-    distances : np.ndarray
-        Distance matrix.
-    """
-    XX = np.einsum('ij,ij->i', X, X)[:, None]
-    if X is Y:
-        YY = XX
-    else:
-        YY = np.einsum('ij,ij->i', Y, Y)[:, None]
-    distances = np.dot(X, Y.T)
-    distances *= -2
-    distances += XX
-    distances += YY.T
-    np.maximum(distances, 0, out=distances)
-    if X is Y:
-        distances.flat[::distances.shape[0] + 1] = 0.
-    # print(distances)
-    # print(distances[distances < 0])
-    # distances[distances < 0] = 0  # set to 0
-    distances = np.sqrt(distances)
-    return distances
-
-
-def compute_neighbors_numpy_chunk(X, Y, n_neighbors):
-    """Compute distance matrix in euclidean norm for chunk.
-    """
-    D = compute_euclidean_distances_using_matrix_mult(X, Y)
-    chunk_range = np.arange(D.shape[0])[:, None]
-    indices_chunk = np.argpartition(D, n_neighbors-1, axis=1)[:, :n_neighbors]
-    indices_chunk = indices_chunk[chunk_range,
-                                  np.argsort(D[chunk_range, indices_chunk])]
-    distances_chunk = D[chunk_range, indices_chunk]
-    # we know that a point has zero-distance to itself:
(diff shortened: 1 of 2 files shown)


falexwolf:
Ah, thank you! This might break some small things, as you replaced an exact search with an approximate neighbor search. It might not affect the tests as this is hard to test.

Nonetheless, you're completely right, there shouldn't be two neighbor functions... I'll briefly check the harder cases and see whether I can reproduce some old notebooks. Please remind me if I don't get back to this very soon.

ivirshup:
I'd be interested in hearing what this might break. While I was checking to see if the euclidean distance implementation here got different results than the sklearn one, my (fairly cursory) tests at the repl passed the `np.allclose` test.

falexwolf:
Apologies for the late response @ivirshup! I married and moved to the US with twin babies last week. And in between, I spilled something over my laptop...

I reproduced all the old notebooks. :smile:

PS: As mentioned, things might "break" as you replaced something exact with something approximate. The difference becomes pronounced on "hard datasets", which are not present in the test of the neighborhood search but only for the "standard clustering tutorial", which doesn't use `knn=False`.

ivirshup:
No problem, those are all much more important. Congratulations!

scanpy/neighbors/__init__.py
```diff
@@ -705,9 +629,14 @@ def compute_neighbors(
         # neighbor search
         use_dense_distances = (metric == 'euclidean' and X.shape[0] < 8192) or knn == False
         if use_dense_distances:
-            # standard eulcidean case for relatively small matrices
-            self._distances, knn_indices, knn_distances = compute_neighbors_numpy(
-                X, n_neighbors, knn=knn)
+            _distances = pairwise_distances(X, metric=metric, **metric_kwds)
```
falexwolf, line 632: @VolkerBergen it's here where the changed implementation sometimes raises an sklearn warning. (FYI @ivirshup)

I'm not sure why this happens. One could suppress the warning in this single line, but of course that's a hack.

ivirshup: Would you mind sharing the warning you get?

VolkerBergen: Apologies for the late response. The warning is the following:

```
./site-packages/sklearn/metrics/pairwise.py:257: RuntimeWarning: invalid value encountered in sqrt
  return distances if squared else np.sqrt(distances, out=distances)
```

It is also an issue for UMAP (https://github.com/lmcinnes/umap/issues/87) which prob traces back to some neighbors computing as well.

ivirshup: Thanks!

There's an [instability issue](https://github.com/scikit-learn/scikit-learn/issues/9354) with pairwise euclidean distances on float32s in sklearn right now, possibly related?

falexwolf: That might be it... it should be a temporary thing... It's a bit ugly to have this warning every now and then but it's fine if it's temporary I'd say; if it doesn't go away we'll make a PR to sckit-learn at some point.

Issue #240, opened 2018-08-23
Louvain clustering on weighted network?


ivirshup:
I'm looking at network based clustering for single cell data, and was wondering about the louvain implementation used here. Is there a reason scanpy doesn't currently allow clustering on a weighted network (like that generated by `sc.pp.neighbors(adata, knn=False, method="gauss")`)? It's my impression that this would be more similar to the cited Phenograph method.

I'd be interested in adding that functionality (and have a proof of concept on [my fork](https://github.com/ivirshup/scanpy/tree/louvain_weights)), but wanted to hear if there was a reason this wasn't implemented currently first.

Thanks for any input!

fidelram:
Do you know what are the advantages of using such network in the louvain clusters?

fidelram:
As a quick test, I tried the weighted version of the louvain method and it was able to identify small clusters that are no identified with the non weighted louvain. However, I did not use `knn=False` as this does not work well with the UMAP representation. Still, I could see differences (eg. cluster 6 and 9 in the top figure):

![image](https://user-images.githubusercontent.com/4964309/44581659-e85ed300-a79e-11e8-8236-cc149e9c17d4.png)

LuckyMD:
@fidelram Can you change the resolution of the non-weighted version to reproduce a clustering similar to the weighted case? Weighting can change the scale of the resolution parameter.

I would assume clustering on a weighted, fully connected network would be a lot more computationally expensive. I'm curious if it's worth the additional cost. It may also perform worse as you are putting more emphasis on the euclidean distances between transcriptomes than you may want (are the values more important or only the order of the distances?).

ivirshup:
Intuitively, I'd think having a more complete graph with weighted edges is more representative of the data than an arbitrary *k* neighbors. Even if you do use a hard cutoff on number of neighbors, I don't see how discounting all distance information would give a more accurate result. I would suspect using a weighted graph could perform better at identifying small subpopulations (where nearest neighbors from other cell types could be common), but that's just conjecture.

From some preliminary attempts of my own, it seems clustering solutions become more stable with respect to parameter choice when I use the `method=gauss, knn=False` weighted network. I'm still in the process of verifying this, however.

I would also note that the documentation for `sc.tl.louvain` references [this](
https://doi.org/10.1016/j.cell.2015.05.047) paper (the Phenograph method), which uses the louvain method on a a weighted graph. If the method is cited, why not allow using it?

[...]

PR #248, opened 2018-08-29
[WIP] Allow weighted network for louvain clustering

ivirshup:
Previously discussed in #240

A few things left to discuss:

## Tests

These are pretty simple, "this doesn't intrinsically throw an error" type tests. Should the tests cover more than that? Should they be more thorough is checking arguments won't throw errors? I'm open to suggestions on other things that could be checked.

Also, is there a place they'd be more appropriate?

## Allowing storage of multiple network representations 

I think this would also be a pretty simple addition, but wanted to check again before implementing it. I'm thinking of adding a `use_network` argument which would allow key access to network stored in the AnnData object – similar to the `use_rep` argument. @LuckyMD mentioned there might be some storage concerns here, though I think the user is ultimately responsible for size in this case.

The value added here is different representations are useful for different analysis, and it'd be useful to not have to have two objects when the rest of the data would be shared.

## Allow more choice of partition method for `louvain-igraph` package

I'm not too fussed on this one. It's just that `"RBConfiguration"` is hard coded when other methods are available, and I'm not aware of a reason it would be the best choice.

### DIFF
--- a/scanpy/tests/clustering.py
+++ b/scanpy/tests/clustering.py
@@ -1,0 +1,19 @@
+import pytest
+import scanpy.api as sc
+
+@pytest.fixture
+def adata_neighbors():
+    adata = sc.read('./data/pbmc3k_raw.h5ad',
+                    backup_url='http://falexwolf.de/data/pbmc3k_raw.h5ad')
+    sc.pp.filter_genes(adata, min_cells=1)
+    sc.pp.normalize_per_cell(adata)
+    sc.pp.log1p(adata)
+    sc.pp.pca(adata)
+    sc.pp.neighbors(adata)
+    return adata
+
+def test_nofail(adata_neighbors): 
+    sc.tl.louvain(adata_neighbors)
+    sc.tl.louvain(adata_neighbors, use_weights=True)
+    sc.tl.louvain(adata_neighbors, use_weights=True, flavor="igraph")
+    sc.tl.louvain(adata_neighbors, flavor="igraph")
--- a/scanpy/tests/clustering.py
+++ b/scanpy/tests/clustering.py
@@ -21,5 +21,4 @@ def test_nofail(adata_neighbors):
 def test_partition_type(adata_neighbors):
     import louvain
     sc.tl.louvain(adata_neighbors, partition_type=louvain.RBERVertexPartition)
-    # Fails due to passing a resolution parameter
-    # sc.tl.louvain(adata_neighbors, partition_type=louvain.SurpriseVertexPartition)
+    sc.tl.louvain(adata_neighbors, partition_type=louvain.SurpriseVertexPartition)
--- a/scanpy/tests/clustering.py
+++ b/scanpy/tests/clustering.py
@@ -17,3 +17,9 @@ def test_nofail(adata_neighbors):
     sc.tl.louvain(adata_neighbors, use_weights=True)
     sc.tl.louvain(adata_neighbors, use_weights=True, flavor="igraph")
     sc.tl.louvain(adata_neighbors, flavor="igraph")
+
+def test_partition_type(adata_neighbors):
+    import louvain
+    sc.tl.louvain(adata_neighbors, partition_type=louvain.RBERVertexPartition)
+    # Fails due to passing a resolution parameter
+    # sc.tl.louvain(adata_neighbors, partition_type=louvain.SurpriseVertexPartition)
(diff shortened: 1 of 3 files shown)


falexwolf:
**Tests**

For the present changes, these tests are sufficient. Only since recently, Scanpy is becoming properly tested via the example notebooks (previously, I always ran everything manually). So, if you want to test this properly, come up with an interesting use case that uses clustering on weighted graphs, make a pull request for a notebook on `theislab/scanpy_usage`, link to it in the docs (https://scanpy.readthedocs.io/en/latest/examples.html) and commit a corresponding test in `tests/notebooks`. That's a lot of work and I think too much for this present case.

Eventually, most of Scanpy's functionality should be tested this way. A lot will also be covered by extending existing notebooks. But I don't think it's possible to meaningfully add the weighted graphs to the present standard Louvain clustering example.

**Multiple network representations**

I don't mind as long as the standard user still uses a single one - otherwise people will get confused. But let's not call these grahs "networks". The notion "network" suggests that nodes "interact". The current graphs in Scanpy, however, are all proxies for manifolds. Edges only represent similarity and, through that, have a topological interpretation. I'd really like to reserve the notion "network" for cases where nodes possibly or actually interact. Hence, a gene regulatory network or a network of cells with cell-cell interactions. Both are completely different things compared to the "neighbors" stuff in Scanpy.

**Allow more choice of partition method for louvain-igraph package**

Fine for me as long as the default stays unchanged. I'm not sure whether it adds a lot of value but it will also not be harmful.

---

Could you also fix the docs for the `flavor` parameter and specify the default? Thanks!

ivirshup:
## Tests

Fair enough.

## More graph/ neighbor representations

For sure, I'm thinking of just

I've taken another look at some of the Neighbors code, and am realizing the amount of code this touches. I still think this is important, but might be out of scope for this pull request. I'm probably going to hack together something for personal use first, and see where I go from there. Just to make sure anything I play around with isn't breaking other parts of `scanpy`, would you mind sharing how you run those notebook tests?

## Partition

Would you mind if I change the requirement in `setup.py` from `louvain` to `louvain>=0.6`? It'd make adding this a bit cleaner.

## Doc changes

Sure! All that needs to change is specifying the default, right?

falexwolf:
Notebook tests: there is only one there in the tests, see https://github.com/theislab/scanpy/tree/master/scanpy/tests/notebooks. I run all other linked examples notebooks run manually... So this is not really an option for you, I'd say.

I think I can add two important further notebooks very soon so that almost all of the functionality is covered.

`setup.py`: yes, definitely, `louvain-igraph>=0.6` is fine! no one should use an earlier version.

ivirshup:
I'm having some trouble debugging whatever is going wrong with the notebook tests here. I get the same results if I run `pytest` on my machine, but don't get a failure if I run the code manually. Additionally, I don't get an error (the `abort`) if I *only* run the notebook tests (`pytest -k "test_pbmc3k"`).

Pretty sure the error is happening on the call to louvain in the notebook tests – an `assert False` fails the tests, one after gives current result – but I can't reproduce the abort interactively.

Any idea what's going on/ how I can get a more helpful error message here?

ivirshup:
Caught it.

I had forgotten that arguments only get evaluated once, so if you mutate them, there is state which is maintained to other calls. I think the unhelpful `abort` message is from `louvain-igraph` expecting a weight vector of the right shape, which ended up with the error:

```
libc++abi.dylib: terminating with uncaught exception of type char const*
```

falexwolf:
Great that you fixed it! :)

I'm in principle happy to merge the pull request! However, it contains a lot of the previous commits to master by other people; maybe you haven't properly rebased your branch to master at some point? Looking at the diff across the whole request, I see mostly old things from the 25 commits before. So, (1) could you point me to the diff across all your commits that you actually did? (2) Are we going to have a messed up history with redundant commits on the master branch if I merge this, I don't think so, but I'm not 100% sure.

Sorry for the additional work, but blindly merging something into master is of course too dangerous, and going through each single commit is too tedious. ;)

Thank you!

ivirshup:
Yeah, I think I messed up a rebase by having `master` point to my fork instead of this repo. The merge looked fine locally, but I wasn't sure what was going on with the diff here. I did a little more rebase-ing and I think the diff should look much more normal now.

One question about numpy style documentation: for the `partition_type` argument, the user should be passing in a class, not an instance. How do I document that? With type annotations I think I would do something like: `Type[louvain.MutableVertexPartition]`, but I'm not sure if that applies here.

falexwolf:
Thank you very much for fixing the diff! It's very easy to track what happened, now. Regarding numpy-style docs; there are few things I still need to fix with the docs anyways. The style is in principle correct, I'll go through this, too.

Thank you for the PR! :smile:

PR #249, opened 2018-08-29
Added function to retrieve gene coordinates

fbrundu:
Support for different genome builds with respect to the latest in Ensembl is provided changing the BioMart host url.


falexwolf:
Thank you very much! :smile:

PR #250, opened 2018-08-30
Added documentation for chunked PCA

ivirshup:
Was playing around with an incremental PCA earlier and saw the arguments were undocumented. Figured I could document it while I was waiting for it to run.

I think I got all the style right, let me know if not.


falexwolf:
Thank you, great!

PR #260, opened 2018-09-07
queries: bugfix and zebrafish support

### LINKED ISSUE #258: `NameError: name 'logg' is not defined` when using `sc.queries.mitochondrial_genes`
I just tried
```python
import scanpy.api as sc
sc.queries.mitochondrial_genes('www.ensembl.org', 'strange_organism')
```
I would expect scanpy complains that it does not know `'strange_organism'`, but I get the error 
```python
NameError                                 Traceback (most recent call last)
<ipython-input-13-6a41b361ab41> in <module>()
      1 import scanpy.api as sc
----> 2 sc.queries.mitochondrial_genes('www.ensembl.org', 'drerio')

~/software/scanpy/scanpy/queries/__init__.py in mitochondrial_genes(host, org)
     34         s.add_attribute_to_xml('mgi_symbol')
     35     else:
---> 36         logg.msg('organism ', str(org), ' is unavailable', v=4, no_indent=True)
     37         return None
     38     s.add_attribute_to_xml('chromosome_name')

NameError: name 'logg' is not defined
```
It seems to me like `queries/__init__.py` misses an `from .. import logging as logg` statement.

Would maybe also make sense to show the the message that an organism is not available at verbosity level 1 instead of 4?

fbnrst:
Here, I fix #258 and add support for zebrafish to the queries.

### DIFF
--- a/scanpy/queries/__init__.py
+++ b/scanpy/queries/__init__.py
@@ -9,9 +9,9 @@ def mitochondrial_genes(host, org):
     ----------
     host : {{'www.ensembl.org', ...}}
         A valid BioMart host URL.
-    org : {{'hsapiens', 'mmusculus'}}
-        Organism to query. Currently available are human ('hsapiens') and mouse
-        ('mmusculus').
+    org : {{'hsapiens', 'mmusculus', 'drerio'}}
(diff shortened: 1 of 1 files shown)


falexwolf:
Awesome! :smile:

PR #266, opened 2018-09-21
Prevent error when no third point is found

flying-sheep:
Else `argmax([])` is called later, which doesn’t make sense. This happens because `Dseg[tips[0]] + Dseg[tips[1]]` are all `inf` for some datasets. Is that simply a consequence of my data or is that a bug?

This PR assumes the former and results in such segments being skipped. In case of my data, this means that I end up with a single DPT group instead of multiple ones.


falexwolf:
Ok, makes sense. Things are infinite when you encounter disconnected clusters. Either increase `n_neighbors`, if it makes sense, or you can only meaningfully run DPT on subsets of the data.

Issue #159, opened 2018-05-21
Where can I find log-fold change in gene expression for the markers


DoruMP:
Does scanpy store log fold change, p-value and marker type flag (negative vs positive) somewhere? It looked at scObj.uns["rank_genes_groups"] but found only 3 fields: params, scores and names.

falexwolf:
No, it currently doesn't. Instead it uses the `scores`... usually some "differential z-score" that goes into the t-test. We will extend differential testing in the future.

a-munoz-rojas:
Just wanted to bring this question back up - it would be great if we could get fold changes and p-values returned from the relevant methods for differential testing in scanpy. Thanks!

falexwolf:
There now is a much more powerful differential testing package `diffxpy`, @davidsebfischer, which easily integrates into Scanpy.

@a-munoz-rojas Would you consider making a pull request that adds log-fold changes for t-test etc. in `rank_genes_groups`? My bandwidth is limited these days, I will certainly do it at some point, but it's faster if you do it.

a-munoz-rojas:
@falexwolf Thanks for pointing out this package, I'll give it a try. I'll also give the pull request a shot - I'm still learning my way around this package but if I can do it I'll submit it for sure.

wflynny:
@falexwolf Is there anyplace where we can read into `diffxpy`?  I've been benchmarking available marker gene detection algorithms and am interested to see what is included in this new package.

a-munoz-rojas:
@davidsebfischer @falexwolf Also, where can we get this package? I tried doing a pip install, and while the package is listed the installation failed.

flying-sheep:
@falexwolf you were a little quick: @davidsebfischer didn’t turn it public yet. maybe david would like to send it to you for beta testing?

falexwolf:
Yes, he told me just yesterday that it's not public yet, but it will soon be.

davidsebfischer:
Hi @a-munoz-rojas, diffxpy will be public very soon, once we finished running all benchmarks that we need for validation. I would be happy to help you set it up if you still want to give it a go then!

a-munoz-rojas:
Hi @davidsebfischer, sounds great! Let me know and I'd be happy to give it a try. Thanks!

PR #270, opened 2018-09-27
P-values added to t-tests and wilcoxon test

a-munoz-rojas:
I updated the rank_genes_groups function to output p-values for t-tests and the wilcoxon rank-sum test, as discussed with @falexwolf  in #159. 
The changes are outlined below:
- The t-test in the original file used a Welch t-test. I kept this, calculated the relevant degrees of freedom for a Welch test and then extracted the corresponding two-tailed p-value for the t-statistic (score). 
- The Wilcoxon test was originally done in chunks. To get the p-values I had to simplify this approach and use the ranksums function in scipy.stats. This caused me to loop through all of the genes being tested, which was fine for my dataset, but might need to be optimized for larger datasets.
- The adjusted p-values (pvals_adj) were calculated with a standard Bonferroni correction.
- All p-values are outputted and sorted the same way as 'names' or 'scores'. The only difference is that the p-values recarrays use float64 as a datatype to avoid converting a lot of the very small p-values to 0.

Hope this is helpful!

Andrés

### DIFF
--- a/scanpy/tools/rank_genes_groups.py
+++ b/scanpy/tools/rank_genes_groups.py
@@ -167,6 +167,7 @@ def rank_genes_groups(
             mean_rest[mean_rest == 0] = 1e-9  # set 0s to small value
             foldchanges = (means[igroup] + 1e-9) / mean_rest
             scores[np.isnan(scores)] = 0
+            #Get p-values
             denominator_dof = (np.square(vars[igroup]) / (np.square(ns_group)*(ns_group-1))) + (
                 (np.square(var_rest) / (np.square(ns_rest) * (ns_rest - 1))))
             denominator_dof[np.flatnonzero(denominator_dof == 0)] = np.nan
@@ -244,8 +245,6 @@ def rank_genes_groups(
 
             # Loop over all genes
             for gene_idx in range(n_genes):
-                # if gene_idx % 20 == 0:
-                #     print('Group {}, Gene regressions:{:.2%}'.format(current_group, gene_idx / n_genes))
                 scores[gene_idx], pvals[gene_idx] = ranksums(X1[:,gene_idx],X2[:,gene_idx])
 
             mean_rest[mean_rest == 0] = 1e-9  # set 0s to small value
@@ -279,18 +278,10 @@ def rank_genes_groups(
             dtype=[(rn, 'float32') for rn in groups_order_save])
         adata.uns[key_added]['pvals'] = np.rec.fromarrays(
             [n for n in rankings_gene_pvals],
-            dtype=[(rn, 'float32') for rn in groups_order_save])
+            dtype=[(rn, 'float64') for rn in groups_order_save])
         adata.uns[key_added]['pvals_adj'] = np.rec.fromarrays(
             [n for n in rankings_gene_pvals_adj],
-            dtype=[(rn, 'float32') for rn in groups_order_save])
-
-    # if method == 'wilcoxon':
-    #     adata.uns[key_added]['pvals'] = np.rec.fromarrays(
-    #         [n for n in rankings_gene_pvals],
-    #         dtype=[(rn, 'float32') for rn in groups_order_save])
-    #     adata.uns[key_added]['pvals_adj'] = np.rec.fromarrays(
-    #         [n for n in rankings_gene_pvals_adj],
-    #         dtype=[(rn, 'float32') for rn in groups_order_save])
+            dtype=[(rn, 'float64') for rn in groups_order_save])
     
     logg.info('    finished', time=True, end=' ' if settings.verbosity > 2 else '\n')
     logg.hint(
@@ -298,8 +289,8 @@ def rank_genes_groups(
         '    \'names\', sorted np.recarray to be indexed by group ids\n'
         '    \'scores\', sorted np.recarray to be indexed by group ids\n'
         .format(key_added)
-        + ('    \'logfoldchanges\', sorted np.recarray to be indexed by group ids'
-           if method in {'t-test', 't-test_overestim_var'} else '')
-        + ('    \'pvals\', sorted np.recarray to be indexed by group ids'
-           if method in {'wilcoxon'} else ''))
+        + ('    \'logfoldchanges\', sorted np.recarray to be indexed by group ids\n'
+           '    \'pvals\', sorted np.recarray to be indexed by group ids\n'
+           '    \'pvals_adj\', sorted np.recarray to be indexed by group ids'
+           if method in {'t-test', 't-test_overestim_var', 'wilcoxon'} else ''))
     return adata if copy else None
--- a/scanpy/tools/rank_genes_groups.py
+++ b/scanpy/tools/rank_genes_groups.py
@@ -240,8 +240,14 @@ def rank_genes_groups(
             scores = np.zeros(n_genes)
             pvals = np.zeros(n_genes)
 
-            X1=X[mask].todense()
-            X2=X[mask_rest].todense()
+            X1 = X[mask]
+            X2 = X[mask_rest]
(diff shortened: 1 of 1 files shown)


LuckyMD:
Is it generally a good idea to output P-values for marker gene identification? As the null model is not valid for this setup, P-values are not really a measure of significance, but should only serve to order genes. I can see the point of outputting log fold changes, but outputting P-values may be misleading.

gokceneraslan:
Why is the null hypothesis invalid? Testing whether two cell types have the same mean for the expression of a given gene seems valid, no? One-vs-rest might be problematic, though.

LuckyMD:
The null hypothesis assumes no difference in gene expression. However the groups are determined by differences in gene expression due to clustering. In random data, clustering would also separate clusters by 'random' gene expression signatures and thus find marker genes with p-value < 0.05.

gokceneraslan:
Hmm, I think it's more complicated than that. First, cell type labels may not come from clustering. Even if it's the case, the design of Louvain does not rely on univariate mean difference between groups due to the distance metrics used in kNN graph, which take all genes into account at once. 

Besides, I believe that how cell types are defined e.g. by biologist's manually annotation of each cell, Louvain, bulk comparison etc. doesn't really make the null hypothesis invalid. t-test just tests if the means are the same or not.

Regarding false positives with random clustering, that's why we have Bonferroni, no? But this is now a different story about p-values and multiple testing correction in general. So it's not our duty to solve it in rank_genes_group function. I'd rather prefer reporting a p-value (or maybe t-stat) and letting the user decide how valid/useful it is for his/her research instead of not outputting it with the consideration of the chance that it's not valid.

LuckyMD:
So you should easily be able to see that the null hypothesis is not valid by the distribution of p-values for all genes in one rank_genes_groups test. If it is valid, this distribution should be uniform. In that case it's only a multiple testing problem... I would guarantee you that it's not uniform though.

Cell type labels from expression-independent sources should not have the same confounding effect. However, I would bet you can trace back all biological annotations of cell types back to expression.

a-munoz-rojas:
Thanks for your comments! This is an interesting discussion. I agree that the validity of a p-value depends on the user's data. From the standpoint of an analysis tool, I think it's better to report a p-value (and a t-statistic, which was already reported in the original function) and let the user decide what to do with that information.

LuckyMD:
I just noticed I forgot to answer to your point about multiple-testing-correction. Multiple testing correction deals with the effect that obtaining any false positives become more likely with repeated tests. However, it does not deal with bias. Bias is however present in the test given that the correct null hypothesis should assume some genes are different between groups, and not that all genes are equal. This comes back to the observation that some genes will be always be different between groups in random data, given that clustering defines the groups over which we test.

And as for louvain using whole transcriptome similarities rather than individual gene similarities... these are related. If you defined groups by differences of whole transcriptomes, this necesitates individual genes having different means. The correct p-value would be the output of a permutation test. 

Could discuss this offline if you like @gokceneraslan. I'm preparing a manuscript which makes a point of this... if I am wrong, please convince me of this soon ;).

LuckyMD:
I can understand your line of reasoning @a-munoz-rojas. But is it not also dangerous to allow a user an interpretation which may be incorrect? I think outputting logFC values is great... I just have issues with calling the other output P-values. It is at the very least contentious whether they are measures of significance.

falexwolf:
Sorry about the super late response and thank you for the PR, @a-munoz-rojas. I have not read most of the above discussion, yet. Right now, I just wanted to note that (@a-munoz-rojas), the tests are failing as the wilcoxon test seems to yield different results, now. I don't know whether it's just marginal, but it should be clarified; users should still get the same results as before.

a-munoz-rojas:
Thanks for your reply, @falexwolf. I was looking through the same error and I can't really understand why the results are different - it might be a difference in accuracy levels between the manual wilcoxon method that was used before, and the built-in scipy.stats function I used. I looked at the differences and they indeed look marginal. **Edit:** I actually just went back through the check results, and the comparison between the results before and after are pretty much identical - it could just be a difference in bit-depth. For example, it's tagging (2.292195  ,  5.7448500e-01) as different from  (2.292195  ,  0.574485). 


I also compared a before-after with my dataset and I get very similar marker genes, albeit in slightly different order (see attached images). I don't really know what would be the best way to address these differences - I am simply using the built in spicy.stats function and not changing the output it gives me. Could this marginal difference be caused by the estimation in the "chunk" approach used in the previous version? Even with this marginal difference, I would assume that using the scipy function is more "accurate". Please let me know what you would prefer and what would be the best way to proceed.
figure_1_newwilcox
figure_1_originalwilcox

a-munoz-rojas:
Also, to reply to @LuckyMD, I definitely see your point. I was mostly trying to replicate other single-cell tools that do provide this metric, but I understand why it could be characterized as misleading - maybe we can add a disclaimer in the function description? I'm open to other suggestions!

falexwolf:
@a-munoz-rojas Thanks for checking and thank you very much for the PR :smile! I like your version of the implementation a lot better than the manual one, as the code is much more readable. I don't know why @tcallies added the wilcoxon this way at the time, but I assume for speed and memory reasons.

So, I'm very happy to merge this PR; I'll just briefly give this another check today or tomorrow and update the tests so that they don't fail anymore.

Regarding the general discussion: Yes, let's just add a disclaimer that several assumptions on how meaningful the null hypotheses are both for wilxocon and t-test for single-cell data, should do the job. Then people will interpret the p-values with care.

falexwolf:
So, the main problem I see is that the runtime of the new implementation is a factor 20 slower: 37 s vs. 2 s. 

image

vs.

image

Let me think about what to do, here.

falexwolf:
I'm merging this but will restore the previous Wilcoxon implementation, for speed reasons. The essential problem is that scipy.stats does not have a multi-dimensional implementation; it should be easy to adapt the previous implementation so that it provides pvalues, too; simply via multi-dimensional adaption of https://github.com/scipy/scipy/blob/v1.1.0/scipy/stats/stats.py#L4931-L4974.

a-munoz-rojas:
Sounds great, @falexwolf! I did notice the slow-down and agree it's not great. That's a great suggestion, I'll take a look. Thanks, glad it helped!

falexwolf:
If you @a-munoz-rojas would give it another try, using a multi-dimensional chunked implementation along the current implementation and the comment above, then that would be very cool! If you don't have bandwidth for that, could you pass this on to @Koncopd, who might have bandwidth? It would be nice to clean up the whole module (in particular, split up the long code chunks into three functions `_t_test()`, `_wilcoxon()`, `_logreg()`).

I also made the tests work again; I guess they failed as the current implementation and scipy are handling ties a little different; the results for scores were exactly the same up to the position of a single gene. 

Please make a new PR for any of this.

falexwolf:
Ah, now we just wrote at the same time. ;) Let me know what you think.

falexwolf:
Ok, now the tests are actually passing again, everything is in the three commits prior and including this one: https://github.com/theislab/scanpy/commit/d889faf9a58d8981c0783584b3f333680fc161ce

falexwolf:
And: just as a remark; the chunking along genes is necessary so that everything still fits into memory for really large datasets, which are almost always sparse when fed into `rank_genes_groups`.

a-munoz-rojas:
I was just about to ask about the chunking along genes - you read my mind @falexwolf. I think it might be possible to do a multi-dimensional adaptation of the scipy.stats code you linked to, and still do the math with sparse matrices, similar to how we implemented the t-tests. This way we could possibly avoid the chunking (it might help with readability of the code). Would this be worth pursuing?

I'll give this a quick try, but I am a little limited in bandwidth. I'll let you know soon if it would be best to get some help from @Koncopd (if they have time!)

PR #282, opened 2018-10-05
Fix the parameter explanation for filter_genes

bebatut:
Hi,

I corrected these small mistakes while checking the documentation to write Galaxy wrapper.

Bérénice


falexwolf:
Thank you very much! :smile:

PR #283, opened 2018-10-05
Support running on distributed compute engines like Dask, and Spark via Zap

tomwhite:
These changes enable Scanpy's pre-processing functions to run on distributed engines including Dask and Spark. The Spark integration itself relies on [Zap](https://github.com/lasersonlab/zap) for a distributed version of NumPy. 

The main change is the `materialize_as_ndarray` function that is used at certain points of the computation to materialize intermediate results (not the full matrix). This is a no-op in the non-distributed case.


falexwolf:
This is awesome! Already very elegant and a very good start! :smile:

tomwhite:
Thanks @falexwolf!

PR #284, opened 2018-10-07
add key_added to store bool array of hvg

VolkerBergen:
add option to keep genes and store bool array of highly variable genes

### DIFF
--- a/scanpy/preprocessing/simple.py
+++ b/scanpy/preprocessing/simple.py
@@ -645,7 +645,7 @@ def pca(data, n_comps=None, zero_center=True, svd_solver='auto', random_state=0,
 
     if data_is_AnnData:
         adata.obsm['X_pca'] = X_pca
-        if use_highly_variable:
+        if not use_highly_variable:
             adata.varm['PCs'] = pca_.components_.T
         else:
             PCs = np.empty(shape=(n_comps, adata.shape[1]))
--- a/scanpy/preprocessing/simple.py
+++ b/scanpy/preprocessing/simple.py
@@ -244,6 +244,7 @@ def filter_genes_dispersion(data,
                             n_bins=20,
                             n_top_genes=None,
                             log=True,
+                            key_added=None,
                             copy=False):
     """Extract highly variable genes [Satija15]_ [Zheng17]_.
 
@@ -285,6 +286,9 @@ def filter_genes_dispersion(data,
         Number of highly-variable genes to keep.
     log : `bool`, optional (default: `True`)
         Use the logarithm of the mean to variance ratio.
+    key_added: `str`, optional (default: `None`)
+        Write a bool array for highly-variable genes while keeping all genes 
+        (if not None) else keep highly-variable genes only.
     copy : `bool`, optional (default: `False`)
         If an :class:`~anndata.AnnData` is passed, determines whether a copy
         is returned.
@@ -320,7 +324,10 @@ def filter_genes_dispersion(data,
         adata.var['means'] = result['means']
         adata.var['dispersions'] = result['dispersions']
         adata.var['dispersions_norm'] = result['dispersions_norm']
-        adata._inplace_subset_var(result['gene_subset'])
+        if isinstance(key_added, str):
+            adata.var[key_added] = result['gene_subset']
+        else:
+            adata._inplace_subset_var(result['gene_subset'])
         return adata if copy else None
     logg.msg('extracting highly variable genes',
               r=True, v=4)
--- a/scanpy/preprocessing/simple.py
+++ b/scanpy/preprocessing/simple.py
@@ -638,7 +638,13 @@ def pca(data, n_comps=None, zero_center=True, svd_solver='auto', random_state=0,
 
     if data_is_AnnData:
         adata.obsm['X_pca'] = X_pca
-        adata.varm['PCs'] = pca_.components_.T
+        if adata.shape[1] == adata_hvg.shape[1]:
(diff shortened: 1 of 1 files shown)


LuckyMD:
This is great, thanks! Now just for the neighbours `use_hvgs=` parameter and then we're sorted for using HVGs for downstream analysis without filtering the whole dataset.

LuckyMD:
This is great, thanks! I forgot our conversation the other week... was it `sc.pp.pca()` and not `sc.pp.neighbors()` that needed the `use_hvgs` parameter?

scanpy/preprocessing/simple.py
```diff
@@ -591,34 +593,36 @@ def pca(data, n_comps=None, zero_center=True, svd_solver='auto', random_state=0,
         logg.msg('reducing number of computed PCs to',
                n_comps, 'as dim of data is only', adata.n_vars, v=4)
 
+    adata_hvg = adata[:, adata.var['highly_variable']] if use_hvg and 'highly_variable' in adata.var.keys() else adata
```
gokceneraslan: If use_hvg is False, this naming doesn't make sense. I would keep it as adata even if use_hvg is True.

falexwolf: Agreed!

Also, please make `use_hvg` default to `None` with the same behavior. Throw an error if `use_hvg==True` and `'highly_variable' not in adata.var`.

`keep_all_genes` is a well-chosen name, `highly_variable`, too. I'd say that instead of `use_hvg` one should rather call it `use_highly_variable`, even though it's rather `restrict_to_highly_variable`, but this is maybe a bit long.

scanpy/preprocessing/simple.py
```diff
@@ -541,6 +541,8 @@ def pca(data, n_comps=None, zero_center=True, svd_solver='auto', random_state=0,
     return_info : `bool`, optional (default: `False`)
         Only relevant when not passing an :class:`~anndata.AnnData`: see
         "Returns".
+    use_hvg: `bool`, optional (default: True)
```
falexwolf: Please add a space in front of `:`.

scanpy/preprocessing/simple.py
```diff
@@ -638,7 +638,13 @@ def pca(data, n_comps=None, zero_center=True, svd_solver='auto', random_state=0,
 
     if data_is_AnnData:
         adata.obsm['X_pca'] = X_pca
-        adata.varm['PCs'] = pca_.components_.T
+        if adata.shape[1] == adata_hvg.shape[1]:
+            adata.varm['PCs'] = pca_.components_.T
+        else:
+            hvg = adata.var['highly_variable']
+            PCs = np.zeros(shape=(n_comps, adata.shape[1]))
```
falexwolf: These should rather be `nans` instead of zeros, I think.

scanpy/preprocessing/simple.py
```diff
@@ -593,36 +593,36 @@ def pca(data, n_comps=None, zero_center=True, svd_solver='auto', random_state=0,
         logg.msg('reducing number of computed PCs to',
                n_comps, 'as dim of data is only', adata.n_vars, v=4)
 
-    adata_hvg = adata[:, adata.var['highly_variable']] if use_hvg and 'highly_variable' in adata.var.keys() else adata
+    _adata = adata[:, adata.var['highly_variable']] if use_hvg and 'highly_variable' in adata.var.keys() else adata
```
falexwolf: OK, this name is fine!

falexwolf: Just remembered that in other instances, we have the convention `adata_comp` if the passed AnnData is not the object on which the actual computation is carried out, so that would be nicer. Thank you!

scanpy/preprocessing/simple.py
falexwolf: It just occurred to me that `keep_all_genes` in a function that is called `filter_genes_...` is not awesome.

Can we do, `subset=True` or `do_subset=True` or `merely_annotate=False` or something like this?

falexwolf:
Awesome, thank you for fixing all the minor comments. It's a great addition! :smile:

fidelram:
@VolkerBergen can you type a simple example on how to use this new functionality. I think that I want to use this.

VolkerBergen:
Highly variable genes (hvg) can now be used without removing the non-hvg from your data.

That's simply `sc.pp.filter_genes_dispersion(adata, subset=False, **params)`, which then does not do the actual filtering but just stores the result in `.var['highly_variable']`. 

`sc.pp.pca(adata, **params)` is then performed on the those hvg per default. As all other operations such as neighbors, embeddings etc. are usually performed on PCA space, they implicitly use hvg as well.

fidelram:
@VolkerBergen Thanks a lot for the clarification.

LuckyMD:
I just tested this out... The idea was that if `sc.pp.pca()` has the parameter `use_highly_variable` that be extension `sc.tl.umap()`, `sc.tl.tsne()`, and `sc.tl.draw_graph()` would also be based only on highly variable genes. That however doesn't seem to be the case. When I subset my anndata object to only highly variable genes I get a different result than when I just run it with `sc.pp.pca(adata, use_highly_variable=True)`. The `sc.pl.pca()` is the same, but `sc.pl.diffmap` seems somehow inverted, and umap, tsne, and draw_graph are all slightly different.

falexwolf:
Should be resolved here: https://github.com/theislab/scanpy/issues/324#issuecomment-433262910

PR #288, opened 2018-10-08
Corrected Haghverdi18 reference - the doi portion of the URL was missing

jorvis:
Minor - noticed a broken link when looking through new features.


falexwolf:
Thank you, Joshua! :smile:

PR #289, opened 2018-10-08
Added p-values to Wilcoxon

a-munoz-rojas:
Added p-values output to Wilcoxon tests, as discussed in #270. Left most of the code the same (including the chunking). Wilcoxon test now outputs pvals, adjusted pvals (Bonferroni correction), and log fold changes. The change in speed should be much smaller than before:
Original time:
screen shot 2018-10-08 at 4 04 02 pm

New time:
screen shot 2018-10-08 at 4 02 47 pm

I also fixed a bug where the Wilcoxon test would output the absolute values of the scores when rankby_abs=True, which shouldn't happen according to the function description. 

As a side note, I was unable to find a quick/efficient way to rank the full sparse matrix for the wilcoxon test, which is why I kept the chunking along genes as suggested. Hope this helps!


LuckyMD:
Just a quick comment... Benjamini-Hochberg correction is usually the standard for multiple-testing correction in differential expression testing. Not sure if you want to take it into account, but I thought I should mention it.

a-munoz-rojas:
That's a good point, @LuckyMD. I chose Bonferroni to have a more stringent correction (albeit with a loss in power), particularly due to the increased power inherent in the large sample sizes of single cell data. I might be wrong, but I think the Benjamini-Hochberg standard was established with bulk RNAseq, where limited sample sizes required an approach with more power. 

However, I'm happy to change it to Benjamini-Hochberg if that's the consensus! It's a simple one-liner - we can even provide both and let the user choose by passing a parameter. Whatever is preferred!

LuckyMD:
I definitely don't define the consensus, but I normally prefer FDR correction. It makes a bit more sense to me to correct for a false discovery rate, rather than a test-based error, if you are only interested in the rejected null hypotheses. 

They also test for FDR control in a [recent comparison of differential testing methods](http://www.nature.com/doifinder/10.1038/nmeth.4612).

a-munoz-rojas:
That's a good point! I just added the Benjamini-Hochberg correction to the Wilcoxon code (as well as to the t-tests) and left that as the default. I left the Bonferroni code in there and marked the place with comments. If and when the pull request is complete, they can choose what to keep and remove the extra comments. Either way, since the uncorrected p-values are also outputted, the user can choose whichever correction they want downstream.

falexwolf:
Looks very good to me, thank you very much!

Would you mind adding an option to select for the correction type that defaults to 'benjamini-hochberg' and can be set to 'bonferroni'?

In the best of all world's, you'd also extend the tests for rank_genes_groups so that the p values are tested and not messed up by pull requests in the future. We want people to get the same p values again and again. And as the whole module sort of involves a lot of custom code as the scipy alternatives are not there for mult-dimensional and sparse data, it's easy to mess this up in the future.

Thank you so much for the awesome addition @a-munoz-rojas , I'll add you both to the Scanpy author list and to the release notes.

a-munoz-rojas:
Thank you very much, @falexwolf! I really appreciate the addition to the author list! I'm glad this is useful.

I just now added the option to choose which correction method to use, and set benjamini-hochberg as the default, so that should be all set. With regards to the test, I unfortunately don't have experience building those tests, and have limited bandwidth at the moment. So it would probably be best if someone else could extend those tests.

falexwolf:
OK, I'll add the tests myself. Thank you!

falexwolf:
Which name should I add to the author list?

a-munoz-rojas:
Great. You can use: Andrés R. Muñoz-Rojas. 
Thanks again!

falexwolf:
Done! :smile:

Issue #286, opened 2018-10-08
Function `plot_scatter` is breaking with matplotlib


dilawar:
**Update** Note sure if it is a problem with our script.  However, if colors are not found, can it use default color-map and let us plot after raising a warning?

Log is following. ~Working on a PR.~ This issue is for reference. Could not upload source file since it depends on data file which are huge.

```
(Py36) pragati@wasabi-simons ~/Work/scanpy_exp $ python planaria.py 
scanpy==1.3.2+4.g7c9fb1a anndata==0.6.11 numpy==1.14.6 scipy==1.1.0 pandas==0.23.4 scikit-learn==0.20.0 statsmodels==0.9.0 
... storing 'clusters' as categorical
computing tSNE
    using data matrix X directly
    using the 'MulticoreTSNE' package by Ulyanov (2017)
    finished (0:02:39.15)
Traceback (most recent call last):
  File "/home/pragati/Py36/lib/python3.6/site-packages/matplotlib/colors.py", line 158, in to_rgba
    rgba = _colors_full_map.cache[c, alpha]
KeyError: ('grey80', None)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/pragati/Py36/lib/python3.6/site-packages/matplotlib/axes/_axes.py", line 4210, in scatter
    colors = mcolors.to_rgba_array(c)
  File "/home/pragati/Py36/lib/python3.6/site-packages/matplotlib/colors.py", line 259, in to_rgba_array
    result[i] = to_rgba(cc, alpha)
  File "/home/pragati/Py36/lib/python3.6/site-packages/matplotlib/colors.py", line 160, in to_rgba
    rgba = _to_rgba_no_colorcycle(c, alpha)
  File "/home/pragati/Py36/lib/python3.6/site-packages/matplotlib/colors.py", line 204, in _to_rgba_no_colorcycle
    raise ValueError("Invalid RGBA argument: {!r}".format(orig_c))
ValueError: Invalid RGBA argument: 'grey80'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "planaria.py", line 47, in <module>
    sc.pl.tsne(adata, color='clusters', legend_loc='on data', legend_fontsize=5, save='_full')
  File "/home/pragati/Py36/lib/python3.6/site-packages/scanpy/plotting/tools/scatterplots.py", line 47, in tsne
    return plot_scatter(adata, basis='tsne', **kwargs)
[...]

PR #291, opened 2018-10-09
Fix custom R colors

fidelram:
This PR addresses #286 and adds an extra check for valid color names.

I added a test for the new functionality and took the opportunity to modify the plotting tests to mimic the way they are done for the pbmc3k tests.


falexwolf:
Nice one! :smile:

fidelram:
One bug less...

PR #292, opened 2018-10-11
Phenograph clustering

awnimo:
Hi @falexwolf 

Added a small wrapper for phenograph clustering, similar to phate.


falexwolf:
Thank you! Can you adapt the doc string so that it matches the other tools. We need numpydoc style documentation for it to render properly. You can also check whether it looks good by running `make html` in the docs folder.

awnimo:
All done! docstring corrected to render properly!

Thank you...

falexwolf:
Sorry about this taking so long, I'm waiting until we have the new way of handling extensions in place... It will only be another couple of days and then this is going to be merged and adapted to that... There's nothing to do from your end on this... Thank you!

flying-sheep:
@falexwolf what is that new way? isn’t there a tracking issue where extensions should be discussed and you said we don’t need them at all?

falexwolf:
I simply meant to do what I described in that issue:  https://github.com/theislab/scanpy/issues/271#issuecomment-431634492

falexwolf:
I'm finally merging this as we know have the `scanpy.external` submodule, where this can be properly linked - along with PHATE and all other wrappers to external single-cell packages.

Sorry for that this took so long, I only could do this during the past calm couple of days. Thank you for the PR!

A very happy new year to you!

PR #295, opened 2018-10-12
fixed #293 related to small point sizes and #286 related to R color names

### LINKED ISSUE #293: Scatter plot default point size
@fidelram 

It looks like the automatic point size selection for scatter plots has changed. Personally, I found the previous method more useful as I'm now getting more overplotting issues. For example, here are two plots made with the command `sc.pl.umap(bm, color="biological_sex")`, in different versions of scanpy:

Old result (at or before 1ec7af2):

old_plotting

Current result (at 94c3dc5):

new_plotting

I suspect difference in marker used is related to this. While scatter points were previously points, it seems like they are circles now. Setting the size just seems to make the width of line for each circle decrease. For example, here's a close up of a plot from the current version with `size=0.01` and `size=0.001` respectively:

screen shot 2018-10-11 at 5 52 39 pm
screen shot 2018-10-11 at 5 53 02 pm

fidelram:


falexwolf:
thank you!

PR #305, opened 2018-10-16
Add some missing requirements

bebatut:
Hi,

I was testing the different functions and found out that some requirements were missing and so in the `requirements.txt`. I tried to add the one I could find on Pypi.

Bérénice

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,11 @@ networkx
 natsort
 joblib
 numba
+magic
+dca
+pypairs
+phate
+fa2
+zappy
+dask
+bioservices


falexwolf:
Sorry, all of these packages aren't necessary for Scanpy's core functionality, supposed to be treated as extensions and shouldn't be installed by default.

Hopefully we'll have a way of handling this that makes it more clear in the future.

bgruening:
@falexwolf the problem is that there are functions that do now work without them and our down stream packages do not work ootb. Would it be possible to add a second file, e.g. `requirements-ect.txt` to keep track of this?

This file could also be used to create testing environments easily.

flying-sheep:
here we are:

https://github.com/theislab/scanpy/blob/6c786e70cecc2581a8009f83e3c843c8e998946c/setup.py#L25-L29

Doesn’t look complete though. Would someone like to complete it?

falexwolf:
Yes, we can simply add @bebatut's packages as an `ext` flag in the setup.

falexwolf:
PR welcome! 🙂

PR #308, opened 2018-10-17
This PR adds a dendrogram to heatmap, matrixplot, dotplot and staked_violins plots

fidelram:
This PR computes a hierarchical clustering and draws a dendrogram to some of the plots.  This is particularly useful when looking at marker genes from `scanpy.tl.rank_genes_groups` because the categories and marker genes are reordered according to the dendrogram.   Eg:

```PYTHON
```
image

image

Currently, for all rank_genes_groups_* plots, the dendrogram is set to `True` by default as I think this results in more meaningful images. For `heatmap`, `dotplot`, `matrixplot` and `stacked_violin` plots, I added a new parameter to activate the dendrogram:

```PYTHON
sc.pl.dotplot(pbmc, marker_genes, groupby='bulk_labels', dendrogram=True)
```
image

More example are available at https://gist.github.com/fidelram/2289b7a8d6da055fb058ac9a79ed485c


falexwolf:
That is super beautiful! :smile: Awesome!

fidelram:
So quick! thanks. I have been using the dendrograms for a while so
hopefully not so many bugs appear. Something that I wanted to have for
discussion is on some parameters relevant for the dendrogram, like the
genes used, the correlation method and the linkage method. All this can be
modified but currently is hard coded as I didn't want to add 3 more
parameters to the plotting functions.

Maybe you have faced a similar problem and have an elegant solution. I
thought about setting some variables like the rcParams for matplotlib but I
think is not justified for just 3 parameters and can be very confusing. Or
we can have a function to compute a dendrogram with all parameters
required, and save this in .uns like rank_genes_groups. Then if other
functions find this information they add the dendrogram.

falexwolf:
The last option is the way forward, I'd say. We should have a tool `tl.dendogram` in the clustering section that has a parameter to select the clustering for which one wants a dendogram, typically defaulting to `louvain` (unfortunately, we still don't have a good consistent naming convention across all tools; `groupby` predominates but is not ideal in this setting. `grouping` or `cluster_key` would also be possible).

You can still have the plotting functions call that function with default parameters. But the user wants more control, he or she can run the dendogram tool.

Of course, we also want dendograms for genes. I think the most elegant (but maybe confusing solution) is to do it as in `pl.scatter`, where annotation of observations is selected if the key is in `.obs` and variables annotations are selected if the key is in `.var`. What do you think?

fidelram:
Sounds good. If I find some time soon I will try that approach.

coh-racng:
It seems like the dendrogram feature has been merged into the master branch according to this thread. However, after updating my scanpy to 1.3.2, I still cannot access the dendrogram feature and get the following error when I set dendrogram=True:
AttributeError: Unknown property dendrogram

Here are the versions that I am using:
scanpy==1.3.2 anndata==0.6.10 numpy==1.14.3 scipy==1.1.0 pandas==0.23.0 scikit-learn==0.19.1 statsmodels==0.9.0 python-igraph==0.7.1 louvain==0.6.1

Issue #310, opened 2018-10-18
Bugs with pypairs functions


bebatut:
Hi,

I am testing the `scanpy.api.tl.sandbag` and `scanpy.api.tl.cyclone` functions, after installing the latest version of `pypairs`. 

1. I got an error: `TypeError: 'module' object is not callable`

   It seems that it could be solved by calling `pypairs.cyclone.cyclone` or `pypairs.sandbag.sandbag`

2. In `sandbag`: the argument `fraction` is not used to call the `pypairs` `sandbag` function
3. In `cyclone` and `sandbag`, `gene_names` and `sample_names` should not be `None`

I can fix these issues in a PR

Bérénice

falexwolf:
Yes, thank you. That would be very welcome! 

@rfechtner: could it be that compat with your interface got messed up? It would be nice if you'd maintained the interface when you do so drastic changes to the underlying package. Thanks!

rfechtner:
@bebatut Thank you for pointing that out. I will as well take a look at it. 

@falexwolf I will make sure to keep it in mind for further updates. The restructuring of the package structure came with the update from 1.x to 2.x and was necessary for some major improvements. I'm sorry for caused inconveniences.

@bebatut if you have further questions or issues, please let me know, I'd be happy to help you out

Ron

PR #312, opened 2018-10-18
To fix issue #310 with pypairs

bebatut:
Fix issue #310 
- Import sandbag and cyclone
- Align parameters to the one in `pypairs`
- Add documentation for the parameters

I wanted to add some tests but I was not sure which dataset makes sense. Any idea?

Bérénice


falexwolf:
Thank you! :smile:

PR #321, opened 2018-10-23
fixed #309 (Extra space around multiple scatter plot panels)

### LINKED ISSUE #309: Plotting layout behavior at repl
As mentioned in #299, I'm getting some strange layout behavior when plotting at the repl.

Using the setup:

```python
import scanpy.api as sc
import numpy as np

N = 1000
M = 2000

adata = sc.AnnData(X=np.random.random_sample((N, M)))

sc.pp.pca(adata)
```

I get outputs that look like the following 

```python
sc.pl.pca(adata, color=list("123"))
```

screen shot 2018-10-18 at 4 46 46 pm

```python
sc.pl.pca(adata, color=list("1234"))
```

screen shot 2018-10-18 at 4 47 37 pm

```python
sc.pl.pca(adata, color=list("12"), ncols=1)
```

screen shot 2018-10-18 at 4 50 11 pm

I get similar results with the MacOSX and GR backends (haven't checked others).

I think the issue is due to the implicit `bbox_inches="tight"` occurring with the matplotlib inline backend, and its explicit usage for `scanpy.plotting.utils.savefig`.  In the cases here, the bounding box is not being trimmed.

fidelram:


falexwolf:
Thank you!

One further thing to consider: with all these frequent image updates the repository will at some point explode in size. In all the image-based tests, we should use the smallest sizes possible. Images are already relatively small, but we can further reduce the size in the future. No necessary to remake all of them now, but something to keep in mind for future PRs.

What do you think?

PR #327, opened 2018-10-26
Speed up test collection

### LINKED ISSUE #326: [Suggestion] Speed up test collection
Currently, test collection takes about 11 seconds. This seemed a little long so I played around with the config a bit, and found if all the test files names are prepended with `test_`, and I set `python_files = test_*.py`, collection takes about 1 second.

Mind if I make that change?

ivirshup:
Test collection was taking a while, this should cut that time by about 90%.

For some reason, pytest test collection seems dependent on the number of files the `python_files` parameter matches. Having a more specific path under `testpaths` doesn't – which would be a simpler change.

Should close #326.


falexwolf:
Cool!

PR #329, opened 2018-10-26
Fix #322 wrong axis label for `sc.pl.diffmap`

### LINKED ISSUE #322: Wrong axis label when using components argument in sc.pl.diffmap
Example:
```python
import scanpy.api as sc
adata = sc.datasets.blobs()

sc.pp.neighbors(adata)
sc.tl.diffmap(adata)
sc.pl.diffmap(adata, components='1,2')
```
gives the output:
image
The axis labels should be DC1 and DC2. If running `sc.pl.diffmap(adata)` the axis labels are correct.

fidelram:


falexwolf:
Thank you!

PR #330, opened 2018-10-28
Hvg

Koncopd:


scanpy/preprocessing/recipes.py
```diff
@@ -2,6 +2,7 @@
 """
 
 from . import simple as pp
+from ._deprecated.highly_variable_genes import filter_genes_dispersion
```
falexwolf: I guess the `recipe_weinreb17` was relying on `filter_genes_cv_deprecated`, it's unfortunately not tested, but can you check and maybe add a small test? I fear this would break a lot of people's code that used the weinreb recipe.

scanpy/tests/_scripts/seurat_extract_hvg.r
```diff
@@ -0,0 +1,34 @@
+library(reticulate)
```
falexwolf, line 1: Ah! You're now requiring Seurat in the tests??? That's something that we don't want. No R scripts in the Scanpy tests... The previous version of `filter_genes_dispersion` was benchmarked and the tests should be produced with it; the new function should test against that output.

Seurat might definitely change their function in the future, and we don't want to get in trouble with that. We just want to be consistent within Scanpy.

falexwolf: OK, got it; you just used seurat to produce the test data.

falexwolf: That's of course fine; it's a bit redundant; but a nice-to-have. :smile:

scanpy/tests/highly_variable_genes.py
```diff
@@ -0,0 +1,33 @@
+import pandas as pd
+import numpy as np
+import scanpy.api as sc
+from pathlib import Path
+
+FILE = Path(__file__).parent / Path('_scripts/seurat_hvg.csv')
+
+def test_higly_variable_genes_compare_to_seurat():
```
falexwolf, line 8: Could you add the same test for the deprecated `filter_genes_dispersion`? We want to be sure that nobody touches it again in the future...

Thank you!

falexwolf:
The whole PR is pretty awesome already! I wrote some comments...

Can you also add the function to the docs, cross reference the deprecated and the new function in the Notes section and add it to the [release notes](https://github.com/theislab/scanpy/blob/master/docs/release_notes.rst)?

scanpy/tests/_scripts/seurat_extract_hvg.r
Koncopd: The script can be removed of course, but i thought it would be nice to have in case the dataset changes, for example.

scanpy/tests/highly_variable_genes.py
Koncopd: Yes, ofc.

scanpy/preprocessing/recipes.py
Koncopd: Oh, actually about this recipe, i forgot to import filter_genes_cv_deprecated from ._deprecated.highly_variable_genes. Fixed it.

For the recipes i just changed the point of import, no other things were changed.

Koncopd: https://github.com/theislab/scanpy/pull/330/commits/58323228ca2c040723980d5c508e9a4a21e45dbc

Koncopd:
Yes, i will.

scanpy/preprocessing/recipes.py
falexwolf: yes, I was referring to that omission :)

scanpy/tests/_scripts/seurat_extract_hvg.r
falexwolf: yes! it's cool!

falexwolf:
Thank you!

falexwolf:
I'm merging this now, as we're slowly running in danger of getting conflicts with other people working on this. Please move forward with the remaining issues in a new PR. Thank you!

PR #343, opened 2018-11-05
Hvg

Koncopd:
Added test for filter_genes_dispersion, updated docs

### DIFF
--- a/docs/release_notes.rst
+++ b/docs/release_notes.rst
@@ -7,12 +7,18 @@
 .. role:: smaller
 
 
+On master, :small:`November 5, 2018`
+------------------------------------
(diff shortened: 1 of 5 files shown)


Koncopd:
Just don't understand why build fails...

falexwolf:
It seems to be a stalled build in CI. Something to do with a URL request... if the tests run through on your end, everything should be fine. Do they?

docs/release_notes.rst
```diff
@@ -7,12 +7,18 @@
 .. role:: smaller
 
 
+On master, :small:`November 5, 2018`
```
falexwolf, line 10: No need for a new entry, here, just update the date remove the comma and the statement about "deprecation"; add your name with "thanks to"

falexwolf:
Thank you for this!

PR #364, opened 2018-11-15
kBET

flying-sheep:
Currently it’s almost a MVP.

What do you people think (especially @mbuttner)? Possible improvements:

- [x] **Heuristic for neighborhood size**
- [x] Expose more data: E.g. a column in `.obs` with the corrected p values
- [ ] **Better docs**
- [ ] **groupby**
- [ ] Mean of multiple samples instead of whole data
- [ ] Adapt to cells that appear in no neighborhoods
- [ ] Speedups?

I guess at least the heuristic and the docs have to go in!

I also don’t know if relying on the previous `neighbors` call is a good idea. Its default is 15 Neighbors, and even for a small (200 cells) dataset, the number the heuristic determined was 75.


mbuttner:
Hi,
I have checked your code and it looks good to me, very neat, actually.
Very important from my point of view is the neighbourhood size estimation - 15 neighbours is too few even for small data sets. 
If you do not have a heuristic, yet, I suggest to use a rough estimate for the neighbourhood size similarly to the R implementation as the quarter of the mean batch size: 
`floor(mean(batch_size)/4)`  
I should come up with a better estimate, though.

I can help you with the docs, if you like! 

Lastly, I'd find it important that you can apply kBET group-wise. For example, you may add a `groupby` parameter and then kBET automatically computes a rejection rate (or acceptance rate) for each cluster. I hope that's not too much overhead.

falexwolf:
How are things going?

I think all of this looks fine except for two things:
* we should *not* have a function `pp.kbet_neighbors` as it won't be used by anything else, `pp.neighbors` exists as it's a detrimental preprocessing step for 10 other tools; `tl.kbet` should make the call of `kbet_neighbors` internally and that should be a private function
* the heading *cluster scores* is very confusing; I'd put into a *metrics* section and the notion that has been established for these kind of things (mostly by the Seurat developers, I think) is "alignment metric"; I can easily make the change if we're ready to go...

PS: Why are there no contributions by you, @mbuttner? After all, it's your method and paper. :wink:

mbuttner:
I guess that everyone is still on holidays. 
There are no code contributions from my side, because I am still a beginner in Python. 
My contribution to kBET in Scanpy is on a higher level.

grst:
Hi @flying-sheep,

does the base functionality already work, or would you still advise me to use the R package if I want to use kBET?

flying-sheep:
I’d advise to use the R package. The score is still off and I didn’t figure out why. Sorry.

/edit: it works now. I’d like to have a good toy example for the tests that actually has a batch effect, then I’d merge this

gokceneraslan:
Some(most?) batch correction methods produce outputs in a lower dimensional space (like CCA or scVI), which are more likely to be stored in `.obsm`. Therefore, if you guys decide to call `kbet_neighbors` and `pp.neighbors` internally, do not forget to provide options like `use_rep` in `sc.tl.kbet()` to make things like `sc.pp.neighbors(..., use_rep='X_CCA')` possible.

Another idea is to keep multiple kNN graphs in `.uns['neighbors']` e.g. `.uns['neighbors']['default']` and `.uns['neighbors']['kbet']`.

fidelram:
@flying-sheep What is the status of this PR?

Zethson:
https://scib-metrics.readthedocs.io/en/stable/generated/scib_metrics.kbet.html

@adamgayoso scib-metrics will feature a kbet implementation. Think that this PR is pretty dead so I'll close it in favor of an ecosystem implementation. Feel free to reopen if we should discuss this again.

adamgayoso:
I didn't realize there was a PR for this, there are some things here that we don't have, like adjusting the pvalues and the neighbor search. The former is fast to add (though I didn't see it in the R implementation?) and the latter could be really slow to have in practice

mbuttner:
Yes, we did an attempt of implementing kBET in Python, but we did not follow up with sufficient testing whether the implementation reproduces the R package. The neighborhood search is slow indeed. For better comparability one should fix one neighborhood per dataset. That's also the reasons why we dropped it in the scIB package.
There was no pvalue adjustment in the original test, indeed. I created a statistic of 100 kBET results instead to compute a mean rejection rate and a range. There is a pvalue computation for the rejection rate, but it's almost always significant, and you rarely achieve rejection rates were the significance level is relevant. Moreover, it is a test statistic over a test statistic, so I am not quite sure how pvalues should be adjusted. 

> I didn't realize there was a PR for this, there are some things here that we don't have, like adjusting the pvalues and the neighbor search. The former is fast to add (though I didn't see it in the R implementation?) and the latter could be really slow to have in practice

adamgayoso:
Ok I think I understand. Now that I see the python code I understand that one only needs to run neighbors once for the max K and then run kbet a few times to maximize the acceptance rate.

This would actually not be hard to add and not be slow. It takes about 5 seconds to run kbet for a fixed K on 900k cells

mbuttner:
Yes, I totally agree that creating a fast implementation is probably not straightforward. The major bottleneck IMHO is computing this many neighbors to maximize the rejection rate, especially with 900k cells. In the original paper, we tried to find a range of neighborhood sizes K that return a maximal rejection rate, which is roughly between K = 50 to 0.5 * N where N denotes the number of cells, but there might be also a dependence on the number of batches, which we did not fully explore.

Issue #349, opened 2018-11-06
Figures related to rankings and heatmap could be transposed.


Xparx:
These are some suggestions regarding visualization defaults.

Working with the ranknings plot i realized that it would be much easier to read if the figure where transposed/rotated so that gene names could be read per line rather than per columns. This would also scale better if one wishes to plot more than 10-30 genes.
There is no inherent information on the y axis that could not be layed out on the x axis.

I also ran in to this format when using the heatmap plotting. Gene names are on the columns rather than rows resulting in a thin matrix and hard to read gene names. As the figure is constructed by using several axis one could transpose the figure having the gene names to the right colorbar on the right and groupings below or ontop of the heatmap. As one is most interested in a dendrogram on the cells (I think?) it could get a nice space on top.

falexwolf:
@fidelram should be the expert for this... :smile:

fidelram:
With respect to the heatmap, indeed it is possible to transpose the matrix.
Currently, this option is only available for `stacked_violin`. I thought
about adding this option to other plots like heatmap, matrixplot and
dotplot but I have not find the time and it is always possible to save the
figure and rotate it so it has low priority for me. The changes are not as
trivial as simply rotating the matrix as all other elements need to be
adjusted.

Xparx:
I made a hack solution for this to have something that would work for me. I put it here: https://gist.github.com/Xparx/33026da63dabb1c200b2602bbae0b95c

It's not defensively coded so it might bug out without proper input.

fidelram:
Thanks! Can you post an image to get an idea of your solution?

Xparx:
Sure. Here is one example of my output image

![4hpf](https://user-images.githubusercontent.com/715716/48497219-b68b6480-e801-11e8-9d80-2cf9219af59e.png)

fidelram:
Thanks. I will do a PR soon which allows the heatmap and matrixplot functions to be rotated.

Xparx:
As this issue is not closed I'll add a question here. 

Is it possible, or if not could it be added, that the cells in the heatmaps are sorted within the groupby variables. Either by pseudotime if availible or just clustered simply by hierarchical clustering. This could add a more visually and intuitive pleasing ordering of cells.

For example as in my figure above the groupby miss some property of the data with pattern over cells. If the cell ordering was random or default this pattern could not be seen.

fidelram:
Should be possible to do what you want.

[...]

PR #369, opened 2018-11-20
Figure improvements

fidelram:
This PR introduces the following changes:

 * `swap_axes` option was added to `pl.matrixplot` and `pl.heatmap`. When `swap_axes=True`, the x axis contains cells  and the y axis contains genes (#349).
 * added `show_genes_labels`  to `pl.heatmap`. This allows to have compact heatmaps without overlapping gene labels.
 * added lines to separate categories in  `pl.heatmap`.
 * changed categories colors in `pl.heatmap` by the colors found in  `adata.uns`
 * removed empty space that was present in different plots
 * added a `layer` option to specify which layer to use for plotting. 
 * added a new visualization called `pl.tracksplot`. 
 * changed `if <variable> is True` by `if <variable>` after @flying-sheep remarks. 
 * added `setup()` from matplotlib.testing
 * reduced dpi of test images to 40.
 * added var_groups plot for stacked_violin when `swap_axes=True` 
 * improved layout of stacked_violin (default width and linewidth of violin plots)

As an example, here is how some of the changes look like. I am using the `rank_genes_groups_*` plots because they contain more visual elements. Further examples can be seen here: https://gist.github.com/fidelram/2289b7a8d6da055fb058ac9a79ed485c

**heatmap** with `swap_axes=True, show_gene_labels=False`:
image

**matrixplot** with `swap_axes=True`:

image

**new *tracksplot***: Each *track* contains the var (genes) values sorted and colored according to the categories used:
image

**tracksplot** using the results of  `sc.tl.rank_genes_groups`:
image


flying-sheep:
Wow, this looks great! One remark for future PRs: We’re migrating to a new documentation style using type annotations (see e.g. https://github.com/theislab/scanpy/blob/master/scanpy/tools/louvain.py)

could you use that in future PRs, please? (no need to do it in this one)

scanpy/plotting/anndata.py
```diff
@@ -823,17 +823,21 @@ def stacked_violin(adata, var_names, groupby=None, log=False, use_raw=None, num_
...
-         By setting `swap_axes` then y are the `groupby` categories and x the `var_names`. When swapping
-         axes var_group_positions are no longer used
     var_group_rotation : `float` (default: `None`)
         Label rotation degrees. By default, labels larger than 4 characters are rotated 90 degrees
     row_palette: `str` (default: `muted`)
         The row palette determines the colors to use in each of the stacked violin plots. The value
         should be a valid seaborn palette name or a valic matplotlib colormap
         (see https://seaborn.pydata.org/generated/seaborn.color_palette.html). Alternatively,
         a single color name or hex value can be passed. E.g. 'red' or '#cc33ff'
+    swap_axes: `bool`, optional (default: `False`)
```
flying-sheep, line 798: while you’re moving this parameter in the docs, could you please move it to where it’s in the argument list (after `layer`)?

scanpy/plotting/anndata.py
```diff
@@ -1200,55 +1181,242 @@ def heatmap(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
...
+        if groupby + "_colors" in adata.uns:
+            colors = [adata.uns[groupby + "_colors"][x] for x in dendro_data['categories_idx_ordered']]
+
+    if show_gene_labels is None:
+        if len(var_names) <= 50:
+            show_gene_labels = True
+        else:
+            show_gene_labels = False
+            logg.warn('Gene labels are not shown when more than 50 genes are visualized. To show '
+                      'gene labels set `show_gene_labels=True`')
```
flying-sheep, line 1140: great idea to show a warning here!

LuckyMD:
This looks amazing!

scanpy/plotting/anndata.py
```diff
@@ -1200,55 +1181,242 @@ def heatmap(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
...
+
+        # plot group legends on top of heatmap_ax (if given)
+        if var_group_positions is not None and len(var_group_positions) > 0:
+            gene_groups_ax = fig.add_subplot(axs[0, 1], sharex=heatmap_ax)
+            _plot_gene_groups_brackets(gene_groups_ax, group_positions=var_group_positions,
+                                       group_labels=var_group_labels, rotation=var_group_rotation,
+                                       left_adjustment=-0.3, right_adjustment=0.3)
+
+    # swap axes case
+    else:
```
flying-sheep, line 1222: Wow that’s a huge branching. Could you please think about reducing duplication here, maybe with a helper function?

It’s hard to see what’s happening anyway, but on first glance, e.g. the `if categorical` blocks are nearly the same!

scanpy/plotting/anndata.py
```diff
@@ -1321,11 +1493,17 @@ def dotplot(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
     -------
     A list of `matplotlib.Axes` where the first ax is the groupby categories colorcode, the
     second axis is the heatmap and the third axis is the colorbar.
+
+    Examples
+    -------
+    >>> adata = sc.datasets.pbmc68k_reduced()
+    >>> sc.pl.heatmap(adata, ['C1QA', 'PSAP', 'CD79A', 'CD79B', 'CST3', 'LYZ'], 'bulk_labels',
+    ...               dendrogram=True, swap_axes=True)
```
flying-sheep: Thank you for adding an example!

fidelram:
@flying-sheep I will be happy to start using the proposed documentation style for future PRs!

scanpy/plotting/anndata.py
```diff
@@ -1657,38 +1798,363 @@ def matrixplot(adata, var_names, groupby=None, use_raw=None, log=False, num_cate
...
+        # plot group legends on top of matrix_ax (if given)
+        if var_group_positions is not None and len(var_group_positions) > 0:
+            gene_groups_ax = fig.add_subplot(axs[0, 0], sharex=matrix_ax)
+            _plot_gene_groups_brackets(gene_groups_ax, group_positions=var_group_positions,
+                                       group_labels=var_group_labels, rotation=var_group_rotation,
+                                       left_adjustment=0.2, right_adjustment=0.8)
+
+        # plot colorbar
+        pl.colorbar(pc, cax=color_legend)
+    else:
```
flying-sheep, line 1667: same here: both branches look very similar, please add some abstraction and don’t copy&paste code.

scanpy/plotting/anndata.py
```diff
@@ -1804,30 +2276,55 @@ def _plot_gene_groups_brackets(gene_groups_ax, group_positions, group_labels,
...
+            codes.append(Path.LINETO)
+            codes.append(Path.LINETO)
+
+            try:
+                group_x_center = left[idx] + float(right[idx] - left[idx]) / 2
+                gene_groups_ax.text(group_x_center, 1.1, group_labels[idx], ha='center',
+                                    va='bottom', rotation=rotation)
+            except:
+                pass
+    else:
```
flying-sheep, line 2045: again nearly identical if/else branches

flying-sheep (changes requested):
great job! there’s some duplication going on that should be avoided, otherwise the code looks great and the plots gorgeous!

falexwolf:
This is awesome! Thank you, Fidel! :smile:

falexwolf:
edit: moved to #373

scanpy/plotting/anndata.py
fidelram: I will add the change. Actually, I think that many doc arguments are repeated in the different figures and can be put in one place as is already happening with the arguments for show and save. Any comments on this?

scanpy/plotting/anndata.py
fidelram: I will see how can I reduce the branching further. I think I can move  the `if categorical` block as a separate function as this is also used by the tracks plot.  Let me know if you have any further suggestions.

scanpy/plotting/anndata.py
fidelram: I will check if I can add some abstraction but is annoying as almost all the lines are slightly different. For example: 

image

flying-sheep: Hmm, in this case yes, but is that the case for all three branchings I commented on?

scanpy/plotting/anndata.py
flying-sheep: That’s a good idea! Just a tip: Pay attention with the indentation, if that’s off somewhere, the docs will break

flying-sheep:
I took this unrelated discussion to its own issue :smiley:

fidelram:
Together with the suggested changes, I am also updating my usual notebook containing examples of all the plots (~~not yet updated:~~ https://gist.github.com/fidelram/2289b7a8d6da055fb058ac9a79ed485c). However, don't you think that this could be part of a the scanpy tutorials section?

scanpy/plotting/anndata.py
fidelram: I tried to reduce as much code repetition as possible. For the three branching that you pointed out it is not possible to reduce the code because 'width' and 'height' and 'x' and 'y' need to be changed in almost all lines even though most of the code remains the same.

flying-sheep: Oh, that’s easy! sorry, I should have been clearer on how I imagined it.

you can just name the dimensions after what they contain, and assign width and height accordingly:

```py
def heatmap(...):
    ...
    dendro_size = 1 if dendrogram else 0
    ...
    size_dim_var = heatmap_size + dendro_size + groupby_size
    size_dim_grp = ...
    ...
    # at this point we get concrete what width and height is
    if not swap_axes:
        width = size_dim_var
        height = size_dim_grp
    else:
        width = size_dim_grp
        height = size_dim_var
```

Of course I saw that you fiddled with those, and the dimensions are different depending on the elements being horizontal or vertical. you could fix that similar to this:

```py
class Sizes(NamedTuple):
    h: float
    v: float
    def __add__(self, other: 'Sizes'):
        return Sizes(self.h + other.h, self.v + other.v)

sizes0 = Sizes(0., 0.)
SIZES = dict(
    dendrogram=Size(h=1., v=.8),
    ....
)

def heatmap(...):
    ...
    dendro_sizes = SIZES['dendrogram'] if dendrogram else sizes0
    ...
    sizes_dim_var = heatmap_sizes + dendro_sizes + groupby_sizes
    sizes_dim_grp = ...
    ...
    # at this point we get concrete what width and height is
    if not swap_axes:
        width = sizes_dim_var.h
        height = sizes_dim_grp.v
    else:
        width = sizes_dim_grp.h
        height = sizes_dim_var.v
```

flying-sheep: A disadvantage with my approach is that the dimensions aren’t local anymore. `SIZES` is defined globally and therefore not in the same function where they’re used (and fiddled with if they don’t fit).

you could of course define that dict inside of the function to rectify it, making it a tiny bit longer and slower but increasing locality.

What do you think about the general idea?

fidelram: I like the idea of using a named tuple. I think I could reduce some code that way and this will also help to keep the dimensions in one place. 

However, I think that trying to find generalizable patterns for each semi-duplicated line will unnecessarily increase the complexity of the code. For example, in one case the code is: 
```PYTHON
        heatmap_ax.set_ylim(obs_tidy.shape[0] - 0.5, -0.5)
        heatmap_ax.set_xlim(-0.5, obs_tidy.shape[1] - 0.5)
        heatmap_ax.tick_params(axis='y', left=False, labelleft=False)
```
while for the other case is:
```PYTHON
        heatmap_ax.set_xlim(0, obs_tidy.shape[0])
        heatmap_ax.set_ylim(obs_tidy.shape[1] - 0.5, -0.5)
        heatmap_ax.tick_params(axis='x', bottom=False, labelbottom=False)
```

Furthermore, the the layout also changes. For example, for heatmap showing cells x genes, the the groupby categories are on the left, dendrogram and colorbar are on the right and the 'brackets' are on top (2x4 layout). For the genes x cells case, the group by categories are at the bottom, the dendrogram is at the top and 'brackets' and colorbar are to the right (3x3 layout). 

At the moment I will leave any further code improvements for the future.

flying-sheep: Welp, future=never, but I understand why you decided against it in the shown case.

Are you sure it applies in all cases or are there parts that can be simplified?

fidelram: Honestly, I like the SIZES idea and also the type annotations and I hope to get back to that at some point, but I need to move on to other pressing issues at the moment. 

I thought that I could not simplify any further but you showed me that it is indeed possible, so to answer your question: I don't see how to simplify further but a keen eye may find further improvements.

flying-sheep: OK, then! LGTM

flying-sheep (approved):

scanpy/plotting/anndata.py
```diff
@@ -2000,3 +2239,154 @@ def translate_pos(pos_list, new_ticks, old_ticks):
...
+    for code, (label, value) in enumerate(obs_tidy.index.value_counts(sort=False).iteritems()):
+        ticks.append(value_sum + (value / 2))
+        labels.append(label)
+        value_sum += value
+        label2code[label] = code
+
+    groupby_ax.grid(False)
+
+    if orientation == 'left':
+        groupby_ax.imshow(np.matrix([label2code[lab] for lab in obs_tidy.index]).T, aspect='auto', cmap=groupby_cmap)
```
flying-sheep: One thing I don’t get: The colors in the bar look pretty arbitrarily chosen from the colormap to me:

grafik

grafik

but when i debug, the matrix you construct contains [0,0,0,0,…,1,1,1,1,…,2,2,2,…], so I’d expect the first colors to be chosen instead?

flying-sheep: ah, OK, fixed it in 426f028708cdd203b7d97d48eb558e695090da82

I think you’ll have to recreate the test images though

grst:
@fidelram, I really like your plotting gallery! Would be cool to have that as part of the tutorials or even integrated in the main documentation (enhance each plotting function with an example image?)

flying-sheep:
OK, my changes in 426f028708cdd203b7d97d48eb558e695090da82 didn’t make the tests break!

Do we currently not use the plotting test results?

falexwolf:
> However, don't you think that this could be part of a the scanpy tutorials section?

Of course, this should become a part of the scanpy tutorials section! That would be awesome! I already asked you for that some time ago. :wink: I'd also start adding calls producing images to the doc pages.

Let me build the core infrastructure for having the tutorials run on readthedocs and adding notebooks containing only code to the scanpy main repo.

If you want, to shortcut, you can make a PR to scanpy_usage and upload your notebook there. Or, equally well, we simply link to your notebook from the tutorials page.

falexwolf:
Can I merge this, @fidelram and @flying-sheep?

fidelram:
Green light on my side to merge.

flying-sheep:
See my last comment. After fixing the colormaps in this PR, I didn’t update the images, but the tests still pass. What’s up with that?

fidelram:
@flying-sheep I think that your changes should produce images that are almost equal to the ones on the tests as your changes simply introduce a different way to get the colormap. Btw, what is the advantage of using `ListedColormap` and `BoundaryNorm` instead of `LinearSegmentedColormap` ?

fidelram:
@falexwolf I added a PR for the notebook that I have on the gist to the scanpy_usage (https://github.com/theislab/scanpy_usage/pull/9)

flying-sheep:
The important part is the `BoundaryNorm`. We got a really weird selection of colors without it, since the default is to treat the colormap as a linear space. `max(vec)` gets the last color, `min(vec)` the first one, and everything else some color between. Using the `BoundaryNorm` I defined, numbers in `[0, len(colors)-1]` get the color at the respective index, and everything smaller or bigger would get the first or last color (not ideal, but better than what we had)

I don’t think it really makes a difference, but ListedColormap is a colormap for discrete uses like ours, LinearSegmentedColormap is for interpolating continuous values onto the map.

Before | After
--- | ---
before | after

In the before pic, we map `[0,1,2,3]` onto `[0;19]`, which results in `[0, 5.75, 10.5, 14.25, 19]`, and `[to_hex(tab20.colors[math.ceil(i)]) for i in [0, 5.75, 10.5, 14.25, 19]]` gives us [`#1f77b4`, `#98df8a`, `#8c564b`, `#c7c7c7`, `#9edae5`]

fidelram:
Thanks for the clarification. Your solution guarantees that the same colors present in let's say .uns['louvain_colors'] will be represented accurately. The previous solution would not have achieve that.

falexwolf:
Thank you again! I'm merging this. Release prior to this is 1.3.4.

fidelram:
Great! Thanks.

PR #371, opened 2018-11-20
Norml

Koncopd:
https://github.com/theislab/scanpy/issues/301


scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,204 @@
...
+    after = np.median(counts) if after is None else after
+    counts += (counts == 0)
+    counts /= after
+    if issparse(X):
+        X = sparsefuncs.inplace_row_scale(X, 1/counts)
+    else:
+        X /= counts[:, None]
+    return X if copy else None
+
+def normalize_quantile(data, counts_per_cell_after=None, counts_per_cell=None,
```
falexwolf: Can we have a better name than `counts_per_cell_after`? Maybe `counts_after` or  `normed_value` or `target_norm` or `target_sum` or something? What's the best name for this?

Can we remove the `counts_per_cell` parameter? Previously, it was just there to enable the recursive call.

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,204 @@
...
+                       layers=[], use_rep=None):
+    """Normalize total counts per cell.
+
+    Normalize each cell by total counts over genes, so that every cell has
+    the same total count after normalization.
+
+    Similar functions are used, for example, by Seurat [Satija15]_, Cell Ranger
+    [Zheng17]_ or SPRING [Weinreb17]_.
+
+    Parameters
```
falexwolf: Can we please stick to 80 lines also in the docstrings? On my 13inch this already gets messed up here when looking at the file on Github, so that I cannot properly check this.

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,204 @@
...
+        to the median of the *counts_per_cell* before normalization.
+    counts_per_cell : `np.array`, optional (default: `None`)
+        Precomputed counts per cell.
+    quantile : `float`, optional (default: 1)
+        Only use genes are less than fraction (specified by *quantile*) of the total
+        reads in every cell.
+    min_counts : `int`, optional (default: 1)
+        Cells with counts less than `min_counts` are filtered out during
+        normalization.
+    key_n_counts : `str`, optional (default: `'n_counts'`)
```
falexwolf: Probably, we shouldn't add anything to `adata` by default. There are dedicated functions for doing this in the meanwhile: https://scanpy.readthedocs.io/en/latest/api/scanpy.api.pp.calculate_qc_metrics.html

Also, if allowing for adding something, the naming convention should reflect the discussion around this: https://github.com/theislab/scanpy/pull/358

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,204 @@
...
+        to cells and columns to genes.
+    counts_per_cell_after : `float` or `None`, optional (default: `None`)
+        If `None`, after normalization, each cell has a total count equal
+        to the median of the *counts_per_cell* before normalization.
+    counts_per_cell : `np.array`, optional (default: `None`)
+        Precomputed counts per cell.
+    quantile : `float`, optional (default: 1)
+        Only use genes are less than fraction (specified by *quantile*) of the total
+        reads in every cell.
+    min_counts : `int`, optional (default: 1)
```
falexwolf: I would totally remove the possibility of filtering from the normalization function.

If people want to filter, they should do it before. A normalization function should not change the dimensions of the data matrix. Many people got confused by this.

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,204 @@
...
+        normalization.
+    key_n_counts : `str`, optional (default: `'n_counts'`)
+        Name of the field in `adata.obs` where the total counts per cell are
+        stored.
+    copy : `bool`, optional (default: `False`)
+        If an :class:`~anndata.AnnData` is passed, determines whether a copy
+        is returned.
+    layers : `str` or list of `str`, optional (default: `[]`)
+        List of layers to normalize. Set to `'all'` to normalize all layers.
+    use_rep : `str` or `None`, optional (default: `None`)
```
falexwolf: How come that `use_rep` refers layers? It was used in other instances for providing a key to `.obsm`, which might store a compressed representation of the data (e.g. from PCA or an autoencoder). It doesn't make sense at all to have this for keying layers??? Sorry if I missed this previously. 

I cannot stay like this.

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,204 @@
...
+        If `None`, after normalization, for each layer in *layers* each cell has a total count equal
+        to the median of the *counts_per_cell* before normalization of the layer.
+        If `'after'`, for each layer in *layers* each cell has a total count equal
+        to counts_per_cell_after.
+        If `'X'`, for each layer in *layers* each cell has a total count equal
+        to the median of the *counts_per_cell* of data.X before normalization.
+
+    Returns
+    -------
+    Returns or updates `adata` with normalized version of the original
```
falexwolf: We want to move away from the `copy` parameter. Instead, we want an `inplace` parameter that defaults to `True`. If it's `False`, the function doesn't return an AnnData but only the part of it that was computed in the function. In some respects unfortunately, we cannot have `copy` and `inplace` at the same time. It would only be acceptable in a transition phase, with a warning that is output.

In any case, for new functions, we directly want to go with inplace. Please adapt this and rewrite the returns section of the docstring similar to [this](https://scanpy.readthedocs.io/en/latest/api/scanpy.api.pp.calculate_qc_metrics.html).

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,204 @@
...
+                    min_counts=1):
+    """Normalize total counts per cell.
+
+    Normalize each cell by total counts over all genes, so that every cell has
+    the same total count after normalization.
+
+    Similar functions are used, for example, by Seurat [Satija15]_, Cell Ranger
+    [Zheng17]_ or SPRING [Weinreb17]_.
+
+    Parameters
```
falexwolf: We don't want to replicate the docstring here. Please use the decorator `@doc_params` that is used in wide parts of Scanpy for avoiding replicated docstrings. Both `normalize_total` and `normalize_counts` should be based on the same docstring. Take a look at https://github.com/theislab/scanpy/blob/master/scanpy/plotting/tools/scatterplots.py, for instance.

falexwolf:
Thank you for this! I added several comments in the the code!

scanpy/preprocessing/normalization.py
Koncopd: But the new functions need to wrap the old, no? I left it for wrappers.

scanpy/preprocessing/normalization.py
Koncopd: Yeah, i will remove

scanpy/preprocessing/normalization.py
Koncopd: But the new functions need to wrap the old, no? I left it for wrappers.

scanpy/preprocessing/normalization.py
Koncopd: Ok

scanpy/preprocessing/normalization.py
Koncopd: It is like this in normalize_per_cell...

scanpy/preprocessing/normalization.py
Koncopd: Ok

scanpy/preprocessing/normalization.py
Koncopd: I added this because it was in normalize_per_cell. So is it needed?

scanpy/preprocessing/normalization.py
Koncopd: Will do like this.

Koncopd:
Also one problem is that i don't understand where to put `materialize_as_ndarray` as it used [here](https://github.com/theislab/scanpy/blob/44c038ad7b6488407958ab020858923b25368d97/scanpy/preprocessing/simple.py#L598) and in filtering.

scanpy/preprocessing/normalization.py
falexwolf: No, the new functions do not need to wrap the old functions. If you want to use the same code base for new and old functions, that's cool and better than simply moving the old functions into `_deprecated`. I'd advocate for it in this case, as I don't think it's going to be a lot of work.

But everything should be based on a private `_norm_base` (better name here, please) function that is called by both the old and the new functions.

The new functions should not exhibit any trace of the bad design choices and parameter names  of the old functions, so no `counts_per_cell_after` etc.

scanpy/preprocessing/normalization.py
falexwolf: Thank you!

scanpy/preprocessing/normalization.py
falexwolf: Keep it in there, but it should default to `None` in which case nothing gets added. If it's set, it is used to add the total counts as annotation.

scanpy/preprocessing/normalization.py
falexwolf: That might be true. That's why I wrote "sorry if I missed this previously". I don't know whether you or Volker added it, but it's definitely a serious inconsistency to use this parameter name here.   Please, think of a different name.

falexwolf:
If you're not able to figure out `matieralize_as_ndarray`, that's fine. That's for the distributed part, which can come at a later stage. Important is that the old `normalize_per_cell` function continues to work as expected.

scanpy/preprocessing/normalization.py
Koncopd: @falexwolf 
And what is the intended effect of normalization on cells that were filtered out previously (i.e. have lower number of counts than min_counts). Just leave them as they are?

falexwolf: There should not be any `min_counts` argument in the new function anymore.

The only thing that I would do is raise a warning if some rows add up to zero and hence cannot be normalized. Instead of making them all `nans`, I'd leave them as zeros and output the warning.

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,88 @@
...
+cell_sum_after : `float` or `None`, optional (default: `None`)
+    If `None`, after normalization, each cell has a total count equal
+    to the median of the *counts_per_cell* before normalization.
+min_counts : `int`, optional (default: 1)
+    Cells with counts less than `min_counts` are filtered out during
+    normalization.
+key_n_counts : `str`, optional (default: `None`)
+    Name of the field in `adata.obs` where the total counts per cell are
+    stored.
+inplace : `bool`, optional (default: `True`)
```
falexwolf: `inplace` should appear at the very end of the list. It's a generic argument like `copy` and will appear in almost every function.

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,88 @@
...
+
+Parameters
+----------
+data : :class:`~anndata.AnnData`
+    The annotated data matrix of shape `n_obs` × `n_vars`. Rows correspond
+    to cells and columns to genes.
+cell_sum_after : `float` or `None`, optional (default: `None`)
+    If `None`, after normalization, each cell has a total count equal
+    to the median of the *counts_per_cell* before normalization.
+min_counts : `int`, optional (default: 1)
```
falexwolf: This is no longer needed, right?

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,88 @@
...
+
+Similar functions are used, for example, by Seurat [Satija15]_, Cell Ranger
+[Zheng17]_ or SPRING [Weinreb17]_.
+
+Parameters
+----------
+data : :class:`~anndata.AnnData`
+    The annotated data matrix of shape `n_obs` × `n_vars`. Rows correspond
+    to cells and columns to genes.
+cell_sum_after : `float` or `None`, optional (default: `None`)
```
falexwolf: Can we do `sum_after` instead of `cell_sum_after`? `*counts_per_cell*` shouldn't appear any longer but should be `sum_after`. Can we do "`observation (cell)`" instead of "each cell"?

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,88 @@
...
+min_counts : `int`, optional (default: 1)
+    Cells with counts less than `min_counts` are filtered out during
+    normalization.
+key_n_counts : `str`, optional (default: `None`)
+    Name of the field in `adata.obs` where the total counts per cell are
+    stored.
+inplace : `bool`, optional (default: `True`)
+    Whether to update data or return dictionary with normalized copies of
+    `data.X` and `data.layers`.
+layers : `str` or list of `str`, optional (default: `[]`)
```
falexwolf: The default value here is an anti-pattern: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

Please correct correspondingly.

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,88 @@
...
+    List of layers to normalize. Set to `'all'` to normalize all layers.
+layer_norm : `str` or `None`, optional (default: `None`)
+    Specifies how to normalize layers.
+    If `None`, after normalization, for each layer in *layers* each cell
```
falexwolf: Consider making a list with bullet points, here. Bullet points via `*`.

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,88 @@
...
+doc_norm_quant = """\
+quantile : `float`, optional (default: 1)
+    Only use genes are less than fraction (specified by *quantile*)
+    of the total reads in every cell.\
+"""
+
+doc_ret = """\
+Returns
+-------
+Returns dictionary with normalized copies of data.X and data.layers
```
falexwolf: `data.X` should be in backticks.

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,88 @@
...
+
+Normalize each cell by total counts over genes, so that every cell has
+the same total count after normalization.
+
+Similar functions are used, for example, by Seurat [Satija15]_, Cell Ranger
+[Zheng17]_ or SPRING [Weinreb17]_.
+
+Parameters
+----------
+data : :class:`~anndata.AnnData`
```
falexwolf: As all of this is for an `AnnData` and we use `adata` for an `AnnData` throughout, can you please adapt this?

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,88 @@
...
+    The annotated data matrix of shape `n_obs` × `n_vars`. Rows correspond
+    to cells and columns to genes.
+cell_sum_after : `float` or `None`, optional (default: `None`)
+    If `None`, after normalization, each cell has a total count equal
+    to the median of the *counts_per_cell* before normalization.
+min_counts : `int`, optional (default: 1)
+    Cells with counts less than `min_counts` are filtered out during
+    normalization.
+key_n_counts : `str`, optional (default: `None`)
```
falexwolf: Can we do `key_added` here? It occurs in several other functions, too.

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,88 @@
...
+[Zheng17]_ or SPRING [Weinreb17]_.
+
+Parameters
+----------
+data : :class:`~anndata.AnnData`
+    The annotated data matrix of shape `n_obs` × `n_vars`. Rows correspond
+    to cells and columns to genes.
+cell_sum_after : `float` or `None`, optional (default: `None`)
+    If `None`, after normalization, each cell has a total count equal
+    to the median of the *counts_per_cell* before normalization.
```
falexwolf: Here, the quantile case merits more description, right? In the quantile case, the passed quantile will sum up to this `sum_after` constant.

Shall we call it `target_sum` instead? I guess it's the better name but people will be more confused to see that name... But better names should be preferred. So, if you agree, I'd go for `target_sum`.

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,109 @@
...
+    counts /= after
+    if issparse(X):
+        X = sparsefuncs.inplace_row_scale(X, 1/counts)
+    else:
+        X /= counts[:, None]
+    return X if copy else None
+
+@doc_params(norm_bulk=doc_norm_bulk, norm_quant=doc_norm_quant, ret=doc_ret, ex_quant=doc_ex_quant)
+def normalize_quantile(data, cell_sum_after=None, quantile=1, key_n_counts=None,
+                       inplace=True, layers=[], layer_norm=None):
```
falexwolf: What I noted above in the docstring obviously applies to the function header...

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,109 @@
...
+        counts_per_cell = np.ravel(counts_per_cell)
+
+        gene_subset = (X>counts_per_cell[:, None]*quantile).sum(0)
+        gene_subset = (np.ravel(gene_subset) == 0)
+    else:
+        logg.msg('normalizing by total count per cell', r=True)
+
+    X = X if gene_subset is None else data[:, gene_subset].X
+    counts_per_cell = X.sum(1)
+    #get rid of data view
```
falexwolf: there should always be a space between a `#` and the actual comment

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,109 @@
...
+    counts_per_cell = np.ravel(counts_per_cell).copy()
+    del X
+    del gene_subset
```
falexwolf: No necessity for deleting the objects. It will not free memory anyways unless you call a garbage collector... Simply remove...

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,109 @@
...
+
+    if key_n_counts is not None:
+        data.obs[key_n_counts] = counts_per_cell
```
falexwolf: Please adapt the namings also in the code. `counts_per_cell` should rather be `target_sum` or something...

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,109 @@
...
+
+    cell_subset = counts_per_cell>0
+    if not np.all(cell_subset):
+        logg.warning('Some cells have total count of genes equal to zero')
```
falexwolf: There is no `logg.warning`. It's `logg.warn`, also in the standard Python logging module...

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,109 @@
...
+
+    if layer_norm == 'after':
+        after = cell_sum_after
+    elif layer_norm == 'X':
+        after = np.median(counts_per_cell[cell_subset])
+    elif layer_norm is None:
+        after = None
+    else: raise ValueError('layer_norm should be "after", "X" or None')
```
falexwolf: This should go on the next line.

scanpy/preprocessing/normalization.py
```diff
@@ -0,0 +1,109 @@
...
+
+    return dat if not inplace else None
+
+@doc_params(norm_bulk=doc_norm_bulk, ret=doc_ret, ex_total=doc_ex_total)
+def normalize_total(data, cell_sum_after=None, key_n_counts=None, inplace=True, layers=[],
+                    layer_norm=None, min_counts=1):
+    """\
+    {norm_bulk}
+
+    {ret}
```
falexwolf: `ret` is very cryptic for `return`. Never ever use cryptic abbreviations when coding, always be verbose and clear.

falexwolf (changes requested):
Wonderful pull request and a wonderful contribution. But there are still a few simple things to do.

Thank you!

scanpy/preprocessing/normalization.py
Koncopd: Yes, this only deletes the references. The logic is to delete references which i don't use later. I don't know when garbage collection happen precisely but i believe it may happen during the execution of the function and then it will free the memory (so it is not dragged through the whole function). If it doesn't happen, then it is no worse than not using del...

But i delete it of course, if you think that it clutters the function.

scanpy/preprocessing/normalization.py
Koncopd: Could you please explain why you don't like this name?
I believe this name for the internal variable explains what is going on quite well. This is not a target sum, just total counts vector for each cell...

scanpy/preprocessing/docs.py
```diff
@@ -0,0 +1,89 @@
...
+    to the median of total counts for observations (cells) of `adata.X`
+    before normalization.
+inplace : `bool`, optional (default: `True`)
+    Whether to update `adata` or return dictionary with normalized copies of
+    `adata.X` and `adata.layers`.\
+"""
+
+doc_norm_quant = """\
+quantile : `float`, optional (default: 1)
+    Only use genes are less than fraction (specified by *quantile*)
```
fidelram: I think there is a typo here. Do you mean 'Only use genes whose count is less than fraction ...' ?

Koncopd: Hi, you are right, should be something like this. I'll fix.

scanpy/preprocessing/normalization.py
flying-sheep: I’d rather recommend to use helper functions for the main steps of the long main function.

1. less spaghetti, the main function is basically just a list of steps!
2. get GC without having to sprinkle “del”
3. you can’t miss deleting a variable you don’t need anymore

falexwolf:
Thank you for all the fixes! Is this ready to go from your side, @Koncopd?

Koncopd:
Hi, @falexwolf 
I think so. However, there is a question of what to do with current normalize_per_cell. Should i move it to deprecated? Also, it supports zarr and my version doesn't.

falexwolf:
@tomwhite: Are you planning to continue to develop and maintain the distributed backend of Scanpy? If yes, should we announce it at some point? I already added you to the authors' list but there is still no entry here:  https://scanpy.readthedocs.io/en/latest/#on-master-january-1-2019 There should definitely be. There should also be an announcement on twitter.

@Koncopd: If Tom's answer is "yes", we definitely need support of zarr/dask in the new functions. Sorry, that I wasn't explicit about it...

tomwhite:
@falexwolf @Koncopd Yes. Agree about announcing it - let me chat to the others @lasersonlab about it too.

falexwolf:
@tomwhite OK, I added this to the release notes (https://github.com/theislab/scanpy/commit/cee23dc13cf2b77d8e23ee0f91eb55fac0e35ed8, sorry confounded with some style change); it would be nice to have a link to your performance benchmarks... Let me know when we should announce it on twitter. I'm also happy to retweet...

Koncopd:
Ok, i will add the support for them.

Koncopd:
Hi, @falexwolf .
Now the functions work with dask and zappy.

falexwolf:
This is awesome, Sergei! :D

Issue #370, opened 2018-11-20
use_raw=True has no effect for plotting.


Xparx:
scanpy.pl.tsne or any of the variants only distinguish use_raw=None and other.
All other values than None results in using the processed X matrix rather than the assigned raw data even if using use_raw=True

falexwolf:
That's quite a bug... Let me check, it can only have stayed unnoticed as the default shows the *raw* data, if `.raw` has been set.

Xparx:
I should add.
scanpy version = '1.3.2'
Python 3.6.6

Xparx:
I upgraded to 1.3.3 and the bug persists. 

PS: I accidentally closed the issue for some reason.

fidelram:
I can reproduce the problem. Very strange. I will submit a PR to fix it.

PR #372, opened 2018-11-21
fix issue #370 related to use_raw

fidelram:


fidelram:
@flying-sheep Can you merge this PR? This is really a bug that should be solved fast.

falexwolf:
Thank you!

PR #380, opened 2018-11-28
This fixes problems with using gene_symbol arg in rank_genes_groups_violin

briangottfried:
This PR aims to fix violin plot labels and KeyError when using the gene_symbol argument to rank_genes_groups_violin.


falexwolf:
Thank you! 🙂

jorvis:
As an added note, it would be great to see this 'gene_symbol' argument used uniformly across the plotting functions.  We've had to handle it in pretty hacky ways to make it work throughout the pipeline.

falexwolf:
I completely agree! We had this discussion already at some point in another issue (unfortunately don't find that issue right now).

I just made a dedicated issue for that: https://github.com/theislab/scanpy/issues/385

PR #382, opened 2018-11-30
Cellbrowser exporter for Scanpy

maximilianh:
This is inspired by the Spring exporter. It also fills a directory with output files, but in addition to the matrix, adds cluster markers, meta data and embeddings. I keep the conversion code in the cellbrowser package, so I can change it easily in the future. We could also copy the whole function over, but I don't know if this makes a lot of sense, as people will need the cellbrowser package anyways and currently it doesn't have any dependencies, so should be easy to install.

This function can optionally convert to html files, start a webserver and serve the result.


maximilianh:
It would be great is someone could look over my matrix exporter quickly. I am not sure if it uses the fastest way to export the matrix to text and whether it wastes memory. First I do ad.X.transpose().tocsr() then I loop over the gene indices and do row = mat.getrow(i).todense(). In retrospect transpose().tocsr() seems weird.

Here is the code:
://github.com/maximilianh/cellBrowser/blob/master/src/cbPyLib/cellbrowser/cellbrowser.py#L2613

In both places, I need if-then-else constructs, as the anndata matrix can also be not-sparse, and then the tocsr() and getrow() functions are not defined (this may be a bug in the current spring exporter, it probably doesn't work with not-sparse matrices)

maximilianh:
When the anndata object was created from a 10X file, are the gene symbols stored somewhere? If yes, I probably should export them, too.

falexwolf:
Yes, you have the choice of either having 'gene_symbols' as your index or 'gene_ids', what is not in the index is available as a column in `.var`.

falexwolf:
Thank you for the PR! It looks good to me. Also the function underlying, as far as I can tell. If there are performance problems, we can still address them in an update.

80 character lines would be nice also for the docstring. Then I could see whether they make sense. I'm seeing this on a 13-inch screen and the docstring looks like a mess through that. ;)

maximilianh:
Thanks! Using the most formal of logic :-), line length is now 80 chars.

maximilianh:
I wonder if cellbrowser should be a suggested extra package for scanpy now... but spring isn't one either, so probably not.

falexwolf:
Thank you, I'll merge this!

We'll get back to the "extra" package question, soon... :wink:

PR #390, opened 2018-12-05
dotplot can take vmin vmax arguments from user

timrandg:
Hi Fidel, 
Here is the pull request for vmin vmax in dotplot...
I am guessing that there could be similar issues with other plotting functions and other plotting keywords.
In general, it would be best to check plotting methods for plotting keywords and use what is provided in kwds by default rather than setting the variable without checking whether it was provided. 
Thanks for your groups great work!
Tim

### DIFF
--- a/scanpy/plotting/anndata.py
+++ b/scanpy/plotting/anndata.py
@@ -1465,7 +1465,16 @@ def dotplot(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
     size = (frac * 10) ** 2
 
     import matplotlib.colors
-    normalize = matplotlib.colors.Normalize(vmin=min(mean_flat), vmax=max(mean_flat))
+
+    #if user defined, then use the vmax, vmin keywords, else use data to generate them...
+    if ('vmax' in kwds) and ('vmin' in kwds):
+        _vmax = kwds['vmax']
+        _vmin = kwds['vmin']
+    else:
+        _vmax = max(mean_flat)
+        _vmin = min(mean_flat)
+
+    normalize = matplotlib.colors.Normalize(vmin=_vmin, vmax=_vmax)
     colors = [cmap(normalize(value)) for value in mean_flat]
 
     dot_ax.scatter(x, y, color=colors, s=size, cmap=cmap, norm=None, edgecolor='none', **kwds)
--- a/scanpy/plotting/anndata.py
+++ b/scanpy/plotting/anndata.py
@@ -1466,20 +1466,8 @@ def dotplot(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
 
     import matplotlib.colors
 
-    #if user defined, then use the vmax, vmin keywords, else use data to generate them...
-    if 'vmax' in kwds:
-        _vmax = kwds['vmax']
-    else:    
-        _vmax = max(mean_flat) 
-
-    if 'vmin' in kwds:
-        _vmin = kwds['vmin']
-    else:    
-        _vmin = min(mean_flat)  
-
-    #normalize = matplotlib.colors.Normalize(vmin=min(mean_flat), vmax=max(mean_flat))
-    normalize = matplotlib.colors.Normalize(vmin=_vmin, vmax=_vmax)
-    colors = [cmap(normalize(value)) for value in mean_flat]
+    normalize = matplotlib.colors.Normalize(vmin=kwds.get('vmin'), vmax=kwds.get('vmax'))
+    colors = cmap(normalize(mean_flat))
 
     dot_ax.scatter(x, y, color=colors, s=size, cmap=cmap, norm=None, edgecolor='none', **kwds)
     y_ticks = range(mean_obs.shape[0])
--- a/scanpy/plotting/anndata.py
+++ b/scanpy/plotting/anndata.py
@@ -1467,13 +1467,17 @@ def dotplot(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
     import matplotlib.colors
 
     #if user defined, then use the vmax, vmin keywords, else use data to generate them...
-    if ('vmax' in kwds) and ('vmin' in kwds):
+    if 'vmax' in kwds:
         _vmax = kwds['vmax']
-        _vmin = kwds['vmin']
-    else:
-        _vmax = max(mean_flat)
-        _vmin = min(mean_flat)
+    else:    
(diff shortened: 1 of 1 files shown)


falexwolf:
This looks good to me, you agree @Fidel?

fidel:
Could you please stop mentioning me? I’m not a part of this project. You did it several times already.

flying-sheep:
It’s @fidelram. The problem is probably that GitHub suggests you, @fidel, in autocompletions even though you’re not a contributer. Maybe adding @fidelram to the contributors list of this project would help?

scanpy/plotting/anndata.py
```diff
@@ -1465,7 +1465,16 @@ def dotplot(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
     size = (frac * 10) ** 2
 
     import matplotlib.colors
-    normalize = matplotlib.colors.Normalize(vmin=min(mean_flat), vmax=max(mean_flat))
+
+    #if user defined, then use the vmax, vmin keywords, else use data to generate them...
+    if ('vmax' in kwds) and ('vmin' in kwds):
```
fidelram: Frequently, I only set the vmax because I am fine with vmin (eg. 0). However, in the current code nothing will happen and this can be very confusing. 

Can you add a conditional for vmax and vmin separately?

timrandg: Hi Fidel, I pushed a new commit that takes vmin vmax as individual optional arguments and generates either from the data when they are not provided. Please see commit c0c28ee from my fork.

falexwolf:
@fidel Sorry about this, I'm well aware of it now and autocomplete shouldn't fool me anymore.

@flying-sheep @fidelram already is a collaborator in the repo settings. Can we do more?

fidelram:
Thanks @timrandg, that will do it. @falexwolf I think we can merge this PR

flying-sheep:
> @flying-sheep @fidelram already is a collaborator in the repo settings. Can we do more?

I did that immediately after writing about it.

scanpy/plotting/anndata.py
```diff
@@ -1465,7 +1465,20 @@ def dotplot(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
     size = (frac * 10) ** 2
 
     import matplotlib.colors
-    normalize = matplotlib.colors.Normalize(vmin=min(mean_flat), vmax=max(mean_flat))
+
+    #if user defined, then use the vmax, vmin keywords, else use data to generate them...
+    if 'vmax' in kwds:
+        _vmax = kwds['vmax']
+    else:
```
flying-sheep: there’s trailing whitespace everywhere. please make your editor get rid of it on save.

scanpy/plotting/anndata.py
```diff
@@ -1465,7 +1465,20 @@ def dotplot(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
...
+    else:    
+        _vmax = max(mean_flat) 
+
+    if 'vmin' in kwds:
+        _vmin = kwds['vmin']
+    else:    
+        _vmin = min(mean_flat)  
+
+    #normalize = matplotlib.colors.Normalize(vmin=min(mean_flat), vmax=max(mean_flat))
+    normalize = matplotlib.colors.Normalize(vmin=_vmin, vmax=_vmax)
```
flying-sheep: any reason the whole PR can’t be replaced by the following?

```python
normalize = matplotlib.colors.Normalize(vmin=kwds.get('vmin'), vmax=kwds.get('vmax'))
colors = cmap(normalize(mean_flat))
```

`Normalize`, when `None` is passed, uses the min and max of its first input to determine the boundaries. When we give it `mean_flat`, that means that one line behaves exactly the same as your like 10 lines.

both `Colormap.__call__` and `Normalize.__call__` are of course vectorized, so no need to make everything slow by introducing list comprehensions or other loops.

flying-sheep (changes requested):
I think the code was too complex before and is now even more complex…

scanpy/plotting/anndata.py
fidelram: isn't `None` in that case the default value when 'vmin' is not in `kwds`?

flying-sheep: exactly. specifying `None` (or nothing) for `vmin` means that `None` is passed to `Normalize`, which means that `min(mean_flat)` is used in the next line. (same for `vmax`)

fidelram: ok, now I see it.  The code was redundant before the PR when `Normalize` was called. In theory your suggestion should not change anything which makes me wonder why I initially put `vmin=min(mean_flat)`. @timrandg can you make the suggested change by @flying-sheep and we check that the test does not fails?

timrandg:
Okay. The improved solution suggested by @flying-sheep works for me. I
changed to that solution and pushed.
Tim

timrandg:
I submitted both the exact changes suggested by @flying-sheep and 

    normalize = matplotlib.colors.Normalize(vmin=kwds.get('vmin'), vmax=kwds.get('vmax'))
    colors = [cmap(normalize(value)) for value in mean_flat]

as a second submission, trying to pass the tests. But it looks like the tests are failing for some other reason outside of the anndata.py file (only file modified in my commit). 

I don't think the tests are failing from the changes made here.

flying-sheep:
Exactly, the error was introduced by some third party update or so.

Therefore there was no need for 6e797fa, and it even is is wrong, the second line *needs* to be

```py
colors = cmap(normalize(mean_flat))
```

It’s both faster and necessary: `Normalize` determines vmin and vmax from the first time it’s called when they’re not set / set to `None`. And when you call it with `normalize(mean_flat[0])` (what happens in the list comprehension), vmin gets set to `min(mean_flat[0]) == mean_flat[0]` instead of `min(mean_flat)`

please do

```sh
git reset --hard a4b3ccd88f0412461813838d5435ce0cc0b10883
git push -f
```

timrandg:
I reset to previous commit (a4b3ccd
<https://github.com/theislab/scanpy/pull/390/commits/a4b3ccd88f0412461813838d5435ce0cc0b10883>)
and  pushed as instructed.
Tim

flying-sheep (approved):

flying-sheep:
OK great. The same errors as on master happen, so this didn’t introduce any failures. Thank you very much.

PR #403, opened 2018-12-20
Pca for sparse data

Koncopd:
For this - https://github.com/theislab/scanpy/issues/393

Benchmarks
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/pca_for_sparse.ipynb

### DIFF
--- a/scanpy/preprocessing/pca_for_sparse.py
+++ b/scanpy/preprocessing/pca_for_sparse.py
@@ -1,7 +1,8 @@
 from scipy.sparse import csr_matrix
 import numpy as np
 
-class CentSparse(sp.sparse.csr_matrix):
+#need to pass issparse check
+class CentSparse(csr_matrix):
     def __init__(self, A, m=None):
         self.A = A
         self.m = A.mean(0) if m is None else m
@@ -16,6 +17,10 @@ class CentSparse(sp.sparse.csr_matrix):
         return BA - Bm
     def toarray(self):
         return self.A - self.m
+    def astype(self, dtype):
+        if dtype == self.dtype:
+            return self
+        return CentSparse(self.A.astype(dtype), self.m.astype(dtype))
     @property
     def dtype(self):
         return self.A.dtype
@@ -25,3 +30,9 @@ class CentSparse(sp.sparse.csr_matrix):
     @property
     def T(self):
         return CentSparse(self.A.T, self.m.T)
+    @property
+    def data(self):
+        return self.A.data
+    @property
+    def indices(self):
+        return self.A.indices
--- a/scanpy/preprocessing/pca_for_sparse.py
+++ b/scanpy/preprocessing/pca_for_sparse.py
@@ -4,6 +4,7 @@ from packaging import version
 from scipy.sparse import csr_matrix
 from sklearn import __version__ as sklearn_version
 from sklearn.utils import check_random_state
+from sklearn.utils.extmath import randomized_svd
 
 # In older versions sklearn csr_mean_variance_axis0 always copies the whole sparse matrix
 # if its dtype is float64
--- a/scanpy/preprocessing/pca_for_sparse.py
+++ b/scanpy/preprocessing/pca_for_sparse.py
@@ -1,10 +1,27 @@
 from scipy.sparse import csr_matrix
+import numpy as np
 
-class CentSparse(csr_matrix):
-    def __init__(self, A):
+class CentSparse(sp.sparse.csr_matrix):
+    def __init__(self, A, m=None):
         self.A = A
-        self.m = A.mean(0).A1
+        self.m = A.mean(0) if m is None else m
+        self.c_r = self.m.shape[0] == 1
     def __mul__(self, B):
-        return self.A * B - self.m * B
+        AB = self.A * B
+        mB = np.dot(self.m, B) if self.c_r else self.m * B.sum(0)[None, :]
+        return AB - mB
     def __rmul__(self, B):
-        return B * self.A - B * np.ones_like(self.m)[:, None] * self.m
+        BA = B * self.A
+        Bm = B.sum(1)[:, None] * self.m if self.c_r else np.dot(B, self.m)
+        return BA - Bm
+    def toarray(self):
+        return self.A - self.m
+    @property
+    def dtype(self):
+        return self.A.dtype
+    @property
+    def shape(self):
+        return self.A.shape
+    @property
+    def T(self):
+        return CentSparse(self.A.T, self.m.T)
--- a/scanpy/preprocessing/pca_for_sparse.py
+++ b/scanpy/preprocessing/pca_for_sparse.py
@@ -1,16 +1,25 @@
-from scipy.sparse import csr_matrix
-from sklearn.utils.sparsefuncs_fast import csr_mean_variance_axis0
+
 import numpy as np
+from packaging import version
+from scipy.sparse import csr_matrix
+from sklearn import __version__ as sklearn_version
+from sklearn.utils import check_random_state
+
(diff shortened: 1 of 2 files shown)


falexwolf:
This is nice! Thank you!

It appears to me that the benchmarks show that this only becomes relevant for *very* large data. So we need to be mindful to not break backward compatibility for all the small and medium-size datasets that people use (which we do by introducing the tiny difference). Don't you think that in the light of this, it would be better to leave the default as is (densifying) and have an option `sparse_pca` or something similar?

Koncopd:
>It appears to me that the benchmarks show that this only becomes relevant for very large data.

Hm, even for my example it is 77.14 MiB vs 893.92 MiB, so 10 times difference. This seems large to me, no?

falexwolf:
> Hm, even for my example it is 77.14 MiB vs 893.92 MiB, so 10 times difference. This seems large to me, no?

Yes, it's definitely large and it's awesome that you solved this problem! I just meant that it's not hitting people's computational resources limits: your example is 60K x 2K, so quite big already, if you densify you need 800MB, which is easily available even on a laptop. That's what I meant.

What do you think?

Koncopd:
Yeah, this seems important only for large datasets in that sense. So, i will add sparse_pca option with False by default.

Koncopd:
Not sure what kind of test to add for this...

falexwolf:
As discussed, @Koncopd will try to integrate this into scikit-learn itself and not into Scanpy. :smile:

Koncopd:
Similar pull request exists already in sklearn.
https://github.com/scikit-learn/scikit-learn/pull/12841
Will watch.

cyrus303 (approved):

flying-sheep:
> So, i will add sparse_pca option with False by default.

How about make it smartly decide by default, informed by the benchmarks? You can describe in the param docs when it happens, and give a debug message which one was choosen.

scanpy/preprocessing/pca_for_sparse.py
```diff
@@ -0,0 +1,70 @@
+
```
flying-sheep, line 1: File needs to be named starting with _ as the others

scanpy/preprocessing/pca_for_sparse.py
```diff
@@ -0,0 +1,70 @@
...
+import numpy as np
+from distutils.version import StrictVersion
```
flying-sheep, line 3: `distutils.version` is broken. Please use `from packaging import version` and then `version.parse`. (I added `packaging` to the requirements on master)

scanpy/preprocessing/pca_for_sparse.py
```diff
@@ -0,0 +1,70 @@
...
+# if its dtype is float64
+# https://github.com/scikit-learn/scikit-learn/blob/f0ab589f1541b1ca4570177d93fd7979613497e3/sklearn/utils/sparsefuncs_fast.pyx#L72
+
+if StrictVersion(sklearn_version) < StrictVersion('0.20.0'):
+    from sklearn.utils.sparsefuncs_fast import _csr_mean_variance_axis0
+    mean_variance = lambda X: _csr_mean_variance_axis0(X.data, X.shape, X.indices)
+else:
+    from sklearn.utils.sparsefuncs_fast import csr_mean_variance_axis0 as mean_variance
+
+# need to pass issparse check
```
flying-sheep, line 19: I don’t understand this comment.

scanpy/preprocessing/pca_for_sparse.py
```diff
@@ -0,0 +1,70 @@
...
+
+        m, var = mean_variance(X)
+        m = np.asmatrix(m)
+        var = var.sum()
+        C = CentSparse(X, m)
+
+        if k >= n_features:
+            raise ValueError('n_components must be < n_features;'
+                             ' got %d >= %d' % (k, n_features))
+        U, Sigma, VT = randomized_svd(C, k, n_iter=self.n_iter,
```
flying-sheep, line 63: please don’t use uppercase letters in variable names.

scanpy/preprocessing/simple.py
```diff
@@ -354,7 +354,9 @@ def pca(
     data: Union[AnnData, np.ndarray, spmatrix],
     n_comps: int = N_PCS,
     zero_center: Optional[bool] = True,
+    sparse_pca: Optional[bool] = False,
```
flying-sheep, line 357: Maybe make it `None` by default, and, based on the benchmarks, use it when the data is sufficiently large.

In this case, make sure you document this in the parameter docs and log a debug message.

flying-sheep (changes requested):
Sorry about causing merge conflicts. Squashing all your commits and then rebasing on master is probably going to be easier than rebasing all the commits individually

gokceneraslan:
Shall we replace sparse_pca with something else like implicit_centering or so? 

sparse_pca might give rise to confusion since there is indeed a PCA technique called sparse PCA.

Koncopd:
Hm, it was decided to suspend this pr earlier.
There is an analogous pr in scikit-learn, but i'm not sure it will got forward.
I'm not sure what to do with this pr...

flying-sheep:
Ah, then we should leave it open and unchanged until there’s a decision on the sklearn PR. Sorry I missed the decision.

VolkerBergen:
What's the current state here?

This just came out 
https://genomebiology.biomedcentral.com/articles/10.1186/s13059-019-1900-3
favoring sklearn's PCA implementation while stating that it cannot yet handle sparse matrices. Is that still true?

Koncopd:
Last time i checked it was true. Need to check again, but i bet it is still the same.

ivirshup:
That's still the case (at least for randomized PCA @Koncopd linked above), though it looks like there may be a another path forward using other solvers: https://github.com/scikit-learn/scikit-learn/issues/12794. Still needs an implementation though.

atarashansky:
You could just add a `sparse` argument to `pca`. If True, just call this function instead of scikit-learn's PCA:

```
def sparse_pca(X,npcs,mu = None):
    # X -- scipy sparse data matrix
    # npcs -- number of principal components
    # mu -- precomputed feature means. if None, calculates them from X.

    # compute mean of data features
    if mu is None: 
        mu = X.mean(0).A.flatten()[None,:]

    # dot product operator for the means
    mmat = mdot = mu.dot 
    # dot product operator for the transposed means
    mhmat = mhdot = mu.T.dot 
    # dot product operator for the data
    Xmat = Xdot = X.dot 
    # dot product operator for the transposed data
    XHmat = XHdot = X.T.conj().dot 
    # dot product operator for a vector of ones
    ones = np.ones(X.shape[0])[None,:].dot 

    # modify the matrix/vector dot products to subtract the means
    def matvec(x): 
        return Xdot(x) - mdot(x)
    def matmat(x): 
        return Xmat(x) - mmat(x)
    def rmatvec(x): 
        return XHdot(x) - mhdot(ones(x))
    def rmatmat(x): 
        return XHmat(x) - mhmat(ones(x))
    
    # construct the LinearOperator
    XL = sp.sparse.linalg.LinearOperator(matvec = matvec, dtype = X.dtype,
                                         matmat = matmat,
                                         shape = X.shape,
                                        rmatvec = rmatvec, rmatmat = rmatmat)
     
    u,s,v = sp.sparse.linalg.svds(XL,solver='arpack',k=npcs)
    
    # i like my eigenvalues sorted in decreasing order
    idx = np.argsort(-s)
    S = np.diag(s[idx])
    # principal components
    pcs = u[:,idx].dot(S) 
    # equivalent to PCA.components_ in sklearn 
    components_ = v[idx,:] 
    return pcs,components_
```

This only works for the `arpack` solver. It's a bit slower than PCA on dense matrices (since arpack is slower than randomized), but it's super memory efficient.

PR #406, opened 2018-12-27
Getting rid of scanpy.api

falexwolf:
This distributes the reexports in `scanpy/api/` in two modules: the core functionality goes into the root (`scanpy/__init__.py`) and the external functionality goes into `scanpy/ext/`. The structure of the latter mirrors the one of `scanpy.api` so that users simply need to adjust their imports
```
import scanpy as sc
import scanpy.ext as sce
```
instead of previously
```
import scanpy.api as sc
```

Backward compatibility for use of `scanpy.api` is fully maintained, but from now on, its use and development are deprecated.


falexwolf:
Happy to see your review, @flying-sheep!

Note that the rehauled docs are here: https://icb-scanpy.readthedocs-hosted.com/en/beyondapi/api/index.html

falexwolf:
@fidelram FYI, this defines an explicit plotting submodule in `plotting/__init__.py`, which also contains the docs for it. There is no need to reexport to `api/pl.py` anymore. Everything is backwards compat and if you want to use new functionality, simply `import scanpy` instead of `import scanpy.api`...

flying-sheep:
can you give me access to https://icb-scanpy.readthedocs-hosted.com?

I just have access to the public readthedocs.org

falexwolf:
I sent you an invitation for readthedocs.com about 2 months ago already - I just resent it. :)

falexwolf:
I've tested this quite a bit in the past 5 days and am merging it into master. In essence, it's a superficial change.

flying-sheep:
> I sent you an invitation for readthedocs.com about 2 months ago already - I just resent it. :)

Well, doesn’t seem like it worked in the past: What I got now was not an invitation that I needed to click, but simply a notification that I’m now member of the team on rtd.com (which I wasn’t before)

The changes look good! I would however prefer to do things via `.. include::` instead of duplicating code for the `scanpy` and `scanpy.api` sections. Except if you plan to not update the `scanpy.api` module and docs section.

falexwolf:
> Except if you plan to not update the scanpy.api module and docs section.

Yes, that's the plan. `scanpy.api` is completely phased out an simply there for backwards compatibility.

flying-sheep:
OK, great!

PR #410, opened 2019-01-04
Fix upper limit of Bonferroni adjusted p-values to 1.0

gokceneraslan:
In some cases Bonferroni adjustment might lead to p-values higher than 1.0. For R implementation, see: https://github.com/SurajGupta/r-source/blob/master/src/library/stats/R/p.adjust.R#L40


gokceneraslan:
Here is the multitest Python implementation: https://github.com/statsmodels/statsmodels/blob/master/statsmodels/stats/multitest.py#L251

falexwolf:
Thank you! 🙂

PR #412, opened 2019-01-04
Make tool, preprocessing and plotting submodules private

falexwolf:
Now that we changed from the reexports in `scanpy.api` whose sole purpose were to reexport top-level functions to the user in a clean way, we run into two problems;
* one can no longer access the submodules as the names are occupied by the functions
* the users sees private modules such as `pp.distributed` and similar

Therefore, I'd change everything that is not intended to be used by the user to a private submodule. This actually was the very first design before `scanpy.api` (inspired by `statsmodels.api` and `matplotlib.pyplot`) was introduced.

Comments are welcome: @flying-sheep and everyone else.


flying-sheep:
looks good! also tells users not to use any of those things

falexwolf:
Merging this to be able to proceed with https://github.com/theislab/scanpy/pull/398.

PR #423, opened 2019-01-11
Add .readwrite.write to the new API

gokceneraslan:
I'm not sure if `sc.write` is forgotten or left out on purpose. Leaving it out make sense as it only offers `sc.write('file.csv', adata)` over `adata.write()`. Alternatively, extending `sc.write` functionality to loom and zarr and keeping it in the new API might make it more useful.

I was using it just to make the code more symmetric :)  i.e. `sc.read` and `sc.write`, but I don't mind if it's removed.


falexwolf:
No, you're right. It's fine to continue having this. Initially, I wanted to get rid of it at some point... but also numpy has this redundancy between many array attributes which pop up as numpy level functions everywhere.

Thank you!

PR #427, opened 2019-01-15
fix docstring neighbors

fbnrst:
the docstring mentioned ``method=='gauss'``, which is not a valid option.


flying-sheep:
Thank you very much!

falexwolf:
Thank you! 🙂

PR #430, opened 2019-01-15
Add `gene_symbols` argument to scatter plots

ivirshup:
Added a `gene_symbols` argument to the scatter plot functions so indices can be standardized and unique while plots have interpretable titles.

Usage:

```python
sc.pl.umap(adata, colors=["Sox2"], gene_symbols="name")
```

One question I have: should the user be allowed to pass both gene ids and gene symbols? Currently this is not allowed, but easily could be.


falexwolf:
Thank you! This is great and very helpful! :smile:

Yes, it would be nicer if the user were allowed to pass both. But isn't it possible simply by virtue of not passing `gene_symbols`?

ivirshup:
No problem! I liked the option for all the other plots, but call this function way more.

By passing both, I mean both in the same call, like:

```python3
sc.pl.umap(adata, colors=["Sox2", "ENSMUSG00000042501"], gene_symbols="name")
```

This could be allowed pretty easily, but I didn't do it at first since I like having less potential ambiguity.

PR #439, opened 2019-01-22
Use np.asarray() rather than Zappy-specific code.

tomwhite:
This uses the `__array__` method on ndarray-like classes to convert from
a distributed array to a regular NumPy ndarray when `materialize_as_ndarray` is called. This was prompted by an [improvement in Zappy](https://github.com/lasersonlab/zappy/pull/7) (released in 0.2.0) that makes Zappy arrays implement the `__array__` method. As another benefit, this change should make it easier to use other implementations of the ndarray interface in Scanpy in the future.

Note that there is still a Dask-specific call for the case of a sequence of arrays, since Dask can materialize these in one call to `compute`.

### DIFF
--- a/scanpy/preprocessing/_distributed.py
+++ b/scanpy/preprocessing/_distributed.py
@@ -1,26 +1,15 @@
+import numpy as np
+
 # install dask if available
 try:
     import dask.array as da
 except ImportError:
     da = None
 
-# install zappy (which wraps numpy), or fall back to plain numpy
-try:
-    import zappy.base as np
-except ImportError:
-    import numpy as np
-
-
 def materialize_as_ndarray(a):
     """Convert distributed arrays to ndarrays."""
     if type(a) in (list, tuple):
         if da is not None and any(isinstance(arr, da.Array) for arr in a):
             return da.compute(*a, sync=True)
-        elif hasattr(np, 'asarray'): # zappy case
-            return tuple(np.asarray(arr) for arr in a)
-    else:
-        if da is not None and isinstance(a, da.Array):
-            return a.compute()
-        elif hasattr(np, 'asarray'): # zappy case
-            return np.asarray(a)
-    return a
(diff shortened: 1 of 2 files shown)


tomwhite:
BTW I have successfully run the distributed tests with this change (`pytest scanpy/tests/test_preprocessing_distributed.py`).

falexwolf:
Nice! Tests should also be run by Travis, shouldn't they? Or have we missed out on demanding dependencies and your tests won't run through for that reason? If so, please point me to it and I'll make sure that Travis actually runs the tests.

tomwhite:
Thanks @falexwolf. The tests are not run by default since dask etc are not installed. I install them with the following to get them to be picked up:

```
pip install dask[array] zappy zarr
pip install pytest
```

falexwolf:
@flying-sheep: Do you agree that we should add this to the travis setup? I thought about creating a `requirements_tests.txt` as for `anndata` and simply adding the line to `.travis.yml`. Good solution? Maybe it's even your solution, I don't remember. :smile:

flying-sheep:
I’m not a fan of duplicating things. We already install optional requirements via the list of extras here:

https://github.com/theislab/scanpy/blob/f428848ece1d7a4794090eb70a34a3b8f1953dee/.travis.yml#L8

so we should simply add them to the `test` extra:

https://github.com/theislab/scanpy/blob/f428848ece1d7a4794090eb70a34a3b8f1953dee/setup.py#L35

or add more extras (e.g. `dask=['dask[array]'],`) and add them to the list of extras to be installed in .travis.yml

falexwolf:
I completely agree. It should simply go in the `test` extra. @tomwhite, would you do that? It might that the tests don't run through on Travis for some reason and then, I guess, it would be great if you could look into it (would for sure be a problem that would pop elsewhere, too).

tomwhite:
@falexwolf @flying-sheep Thanks for the suggestion. I'll take a look at this.

PR #441, opened 2019-01-22
Consistent hint for the distance matrix

gokceneraslan:
Define `.uns['neighbors']['distances']` as the distance matrix instead of the weighted adjacency matrix in logg.hint.

By the way, when `knn=True`, distances to non-neighbor data points are stored as zero in this matrix to keep it sparse, right? But technically these are not zeros, but rather "unknown values". Does it make sense to add a note in the documentation about that? Because for `knn=False`, the semantics of zeros change completely.

### DIFF
--- a/scanpy/neighbors/__init__.py
+++ b/scanpy/neighbors/__init__.py
@@ -96,7 +96,7 @@ def neighbors(
     logg.info('    finished', time=True, end=' ' if settings.verbosity > 2 else '\n')
     logg.hint(
         'added to `.uns[\'neighbors\']`\n'
-        '    \'distances\', weighted adjacency matrix\n'
+        '    \'distances\', distances for each pair of neighbors\n'
         '    \'connectivities\', weighted adjacency matrix')
     return adata if copy else None
 


falexwolf:
Thanks!

Your comment about distances is exactly true and it's the canonical usage. I'm fine with adding a note about this.

flying-sheep:
Yeah, I think I invented that convention in the early days of destiny, and it sloshed over to here.

It recently completely confused a destiny user, which made me realize that this is *not* canonical and intuitive use, but just something I came up with so long ago that I forgot.

falexwolf:
Haha, this convention has nothing to do with destiny. :wink: Also the code in Scanpy solely started off Laleh's Matlab version. Unfortunately, I never managed to actually dig into the destiny code...

The convention is the typical convention used since the early days of knn graphs and spectral clustering, hence I say "canonical usage". Fun that the [following reference](https://scholar.google.com/scholar?cluster=8671001140798821162&hl=en&as_sdt=0,5&sciodt=0,5), which is cited >6K times and which I've sent around 100s of times since I stumbled across it in fall 2015, isn't yet fully acknowledged in the field. We also put it into the DPT paper. It starts with outlining the "different types of similarity graphs" = the convention mentioned above. 🙂

flying-sheep:
> Also the code in Scanpy solely started off Laleh's Matlab version.

In which I introduced that convention when helping Laleh to make it more efficient :wink: 

I don’t know if others came up with it independently before I did in early 2015, but it wouldn’t surprise me :laughing:

gokceneraslan:
The convention I know is to return two `n × k` matrices, one with neighbor indices and one with neighbor distances. I think this is more explicit, no confusion about zeros, but the sparse matrix is more compact for sure.

flying-sheep:
It's only more compact when you have a specific symmetric sparse matrix class. Otherwise it's pretty much the same. The difference is that it's easier to operate on an adjacency matrix for many tasks, while the dual matrix storage is only good for determining the nth nearest neighbor per cell directly.

falexwolf:
> In which I introduced that convention when helping Laleh to make it more efficient

Cool, I didn't know that! Should have made it a lot more efficient! :smile:

> The convention I know is to return two n × k matrices.

Right, this is the default in sklearn. But yes, in the end, we want some sort of adjacency matrix for convenience and direct integration with all the graph stuff.

PR #442, opened 2019-01-23
Easier 10x reading

ivirshup:
Removes the need to specify a genome string in 10x h5 files by default. This removes a personal annoyance of mine, where I had to figure out what the reference was called when there's often only one reference used per file.

For cellranger `v3.0.0+` files, specifying a genome acts as a filter on input, as it did already. However, it doesn't only act if `gex_only` is `True`. Additionally, the behavior of `gex_only` has been changed to fit the documentation, i.e. it just filters for gene expression variables.

For legacy files:

* If the file only has one genome group, that one is used by default
* If multiple genomes are found and the user did not specify one, an error will be thrown. This is because there are no structural assurances the genomes will match to the same samples.

As the behavior of the function has meaningfully changed, this is a breaking change (though I'd be surprised if it affected many people). Personally, I haven't seen many 10x files which contain multiple genomes, so I'd appreciate feedback or examples from people who have.

### DIFF
--- a/scanpy/readwrite.py
+++ b/scanpy/readwrite.py
@@ -89,15 +89,16 @@ def read(filename, backed=False, sheet=None, ext=None, delimiter=None,
     return read_h5ad(filename, backed=backed)
 
 
-def read_10x_h5(filename, genome='mm10', gex_only=True):
+def read_10x_h5(filename, genome=None, gex_only=True):
     """Read 10x-Genomics-formatted hdf5 file.
 
     Parameters
     ----------
     filename : `str` | :class:`~pathlib.Path`
         Filename.
-    genome : `str`, optional (default: ``'mm10'``)
-        Genome group in hdf5 file.
+    genome : `str`, optional (default: ``None``)
+        Filter expression to this genes within this genome. For legacy 10x h5
+        files, this must be provided if the data contains more than one genome.
     gex_only : `bool`, optional (default: `True`)
         Only keep 'Gene Expression' data and ignore other feature types,
         e.g. 'Antibody Capture', 'CRISPR Guide Capture', or 'Custom'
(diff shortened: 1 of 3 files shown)


falexwolf:
Wow, awesome! @gokceneraslan wanted this from the beginning. Thank you so much! :smile:

flying-sheep:
I don’t consider it breaking. If I understod you right, the only change in behavior are that not specifying a genome works now in cases where there’s only one.

No longer throwing an error is a perfectly fine change!

ivirshup:
I thought it was breaking due to a couple behavior changes:

One case where the results would be different is the call `sc.read_10x_h5(h5pth)`, where the file at `h5pth` is a legacy formatted file which contains `mm10` and `hg38` genomes. Prior to this PR, only the `mm10` genome would be read in. Now, an error is thrown. 

If the file had the `v3` format (also containing two genomes), now values for features from both genomes would be read in, instead of just `mm10`.

Even better than removing an error, previously `v3` files would get all the vars filtered out and not throw an error!

flying-sheep:
OK! You explained that the behavior you have for v3 isn’t possible to replicate for legacy, so I’m sure this is fine.

ivirshup:
It's definitely possible to replicate, I don't know if it makes sense as a default. If the legacy files with multiple genomes almost always had the same cells aligned against all those genomes, it could make sense to read everything in by default. I'm not sure if that's the case though.

flying-sheep:
Ha, that’s what I meant, that you said it doesn’t make sense.

> If multiple genomes are found and the user did not specify one, an error will be thrown. This is because there are no structural assurances the genomes will match to the same samples.

I very much agree!

falexwolf:
Oh, I was too hasty in merging this. Thanks for clarifying more of this. I think it's perfectly fine to have this better and more stringent behavior. 

Added a note in the release notes: https://github.com/theislab/scanpy/commit/f428848ece1d7a4794090eb70a34a3b8f1953dee

Btw: I think we should have much nicer release notes with batches both for subversions and author contributions. I'll try improving them very soon.

PR #444, opened 2019-01-23
Better 10x read errors

ivirshup:
An addendum to #442, now if you try to read a 10x file and pass a genome it doesn't have, the error tells you which file it was and the genomes it does have:

```python
In [1]: import scanpy as sc 
   ...: sc.read_10x_h5("scanpy/tests/_data/10x_data/1.2.0/filtered_gene_bc_matrices_h5.h5", "not a genome")                                                                          

...

ValueError: Could not find genome 'not a genome' in 'scanpy/tests/_data/10x_data/1.2.0/filtered_gene_bc_matrices_h5.h5'. Available genomes are: ['hg19_chr21']
```

Previous behavior:

```python
In [1]: import scanpy as sc 
   ...: sc.read_10x_h5("scanpy/tests/_data/10x_data/1.2.0/filtered_gene_bc_matrices_h5.h5", "not a genome")                                                                          
NoSuchNodeError                           Traceback (most recent call last)

...

Exception: Genome not a genome does not exist in this file.
```

### DIFF
--- a/scanpy/readwrite.py
+++ b/scanpy/readwrite.py
@@ -132,14 +132,20 @@ def _read_legacy_10x_h5(filename, genome=None):
     """
     with tables.open_file(str(filename), 'r') as f:
         try:
+            children = [x._v_name for x in f.list_nodes(f.root)]
             if not genome:
-                children = [x._v_name for x in f.list_nodes(f.root)]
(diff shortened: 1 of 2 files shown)


falexwolf:
Great!

PR #457, opened 2019-02-03
Updated PyPairs to v3.0.9

rfechtner:
PyPairs v3.0.9 is fully rewritten and will now directly accept AnnData objects. Further it will respect the settings from scanpy regarding verbosity, n_jobs, and output directories. The full documentation can be found at: pypairs.rtfd.io

### DIFF
--- a/scanpy/tools/_pypairs.py
+++ b/scanpy/tools/_pypairs.py
@@ -1,22 +1,14 @@
 """Calculate scores based on relative expression change of maker pairs
 """
 
-import pandas as pd
-
-
 def sandbag(
         adata,
-        categories,
+        annotation,
         gene_names,
         sample_names,
-        fraction=0.5,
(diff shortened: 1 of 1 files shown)


falexwolf:
Looks great!

fidelram:
@falexwolf any update on the plug-ins idea for scanpy?

falexwolf:
The reorganization of using the "external API" (shallow interfaces) via an `import scanpy.external as sce` and the "internal API" as accessible via `import scanpy as sc`, sort of, provided a solution to what bothered people the most: expecting the "internal API" to run through at a single install, be properly maintained etc. and the interfaces to external packages be clearly marked.

I think this is a sustainable, long-term solution, which scales and is convenient for contributors. @flying-sheep agreed as I understood it. Do you think we need more?

Issue #429, opened 2019-01-15
counts_per_cell has no effect on pp.normalize_per_cell(adata)


VolkerBergen:
`counts_per_cell` has no effect on `pp.normalize_per_cell(adata)` as it is overwritten here:

https://github.com/theislab/scanpy/blob/e0d2ea60fa2394f1158618bc69dff4fc639b51c4/scanpy/preprocessing/_simple.py#L599

Would be great if one could use, e.g. the initial cell size (from before filtering). Note, that this issue only holds when passing `AnnData`. With `AnnData.X` it works fine.

falexwolf:
@Koncopd: will this still be an issue in the new implementations of `normalize_total` and `normalize_quantile`? Would be great if you checked!

Koncopd:
@falexwolf 
Hm, new versions don't have `counts_per_cell ` at all.

falexwolf:
@VolkerBergen, is this really important? I had the intend of allowing passing a precomputed  `counts_per_cell` vector, but I think it wasn't really ever used... So, for a simpler function and cleaner code, it would be nice to get rid of it; as @Koncopd did for the new version. Any objections?

VolkerBergen:
Totally fine. Was just wondering whether we can make it work as expected for the deprecated normalize_per_cell module, which is just modifying a single line (PR?).

My use case: I have found that, in the context of RNA velocity analysis, normalizing by initial sizes (cell sizes before filtering) can be quite important. That is where I used to pass the `counts_per_cell` attribute.

Here, I made it the default:
https://github.com/theislab/scvelo/blob/c86e4530485e6e62a055c5a9285177b9554613ce/scvelo/preprocessing/utils.py#L256

falexwolf:
Yes, of course, happy to see a PR! Thank you!

Also, thank you for the details!

Koncopd:
Shoild i add `counts_per_cell` to the new normalization functions?

VolkerBergen:
I find normalising by initial cell sizes quite important. That's the only use case for me where I use `counts_per_cell`. And I believe that applies to many others as well. Thus, one might think of having a bool option such as `by_initial_size` instead. What do you think, @falexwolf ?

In that case, one would have to store the initial cell size before filtering (either when doing the gene filtering or even when reading the data).

falexwolf:
[...]

PR #459, opened 2019-02-04
addressing #429 (counts_per_cell)

VolkerBergen:
to make these two equivalent:
`normalize_per_cell(adata.X, counts_per_cell=counts_per_cell, copy=True)`
`normalize_per_cell(adata, counts_per_cell=counts_per_cell, copy=True).X`


falexwolf:
Thank you! 🙂

Issue #460, opened 2019-02-04
Additional parameters in some functions - new functions


andreatangherloni:
Dear all,

I am writing to ask you some other functionalities.
I have just moved from Seurat to Scanpy and I am finding Scanpy a very nice and well done Python package. 

1. I wrote a function to show the 3D plot of the UMAP, tSNE and PCA spaces. In the `scanpy.tl.tsne` function is not possible to change the number of components, it calculates only the first two components, even if the `scanpy.pl.tsne` function has a parameter `component`. May you add a parameter like the `n_components` of the `scanpy.tl.umap` function?

2. In the `rank_genes_groups` function the log2FC values are provided only for ‘t-test’ based methods. May you return the log2FC values (maybe named log2FC) for all the implemented statistical methods?

3. I think that two parameters in the `rank_genes_groups` function should be added.
    - `min_pCells` to test only the genes that are detected in a minimum fraction of cells of either of the two populations (e.g., cluster 0 vs rest). For instance, min_pCells=0.3 means that at least 30% of the cells must express that gene.
   - `positive`, if it is True, the function should return only positive marker genes for each population.

4. A function showing the volcano plots (based on the log2FC) can help (I can write it if the log2FC values are provided).

Thank you in advance.
Best,
Andrea

falexwolf:
Thank you for the constructive wish list! All of it sounds doable. Would you feel comfortable with adding this functionality yourself? A pull request would be welcome!

andreatangherloni:
Thank you for the prompt reply,  I am going to open a pull request.
I can work on points 1 and 4. May you help me with the other 2 points?

Thanks!

MichaelPeibo:
@andrea-tango 
Really awesome!
I am also wondering to find some parameters to tune in scanpy's `rank_genes_groups` like in Seurat. 

Because I found there is some difference in makers by scanpy's default(using `wilcoxon` ) and Seurat's default parameters` only.pos = TRUE, min.pct = 0.25, logfc.threshold = 0.25`. (Seurat's default method is wilcoxon), in this case, I can find interesting markers calculated by Seurat but not in Scanpy's.

However, when I tried scanpy's `logreg` method, I found many overlap DEGs between two calculations, aka, scanpy's `logreg` and Seurat's `wilcox`.

Have you ever came into similar results?

falexwolf:
Interesting, @MichaelPeibo! Would you share these results somewhere publicly? A notebook on GitHub?

[...]

PR #461, opened 2019-02-05
Additional parameters in some functions - new functions #460

andreatangherloni:


andreatangherloni:
I tested the code, now it is possible to calculate more components for the tSNE embedding method

falexwolf:
These are a lot of different things in one PR.

I'm providing some comments in the code.

scanpy/utils.py
```diff
@@ -667,8 +667,22 @@ def select_groups(adata, groups_order_subset='all', key='groups'):
     if groups_order_subset != 'all':
         groups_ids = []
         for name in groups_order_subset:
-            groups_ids.append(
-                np.where(adata.obs[key].cat.categories.values == name)[0][0])
+            # deal with case that all group names were originally integers,
```
falexwolf, line 670: All of this is a lot of code that doesn't handle Scanpy's deprecation of non-string categories in a good way. So, I don't like to merge this.

Instead, we should print a warning if someone operates with non-string categories. If you just make it a warning, here, `logg.warn`, that's fine.

falexwolf:
OK, wasn't too much:
- happy to merge a PR around tSNE
- there is a PR on the wilcoxon issue, here: https://github.com/theislab/scanpy/pull/519 I know it came later than yours (the result of my sickness and absence here...), but I guess @a-munoz-rojas, who contributed a lot to rank_genes_groups already should make these changes

andreatangherloni:
@falexwolf many thanks for the work you are doing.

falexwolf:
Thank you for contributing, @andrea-tango!

PR #462, opened 2019-02-06
Update numba usage in qc metrics

ivirshup:
This is a small update to use numba more effectively and slightly decrease test times for calculating qc metrics.

* I've enabled no python mode for `top_segment_proportions_sparse_csr`
* I've removed `numba` from functions currently only used for testing
  * This mainly reduces test time. If anyone wants to use the `top_proportions` function to make `plotScater` type plots, maybe these should get re-enabled.

Test times are still not great, but since it's due to numba compilation I'm not sure much can be done about it. The ideal solution is it becoming possible to have numba functions which are both parallel and cached.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,4 @@ patsy
 networkx
 natsort
 joblib
-numba>=0.40.0
+numba>=0.41.0
--- a/scanpy/preprocessing/_qc.py
+++ b/scanpy/preprocessing/_qc.py
@@ -157,7 +157,7 @@ def top_proportions(mtx, n):
         if not isspmatrix_csr(mtx):
             mtx = csr_matrix(mtx)
         # Allowing numba to do more
-        return top_proportions_sparse_csr(mtx.data, mtx.indptr, n)
+        return top_proportions_sparse_csr(mtx.data, mtx.indptr, np.array(n))
     else:
         return top_proportions_dense(mtx, n)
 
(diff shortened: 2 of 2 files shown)


flying-sheep:
Thank you! I think we should only use `@njit` anyway. I don’t understand why `@jit` exists if it can silently fail. Could you please elaborate on the following?

> The ideal solution is it becoming possible to have numba functions which are both parallel and cached.

So am I deducing correctly that numba can parallelize code and usually caches functions to reduce compilation times, but can’t do both for the same function yet?

ivirshup:
`@jit` can be fine for supporting a greater range of `numba` versions or just compiling parts of the function through lifted loops (which this was using before).

I don't think caching is on by default, but you can cache compiled functions to reduce compilation times ([docs](https://numba.pydata.org/numba-doc/dev/user/jit.html#cache)). However, I don't think you can use `@jit(parallel=True, cached=True)`. Here's an issue for it: https://github.com/numba/numba/issues/2712

flying-sheep:
OK got it! I still think `@jit` is too opaque –  how should you know that some innocent-looking change results in a loop no longer being compiled? I think we should use `@njit` to be sure we have compiled performance-critical parts going forward.

ivirshup:
I believe numba will always throw a warning if some part of the requested compilation failed. We could add tests for compilation based on this.

ivirshup:
Just checking in, do I need to do anything else for this PR before it can be merged?

flying-sheep:
No, looks good and we test the QC metrics.

Issue #469, opened 2019-02-08
Wilcoxon in rank_gene_group ignores genes on chunk borders


ahsv:
Using `method=wilcoxon` in the `rank_genes_groups` function chunks the genes into groups (to avoid using too much memory).  The scores for the genes on the endpoints of these chunks are never computed, however.

This can easily be observed by looking at a dataset containing 5000 genes (so that the chunk size is 2000) and setting `rankby_abs=True`.  The genes with indices 2000, 4000, 6000, .... will all be selected as marker genes with p-values of 0.  This is due to the fact that the rank-sums for these genes are left at 0 (they are never computed), which is then judged to be EXTREMELY significant.  

I will submit a PR with a fix for this in a minute :)

ivirshup:
@ahsv and @a-munoz-rojas, has this been fixed? It looks like you've fixed it, but I haven't had a chance to check yet.

falexwolf:
Yes, it should be fixed!

ivirshup:
Great!

PR #471, opened 2019-02-08
Add fixes for bugs in rank_genes_groups

### LINKED ISSUE #470: rankby_abs does nothing for logistic regression in rank_genes_groups
The `rankby_abs` parameter is ignored for `method=logreg` in the `rank_genes_groups` function.

ahsv:
This includes fixes for both #469 and #470 

#469 was a small indexing error

To fix #470, a `rankby_abs` check is included in the `logreg` section of the method that mirrors the `rankby_abs` checks in the other two methods.

This PR additionally updates `select_groups` function in `scanpy/utils.py.`  I was having some issues when the clusters that I was using were labelled by integers (i.e. when `adata.obs[key].cat.catagories.values.dtype` was some form of integer) AND when I was looking at a subset of the clusters (e.g. `groups=[0,1]`, not when `groups='all'`).  At the start of the `rank_genes_groups` function, these cluster labels are converted into strings in the `groups_order` variable.  In the `select_groups` function (line 667 of the original utils.py file), however, we call 
``` 
np.where(adata.obs[key].cat.categories.values == name)[0][0]
```
which fails with an error (since `name` is a string from `select_groups`  and the elements of `adata.obs[key].cat.categories.values` are integers).  Thus, this PR includes a check for the `dtype` of `adata.obs[key].cat.categories.values` - if it is numeric, we instead look at 
``` 
np.where(adata.obs[key].cat.categories.values == float(name))[0][0]
```

This error should only appear if the cluster labels are integers (since this is the only time that the cluster labels are converted to strings for `groups_order` in `rank_genes_groups`) but the above fix should also work if the cluster labels are any floating point numbers (just in case the `rank_genes_groups` is ever generalized in this way).

([Here](https://docs.scipy.org/doc/numpy-1.13.0/reference/arrays.scalars.html) is a link to the numpy type hierarchy)

Edit: added a line number

### DIFF
--- a/scanpy/tools/_rank_genes_groups.py
+++ b/scanpy/tools/_rank_genes_groups.py
@@ -233,8 +233,9 @@ def rank_genes_groups(
                 scores = scores_all[0]
             else:
                 scores = scores_all[igroup]
-            partition = np.argpartition(scores, -n_genes_user)[-n_genes_user:]
-            partial_indices = np.argsort(scores[partition])[::-1]
+            scores_sort = np.abs(scores) if rankby_abs else scores
+            partition = np.argpartition(scores_sort, -n_genes_user)[-n_genes_user:]
+            partial_indices = np.argsort(scores_sort[partition])[::-1]
             global_indices = reference_indices[partition][partial_indices]
             rankings_gene_scores.append(scores[global_indices])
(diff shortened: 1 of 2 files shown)


falexwolf:
What I wrote for https://github.com/theislab/scanpy/pull/461 also applies here. Sorry for the long absence...

Issue #476, opened 2019-02-11
UMAP color pallet when plotting a gene


tsotnech:
Hi guys, 

I can't change the color palette when plotting a gene with sc.pl.umap functions. 
if I plot louvain or certain categories, it does change the color scheme
e.g.
`sc.pl.umap(adata, color=['louvain'], palette='tab20') `

but If I plot a gene 
`sc.pl.umap(adata, color=['NANOG'], palette='plasma')`

it doesn't change the color scheme and I always get 'viridis' color palette. 

Is there any way I can adjust this? 

Thanks a lot,

Tsotne

aditisk:
@tsotnech I ran into the same issue a few months ago but found a fix by using the sc.pl.scatter() function instead. I have pasted the command here and it worked for me.

from matplotlib import pyplot as plt
sc.pl.scatter(adata,color='NANOG',legend_loc='none',color_map=plt.cm.Reds,size=5,basis='umap')

Hopefully this works for you too.
-Aditi

tsotnech:
awesome, thanks a lot @aditisk, that works well

ivirshup:
Continuous color schemes are given with the `color_map` argument, categorical schemes are given with `palette`. All the scatter plots (`scatter`, `pca`, `tsne`, `umap`, etc...) share these arguments.  Here's an example:

```python
import scanpy as sc
import matplotlib as mpl
adata = sc.datasets.pbmc68k_reduced()
sc.pl.umap(adata, color=["louvain", "HES4"])
sc.pl.umap(adata, color=["louvain", "HES4"], palette="Set2", color_map=mpl.cm.Reds)
```

It's not that clearly documented for `umap`, and is pretty easy to miss.

jlehrer1:
This is still not documented.

PR #477, opened 2019-02-12
Document color_map argument for scatter plots

ivirshup:
Documenting `color_map` argument for scatter plots. This should reduce confusion about how to provide a continuous palette (#476).

I'd also be up for having the arguments have names like `cont_palette` and `cat_palette`, which could be more clear.

### DIFF
--- a/scanpy/plotting/_docs.py
+++ b/scanpy/plotting/_docs.py
@@ -54,11 +54,14 @@ size : `float` (default: `None`)
 color_map : `matplotlib.colors.Colormap` or `str`, optional (default: None)
     Color map to use for continous variables. Anything that works for `cmap`
     argument of `pyplot.scatter` should work here (e.g. `"magma"`, `"viridis"`,
-    `mpl.cm.cividis`).
+    `mpl.cm.cividis`). If `None` value of `mpl.rcParams["image.cmap"]` is used.
 palette : `str`, list of `str`, or `Cycler` optional (default: `None`)
     Colors to use for plotting categorical annotation groups. The palette can be
     a valid `matplotlib.pyplot.colormap` name like `'Set2'` or `'tab20'`, a list
-    of colors like `['red', '#ccdd11', (0.1, 0.2, 1)]` or a Cycler object.
+    of colors like `['red', '#ccdd11', (0.1, 0.2, 1)]` or a Cycler object. If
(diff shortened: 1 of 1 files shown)


flying-sheep:
Could you please also mention what the default of `None` does? If I recall correctly:

- `color_map`: the default is `rcParams['image.cmap']`
- `palette`: the default are the colors in `rcParams['axes.prop_cycle']`

ivirshup:
It sounds like you have a better idea than me. But I'll give it a shot. How does this look?

```
color_map : `matplotlib.colors.Colormap` or `str`, optional (default: None)
    Color map to use for continous variables. Anything that works for `cmap`
    argument of `pyplot.scatter` should work here (e.g. `"magma"`, `"viridis"`,
    `mpl.cm.cividis`). If `None` value of `mpl.rcParams["image.cmap"]` is used.
palette : `str`, list of `str`, or `Cycler` optional (default: `None`)
    Colors to use for plotting categorical annotation groups. The palette can be
    a valid `matplotlib.pyplot.colormap` name like `'Set2'` or `'tab20'`, a list
    of colors like `['red', '#ccdd11', (0.1, 0.2, 1)]` or a Cycler object.
    If `None`, `mpl.rcParams["axes.prop_cycle"]` is used unless categorical
    variable already has colors stored in `adata.uns["{var}_colors"]`.
```

I could maybe also mention that passing an argument for palette overwrites the values in `adata.uns["{var}_colors"]`?

flying-sheep:
Great! Yes, that’s probably all very helpful!

fidelram:
@flying-sheep @ivirshup  The default continuous color map is usually set using `sc.set_figure_params` in the example tutorials. I am not totally sure but I think that internally the `rcParams` are modified. Maybe you can also mention this on the improved documentation.

flying-sheep:
Sure! the docs `set_figure_params` should explain that the basically just modify the `rcParams`.

ivirshup:
I'm of two minds how to document that. Either every plotting function gets a note saying that default values for some arguments can be changed globally with `sc.set_figure_params`, or add that note to the top of the plotting docs on the website.

flying-sheep:
Hmm, the docs for `set_figure_params` [already mention](https://scanpy.readthedocs.io/en/latest/api/index.html#settings) `rcParams`, but maybe it should be mentioned directly in the docs for `scanpy.plotting`, what do you think?

PR #494, opened 2019-02-19
[Bugfix] I had forgotten what `all(dataframe)` does

ivirshup:
Just spent an hour re-figuring out that `all(dataframe)` does `all(bool(c) for c in dataframe.columns)` 😢. Now finding out where I've written that.


flying-sheep:
Great! so that test succeeded because it was testing the wrong thing?

ivirshup:
It would've succeeded anyways, it just wasn't testing the right thing

flying-sheep:
yep, that’s what I thought. Thank you

PR #503, opened 2019-02-25
Harmony framework for connecting scRNA-seq data from discrete time points

awnimo:
@falexwolf
I'm adding a wrapper for Harmony by [Setty et al. (2018)](https://doi.org/10.1101/471078)
Please let me know if you have any comments


wflynny:
Perhaps the name should be more specific to avoid confusion with [harmony package for batch correction](https://github.com/immunogenomics/harmony)?  That code has a C++ backend for which a Python front end might be written.

falexwolf:
Thank you! The same comments as for https://github.com/theislab/scanpy/pull/493 apply.

falexwolf:
Agreed on the name. It is very generic. Could you make it `harmony_something`? Also, can, you briefly explain what harmony is? I know palantir, but haven't heard about harmony. Also, this seems very recent research.

awnimo:
hi @falexwolf,
Is there a reason why this branch not merged yet?
Thank you!

ivirshup:
Hey @awnimo, I don't think we realized that this was ready for review again. Sorry about the confusion!

To me, this looks pretty close to ready. Just a few things to address:

* Could you add a test, in a new module `test_external.py`? That way we can know if some change we make would break this.
* If a list of csv files are valid input, could you document some more about the format of the files? E.g. do you expect cells to be rows or columns?
* Any thoughts on solutions for the name collision?

Thanks!

awnimo:
> Hey @awnimo, I don't think we realized that this was ready for review again. Sorry about the confusion!
> 
> To me, this looks pretty close to ready. Just a few things to address:
> 
> * Could you add a test, in a new module `test_external.py`? That way we can know if some change we make would break this.
> * If a list of csv files are valid input, could you document some more about the format of the files? E.g. do you expect cells to be rows or columns?
> * Any thoughts on solutions for the name collision?
> 
> Thanks!

Hey @ivirshup, I created a test module as requested. I also documented more details on the format of csv files. With regard to the name I spoke with the my PI, Dr. Dana Pe'er, and the author of the paper, Dr. Manu Setty, and we decided to keep the name as is. The publication can be found here https://www.nature.com/articles/s41586-019-1127-1
Please let me know if there is anything else needed to merge.
Thank you!

awnimo:
hi @falexwolf and @ivirshup,
Please let me know if there is any further actions need to be done in order to get this PR merged with Scanpy.
Please review my recent response 8 days ago, as well.
Thank you!

awnimo:
hi @falexwolf and @ivirshup,
Please let me know if there is any further actions need to be done in order to get this PR merged with Scanpy.
Please review my recent response 8 days ago, as well.
Thank you!

LuckyMD:
I wonder if we need to find a canonical way of dealing with name collisions. "First come, first served" might not work well for long term cases. I thought I read on Twitter that one of the harmony frameworks had decided to change name?

ivirshup (changes requested):
Thanks for the updates! A few points to address:

# Tests

First, CI won't run while this branch conflicts with master. Could you resolve those conflicts so we can make sure the tests pass?

Additionally, the test data is a little large (roughly the size of the entire repository). Could you cut that down by:

* Reducing the size of the example
* Compressing the example files
* And removing the large files from this PR's commit history

I'd aim for a total size of around 1mb.

# Documentation

Have you checked if the documentation renders the way you want it to? I noticed the parameters were moved to an internal function's doc-string for palantir. I don't think they'll show up in the rendered docs.

# Naming convention

I'm not asking that the name of the method change, but I'd like it to be clear which of the Harmony methods this is. Would it be alright if this function was called `harmony_timeseries` or something to that effect?

# Formatting

Before this merges, could you run `black` over the files you've edited?

PR #508, opened 2019-03-01
Replace `is` for `==` in string testing.

gamazeps:
Using `is` in a test, tests for identity and not equality, this leads to
inconsistent behaviours when testing for equality with strings.

See https://stackoverflow.com/questions/1504717/why-does-comparing-strings-using-either-or-is-sometimes-produce-a-differe


flying-sheep:
Oh wow, yes. @tcallies I hope you didn’t do that since, it’s 100% a bug.

PR #512, opened 2019-03-03
Add row or column scaling option to matrixplot

gokceneraslan:
Especially when we visualize raw counts, sometimes it's hard to see the differences between the expression of one gene across cell types in the heatmap since one value can simply dominate the dynamic range of expression. I think we can add a scaling option to matrixplot, which squashes expression values between 0 and 1 to make markers more pronounced.

Heatmap of the raw values:

image

Heatmap of the logarithmized values (which also helps a bit but not for all genes):

image

Heatmap of the col-normalized values:

image

PS: The option is actually borrowed from Seaborn (https://seaborn.pydata.org/generated/seaborn.clustermap.html). 

PPS: There is an edge case such as division by zero. Also, `swap_axes` option makes 'row'/'col' naming a bit confusing. Let me know if you have suggestions about these or the standardization idea in general.


falexwolf:
Happy to merge this if @fidelram is happy with it, too! :)

scanpy/plotting/_anndata.py
```diff
@@ -1623,6 +1627,17 @@ def matrixplot(adata, var_names, groupby=None, use_raw=None, log=False, num_cate
 
     mean_obs = obs_tidy.groupby(level=0).mean()
```
fidelram, line 1633: To avoid confusion of what is row and what is column I can think of two solutions. One is to duplicate the code in each of the large swap blocks. The other is to add here something like:
```PYTHON
if swap_axes:
    standandar_scale = 'row' if standard_scale == 'col' else 'col'
```

fidelram:
This is very useful! Thanks a lot.

Would you mind adding a test in `scanpy/tests/test_plotting.py`?

scanpy/plotting/_anndata.py
gokceneraslan: OK, I think second way is a bit cleaner. Let me know if it's ok now.

gokceneraslan:
I was looking at the tests, and something made me think. If `use_raw=False` is given, we use a diverging colormap on purpose, right? Like this:

image

Now with the new standardization option, values are squashed between 0 and 1 but the color scale is still diverging:

image

@fidelram Do you think that's ok, or should we switch back to viridis when `standard_scale` is given, regardless of `use_raw` state?

gokceneraslan:
Added tests, too. (Tests fail in my setup, though. RMS is usually around 50-60, rather than 15.)

scanpy/plotting/_anndata.py
fidelram: Actually, what may be better is to set the options for standar_scale as `category` or `var` as this is always clear what to scale no matter the orientation. What do you think?

fidelram:
@gokceneraslan Tests with images are always problematic. I will take a look later. With respect to `use_raw=False`, what happens is that usually, `adata.X`  is scaled and centered around zero. Thus, a default diverging scale was used with parameters `vmin=-3, vmax=3, cmap='bwr'`. This has been useful to me, but if data is being processed differently, this could be confusing. What is your opinion?

scanpy/plotting/_anndata.py
gokceneraslan: This is a great idea! Now I changed it to var and group.

gokceneraslan:
> @gokceneraslan Tests with images are always problematic. I will take a look later. With respect to `use_raw=False`, what happens is that usually, `adata.X` is scaled and centered around zero. Thus, a default diverging scale was used with parameters `vmin=-3, vmax=3, cmap='bwr'`. This has been useful to me, but if data is being processed differently, this could be confusing. What is your opinion?

Default values of vmin/vmax looks ok for most datasets, but when we squash everything between 0 and 1, the negative scale doesn't make sense any more. Furthermore, even for max values (i.e. 1.0) colors will look dimmer than the color representing the max value. For example:

```python
sc.pl.rank_genes_groups_matrixplot(adata, n_genes=1, standard_scale='var', use_raw=False)
```

generates

image

Color scale might be a bit confusing for some, IMHO. I think it makes more sense to switch back to default sequential colormap if `standard_scale in ('var', 'group')`, but it's totally up to you :) 

We can merge this and play around with different datasets etc, too.

gokceneraslan:
Any thoughts @fidelram?

fidelram:
I think we should remove the that uses the divergent colormap and keep things simpler.

gokceneraslan:
@fidelram Shall I do it in this PR or shall we merge this and address it in another PR?

fidelram:
Let me see if I can make changes on top of your changes. But I think I need to make a new PR.

Issue #517, opened 2019-03-06
Log2 fold changes in rank_genes_groups are calculated from log-transformed data


a-munoz-rojas:
Hi all. I was looking through the `_rank_genes_groups` function and noticed that the fold-change calculations are based on the means calculated by `_get_mean_var`. The only problem with this is that (usually) the expression values at this point in the analysis are in log scale, so we are calculating the fold-changes of the log1p count values, and then further log2 transforming these fold changes. I know that different programs do it differently, but I think it's more intuitive to convert the matrix back to counts, calculate the fold change, and then report the log2 fold change. Any thoughts?

For the actual differential testing, I think it's ok to run the tests on the log1p transformed data, as that seems to be the norm for many pipelines using the types of tests we are using. However, some pipelines do use raw count data, which might be interesting to implement if we want. Either way, I think it's a little unintuitive to report a log2 fold change of log expression values. I can submit a pull request to implement this if this is something you agree with, and can add a parameter to let the user decide whether to use log-transformed or raw-count data. Let me know what you think!

LuckyMD:
I second the notion of double logs not making sense. You could use differences in mean expression on the log-scale to give log-fold changes though I think.

If you use count data, you should probably use an NB model (or at least Poisson). The t-test assumes normality, which is at least approximated by log-normalized data.

a-munoz-rojas:
@LuckyMD Yeah, we can definitely do the differences in mean expression on the log-scale as a quick fix. However, I think it's a little bit more intuitive to express fold-changes with a base of 2, as opposed to a base of e, which is the transformation used for the data matrix. This also mimics what other packages usually report, but I'm happy either way!

For the differential testing itself, I completely agree with you about the assumption of normality with t-tests, so we definitely shouldn't change that. For the wilcoxon test, however, normality isn't assumed, so do you think it would be ok to run on raw counts?

Either way, we should definitely fix the double-log for the fold-change reporting. Is there a preference on whether to keep it was loge(base e) difference or log2 difference? For simplicity, this change in scale would only affect the fold-change reporting, all other tests I think we should keep in loge to keep it consistent.

One final thought I wanted to ask your opinion on: log-mean vs mean-log. (e.g. log(mean(count)) vs mean(log(count)). For t-tests I think it makes sense to use mean-log, however, when calculating fold-change differences in mean expression, I'm not entirely sure what's preferred (particularly for methods like wilcoxon that don't use the mean expression to calculate scores). Thoughts?

[...]

PR #519, opened 2019-03-07
Fix for double log in fold-changes and for indexing in wilcoxon test

a-munoz-rojas:
This is a fix for the issue discussed in #517. I implemented it by adding a new parameter to indicate whether the data in `adata.X` is log-transformed or not, and set it to True as a default, assuming most users will have log-transomed data at this point. This fix is assuming the transformation was done with log1p, as in `sc.pp.log1p.` The matrix is transformed back to raw counts only for the fold-change calculations - everything else is left the same to maintain the tests in log-space.

I also included a fix for a bug in the indexing that was implemented in the original chunking for the wilcoxon test. This was noted in #469 and the same fix is implemented in PR #471 - thanks for catching that bug!


falexwolf:
Hey @a-munoz-rojas and @LuckyMD! Thank you for this!

However, I think this adds a lot of code and computational burden. I'd prefer to make an approximation, which is similar to interchanging log mean with mean log.

`x1` and `x2` are two arrays storing count values.
```
log2FC(x1, x2) = log2(mean(x1) / mean(x2)) = log2(mean(x1)) - log2(mean(x2)) = (log(mean(x1)) - log(mean(x1))) / log(2)
```
In Scanpy, we typically store `log1p(x)` in the data matrix. For the tests, we have already computed `mean(log1p(x))`. Hence, the following expression is a one-line edit. 
```
log2FCapprox = log2((expm1(mean(log1p(x1))) / expm1(mean(log1p(x2))))
```
Obviously, this is an approximation; high values in `x` are down-weighted by virtue of computing a mean that's weighted with a `log1p`. The resulting `mean(log1p(x))` is therefore rigorously smaller than `log1p(mean(x))`; if you don't have a lot of outliers, then there will be little difference. If there is a lot of variance, there will be a difference. As one has the same effect both in the numerator and denominator, some of this will cancel out. Hence, one gets an estimator for "effect strength" (after all, this is what we are interested in, that is very similar to `log2FC`, but more robust to outliers.

I'd prefer this, I have to admit, together with a note in the docs.

LuckyMD:
I reckon that's a fair consideration. In the end we don't use `sc.tl.rank_genes_groups()` for complex DE tests that require this amount of detail, but instead for marker gene calculations where it's mainly about ranking genes. It would be interesting to see the error of the approximation though.

a-munoz-rojas:
Hey @falexwolf, thanks for your note. I apologize for the late reply - I was away at a conference and had little time to work on this. 
I agree that this is a cleaner way of doing this and will be more robust to outliers, even if it's at the expense of the approximation. Like @LuckyMD mentioned, this function is generally used for more basic differential testing, so further exploration of differential expression can rely on more complicated downstream analysis. Out of curiosity, I'll also calculate what the error of the approximation is, just to have an idea.

I'm a little limited in bandwidth at the moment, but this is an easy change that I should be able to implement this week. Sorry for the delay - I'll submit the changes soon!

falexwolf:
Thank you very much, @a-munoz-rojas!

a-munoz-rojas:
Hey, sorry for the delayed response. I just pushed the new version with the changes we discussed. I also checked what the error looked like - it's pretty close for most of the cases (I'm attaching a plot with estimated (mean-log) vs "actual" (log-mean) fold changes - as expected, only in the extremes do we start to get a little bit of an underestimation. In my dataset, I got a mean error of about 3%, with the largest error seen in genes with extreme outliers as we expected (the attached violin plot has the largest error, ~70% lower.  
[test_meanlogerror.pdf](https://github.com/theislab/scanpy/files/3000658/test_meanlogerror.pdf)
[violintest_outlier.pdf](https://github.com/theislab/scanpy/files/3000659/violintest_outlier.pdf)

falexwolf:
Thank you, this looks good! I'm merging it. :)

gokceneraslan:
@a-munoz-rojas I'm getting the following warning:

```
/home/user/Code/scanpy/scanpy/tools/_rank_genes_groups.py:203: RuntimeWarning: overflow encountered in expm1
  foldchanges = (np.expm1(means[igroup]) + 1e-9) / (np.expm1(mean_rest) + 1e-9) #add small value to remove 0's
```

Any ideas?

gokceneraslan:
Here is another one:

```
/home/user/Code/scanpy/scanpy/tools/_rank_genes_groups.py:203: RuntimeWarning: invalid value encountered in true_divide
  foldchanges = (np.expm1(means[igroup]) + 1e-9) / (np.expm1(mean_rest) + 1e-9) #add small value to remove 0's
```

gokceneraslan:
Also `log_transformed=True` parameter is not listed in the documentation.

gokceneraslan:
Wait, where exactly is `log_transformed` used in the code?

falexwolf:
`log_transformed` disappeared in the last commit here...

Better: `pp.log1p` should write an attribute to `.uns`, say simply `.uns['log1p'] = True`. Depending on that attribute, log2fc is computed by rexponaniating or not.

Also: If trying to call a t-test with non-logarithmized data, a warning should be written.

The overflow and 0 warnings: are you sure you used logarithmized data, Gökcen?

gokceneraslan:
> Better: `pp.log1p` should write an attribute to `.uns`, say simply `.uns['log1p'] = True`. Depending on that attribute, log2fc is computed by rexponaniating or not.

Fully agree.

> 
> Also: If trying to call a t-test with non-logarithmized data, a warning should be written.
> 

Also agree.

> The overflow and 0 warnings: are you sure you used logarithmized data, Gökcen?

Oh true, it wasn't log transformed, true.

falexwolf:
Ok, good to read that it wasn't log-transformed!

@Koncopd, could you quickly implement these simple changes? Before continuing to work on the UMAP? These simple changes are for 1.4.1, the UMAP will be for 1.5. Thank you!

PR #524, opened 2019-03-08
Add `smallest_dot` parameter to dotplot

sjfleming:
Allow greater customizability and visibility when using dotplot by allowing the user to set the smallest dot size.  This change updates the dots along with the dot size legend.


falexwolf:
@fidelram, are you happy with this? If you agree, I would also just make you a maintainer of Scanpy, so that you can merge PRs related to the plotting API yourself. :)

fidelram:
Thanks for this PR @sjfleming . If you don't mind I will integrate this functionality into a new PR that also covers #512 and #525

falexwolf:
Merged through https://github.com/theislab/scanpy/pull/528.

gokceneraslan:
Does this feature work with the latest scanpy? I don't see any change when I specify different values for the smallest_dot option.

PR #525, opened 2019-03-10
Fix stacked_violin plot bug

### LINKED ISSUE #405: sc.pl.stacked_violin:   IndexError, list index out of range
Scanpy vs. 1.3.6
installed using pip3
OSX 10.10.5
Jupyter lab

code:
`list_of_list_of_marker_genes = [mg1, mg2, mg3]
for mg in list_of_list_of_marker_genes:
    sc.pl.stacked_violin(adata, mg, groupby = 'louvain’, rotation=90)
    print(mg)`

This works for some mg's but not for all; sc.pl.matrixplot works for all of them; the same for sc.pl.violin; it works also if I would combine all marker genes into one list and then run
sc.pl.stacked_violin. When I say it does not work: it typically generates one stacked plot and then
'hangs up' with an error message for subsequent plots:

Error message:

IndexError: list index out of range

When I run sc.pl.stacked_violin for the individual mg’s, some work others don’t. It is a reproducible results.

I explicitly tested whether the genes in the respective marker gene lists are present in ’n_cells’. If not, I update the lists.

What am I missing here?

outlace:
`num_rows` was incorrectly calculated when `swap_axes=False` in the `stacked_violin` plot. This fixes issue #405

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -878,7 +878,7 @@ def stacked_violin(adata, var_names, groupby=None, log=False, use_raw=None, num_
         else:
             width, height = figsize
 
-        num_rows = len(categories)
+        num_rows = len(var_names)
         height_ratios = None
         if has_var_groups:
             # add some space in case 'brackets' want to be plotted on top of the image


falexwolf:
Thank you for this! Unfortunately, it breaks the tests. It's such a small change, but I guess @fidelram had something in mind when setting it the way he did...

outlace:
I’ll have another look. But it doesn’t work the way it is now either

fidelram:
@outlace I will check the problem with the test and integrate your changes in a new PR that addresses #512 and #524 if this is OK with you.

outlace:
@fidelram Sounds good!

fidelram:
@outlace Curiously, your change causes an error. Without your change I can run the tests correctly without a problem. I remember that I fixed a bug similar to this one that was recently integrated into master (see https://github.com/theislab/scanpy/pull/425/files#diff-b5175ed1415cdbf853646e523cbe8ae0L902). Could it be that you didn't have the latest pull from scanpy and that was causing the error?

outlace:
Possible, I will try pulling the latest master and see if I still get the problem

fidelram:
@outlace do you think I can close this PR?

ivirshup:
@fidelram, did that PR happen? I assume #405 got fixed but am not actually sure.

fidelram:
yes 405 got fixed. The code was moved to `_stacked_violin.py`

ivirshup:
Great, thanks!

PR #527, opened 2019-03-11
Quick fix for typo

ivirshup:
Looks like there was a typo at the bottom of `scanpy/preprocessing/_dca.py` that was causing a parse error. This should fix it.


flying-sheep:
Thank you!

PR #538, opened 2019-03-19
Add an explicit expression threshold to dotplot

gokceneraslan:
This also adds an option to calculate the mean only over the cells expressing given genes. I think these options give users more flexibility to binarize expression. For example, for z-score normalized data, a combination of `expression_threshold=-1` and `use_raw=False` binarizes the expression using a cutoff value corresponding to one sd away from the mean.

Also the `obs_bool = obs_tidy.astype(bool)` was erroneous for scaled data without the .raw since both positive and negative values were mapped to True.

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -1391,6 +1391,11 @@ def dotplot(adata, var_names, groupby=None, use_raw=None, log=False, num_categor
     # transform obs_tidy into boolean matrix using the expression_cutoff
     obs_bool = obs_tidy > expression_cutoff
 
+    # compute the sum per group which in the boolean matrix this is the number
+    # of values > expression_cutoff, and divide the result by the total number of values
+    # in the group (given by `count()`)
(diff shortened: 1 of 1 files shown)


falexwolf:
@fidelram; I gave you write access to the repo so that you can quickly merge PRs like this to your plotting API. It would still be great if bigger chunks of new functionality continued to come in PRs, which you could then merge yourself upon a brief check by any of us. Does that sound right?

fidelram:
@gokceneraslan Thanks for the PR. I hesitated in the past to add such functionality as this makes the definition of expression very *ad hoc*. However, I also noticed the pitfalls when you have only normalized data and the dotplot does not makes sense. Are you planning any further changes?

@falexwolf Thanks for giving me write access. I will use the new status responsible.

gokceneraslan:
@fidelram Thanks, no further changes. I agree about the arbitrariness but I think it's good to provide a configurable implementation with reasonable defaults. This, I believe, is also what we do in other functions like pp.neighbors and pp.highly_variable_genes.

falexwolf:
Cool! Great to have you with write access, @fidelram! :smile:

PR #539, opened 2019-03-19
Enable autocompletion for scatterplot functions

### LINKED ISSUE #535: Argument tab completion for sc.pl.umap-like functions?
It'd be nice if some level of tab completion of argument names could be maintained for functions like `sc.pl.umap`. This came up recently in #455, and I find myself frequently misspelling (mostly mis-pluralizing) argument names like `color/ colors` and `gene_name/ gene_names`. I've given this a shot using `functools.wraps`, but no luck yet.

Any ideas on if we could do this @flying-sheep?

flying-sheep:
Fixes #535

Sorry for the messy diff, I had to put the API functions after `plot_scatter` since the decorator is called at import time and needs `plot_scatter` to be defined beforehand.

This also fixes the docs for `pca` and adds parameter docs for `arrows_kwds` and `return_fig`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -71,6 +71,7 @@ intersphinx_mapping = dict(
     leidenalg=('https://leidenalg.readthedocs.io/en/latest/', None),
     louvain=('https://louvain-igraph.readthedocs.io/en/latest/', None),
     matplotlib=('https://matplotlib.org/', None),
+    networkx=('https://networkx.github.io/documentation/networkx-1.10/', None),
     numpy=('https://docs.scipy.org/doc/numpy/', None),
     pandas=('http://pandas.pydata.org/pandas-docs/stable/', None),
     python=('https://docs.python.org/3', None),
--- a/scanpy/plotting/_docs.py
+++ b/scanpy/plotting/_docs.py
@@ -3,91 +3,97 @@
 
 
 doc_adata_color_etc = """\
-adata : :class:`~anndata.AnnData`
+adata
     Annotated data matrix.
-color : string or list of strings, optional (default: `None`)
+color
     Keys for annotations of observations/cells or variables/genes, e.g.,
     `'ann1'` or `['ann1', 'ann2']`.
-gene_symbols : string, optional (default: `None`)
+gene_symbols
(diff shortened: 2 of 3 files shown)


falexwolf:
Looks great! Let's wait for the tests to complete and merge!

flying-sheep:
I think putting “basis” first and making it required is a valid change since `plot_scatter` should never be called directly, right?

The only problem with the whole thing is that some parameters (e.g. the `arrows` family) isn’t valid for all functions, but will still autocomplete and be passed on to `plot_scatter`. Changing that is more work though.

falexwolf:
Putting `basis` first is fine as `plot_scatter` is an internal function that doesn't appear in the API.

The `arrows` should evtl appear in all embeddings; so that's fine. All of these scatter functions should have exactly the same arguments. So, it's not the fault of your PR but my fault, as I haven't pulled the change through. Due to `scvelo`, `arrows` might also disappear... So, let's not worry about it.

flying-sheep:
OK, great!

fidelram:
@flying-sheep @falexwolf I ended up calling `plot_scatter` directly, and maybe should be part of the API as this is useful to plot data whose `basis` is none of the predefined ones. 

Recently, I was  storing different data in `.obsm` to reflect different parameters used for example for umap or drawgraph and then plotting with `plot_scatter`.

Other solution would be to add parameters like `key_added` and for the `tool` functions like `pca`, `umap`, `tsne` etc. And a  `key` parameter for the corresponding `plotting` functions.

falexwolf:
Thanks for the note, @fidelram. But, putting `basis` first will still allow you to call `plot_scatter` directly, right? We don't have problem, here.

I see your use case and it's a valid one. Let's discuss about putting `plot_scatter` into the API in another PR, though.

flying-sheep:
I think it should be named differently then.

It expects `adata.obsm['X_' + basis]` to exist and optionally also `adata.obsm['Delta_' + basis]`, so maybe `scatter_obsm` or `scatter_dimred`? Do we have a term for the `X_*` arrays?

falexwolf:
Yes, agreed that we'll want a different name in that case. That's one of the reasons why we want another PR.

PR #540, opened 2019-03-19
fix #536 

### LINKED ISSUE #536: Scatter plot of numerical observation with missing data
I have a dataset for which I have an observation that is only available for some cells. When I make a scatter plot that I color code for this observation not all cells are plotted:
```python
import random
import scanpy as sc

adata = sc.datasets.blobs()
adata.obs['property'] = 630 * [float("nan")] + 10 * [1] 

sc.tl.pca(adata)
sc.pl.pca(adata, color='property', size=50)
```
While this should plot 10 cells it only shows one cell:
image
I can get the plot I want by filtering cells first:
```python
sc.pl.pca(adata[adata.obs['property'] == 1], color='property', size=50)
```
image
Would you agree that scanpy should plot all cells that have a valid observation?

fidelram:
fix a problem when data to plot contains NaNs.

The fix is to return a numpy array instead of a pandas.Series

### DIFF
--- a/scanpy/plotting/_tools/scatterplots.py
+++ b/scanpy/plotting/_tools/scatterplots.py
@@ -651,7 +651,7 @@ def _get_color_values(adata, value_to_plot, groups=None, palette=None, use_raw=F
                 # that are not in the groups
                 color_vector[~adata.obs[value_to_plot].isin(groups)] = "lightgray"
         else:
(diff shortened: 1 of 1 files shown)


falexwolf:
The failing test seems unrelated. I'm merging this. [@fidelram, you could have merged it yourself. :)]

PR #543, opened 2019-03-20
Density calculation and plotting

LuckyMD:
This pull request is for calculating and plotting cell densities on an embedded representation. This is especially useful together with an `.obs` covariate to calculate and visualize cell densities over conditions.

Code is adapted from raw version by @sophietr .

Still work in progress...

### DIFF
--- a/scanpy/plotting/_tools/__init__.py
+++ b/scanpy/plotting/_tools/__init__.py
@@ -646,7 +646,7 @@ def embedding_density(
         group: Optional[str] = None,
         color_map: Union[Colormap, str] = 'YlOrRd',
         bg_dotsize: Optional[int] = 80,
-        fg_dotsize:  Optional[int] = 160,
+        fg_dotsize:  Optional[int] = 180,
         vmax:  Optional[int] = 1,
         vmin:  Optional[int] = 0,
         save: Union[bool, str, None] = None,
@@ -675,7 +675,7 @@ def embedding_density(
         Matplolib color map to use for density plotting.
     bg_dotsize : `int`, optional (default: `80`)
         Dot size for background data points not in the `group`.
-    fg_dotsize : `int`, optional (default: `160`)
+    fg_dotsize : `int`, optional (default: `180`)
         Dot size for foreground data points in the `group`.
     vmax : `int`, optional (default: `1`)
         Density that corresponds to color bar maximum.
@@ -702,7 +702,7 @@ def embedding_density(
         basis = 'draw_graph_fa'
 
     if 'X_'+basis not in adata.obsm_keys():
-        raise ValueError('Cannot find the embedded representation `adata.obsm[X_{!r}]`.'
+        raise ValueError('Cannot find the embedded representation `adata.obsm[X_{!r}]`. '
                          'Compute the embedding first.'.format(basis))
 
     if key not in adata.obs:
@@ -712,18 +712,20 @@ def embedding_density(
         raise ValueError('Please run `sc.tl.embedding_density()` first and specify the correct key.')
 
     if 'components' in kwargs:
-        logg.warn('Components were specified, but will be ignored. Only the'
+        logg.warn('Components were specified, but will be ignored. Only the '
                   'components used to calculate the density can be plotted.')
+        del kwargs['components']
 
     components = adata.uns[key+'_params']['components']
     groupby = adata.uns[key+'_params']['covariate']
 
     if (group is None) and (groupby is not None):
-        raise ValueError('Densities were calculated over an `.obs` covariate.'
+        raise ValueError('Densities were calculated over an `.obs` covariate. '
                          'Please specify a group from this covariate to plot.')
 
     if (group is not None) and (group not in adata.obs[groupby].cat.categories):
-        raise ValueError('Please specify a group from the `.obs` category over which the density was calculated.')
+        raise ValueError('Please specify a group from the `.obs` category over which the density '
+                         'was calculated.')
 
     if (np.min(adata.obs[key]) < 0) or (np.max(adata.obs[key]) > 1):
         raise ValueError('Densities should be scaled between 0 and 1.')
@@ -751,10 +753,17 @@ def embedding_density(
     adata_vis = adata.copy()
     adata_vis.obs['Density'] = dens_values
 
-    # colour transformation is not really working! Probably need to layer the plots...
     norm = colors.Normalize(vmin=vmin, vmax=vmax)
     cmap.set_over('black')
     cmap.set_under('lightgray')
     
+    # Ensure title is blank as default
+    if 'title' not in kwargs:
+        title=""
+    else:
+        title = kwargs.pop('title')
+
     # Plot the graph
-    return plot_scatter(adata_vis, basis, components=components, color='Density', color_map=cmap, norm=norm, size=dot_sizes, vmax=vmax, vmin=vmin, save=save, **kwargs)
+    return plot_scatter(adata_vis, basis, components=components, color='Density',
+                        color_map=cmap, norm=norm, size=dot_sizes, vmax=vmax,
+                        vmin=vmin, save=save, title=title, **kwargs)
--- a/scanpy/plotting/_tools/__init__.py
+++ b/scanpy/plotting/_tools/__init__.py
@@ -685,10 +685,10 @@ def embedding_density(
     if key not in adata.obs:
         raise ValueError('Please run `sc.tl.density()` first and specify the correct key.')
 
-    if key+'_param' not in adata.uns:
+    if key+'_params' not in adata.uns:
         raise ValueError('Please run `sc.tl.density()` first and specify the correct key.')
 
-    if components:
+    if 'components' in kwargs:
         logg.warn('Components were specified, but will be ignored. Only the'
                   'components used to calculate the density can be plotted.')
 
--- a/scanpy/plotting/_tools/__init__.py
+++ b/scanpy/plotting/_tools/__init__.py
@@ -683,7 +683,7 @@ def embedding_density(
 
     components = [0,1]
 
-    if basis == 'dm':
+    if basis == 'diffmap':
         components = [1,2]
 
     if key not in adata.obs:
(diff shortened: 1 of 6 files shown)


falexwolf:
This looks good so far! Thank you!

Can we call it `embedding_density` or something similar? `density` is a little generic; we might have other methods in the future...

Some form of test would also be great! ;)

Let me know when this should be merged.

scanpy/tools/_density.py
```diff
@@ -0,0 +1,139 @@
...
+    adata : :class:`~anndata.AnnData`
+        The annotated data matrix.
+    embedding : `str`
+        The embedding over which the density will be calculated. This must
+        be one of:
+        'umap' : UMAP
+        'dm' : Diffusion map
+        'pca' : PCA
+        'tsne' : t-SNE
+        'fa' : Force-directed graph layout by Force Atlas 2
```
gokceneraslan: I think this is same as the `basis` option we have in other plotting functions (https://github.com/theislab/scanpy/blob/master/scanpy/plotting/_anndata.py#L79). There is no need to hard-code basis, it shouldn't be limited to predefined basis as in here. We can use spatial basis, dca or scvi basis etc., if they're stored in obsm.

You can use the same processing here: https://github.com/theislab/scanpy/blob/master/scanpy/plotting/_anndata.py#L300

There is also a trick to skip the useless dimension of diffmap there.

LuckyMD: This is indeed the same... I should rename the argument basis. And yes, the adaptation for diffmap is very helpful.

In terms of not hard-coding I guess I can just throw an error to saw that `adata.obsm['X_'+basis]` does not exist then?

LuckyMD:
@fidelram I guess you are the right person to ask for help with this... I'm struggling to work nicely with `plot_scatter()`. I am trying to generate a plot where density values for non-selected conditions are grey, while density values for the selected condition are on 'YlOrRd' or another color map. It seems this is not ideal with a single `plot_scatter()` call (which I was hoping to use as the facet wrapping is already done there). For the grey values I am using a color value of -1, while the others are between 0 and 1. However, when I define a color map that is symmetric around 0, positive values near 0 are mapped to grey instead of colours... any idea why?

scanpy/tools/_density.py
gokceneraslan: Sounds good.

scanpy/plotting/_tools/__init__.py
```diff
@@ -633,3 +636,110 @@ def sim(adata, tmax_realization=None, as_heatmap=False, shuffle=False,
...
+    """
+    sanitize_anndata(adata)
+    
+    # Test user inputs
+    basis = basis.lower()
+
+    if basis == 'fa':
+        basis = 'draw_graph_fa'
+
+    if 'X_'+basis not in adata.obsm.dtype.names:
```
gokceneraslan: `adata.obsm_keys()` might be better here.

scanpy/plotting/_tools/__init__.py
```diff
@@ -633,3 +636,110 @@ def sim(adata, tmax_realization=None, as_heatmap=False, shuffle=False,
...
+        raise ValueError('Cannot find the embedded representation `adata.obsm[X_{!r}]`. Compute the embedding first.'.format(basis))
+
+    components = [0,1]
+
+    if basis == 'dm':
```
gokceneraslan: I think this is `X_diffmap`, see https://github.com/theislab/scanpy/blob/master/scanpy/tools/_diffmap.py#L32.

falexwolf:
Can't you use `vmin` and `vmax`? I remember there was a similar issue asking for greying out some data points a couple of weeks ago. And the person asking seemed to be happy with that answer...

fidelram:
@LuckyMD  You can do something like this:

```PYTHON
import matplotlib as mpl

adata = sc.datasets.blobs()
adata.obs['n_blob'] = np.array(adata.obs.blobs.apply(lambda x: int(x)))
sc.tl.pca(adata)


cmap = plt.get_cmap('YlOrRd')
norm = mpl.colors.Normalize(vmin=1, vmax=3)
cmap.set_over('blue')
cmap.set_under('lightgray')

sc.pl.pca(adata, color='n_blob',  cmap=cmap, norm=norm)
```
image

In this case any value that is less than 1 get a `lightgray` color and any value greater than 2 gets a `blue` color. You can pass `cmap` and `norm` to any scatter plot function including `plot_scatter`

LuckyMD:
I had no idea that I could define out-of-range colours like this. That's amazing! Thanks a lot!

scanpy/plotting/_tools/__init__.py
LuckyMD: Thanks! Wasn't aware of that.

scanpy/plotting/_tools/__init__.py
LuckyMD: good spot :)

scanpy/plotting/_tools/__init__.py
```diff
@@ -633,3 +636,110 @@ def sim(adata, tmax_realization=None, as_heatmap=False, shuffle=False,
...
+        greysize: Optional[int] = 20,
+        colorsize:  Optional[int] = 40,
+        vmax:  Optional[int] = 1,
+        vmin:  Optional[int] = 0,
+        save: Union[bool, str, None] = None,
+        **kwargs):
+    """Plot the density of cells in an embedding (per condition)
+
+    Plots the gaussian kernel density estimates (over condition) from the
+    `sc.tl.density()` output.
```
fidelram: I think here should be `sc.tl.embedding_density`

scanpy/plotting/_tools/__init__.py
```diff
@@ -633,3 +636,110 @@ def sim(adata, tmax_realization=None, as_heatmap=False, shuffle=False,
...
+    ----------
+    adata : :class:`~anndata.AnnData`
+        The annotated data matrix.
+    basis : `str`
+        The embedding over which the density was calculated. This embedded
+        representation should be found in `adata.obsm['X_[basis]']``.
+    key : `str`
+        Name of the `.obs` covariate that contains the density estimates
+    group : `str`, optional (default: `None`)
+        Category of the observed covariate which will be plotted.
```
fidelram: is not clear to me from the description what group means or does. Is a categorical value from `.obs` ? If this is the case, in other tools we tend to use `groupby`.

scanpy/plotting/_tools/__init__.py
```diff
@@ -633,3 +636,110 @@ def sim(adata, tmax_realization=None, as_heatmap=False, shuffle=False,
...
+    """
```
fidelram, line 695: can you add an example here? 

Something like (I am copying from the `sc.tl.dendrogram` function.
```
    Examples
    --------
    >>> adata = sc.datasets.pbmc68k_reduced()
    >>> sc.tl.dendrogram(adata, 'bulk_labels')
    >>> sc.pl.dendrogram(adata, 'bulk_labels')
```

scanpy/plotting/_tools/__init__.py
```diff
@@ -633,3 +636,110 @@ def sim(adata, tmax_realization=None, as_heatmap=False, shuffle=False,
...
+    if 'X_'+basis not in adata.obsm.dtype.names:
+        raise ValueError('Cannot find the embedded representation `adata.obsm[X_{!r}]`. Compute the embedding first.'.format(basis))
+
+    components = [0,1]
+
+    if basis == 'dm':
+        components = [1,2]
+
+    if key not in adata.obs:
+        raise ValueError('Please run `sc.tl.density()` first and specify the correct key.')
```
fidelram: here should also be  `sc.tl.embedding_density` and in all other lines.

scanpy/tools/_embedding_density.py
```diff
@@ -0,0 +1,143 @@
...
+        adata: AnnData,
+        basis: str,
+        groupby: Union[str, None] = None,
+        key_added: Union[str, None] = None):
+    """Calculate the density of cells in an embedding (per condition)
+
+    Gaussian kernel density estimation is used to calculate the density of
+    cells in an embedded space. This can be performed per category over a
+    categorical cell annotation. The cell density can be plotted using the 
+    `sc.pl.density()` function.
```
fidelram: change to `.tl.embedding_density`

scanpy/plotting/_tools/__init__.py
LuckyMD: It's a specific category from a categorical covariate in `.obs`. I struggle to define this in a nicer way.

sophietr: maybe similar to sc.pl.umap (and other plotting functions)
groups : str, None
Restrict to a few categories in categorical observation annotation. The default is not to restrict to any groups.

scanpy/plotting/_tools/__init__.py
LuckyMD: done, thanks for the tip!

LuckyMD:
@fidelram I think I found a bug in `plot_scatter`. While the `sort_order` parameter is reordering the data points to plot the highest obs covariate last, which is great... I don't think this is happening correctly with the dot sizes. Thus, with `sort_order=True` you cannot really use dot sizes. I don't have a minimal example for this yet, but I think this is what I see while calibrating defaults for dot sizes with this code. I will confirm later, but let me know if you think this is possible.

scanpy/plotting/_tools/__init__.py
LuckyMD: Good idea, thanks. I don't think I can use `plot_scatter` to plot multiple categories like this though, because you need different dot sizes, and density values for each category. I don't think I can just wrap all those things in a list for multiple plots.

fidelram:
@LuckyMD plot_scatter does not sort size.. Let me submit a PR for this quickly.

LuckyMD:
@fidelram Thanks a lot! I was getting confused why my dot sizes weren't working (only sometimes, strangely...). If the dot sizes are sorted according to the `color` variable it should hopefully work.

LuckyMD:
I think everything should be there now pending @fidelram's fix. If that solves my dot size issues, then it's ready to merge. So I'm waiting for #546 to be merged.

fidelram:
@LuckyMD  #546 is merged.

falexwolf:
Wow! This looks absolutely fantastic! So I guess we can merge and make a 1.4.1?

LuckyMD:
Let me check if dot sizes work now. Also, I still have another function I would like to add, which probably requires less work. Just some marker gene overlap test that takes a dictionary as input. Could you give til next week Wednesday for that?

LuckyMD:
Dot sizes now work, this is ready to be merged... assuming you are happy with how I integrated it into the documentation (I added a separate "density" subsection for the plotting tools as I felt it shouldn't really be put in the same category as embeddings.)

falexwolf:
Great! :smile:

Sure, next Wednesday is fine. Also, we can always simply make 1.4.2. I just need to check one tiny thing before we actually make the release.

fidelram:
@LuckyMD why don't you add a test to `scanpy/tests/test_plotting.py`. Thus we can guarantee that future changes do not break your code.

LuckyMD:
I added two tests in `scanpy/tests/test_embedding_density.py`... one of them just to test if the plotting functions run. Should this have been placed in a different file?

PR #546, opened 2019-03-21
fix an error when figsize is given

fidelram:


LuckyMD:
This is great... are you planning to merge this soon? Then I can check if it fixes my issue.

PR #549, opened 2019-03-22
Marker gene overlap scoring function

LuckyMD:
Introduces a function to calculate marker gene overlaps between a reference set of marker genes provided as a dictionary, and data-derived marker genes as calculated by `sc.tl.rank_genes_groups()`.

Currently implemented overlap functions are: overlap counts (with row or column normalization), overlap coefficient, and jaccard index.

Still to do:
- write a test
- finish documentation
- allow p-value thresholding when available
- allow using top X marker genes rather than all calculated markers
- test that it works properly...


falexwolf:
This looks very good already!  Nice code! :) Thanks for also adding the things you're still planning to add.

@flying-sheep might not like that you spell out the types in the docs in addition to the type annotations but I think it's good: the string selection values for `method` can't be represented as a type annotation, I think; and I still like browsing the type annotations with shift+tab at the same time when reading the rest of the parameter explanation. So, I'd say, just keep it as it is.

flying-sheep:
Could you please do `Dict[KeyType, ValueType]` instead of `dict` in the type annotations?

@falexwolf you forgot that the types will be added to the shift-tab info too: https://github.com/theislab/scanpy/blob/10f8a3c8aa5cfa4431db2a10f1f3cc088072e788/scanpy/__init__.py#L42

So yes, @LuckyMD please remove all *redundant* type info in the docs. The info for `method` and `normalize` should stay, the rest can go with absolutely no loss of information anywhere.

falexwolf:
OK, cool; thanks @flying-sheep!

The function looks otherwise good!

scanpy/tools/_marker_gene_overlap.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Returns
+    -------
+    Updates `adata.uns` with an additional field specified by the `key_added`
+    parameter (default = 'marker_gene_overlap'). 
+
+    Examples
+    --------
+    >>> adata = sc.datasets.pbmc68k_reduced()
+    >>>
```
fidelram: don't forget to finish this

LuckyMD: of course... it was just a copy-paste from embedding density for now ;).

LuckyMD:
> Could you please do `Dict[KeyType, ValueType]` instead of `dict` in the type annotations?

@flying-sheep When I put in `key: dict[str, set]` it tells me type annotations are not subscriptable. How can I state the dict key-value types required?

LuckyMD:
Finished most of the features... just need to add the documentation... would appreciate if someone could take a look at the code @flying-sheep, @falexwolf...

LuckyMD:
From my side it's ready to be merged. I have left a note in the comments about extending this to enrichment scores, and that this would be difficult. I though it might be good to leave that in there in case anyone wants to extend it later. Thoughts on that?

scanpy/tools/_marker_gene_overlap.py
```diff
@@ -0,0 +1,251 @@
...
+        
+    #Note:
+    # Could add an 'enrich' option here (fisher's exact test or hypergeometric test),
+    # but that would require knowledge of the size of the space from which the reference
+    # marker gene set was taken. This is at best approximately known.
+        
+    # Create a pandas dataframe with the results
+    marker_groups = list(reference_markers.keys())
+    clusters = list(cluster_ids)
+    marker_matching_df = pd.DataFrame(marker_match, index=marker_groups, columns=clusters)
```
falexwolf: Currently, we cannot write DataFrames using `adata.write`... :(

What I would do, is add an `inplace` parameter that defaults to `False`. If users set it to `True`, they get a `NotImplementedError`.  Then, instead of adding the `marker_matching_df` `inplace` to `adata`, you simply return it.

At some point, we will be able to store dfs within `.h5ad`. Then we can also implement the `inplace` version of this function.

@flying-sheep Any chance that you would have fun implementing this. It's just some reorganization fo the code as I implemented the functionality already for `.obs` and `.var` via https://github.com/theislab/anndata/blob/473413a76ebe741754b63558c4c5a34861451c82/anndata/base.py#L193-L222

It could be a 30 min job. Upon write, each dataframe is transformed to a record array, upon read, it's transformed back to a dataframe.

Having this would also allow a much nicer return value of `rank_genes_groups`...

falexwolf:
This looks great, Malte! Only thing see my comment above. You should get on the author list! :)

LuckyMD:
I'd love to get on the author list :).

Will add the inplace option now with the `NotImplementedError` raised. The only thing remaining is what @flying-sheep suggested with the typing for the dictionary, which I can't seem to do.

LuckyMD:
Found the `Dict[str, set]` option in the `typing` module. I've left the `dict` typing for the hidden functions though... I guess they're not user exposed anyway.

I left the code in there for `inplace=True`... so when the pandas writing option is implemented, the `if inplace: raise NotImplementedError('...')` lines just need to be removed.

Ready to be merged from my side now.

LuckyMD:
Not sure what's up with Travis... the tests pass on my machine, and they were passing on Travis the whole time... my last commit hasn't really changed anything that would cause this fail.

falexwolf:
Great! Travis looks like stalled build...

falexwolf:
Damn, there seems to be some Python 3.5 related thing going on: https://travis-ci.org/theislab/scanpy/jobs/513835981

@LuckyMD, any idea why this returns something completely different?

LuckyMD:
I have no idea... I was trying to understand how this would be different... the matrix I expect is `[[3,1],[1,3]]`... And that's what I get for python 3.6. I don't know what changed from 3.5. Maybe pandas `.iloc[]` does something else?

LuckyMD:
It started to go wrong with ff26149 and all I changed there was that I moved the `if inplace:` statement up to the user tests instead of down to where the outputs were written.

Is python 3.5 somehow sensitive to whitespace after if statements? I found a whitespace after the `if inplace: `... maybe that's it?

LuckyMD:
Does anyone have a nice instruction set on how I can reproduce the travis python 3.5 environment? @flying-sheep? Maybe I can hunt down the cause of the error then... although it's apparently only sporadic according to @ivirshup :/

flying-sheep:
I don’t think there’s an easy way to reproduce it exactly, sorry.

ivirshup:
iirc, I think there is a docker image which is pretty close to what Travis runs.

flying-sheep:
Maybe travis’ [ubuntu-systemd](https://hub.docker.com/r/travisci/ubuntu-systemd)?

Then you’d need to use [travis-build](https://github.com/travis-ci/travis-build) (for which there also is a [docker image](https://hub.docker.com/r/travisci/travis-build)) to convert the .yml to a .sh and execute that on the ubuntu container. (travis-build exists to do all this but that’s maybe overkill)

LuckyMD:
In #583 I have amended this PR a bit... corrected the normalization to make more sense, made the input dict value type more malleable, and I'm playing with the Travis build there.

For some reason it already stopped failing when I just changed the order of `assert` statements. I thought there might have been too many `assert` statements in a single function, so I split them up into separate testing functions. I think that's probably better coding practice anyway... but now they fail again for python 3.5... Maybe move the discussion there?

Issue #435, opened 2019-01-18
TypeError: ufunc 'log1p' output (typecode 'd') could not be coerced to provided output parameter (typecode 'l') according to the casting rule ''same_kind''


ajitjohnson:
Hi there,

I seem to have trouble analyzing a dataset.

https://www.dropbox.com/s/og5lw42chh2qujm/Trial_data1.csv?dl=0

If I run sc.pp.log1p (adata), I get the following error. 

TypeError: ufunc 'log1p' output (typecode 'd') could not be coerced to provided output parameter (typecode 'l') according to the casting rule ''same_kind''

I could normalize it on my own and say if I do a PCA analysis and try to plot the results, I get the following error.

sc.pl.pca(adata, color = 'DAPI')
TypeError: object of type 'numpy.int64' has no len()

If I plot it without the color, it does work though. 

Below is how I go from CSV to AnnData

# Read File
x = pd.read_csv('Trial_data1.csv', delimiter=',', index_col=0)

# Convert to AnnData
file_url = 'https://raw.githubusercontent.com/ajitjohnson/Jupyter-Notebooks/master/py_scripts/mi_pp_anndata.py'
exec(open(wget.download(file_url)).read())
adata = mi_pp_anndata (x)

Any help is appreciated. Thank you.

falexwolf:
This is really strange, have you had problems with other datasets? Is your dataset corrupted in some way?

ajitjohnson:
Hi Alex,
I managed to get the log working by using your function to convert to AnnData rather than mine. (adata = sc.AnnData(x))

However, coloring the plots still does not work. I get the following error.
TypeError: object of type 'numpy.int64' has no len()

You can reproduce the error by the following
### Load Data
x = pd.read_csv('Trial_data.csv', delimiter=',', index_col=0)
### Drop DAPI
x = x.drop(list(x.filter(regex='DAPI.', axis=1)), axis=1)
### Convert to AnnData
adata = sc.AnnData(x)
### Filter cells
sc.pp.filter_cells(adata, min_genes=1)
sc.pp.filter_genes(adata, min_cells=1)
adata.obs['n_counts'] = adata.X.sum(axis=1)
### Normalize data
sc.pp.log1p(adata)
### PCA
sc.tl.pca(adata, svd_solver='arpack')
sc.pl.pca(adata)
sc.pl.pca(adata, color='CD3D')

I also tried it on a different dataset.

falexwolf:
[...]

PR #551, opened 2019-03-23
[WIP] Make log1p work with integers/ output of downsample_counts

ivirshup:
Addressing #435 

I've opened this with a not-pretty version. I think this function could be simplified a lot by not requiring it to work on arrays. I'm willing to leave it ugly for now.

Left to resolve: return type of `downsample_counts`.

### DIFF
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -928,7 +928,7 @@ def downsample_counts(
     """
     Downsample counts from count matrix.
 
-    If `counts_per_cell` in specified, each cell will downsampled. If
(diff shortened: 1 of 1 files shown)


falexwolf:
Ah sorry, I should have told you that. All of the preprocessing functions can be migrated to only work with AnnData; there is no important setting in which you want to pass an array or a sparse matrix. That's also remniscient from the early days when I thought people might not like to use AnnData. But that's of course stupid, they wouldn't use Scanpy in that case, anyway.

I'm merging this for now so that we have something working for 1.4.1, but if you want to simplify and reduce this to AnnData-only, happy to merge a PR on that.

PR #554, opened 2019-03-25
DOC: one can also run bbknn before louvain

fbnrst:


ivirshup:
It's definitely good to mention that the connectivity matrix can be provided another way. However, I'm not sure this function actually requires either of those, since you can pass the adjacency matrix.

flying-sheep:
Yup. I clarified that as well and merged it.

PR #557, opened 2019-03-25
DOC: remove not accepted arguments from pl.scatter

### LINKED ISSUE #458: Documentation for pl.scatter
Hi,

Some of the parameters documented online for `pl.scatter` are not accepted by the method: `ncols`, `wspace`, `hspace`. And some are not documented: `right_margin`, `left_margin`.

The difference seems to come from the parameters described in `{scatter_bulk}` and used for `pl.scatter` documentation.

What should be fixed `scatter_bulk` or inputs for `pl.scatter`?

Bérénice

fbnrst:
fixes #458

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -17,7 +17,7 @@ from .. import logging as logg
 from . import _utils as utils
 from ._utils import scatter_base, scatter_group, setup_axes
 from ..utils import sanitize_anndata, doc_params
-from ._docs import doc_scatter_bulk, doc_show_save_ax, doc_common_plot_args
+from ._docs import doc_scatter_temp, doc_show_save_ax, doc_common_plot_args
 
 VALID_LEGENDLOCS = {
     'none', 'right margin', 'on data', 'on data export', 'best', 'upper right', 'upper left',
@@ -26,7 +26,7 @@ VALID_LEGENDLOCS = {
 }
 
 
-@doc_params(scatter_bulk=doc_scatter_bulk, show_save_ax=doc_show_save_ax)
+@doc_params(scatter_temp=doc_scatter_temp, show_save_ax=doc_show_save_ax)
 def scatter(
         adata,
         x=None,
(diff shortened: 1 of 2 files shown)


flying-sheep:
That’s super redundant now.

Please extract all that text from `doc_scatter_bulk` into another variable and import and use that one instead.

fbnrst:
In #458 @fidelram suggested that this would be the way to go. If I put the text into another variable, this variable will only be used once. Does this still make sense? Anyways, I think this is just temporary until `pl.scatter` is in a better shape if I follow @falexwolf correctly.

flying-sheep:
Deduplication always makes sense. I often use speaking variable names instead of comments to clarify what I’m doing.

Here the 6 reasons why I’m convinced of the way I suggested:

1\) If I look at your change as it is, I have no idea what parameters are missing compared to `doc_scatter_bulk`. If you used variables, we could see it at a glance. 2) You could add a comment explaining that this one is just temporary (Hard to do in-line in a docstring).

If one makes changes to the parameters, 3) they only have to make them once and 4) can’t forget to make them twice. 5) Also the diff becomes much smaller and 6) git will be able to track doc changes that are made in the mean time.

So do it please.

fbnrst:
@flying-sheep what about this then?

flying-sheep:
There’s still duplication. There shouldn’t be any duplicated parameter docs. I’d expect something like this:

The diff will be 8 changed lines or so as opposed to dozens of added duplicate lines.

```py
_doc_scatter_common = '''
sort_order
...
'''

_doc_scatter_panels = '''
ncol
...
'''

_doc_scatter_meta = '''
title
...
'''

doc_scatter_temp = _doc_scatter_common + _doc_scatter_meta
doc_scatter_bulk = _doc_scatter_common + _doc_scatter_panel + _doc_scatter_meta
```

fbnrst:
OK, I think, now I got you point.

fbnrst:
@flying-sheep Thanks for your help. Like this, maybe?

flying-sheep:
beautiful, thank you

PR #572, opened 2019-03-30
Add 'equal_frequency' option to highly_variable_genes

### LINKED ISSUE #415: equal_frequency bins in highly_variable_genes
Hi,

Using Seurat, in their variable gene function I've had some success using the `equal_frequency` option, where each bin contains an equal number of genes. Would it possible to implement this option in scanpy? 

If you'd like I could submit a PR to implement this feature. I think it could be as simple as using `pd.qcut` instead of `pd.cut` or you could use a similar style as in the `cell_ranger` flavor with `pd.cut(df['mean'], np.r_[-np.inf,
np.percentile(df['mean'], np.arange(10, 105, 5)), np.inf])`. I don't know how useful it would be, but I could also add the option to have more bins in the `cell_ranger` flavor by replacing `np.arange(10,105,5)` with `np.linspace(10, 100, n_bins - 1)`.

Best,
David

davidhbrann:
This fixes #415, by allowing one to find variable genes using the `equal_frequency` option. It also adds and option to change the number of bins for cell ranger flavor.

I originally tried to copy the implementation in Seurat, which would allow a test similar to what's already present for the `equal_width` implementation. However the Seurat code has an error:
```R
else if (binning.method=="equal_frequency") {
        data_x_bin <- cut(x = gene.mean, breaks = c(-1,quantile(gene.mean[gene.mean>0],probs=seq(0,1,length.out=num.bin))))
}
```
The `-1` in the code makes it such that there is always only one value in the first bin, which goes from -1 to the minimum value. Not sure why they have this, but then we get different answers since the Scanpy code in `highly_variable_genes` always makes bins that have only one gene significant (to correct the other error from Seurat that normally excludes these bins/genes, which often contain some highly-expressed genes). Additionally, the `cut` function in R sometimes returns bin edges with different rounding than the Seurat implementation since Seurat does not modify the default `dig.lab = 3`. In contrast, I believe pandas uses the actual cutoffs in the data.

### DIFF
--- a/scanpy/preprocessing/_highly_variable_genes.py
+++ b/scanpy/preprocessing/_highly_variable_genes.py
@@ -114,8 +114,7 @@ def highly_variable_genes(
         if binning_method == 'equal_width':
             df['mean_bin'] = pd.cut(df['mean'], bins=n_bins)
         elif binning_method == 'equal_frequency':
-            # be consistent with Seurat, even if qcut would also work
-            df['mean_bin'] = pd.qcut(df['mean'], q=n_bins)
+            df['mean_bin'] = pd.qcut(df['mean'], q=n_bins, duplicates='drop')
         else:
             raise ValueError('`binning_method` needs to be "equal_width" or "equal_frequency"')
         disp_grouped = df.groupby('mean_bin')['dispersion']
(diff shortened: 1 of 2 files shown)


falexwolf:
Great, thank you! 

It will be merged very soon; the change in the cell ranger option might be a backwards breaking change, hence, I added this here: https://github.com/theislab/scanpy/issues/453.

  Alex

gokceneraslan:
Hey @falexwolf, if starting the percentile bins from 10 is not intentional (as mentioned in #624 ) I can resolve the conflicts and merge this along with the fix in #624 .

falexwolf:
Sounds good, @gokceneraslan!

cyrus303 (approved):

Issue #558, opened 2019-03-26
Where should `sc.datasets` put data?


ivirshup:
I'm adding that expression atlas downloader now (#489), and wondering where the files should go.

`pbmc68k_reduced` and `toggleswitch` put the datasets relative to where scanpy is installed (via `__file__`). All other functions place the data relative to where the python process was started.

While I like not storing the same files all over a filesystem, I'm not sure in the `scanpy` installation directory is the right place to be storing data.

Thoughts?

flying-sheep:
The scanpy install directory is super wrong, as it’s not writable for many people.

There’s exactly one correct way of determining a global place for cache* files like this: [`appdirs.user_cache_dir(...)`](https://pypi.org/project/appdirs/)

Alex and me talked in the past and decided for a visible directory in the working directory. I’d be up for changing it to `user_cache_dir(…)` for the data.

*the data are cache files since reexccuting their function after deleting the files will redownload them without loss of information.

maximilianh:
Hi Isaac, I have a related question: does your expression atlas
downloader also store the coordinate and all the meta data?

ivirshup:
@flying-sheep That does look useful, though I definitely see a lot of toolkits using something like `~/.{toolkitname}` on my mac. Personally, I'd look under my home directory first.

I think if things are going to move, there should be a sensible default, but the location should be user configurable (probably through an environment variable). For example, the main hpc I'm on only gives me about 1gb of space where `appdirs` would put these files.

@maximilianh Currently I don't get coordinates along with some other computed values like SC3 clustering. I also believe it's only a subset of the metadata. I think there's more through the array express api, but that was more of a pain to navigate.

Was there anything in particular you wanted to see included?

maximilianh:
Well, I know nothing, maybe you already have everything, but I could
look at an example? The advantage of the expression atlas is that they
have really good meta data. That's provided through their downloadable
files, as far as I remember. So if you got everything that is in their
downloadable meta data files, then you certainly have everything of
interest.

flying-sheep:
> though I definitely see a lot of toolkits using something like ~/.{toolkitname} on my mac. 

There’s two possible reasons: 1.: We’re talking about something from the 80’s like SSH or BASH. They earned their right to things their way because they’re older than the standards they’re not following. Or 2.: Whoever designed this didn’t do their research and just hacked in first thing that came to mind. This is not an excuse.

[...]

Issue #489, opened 2019-02-17
[Suggestion] Expression atlas datasets?


ivirshup:
I've written up a little script for making an `AnnData` given an accession for the EBI single cell expression atlas ([gist](https://gist.github.com/ivirshup/e7f0c435474d9b06b622c63d4221afe6)), and was wondering if it'd be useful to add under `sc.datasets`. It could be nice to make it easier to try out methods/ have examples which weren't for pbmcs.

flying-sheep:
Sure, that seems like a great idea!

falexwolf:
Awesome idea! :)

Sorry that I haven't merged all your other PRs, yet. I got back to work on Scanpy yesterday or so after being sick for 2 weeks and another 2 weeks of complete chaos (sick babies)... ;)

You'll get feedback very soon, but at first sight, all of them looked fine, anyway. :)

Issue #472, opened 2019-02-10
Tracking operations performed on AnnData objects


afrendeiro:
When exploring various options of preprocessing data, I try to avoid having several copies of AnnData objects in memory if they're not sparse, so I save them to h5ad at key steps. Sometimes alas, after a few iterations I re-write stuff and forget what operations have been performed in my "X" (particularly in the preprocessing steps).

So, because being lazy makes me creative, I started tracking these in the object itself (see example https://gist.github.com/afrendeiro/7ccaf324bfdbff042ae36f734f544860) by decorating the preprocessing functions post hoc (this could even easily be used to save the values of kwargs passed potentially).

I wonder if an internal implementation of this would be of broad interest, particularly for functions which modify "X" inplace?
Of course this would be no replacement for proper documentation of one's steps, etc but I thought it could be an interesting addition to scanpy in any case.

ivirshup:
I like this idea, but think it could be expanded on a bit. I think there are benefits to approaching this through logging. Some advantages of doing this through logging:

* Global record. If a copy is made or an AnnData split, you could figure out which object came from where.
* Control over level of detail. What kind of information is recorded can be customized. Maybe the user wants provenance, but maybe they want performance information.

What if tracking was done through logging? Here's a couple quick examples of what I mean:

<details>

<summary>Simple example. Logs `anndata` used, function called, time elapsed </summary>

```python
from anndata import AnnData
from datetime import datetime
from functools import wraps
from structlog import get_logger
from time import sleep
import uuid

logger = get_logger()

def logged(func):
    @wraps(func)
    def func_wrapper(*args, **kwargs):
        call_id = uuid.uuid4() # So we can always match call start with call end
        call_start_record = dict(call_id=call_id, called_func=func.__name__)
        if type(args[0]) is AnnData:
            call_start_record["adata_id"] = id(args[0])
        logger.msg("call", **call_start_record)

        t0 = datetime.now()
        output = func(*args, **kwargs)
        dt = datetime.now() - t0

        call_finish_record = dict(called_func=func.__name__, elapsed=dt)
        if type(output) is AnnData:
            call_finish_record["returned_adata_id"] = id(output)
        logger.msg("call_finish", **call_finish_record, call_id=call_id)
        return output
    return func_wrapper

# Usage

@logged
def foo(adata, x, copy=False):
    sleep(0.5)
    if copy: return adata.copy()

import scanpy as sc
pbmcs = sc.datasets.pbmc68k_reduced()

[...]

PR #573, opened 2019-03-30
Expression atlas

ivirshup:
Adding an expression atlas downloader to `sc.datasets` (proposed in #489).

I've punted on replacing where datasets are downloaded by just making it a variable in settings, since it seems contentious where datasets should be downloaded by default #558.

@flying-sheep when I build the docs locally, the link to the expression atlas doesn't format properly on the main `API` page, but does on it's own page. Any ideas on if we can get that to work?


ivirshup:
Changing to WIP since I'd like to improve failure handling before this gets merged.

ivirshup:
That should do it. Ready for review

scanpy/datasets/_expression_atlas.py
```diff
@@ -0,0 +1,174 @@
...
+        if filter_boring:
+            adata.obs = _filter_boring(adata.obs)
+        return adata
+    except OSError:
+        # Dataset couldn't be read for whatever reason
+        pass
+
+    download_experiment(accession)
+
+    print("Downloaded {} to {}".format(accession, experiment_dir.absolute()))
```
falexwolf: Hm, the default logging level is `warning`, not `info`.

It's strange to use `logg.warn` as we don't want to print `WARNING`, so, I guess, simply using print is fine... (it won't appear in a logging file, but no one is using this functionality anyway, so we can drop it...)

Can you add a comment on that one can change `settings.datasetsdir`?

falexwolf:
Isaac,

this is great, thank you so much!

Regarding the default for the dataset directory. I like this solution!

Very small edits in addition to what I commented in the code:
* Can we call this `epi_sc_expression_atlas` instead of `expression_atlas`?
* For the time being, can we make this `settings.datasetsdir` instead of `settings.dataset_dir` and add it here: https://github.com/theislab/scanpy/blob/97c8b54ec884ac8e8396a80b6782a0d59a17a874/scanpy/api/__init__.py#L272
* Can we point it to the home directory by default, I'd say `~/scanpy-datasets/`?

Notes:
* By having a datasets dir, which is separate from the cache dir (which make sense), I guess, we can also use `user_cache_dir(…)` as the default for the cache dir (which would hopefully choose something in `~/.cache` on a Linux system, probably via `~/.cache/scanpy/`, I think this what you, also Phil (!) and Gökcen favored if I'm correctly summarizing the long thread?
* We already had a Scanpy config in the beginning (was an `.ini`) and we can reintroduce it in the future, and it should probably go into `~/.config/scanpy.ini` (or `.json` or `.yaml`). No reason not to have it. No need to have a CLI for this purpose.
* We can replace `.settings` with an instance of a class `._settings.Settings`. By that, attributes get auto-documented, we can do nice checks on setting attributes via properties, and we can also directly write to a `~/.config/scanpy.ini` file...
* `pyplot.rc_context` sounds awesome.
* Precedence for settings is correct as stated in https://github.com/theislab/scanpy/issues/558#issuecomment-478214932, this is also how I had it before removing the config file...

scanpy/datasets/_expression_atlas.py
ivirshup: I'd gone with printing for downloading since I believe the progress bars go to stdout anyways, so it's not adding noise to a previously silent operation.

Also, I had been thinking about trying to make logging a bit more consistent and easy to implement (#472), and might pick that up once I've got bandwidth.

Should the comment about `settings.datasetdir` go into the function docs or the printed message?

ivirshup:
> Can we call this epi_sc_expression_atlas instead of expression_atlas?

Definitely agree it's good to specify it's from EBI. Would `ebi_expression_atlas` be alright? `ebi_sc_expression_atlas` feels a little verbose for me. I think it's implied it's single cell, plus it's explicit in the doc-string.

> For the time being, can we make this settings.datasetsdir instead of settings.dataset_dir and add it here:

Changed the name. It looks like the main docs aren't being generated from `scanpy/scanpy/api/__init__.py`, but from `docs/api/index.rst` instead. Which is correct?

> Can we point it to the home directory by default, I'd say ~/scanpy-datasets/?

Changed.

Does this mean config changes should also happen in this PR? I think this may cause trouble (HPC environments with small `~` allocations) without allowing default configuration at the same time. I had previously figured that setting up a config file could be factored out to a separate PR.

To be able to put off adding the config for now, we could temporarily special case a `SCANPY_DATASETDIR` environment variable.

ivirshup:
Oh, I also added tests for example dataset loading since checking they worked manually was a pain. These won't run by default (they take a while, and can fail for network access reasons), but will run with `pytest --internet-tests`. Note that `test_burczynski06` will fail until this get's rebased on master. Thoughts?

Also travis failed this for `scanpy/tests/test_marker_gene_overlap.py` failing an assertion on the first time around, but passed when I triggered a new build. Not sure what's up with that.

scanpy/settings.py
```diff
@@ -51,6 +52,10 @@
 """Default cache directory.
 """
 
+datasetdir = Path.home() / "scanpy_datasets"
```
flying-sheep: Veto, see #558. I consider creating `~/something-visible` a completely unacceptable default for a library.

> My personal hell certainly includes dozens of libraries and applications putting all kinds of crap in unhidden directories in my home. All of them have a different way to configure that location or none at all. Chills me right to the core.

I think it would be best to leave this change out of this PR. We can do a separate one once we have decided on how to tackle this.

flying-sheep (changes requested):

scanpy/settings.py
ivirshup: Sure, I also think it would be a good idea to leave changing the default to a separate PR.

You'll see I initially made this a non-breaking change (`datasetdir = Path("./data")`), just changed to this going off feedback from Alex.

I'm gonna wait on actually modifying the code for a minute to make the eventual rebase easier.

flying-sheep: OK, great. I think the changes to the default (`./data` → `~/???`) should be in a PR that also changes what’s printed on download/read of such files, and we can figure out there how to detect if we’re on a linux server (where `~/scanpy_data` would make sense) or on a desktop system (where it doesn’t)

As an aside: Windows Explorer’s default view doesn’t show you your `%HOME%`, so `~/scanpy_datasets` wouldn’t even be visible there. @falexwolf, @ivirshup: In summary: This topic requires discussion, it’s more complex than it seems, and I’ve read up a *lot* on it, so I think you should either leave the decision to me or let me explain it to you in all its tiring depth. I’ll do it if you think it’s worth your time.

ivirshup:
@falexwolf, @flying-sheep 

Just to recap what's left to be resolved here

* I'll reset the default value of `datasetdir` to the current value "./data"
* Related, how about `datasetdir` instead of `datasetsdir`? It matches more to `cachedir` and `figdir`. Also, by analogy, it's "potato sack" not "potatoes sack" so "dataset directory" sounds more natural that "datasets directory" to me.
* `ebi_expression_atlas` vs `ebi_sc_expression_atlas`
* Potentially adding a class for settings right now?
  * I think this becomes more important if `datasetdir` is documented. I bet people will set it with a `str` instead of a `Path` and that'll break things.

docs/api/index.rst
```diff
@@ -249,6 +250,7 @@ Datasets
    :toctree: .
 
    datasets.blobs
+   datasets.ebi_expression_atlas
```
ivirshup, line 253: @flying-sheep the link to the expression atlas doesn't format correctly on the API page (it does on the functions own page). Any suggestions for how to fix this?

scanpy/settings.py
ivirshup: Would you mind sharing some links which explain why this is important? I'll admit, it does not seem very complex to me at the moment, but I also don't know much about this topic.

Also, a more general question, does scanpy work on windows?

flying-sheep: Sadly there’s no writeup. It’s a lot of different observations I made and discussions in different projects. If you search XDG_CONFIG_HOME and [XDG_CACHE_HOME](https://github.com/search?q=xdg_cache_home&type=Issues) you’ll find a lot of people requesting that (e.g. mwaskom/seaborn#1530)

The question “why not” is always answered with “we didn’t think about it” or “we followed what others did (without thinking about it)”, and the answer to “why don’t you change” is always “we have been doing it for too long, people will get confused”. So I won’t wait until somebody is able to say the latter about scanpy.

So yeah: What `appdirs` and similar modules do as well as the discussions in those modules’ issue trackers is pretty much the best one can get as a writeup. One then cross-references what one’s library/app needs to write to some global location (executables? config? cache? logs? documents?) with the locations those modules provide. Then one saves those files there instead of coming up with something themselves because many people have thought a long time about this to create that module.

----

About Windows: Yes! I think it’s important to keep it that way. I recently read [this blog post](https://samsaffron.com/archive/2019/03/31/why-i-stuck-with-windows-for-6-years-while-developing-discourse) and fully agree with the author:

> […] a Ruby that works well on native Windows will help Ruby adoption a lot and eliminate drain to other frameworks. Installing a whole brand new OS is just too much of a barrier. Just install Linux is not a solution.

The same applies to everything of course. Many just start out and aren’t fed up by Windows enough to ditch it, so making the start on Windows as smooth as possible helps adoption by new people a lot.

docs/api/index.rst
flying-sheep: Could you please paste ascreenshot here? Maybe that saves me from checking this out, building the docs, …

ivirshup: ## Where it isn't working

Screen Shot 2019-04-03 at 11 56 33 AM

Here's the html:

```html
<td>
    Load a dataset from the
    <a href="#id39">
        <span class="problematic" id="id40">
            `EBI Single Cell Expression Atlas`_
        </span>
    </a>
    .
</td>
```

## Where it is working

Screen Shot 2019-04-03 at 12 00 17 PM

```html
<p>
    Load a dataset from the 
    <a class="reference external" href="https://www.ebi.ac.uk/gxa/sc/experiments">
        EBI Single Cell Expression Atlas
    </a>
    .
</p>
```

scanpy/settings.py
ivirshup: I'll check out that issue.

I have nothing against windows support, I just can't help on any issues related to it. I was mostly wondering if all the dependencies worked on windows, and if there's a reason there aren't appveyor builds.

scanpy/datasets/_expression_atlas.py
falexwolf: I thought about the printed message...

Making logging more consistent would be terrific. :D

falexwolf:
OK guys, I'm happy with all that!

I suggested `datasetsdir` over `datasetdir` because the module is called `datasets` and it's a place where many datasets end up. But, Isaac, you're the native speaker, so you're choice.

Regarding making a settings a class: happy to if you feel you want to do that already in this PR.

Really just change `scanpy/settings.py` to `scanpy/_settings.py`, put a `Settings` class in that file and generate an instance upon importing Scanpy. You could also just make `datasetdir` the only property and add all other attributes of the current `scanpy/settings` module as simple attributes. Then, all of this should be 10 min of work.

docs/api/index.rst
flying-sheep: Wild guess: the code to create that table copies the first line and embeds it in the other document. That doesn’t include the link target. You could probably convert it into an inline link and it’ll work.

Or include the link target in the references file, which is always included (which is the reason for the reference working which is also in your table screenshot)

falexwolf:
Wow, you already did it! Fantastic! :smile:

ivirshup:
Went with properties for everything except private variables. Took a little longer than 10 minutes, but I think it's mostly there.

Got all the tests to pass on my machine, but I bet other things will fail. I also haven't tested what'll happen with the docs, though I did have to modify some of the documentation code.

* The `verbosity` settings might be trouble. I changed the value again... but this should stop an error I'm getting with bbknn and be consistent with the python logging module. 
* The settings imports are ugly. Wasn't sure how to import a variable from a parent module. Is this possible?

scanpy/settings.py
ivirshup: This requested change is resolved now, right?

ivirshup:
Assuming this build passes, is this good to go?

docs/api/index.rst
ivirshup: It didn't seem to work in the references file. I ended up just putting it in the references file.

I did notice that some of the pages weren't being updated with `make html` and I had to clear the build cache. Could that be changed? Building everything can be kinda slow.

falexwolf:
> Went with properties for everything except private variables. Took a little longer than 10 minutes, but I think it's mostly there.

Yes, now that you made everything a property, I would have expected it to take much longer than 10 minutes. It's great that you did!

I'll make some tiny additions and merge.

scanpy/__init__.py
```diff
@@ -30,13 +30,15 @@
...
+from . import datasets, logging, queries, external
 
 from anndata import AnnData
 from anndata import read_h5ad, read_csv, read_excel, read_hdf, read_loom, read_mtx, read_text, read_umi_tools
 from .readwrite import read, read_10x_h5, read_10x_mtx, write
 from .neighbors import Neighbors
-from .settings import set_figure_params
+from ._settings import settings
+
+set_figure_params = settings.set_figure_params
```
falexwolf, line 41: I'll fix this.

falexwolf: Ok, got, you made it a member function...

falexwolf:
Ah, I cannot push to your fork it seems. I think you would have had to set "allow maintainers to edit" or something on the right-hand side. I'm simply merging this and editing after that.

falexwolf:
Here we go: https://github.com/theislab/scanpy/commit/a06c30499967b135b1138805db42e51eeb513e68

scanpy/__init__.py
ivirshup: I made it a method for the class in anticipation of making the updated rc_params a dict in the object, then using that dict with `mpl.rc_context`

ivirshup:
Oh, huh, I thought allowing modifications was on by default. My bad. 

> Yes, now that you made everything a property, I would have expected it to take much longer than 10 minutes. It's great that you did!

I figured it'd be good to with everything being consistent. Plus documentation for the settings is now available through `?sc.settings.{setting}`!

ivirshup:
Oh! I keep forgetting to ask: Why did the original scanpy config file get removed?

falexwolf:
> I figured it'd be good to with everything being consistent. Plus documentation for the settings is now available through ?sc.settings.{setting}!

Of course, it's much better to be consistent. I was just suggesting a quick solution that wouldn't have been worse than the current one... ;)

A complete automatic documentation is now also available from 
https://scanpy.readthedocs.io/en/latest/api/scanpy._settings.ScanpyConfig.html
under
https://scanpy.readthedocs.io/en/latest/api/index.html#settings

scanpy config: At some point almost 2 years ago, I removed a lot of stuff that I didn't think was essential to clean up the project. It was just an element of that. I'm very happy to introduce it again; these days, the project is much more major and indeed has many config options (there were only few at the time) and hence it would indeed merit having a config file.

falexwolf:
Only strange thing is the `getdoc` function.

To be more verbose, one could also change all the cryptic abbreviations to:
```
directory_cache
directory_figures
directory_datasets
directory_write (but this should be deprecated anyway, wasn't a great idea... reminiscent from the very early CLI time of scanpy)
```
or something. It's great that this is super easy without breaking backwards compat now that everything is properties. :)

ivirshup:
>Only strange thing is the getdoc function.

It looks like instance methods can't have new attributes assigned (probably have slots). It's possible the `.getdoc` attribute could be added to the classes method (not sure if that's the right way to say that, here's an example):

```python
class Foo(object):
    def bar(self):
        return 1

# Setting an attribute on the method of an instance raises an error
Foo().bar.x = 1
# AttributeError: 'method' object has no attribute 'x'

# Setting an attribute on the method of a class seems fine:
Foo.bar.x = 1 
Foo().bar.x
# 1
```

PR #574, opened 2019-03-30
Fix read softgz

ivirshup:
While testing my changes to dataset code, I saw that `sc.datasets.burczynski06()` raised the error:

```python
ValueError: `X` needs to be of one of ndarray, MaskedArray, spmatrix, ZarrArray, ZappyArray, not <class 'dict'>.
```

But it had a pretty easy fix.


falexwolf:
Thanks! 🙂

PR #576, opened 2019-03-30
Remove frozen umap

### LINKED ISSUE #522: Getting rid of legacy umap code
Now that UMAP is mature, we can get rid of the temporary frozen umap copy, I introduced a year ago for various reasons (see below).

Meanwhile, the following two functions can maybe direcly imported from UMAP, if not, we could make a PR there or use pynndescent?
https://github.com/theislab/scanpy/blob/07606455c524b38c4efec475a0d7ba87251754bb/scanpy/neighbors/__init__.py#L105
https://github.com/theislab/scanpy/blob/07606455c524b38c4efec475a0d7ba87251754bb/scanpy/neighbors/__init__.py#L258

There is a chance that this can also be solved with an import from UMAP.
https://github.com/theislab/scanpy/blob/07606455c524b38c4efec475a0d7ba87251754bb/scanpy/tools/_umap.py#L107

As just discussed, @Koncopd, can you look into this and make a PR that gets rid of the umap legacy code?

Thank you so much!
  Alex

PS: Just wrote an explanation for the reasons why I intorduced the duplicated code in the first place.
> The duplicated code in Scanpy came about as I wanted to very quickly move forward with a version 1.0 of Scanpy about a year ago. UMAP was just becoming available on GitHub and there wasn’t even a preprint, I think. It changed very quickly and there were dramatic bugs every now and then. Nonetheless it was clear that it’s a major improvement over existing solutions, both in terms of computational performance, quality of the result and ease of installation and use. I wanted to achieve two things: (i) I had to rewrite some parts of UMAP so that I could decompose it a neighbors computing and a dedicated embedding step; you know that in Scanpy, the neighborhood graph is used for many other things other than for the embedding (clustering and trajectory inference). I also added the Gaussian kernel solution that I had before switching to a “UMAP backend” for `pp.neighbors`; which was needed so that results for DPT could be reproduced. All of this would have been quite a discussion with Leland. Until we would have had settled on the “Scanpy needs” that certainly weren’t aligned with the development of an independent young package, PRs would have been integrated to much time would have been lost. Finally, I wanted absolute reproducibility for Scanpy users, which could only be achieved by “freezing the code”.  So, I asked Leland whether he is OK if I add a frozen version of umap as an intermediate solution.

Koncopd:
Still need to figure out why paga test fails.

Also `simplicial_set_embedding` from umap requires data and metrics. Data is `adata.X` and i set `metrics='euclidean'`, but this is not clear.

Fixes #522


falexwolf:
Great!

Some tests should fail as there are probably differences in the neighbor algorithm.

This is also why this is a backwards-compat breaking change. Can you just visually inspect https://scanpy-tutorials.readthedocs.io/en/latest/paga-paul15.html and see what's going on?

This is another notebook that should still do something meaningful after the change: https://nbviewer.jupyter.org/github/theislab/scanpy_usage/blob/master/170502_paul15/paul15.ipynb

And finally, of course, https://scanpy-tutorials.readthedocs.io/en/latest/pbmc3k.html should give somewhat consistent results. But I expect slight variations and no perfect consistence... Actually, I'd expect the associated tests (https://github.com/theislab/scanpy/blob/master/scanpy/tests/notebooks/test_pbmc3k.py) to fail. Can you check?

falexwolf:
Finally, I want to mention that we also need to export the following functions from UMAP: https://github.com/theislab/scanpy/blob/97c8b54ec884ac8e8396a80b6782a0d59a17a874/scanpy/tools/_umap.py#L107

Koncopd:
@falexwolf 

Hi, Alex.
Yes, i'm checking these.

Actually, it somehow passes [test_pbmc3k](https://github.com/theislab/scanpy/blob/master/scanpy/tests/notebooks/test_pbmc3k.py). Only [test_paga_paul15_subsampled](https://github.com/theislab/scanpy/blob/master/scanpy/tests/notebooks/test_paga_paul15_subsampled.py) fails. It seems that adjacency matrix is a bit different after the change and this affects paga connectivities. But it is preliminary, i'm checking still.

Koncopd:
> Finally, I want to mention that we also need to export the following functions from UMAP:
> 
> [scanpy/scanpy/tools/_umap.py](https://github.com/theislab/scanpy/blob/97c8b54ec884ac8e8396a80b6782a0d59a17a874/scanpy/tools/_umap.py#L107)
> 
> Line 107 in [97c8b54](/theislab/scanpy/commit/97c8b54ec884ac8e8396a80b6782a0d59a17a874)
> 
>  from ..neighbors.umap.umap_ import find_ab_params, simplicial_set_embedding

This is already in the PR.

falexwolf:
Great!

Yes, I would have expected that the adjacency matrix will differ slightly and hence, `test_paga_paul15` fails. We'll need to rerun and upload https://scanpy-tutorials.readthedocs.io/en/latest/paga-paul15.html with the new version in that case and also update the tests.

falexwolf:
One other thing.

We'd like to have two new convenience functions:

```
def neighbors_update(adata, adata_new)
def umap_update(adata, adata_new)
```

The first maps the new data into the existing neighbor graph based on the chosen latent representation.

The second maps the new data into the existing UMAP embedding.

For the second function, one just needs to find a good way of wrapping
```
model = umap.UMAP(seed=1234)
model.fit(X)
model.transform(new_X)
```
For the first, I'm not quite sure how easy it is easy. I'm using `pynndescent` for it, which will become UMAP's dependency at some point, but isn't yet.

Maybe what UMAP does internally is already sufficient, but I don't know.

Can you investigate and if it's easy cover in this PR? If it's tricky, let's wait for another PR.

tomwhite:
This looks good to me.

Do you intend to remove `scanpy/neighbors/umap` as a part of this PR?

falexwolf:
Yes, `scanpy/neighbors/umap` should definitely be removed.

scanpy/tools/_umap.py
```diff
@@ -131,7 +132,9 @@ def umap(
         n_epochs,
         init_coords,
         random_state,
-        max(0, verbosity-3))
+        metric="euclidean",
```
falexwolf: I don't think that a metric makes any sense here. We're just embedding the graph that had already been computed before. Are you sure that `simplicial_set_embedding` still takes a metric as an argument?

Koncopd: Yes, unfortunately it does. Should i add metric as an argument to `umap`?

falexwolf: Ok, just looked it up, it's just used for the initialization.

Same as the `data` argument, which, btw, is not correct to be set to `adata.X`. It should be set to the representation chosen in `pp.neighbors` and the metric also needs to be chosen consistently with it.

I'd recommend to stare the params of the `neighbors` call as `.uns['neighbors']['params']` as a dict as in several other tools, too and initizalize from that.

scanpy/tools/_umap.py
```diff
@@ -118,9 +118,10 @@ def umap(
         init_coords = init_pos
     from sklearn.utils import check_random_state
     random_state = check_random_state(random_state)
-    n_epochs = maxiter
+    n_epochs = 0 if maxiter is None else maxiter
```
falexwolf, line 121: 0 epochs will result in no embedding. But `None` is not legit anymore?

Koncopd: It was changed, now 0 is the same as None before.

Koncopd: https://github.com/lmcinnes/umap/blob/a858c6322a3e682d8daf9c17e13ac023f3e18cfa/umap/umap_.py#L921

scanpy/tools/_umap.py
Koncopd: Hm, in the [description ](https://github.com/lmcinnes/umap/blob/a858c6322a3e682d8daf9c17e13ac023f3e18cfa/umap/umap_.py#L856) it says

> data: array of shape (n_samples, n_features)
> The source data to be embedded by UMAP

It is used in [spectral_layout](https://github.com/lmcinnes/umap/blob/9342e43ad56cb93e63089997fb84e6484f8b1f24/umap/spectral.py#L199) only, it uses data and metric only when the number of connected components is bigger than 1. And it also seems from [the code](https://github.com/lmcinnes/umap/blob/9342e43ad56cb93e63089997fb84e6484f8b1f24/umap/spectral.py#L49) that this is exactly `adata.X`. Or maybe not `adata.X`, but the data which was used for embedding, `adata.X` could be changed already from that state. So what can we do, froze `adata.X` which were used for embedding?

falexwolf: What we should do is this
> I'd recommend to store the params of the `pp.neighbors` call as `.uns['neighbors']['params']` as a `dict` as in several other tools, too and initialize from that.

So, you'll have a `use_rep` param in that dict. And that param tells you which representation intends to use for the embedding (used for computing the neighborhood graph/simplicial set).

scanpy/tools/_umap.py
falexwolf: Thanks!

scanpy/tools/_umap.py
Koncopd: Yes, thank you, i see . Will do like this.

Koncopd: Just forgot that there can be another representation to use.

ivirshup:
@falexwolf @Koncopd any chance you've had time to look into the projection of new data into an existing UMAP yet? Additionally, would this be rolled out with an equivalent `pca_update`? I'd be interested in taking some part of implementing this if you haven't worked it out already.

Koncopd:
Hi, @ivirshup.
No, i haven't looked into this yet, so if you have time and ideas, i'll appreciate any help.

ivirshup:
Alright! I've got a little example case I'd probably be using for a test case [here](https://gist.github.com/ivirshup/2a0d9a785339b719e7d372027ae2df31) (doublet prediction by simulation and projection).

My current thoughts:

* Since we need to be working in the same feature space, we'll at least need PCA projection, but this is pretty easy:

<details>
<summary> Basic PCA projection </summary>

```python
def pca_update(tgt, src, inplace=True):
    # TODO: Make sure we know the settings (just whether to center?) from src
    if not inplace:
        tgt = tgt.copy()
    if sparse.issparse(tgt.X):
        X = tgt.X.toarray()
    else:
        X = tgt.X.copy()
    X -= np.asarray(tgt.X.mean(axis=0))
    tgt_pca = np.dot(X, src.varm["PCs"])
    tgt.obsm["X_pca"] = tgt_pca
    return tgt
```

</details>

* Are you planning on storing the UMAP object in the AnnData? That would make transformation easier, but I see how on-disk representation could get complicated.
* What order should we do this in? Would you like everything to be accomplished by this PR or should we break it up?
* Are we introducing a general transfer learning api? Probably worth considering that a bit. Some relevant questions:
  * Does the syntax still work for cases other than 1-to-1 transfer? 
  * How do we deal with concatenation/ joins? The current `concatenate` doesn't join things like `obsm`.
  * Alternatively, does everything have to be in the same AnnData? It would solve issues with having `var` be the same, but could complicate a lot of other code (many functions would need some kind of masking argument).

falexwolf:
@ivirshup, fantastic! Yes, it would be great if you could work on this. We should also keep it separate from this PR, in this case. This PR should remain concerned with getting rid of the duplicate code.

I'll open up an issue.

Koncopd:
Added the initial version of `neighbors_update`, only works with `use_rep='X' ` in `sc.pp.neighbors` for now.

scanpy/neighbors/__init__.py
```diff
@@ -111,6 +112,66 @@ def neighbors(
...
+    else:
+        rp_forest = None
+    train = adata.X
+    test = adata_new.X
+
+    init = initialise_search(rp_forest, train, test, int(k * queue_size), random_init, tree_init, rng_state)
+    result = search(train, search_graph.indptr, search_graph.indices, init, test)
+
+    indices, dists = deheap_sort(result)
+    return indices[:, :k], dists[:, :k]
```
ivirshup, line 173: Just checking, since downstream functions like `leiden` use the `"connectivities"` weighted graph, are you planning on returning that as well?

Koncopd: For now this is just an analog of
```
neighbors = NNDescent(adata1.X)  # construct first graph
neighbors.query(adata2.X)  # lookup second data
```
But calculating `connectivities` should be possible, yes.

falexwolf:
This looks good! :smile:

Storing the forest in the AnnData is good! It should also be compatible with the updates the @tomwhite plans on UMAP and pynndescent (UMAP will depend on pynndescent) as that should be the most basic object to store when to enable queries later on...

But I would not store the "forest" in a default neighbors call. Or do you have any estimate on how large it is?

Great work!

falexwolf:
@Koncopd, can we merge this without the `neighbors_update` function and without writing the `rp_forest` to the AnnData object? Your code is good, but we should put it into another PR.

> Can you investigate and if it's easy cover in this PR? If it's tricky, let's wait for another PR.

Is what I wrote in the beginning. I think it turned out tricky and is a case for https://github.com/theislab/scanpy/issues/562#issuecomment-487409358. So, let's keep this PR really simple and just be about removing the legacy code.

Your statement about "all tests pass except for the PAGA tests" is still true? Did you manually inspect the PAGA notebook and does it look consistent? Just a few cosmetic things should have changed, I guess.

If yes, we'll merge this, now that `1.4.1` is out.

Koncopd:
@falexwolf 
Yes, i inspected the notebook, everything looks the same. Also this PR passes all tests now. So, yes, i think it can be merged

scanpy/neighbors/__init__.py
```diff
@@ -92,8 +93,17 @@ def neighbors(
...
+    adata.uns['neighbors']['params']['metric'] = metric
+    if metric_kwds:
+        adata.uns['neighbors']['params']['metric_kwds'] = metric_kwds
+    if use_rep is not None:
+        adata.uns['neighbors']['params']['use_rep'] = use_rep
+    if n_pcs is not None:
+        adata.uns['neighbors']['params']['n_pcs'] = n_pcs
     adata.uns['neighbors']['distances'] = neighbors.distances
     adata.uns['neighbors']['connectivities'] = neighbors.connectivities
+    if neighbors.rp_forest is not None:
```
falexwolf, line 106: But, as mentioned, I'd not at the `rp_forest` object at this stage. Makes sense? Can you remove it?

falexwolf:
Cool! But, can you address the comment above? And, what about all these strange conflicts? There shouldn't be any in `scanpy/neighbors/` as all of these files get removed. Could you fix the conflict in `scanpy/tools/_umap.py`?

Koncopd:
@falexwolf , yes, i'll remove the commits with rp_forest and update_neighbors and resolve the conflict today. The conflicts are because of [this commit](https://github.com/theislab/scanpy/commit/0401341877e586e36412db1e178601bf1f545037), i believe.

scanpy/tools/_umap.py
```diff
@@ -1,5 +1,5 @@
 from ._utils import get_init_pos_from_paga, choose_representation
-from .. import settings
+from .._settings import settings
```
falexwolf, line 2: This should not have been necessary. It should be `from .. import settings`...

scanpy/neighbors/__init__.py
```diff
@@ -636,9 +636,9 @@ def compute_neighbors(
             if X.shape[0] < 4096:
                 X = pairwise_distances(X, metric=metric, **metric_kwds)
                 metric = 'precomputed'
-            knn_indices, knn_distances, forest = compute_neighbors_umap(
+            knn_indices, knn_distances, _ = compute_neighbors_umap(
                 X, n_neighbors, random_state, metric=metric, metric_kwds=metric_kwds)
-            self._rp_forest = _make_forest_dict(forest)
+            #self._rp_forest = _make_forest_dict(forest)
```
falexwolf, line 641: OK, looks good!

falexwolf:
So, is this good to go? Waiting for your OK!

Koncopd:
@falexwolf 
It is ready for merge.
However before i checked only paul15.ipynb, it is the same.
But for paga-paul15 there are serious differences.
The upper graph is new
image
Denoising the graph
image

scanpy/tools/_umap.py
Koncopd: It is the same as in the master branch
https://github.com/theislab/scanpy/blob/9b522f54e0f839e1a0c9874ca658400bfe79a894/scanpy/tools/_umap.py#L2

falexwolf: Well, it doesn't hurt, I don't who put it there. But it should be `from ..settings import settings`. :) Don't worry...

falexwolf:
I tested myself and obtained exactly the same results. :)

You probably don't have the FA2 package installed, that's why your graph look different... :)

I'm merging this! Awesome work!

falexwolf:
@tomwhite, we're now fully depending on UMAP. :smile:

tomwhite:
That's great! Thanks @Koncopd and @falexwolf

Issue #577, opened 2019-04-01
Scatter plots broken on master


ivirshup:
To recreate:

```python
import scanpy as sc
print(sc.__version__)  # 1.4+222.gdd052ed
pbmcs = sc.datasets.pbmc3k()
sc.pp.pca(pbmcs)
sc.pl.pca(pbmcs, color="MALAT1")
```

<details>
<summary>Traceback:</summary>

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-1-a8a0d93afa77> in <module>
      3 pbmcs = sc.datasets.pbmc3k()
      4 sc.pp.pca(pbmcs)
----> 5 sc.pl.pca(pbmcs, color="MALAT1")

~/github/scanpy/scanpy/plotting/_tools/scatterplots.py in pca(adata, **kwargs)
    413     If `show==False` a `matplotlib.Axis` or a list of it.
    414     """
--> 415     return plot_scatter(adata, 'pca', **kwargs)
    416 
    417 

~/github/scanpy/scanpy/plotting/_tools/scatterplots.py in plot_scatter(adata, basis, color, gene_symbols, use_raw, sort_order, edges, edges_width, edges_color, arrows, arrows_kwds, groups, components, layer, projection, color_map, palette, size, frameon, legend_fontsize, legend_fontweight, legend_loc, ncols, hspace, wspace, title, show, save, ax, return_fig, **kwargs)
    149             adata, value_to_plot, layer=layer,
    150             groups=groups, palette=palette,
--> 151             use_raw=use_raw, gene_symbols=gene_symbols,
    152         )
    153 

~/github/scanpy/scanpy/plotting/_tools/scatterplots.py in _get_color_values(adata, value_to_plot, groups, palette, use_raw, gene_symbols, layer)
    726         elif use_raw and value_to_plot in adata.var_names:
    727             color_vector = adata.raw[:, value_to_plot].X
--> 728         elif value_to_plot in adata.raw.var_names:
    729             color_vector = adata[:, value_to_plot].X
    730 

AttributeError: 'NoneType' object has no attribute 'var_names'
```

</details>

Looks like the `raw` attribute is being checked when it's not there. @fidelram, looks like you might have been working on this recently?

flying-sheep:
whoops, now we both submitted a fix.

PR #578, opened 2019-04-01
Fix for #577

ivirshup:
I think this should fix it – it fixes it on my machine and the tests pass.

There's no reason to access raw if `use_raw` isn't passed, right?

### DIFF
--- a/scanpy/plotting/_tools/scatterplots.py
+++ b/scanpy/plotting/_tools/scatterplots.py
@@ -725,7 +725,7 @@ def _get_color_values(adata, value_to_plot, groups=None, palette=None, use_raw=F
             color_vector = adata[:, value_to_plot].layers[layer]
         elif use_raw and value_to_plot in adata.var_names:
             color_vector = adata.raw[:, value_to_plot].X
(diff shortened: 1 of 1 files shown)


scanpy/plotting/_tools/scatterplots.py
```diff
@@ -725,7 +725,7 @@ def _get_color_values(adata, value_to_plot, groups=None, palette=None, use_raw=F
             color_vector = adata[:, value_to_plot].layers[layer]
         elif use_raw and value_to_plot in adata.var_names:
             color_vector = adata.raw[:, value_to_plot].X
```
flying-sheep, line 727: are you sure that the `.raw` shouldn’t be inserted in the test for this branch instead? (<code>elif use_raw and value_to_plot in adata<strong>.raw</strong>.var_names:</code>). in my fix in #579 I do that.

ivirshup: That's probably right. I just made a change so the bug would stop happening and I could work on other stuff. We can go with yours.

PR #583, opened 2019-04-01
Marker overlap fix

LuckyMD:
Noticed that I did not normalized as intended, and made the input dictionary more flexible.

Now:
1. Normalization is not just performed so that rows/columns sum to 1, but instead over the number of marker genes in the reference/the number of marker genes used from the data.
2. Reference marker dictionaries now accept `Union[Dict[str, set], Dict[str,list]]`. Dictionaries of lists are easier to use in other applications, like scoring based on gene sets.

Still no idea why Travis is failing though :/.

### DIFF
--- a/scanpy/tests/test_marker_gene_overlap.py
+++ b/scanpy/tests/test_marker_gene_overlap.py
@@ -22,11 +22,6 @@ def test_marker_overlap_base():
 
     t1 = sc.tl.marker_gene_overlap(test_data, marker_genes)
 
-    print(t1)
-    print(marker_genes)
-    print(test_data.uns['rank_genes_groups']['names']['c0'])
-    print(test_data.uns['rank_genes_groups']['pvals_adj']['c1'])
-    
     assert t1['c0']['type 1'] == 3.0
     assert t1['c1']['type 2'] == 3.0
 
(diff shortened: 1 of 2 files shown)


LuckyMD:
Okay... I have no idea what else I can do... 

I put print statements into the tests and saw that the assignment of names of the `adata.uns['rank_genes_groups']['names']` recarrays is not in the expected order when the tests fail. That's why I put in explicit names into the test data and started testing by these names. For some reason the tests fail when I take the print statements out, but they pass when the print statements are in there... so I can't even look at why they are failing anymore...

I may continue to play with this when I have some more time tonight, but I need to focus on some other things atm. If you have any ideas @flying-sheep @ivirshup @falexwolf, I'd be super keen to hear them.

Should I just leave print statements in the tests so that it's super verbose when tests fail?

falexwolf:
Malte, don't you have `pytest` installed locally? Debugging using all these `added prints` etc. commits  doesn't help maintain a clean git history. :wink:

Is it possible that there is any ambiguity regarding floating point precision? It's a bit hard for me to debug this.

In case you don't have python 3.5 installed. Simply do `conda create -n py35 python=3.5`. Calling `pytest scanpy/tests/marker_gene_overlap.py` should rapidly reveal what's going on. Or simply debugging this in a notebook.

Thank you and sorry that this causes trouble!

LuckyMD:
I don't have `pytest` installed locally (will change that), and the plan was to emulate the Travis python 3.5 environment, but I'm not sure what versions of all the dependencies are in there.

I've been debugging in a notebook, but it always works there... at least with python 3.6. I'll try just creating a conda python 3.5 env to see what happens when I do that. Chances are it will always work locally as well though... hence my remote debugging. Sorry for that...

Previous print statements have shown that the order of covariates is just different sometimes in the recarrays. So I thought it would all be fixed with 5b602f5.

gokceneraslan:
This might be related to the fact that in Python 3.6 dictionaries keep the keys in the order of insertion (as an implementation detail, which became a real feature in 3.7). If the keys in recarrays are looked up via dictionaries that might be the problem here.

LuckyMD:
@gokceneraslan that might have been an issue before... however I have now specified the tests via column and row names, and generated named recarrays, so I'm still looking now...

PR #585, opened 2019-04-02
Fixes scanpy import error on master

fbrundu:
I am not sure if it has been already addressed.
This should fix the following import error of scanpy from master, due to missing `__init__.py` in external. Probably a more clean solution would be to wrap the import for external in a try/except block.

```python
ModuleNotFoundError                       Traceback (most recent call last)
 in 
----> 1 import scanpy as sc
      2 sc.settings.verbosity = 3  # verbosity: errors (0), warnings (1), info (2), hints (3)
      3 sc.settings.set_figure_params(dpi=200)  # low dpi (dots per inch) yields small inline figures
      4 sc.settings.figdir = out('fig_supp')

~/.pyenv/versions/3.6.5/Python.framework/Versions/3.6/lib/python3.6/site-packages/scanpy/__init__.py in 
     31 from . import preprocessing as pp
     32 from . import plotting as pl
---> 33 from . import datasets, logging, queries, settings, external
     34 
     35 from anndata import AnnData

~/.pyenv/versions/3.6.5/Python.framework/Versions/3.6/lib/python3.6/site-packages/scanpy/external/__init__.py in 
----> 1 from . import tl
      2 from . import pl
      3 from . import pp
      4 
      5 from .. import _exporting as exporting

~/.pyenv/versions/3.6.5/Python.framework/Versions/3.6/lib/python3.6/site-packages/scanpy/external/tl.py in 
      2 from ..tools._phate import phate
      3 from ..tools._phenograph import phenograph
----> 4 from ._tools._palantir import palantir

ModuleNotFoundError: No module named 'scanpy.external._tools'
```


flying-sheep:
This can’t be it, `__init__.py` is [optional since Python 3.3](https://www.python.org/dev/peps/pep-0420/).

I also don’t see that import error, `import scanpy` works perfectly. Can you specify how you got that?

```py
>>> import scanpy, anndata
>>> scanpy.external.tl.palantir(anndata.AnnData())
ImportError: 
please install palantir: 

        git clone git://github.com/dpeerlab/Palantir.git
        cd Palantir
        sudo -H pip3 install .
```

fbrundu:
I had this problem importing `import scanpy as sc`.
I'll update you if this problem persists.

flying-sheep:
Sure! It would need another fix though, as I don’t think it’s possible that what you did here changes anything.

jarny:
@fbrundu I just had the exact same error after installing from cloned master, so just wondering if it's working for you now.

fbrundu:
@jarny I have the same error still, but when testing on travis it doesn't fail so I have no clue. Locally I create the `__init__.py` file though to make it work.

PR #591, opened 2019-04-05
Add pipenv files to .gitignore

ivirshup:
Would anyone mind if these were added? I've used this to manage my `scanpy` dev environment, but would like it to stop showing up in `git status`.


falexwolf:
Of course that's fine. You can simply push such things to master. ;)

PR #592, opened 2019-04-05
Minor fixes, mostly io related

ivirshup:
This is a bit of a grab bag, but is mostly `io` related. This started out as me trying to learn some vscode git integration, but turns out it's not great at figuring out what lines changed. Apologies for any weird stuff in the commits.

Main changes:

* I've made the tests for `sc.datasets` more thorough. Now they actually check the data looks kinda okay, rather than whether they throw an error.
* I've removed cache-ing in a few places
   * The `read_10x_*` tests, where that definitely shouldn't have been happening
   * In a couple of the `sc.datasets`. I'd be willing to go back on this, but we shouldn't let them use the cache during testing.


falexwolf:
Wow, great! You're doing so much hard work. Incredible!

ivirshup:
Haha, this PR was definitely me procrastinating on some writing!

PR #594, opened 2019-04-06
removed normalize_quantile

falexwolf:
Hi @Koncopd, as discussed, we should probably not call this function using this name. As we might have multiple conceptually more different normalization functions in the future, we should probably just be satisfied with this one normalization function for now. The most natural thing is to have a `quantile` param in `normalize_total`. What do you think?

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -36,7 +36,6 @@ For visual quality control, see :func:`~scanpy.pl.highest_expr_gens` and
    pp.log1p
    pp.pca
    pp.normalize_total
-   pp.normalize_quantile
    pp.regress_out
    pp.scale
    pp.subsample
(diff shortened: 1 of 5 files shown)


Koncopd:
Hi, @falexwolf .
Seems right to me. Maybe it is better to rename `quantile` argument to `fraction`?

falexwolf:
You're right `fraction` is better! Can you change it directly on this branch?

Also, I did not find a test for this. Can you add one? You can essentially copy your example and the test that's there for `normalize_per_cell`.

Thank you!

Koncopd:
@falexwolf 
It is strange that there are no tests for this. I thought i added them...
I will try to do all these thing in two days. Sorry for the delays.

falexwolf:
Ok, thank you! Maybe I just didn't find them. If so, please point me to them. Merging this in the meanwhile, you can add the tests in a new PR.

PR #597, opened 2019-04-08
Fix standardscale NA bug

### LINKED ISSUE #590: All-zero genes show up as white columns in matrixplots if standard_scale='var' is given
Looks like this:

image

Not that it's wrong but it'd be prettier to show as regular as a regular grid.

gokceneraslan:
Fixes #590.


flying-sheep:
thank you!

PR #602, opened 2019-04-11
Downsample improvements

ivirshup:
Just a minor improvement. Now you can specify the total counts to downsample to on a per-cell basis by passing an array for `counts_per_cell`. This is so I don't have to split and merge an AnnData object when I want multiple distributions of counts per cell.


falexwolf:
Great, thank you!

PR #608, opened 2019-04-15
Scanpydoc flat return

falexwolf:
https://github.com/theislab/scanpydoc/issues/7


falexwolf:
The central commit is: https://github.com/theislab/scanpy/pull/608/commits/08111b87ca1274138e6c25963c412b44f7ab57c1

Unfortunately, this is not a great solution... But at least it's one that doesn't mess up the `numpydoc`-style-formatted return sections and re-enables return sections that have been written using prose, bulleted lists etc. (those went broke almost half a year ago after changing from `numpydoc` to `sphinx.ext.napoleon`).

flying-sheep:
Don’t copy code please. just instroduce some flag parameter or so.

flying-sheep:
How is this different than the scanpydoc-flat-return branch?

flying-sheep:
How is this different than my original code in the scanpydoc-flat-return branch?

PR #610, opened 2019-04-15
Old, simple method for flat return sections

flying-sheep:
Everything looks like it should:

- https://scanpy.readthedocs.io/en/return-formatting/api/scanpy.pp.calculate_qc_metrics.html

    Here the rst is weird. Just get rid of the `*`!

- https://scanpy.readthedocs.io/en/return-formatting/api/scanpy.pp.normalize_total.html

    Short prose, looks fine.

- https://scanpy.readthedocs.io/en/return-formatting/api/scanpy.tl.leiden.html

    This should be converted into a definition list as well.

- https://scanpy.readthedocs.io/en/return-formatting/api/scanpy.tl.dpt.html

    Mixed. Looks great.

No rtype interfering anywhere

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,16 +47,22 @@ or updates `adata` with normalized version of the original
 ```
 
 Many functions also just modify parts of the passed AnnData object,
-like e.g. [`tl.leiden`](https://scanpy.readthedocs.io/en/latest/api/scanpy.tl.leiden.html)
+like e.g. [`tl.dpt`](https://scanpy.readthedocs.io/en/latest/api/scanpy.tl.dpt.html)
 
 ```rst
 Returns
 -------
-`adata.obs[key_added]`
-    Array of dim (number of samples) that stores the subgroup id (`'0'`, `'1'`, ...) for each cell.
+Depending on `copy`, returns or updates `adata` with the following fields.
 
-`adata.uns['leiden']['params']`
-    A dict with the values for the parameters `resolution`, `random_state`, and `n_iterations`.
+If `n_branchings==0`, no field `dpt_groups` will be written.
+
+dpt_pseudotime : `pd.Series` (`adata.obs`, dtype `float`)
+    Array of dim (number of samples) that stores the pseudotime of each
+    cell, that is, the DPT distance with respect to the root cell.
+dpt_groups : `pd.Series` (`adata.obs`, dtype `category`)
+    Array of dim (number of samples) that stores the subgroup id ('0',
+    '1', ...) for each cell. The groups  typically correspond to
+    'progenitor cells', 'undecided cells' or 'branches' of a process.
 ```
 
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,35 +16,48 @@ We stick to [PEP 8](https://www.python.org/dev/peps/pep-0008) and to this [edito
 
 ### Docs
 
-We use the numpydoc style for writing docstrings. Either take a look at any Scanpy or Numpy function or [here](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).
+We use the numpydoc style for writing docstrings.
+Either take a look at any Scanpy or Numpy function or
+[here](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html).
 
 The `Params` abbreviation is a legit replacement for `Parameters`.
 
-The `Returns` section deserves special attention: you can either use the standard numpydoc way of populating it
-```
+The `Returns` section deserves special attention:
+While you can either use the standard numpydoc way of populating it,
+you can also just write any prose:
+
+```rst
 Returns
 -------
 type of return value 1
     description of return value 1
+
 type of return value 2
     description of return value 2
 ```
-or you can write plain RST-formatted prose, for instance from [`pp.normalize_total`](https://scanpy.readthedocs.io/en/latest/api/scanpy.pp.normalize_total.html)
-```
+
+or like [`pp.normalize_total`](https://scanpy.readthedocs.io/en/latest/api/scanpy.pp.normalize_total.html)
+
+```rst
 Returns
 -------
 Returns dictionary with normalized copies of `adata.X` and `adata.layers`
 or updates `adata` with normalized version of the original
 `adata.X` and `adata.layers`, depending on `inplace`.
 ```
-or from [`tl.leiden`](https://scanpy.readthedocs.io/en/latest/api/scanpy.tl.leiden.html)
-```
+
+Many functions also just modify parts of the passed AnnData object,
+like e.g. [`tl.leiden`](https://scanpy.readthedocs.io/en/latest/api/scanpy.tl.leiden.html)
+
+```rst
 Returns
 -------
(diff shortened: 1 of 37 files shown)


flying-sheep:
Wait, aren’t the docs supposed to be like this?

```py
def myfunc():
    """\
    First line
    ...
    """
```

falexwolf:
No, they aren't. Most of them start with `"""my title`. But we can happily migrate... see Slack...

flying-sheep:
Well, it wasn’t that one anyway. I’ll figure it out

flying-sheep:
Ah, the problem was that the string actually contained that return type!

Theoretically that should have parsed as a 1-element definition list, but I don’t think it’s a problem that it doesn’t.

falexwolf:
> Ah, the problem was that the string actually contained that return type!

This is the standard way of writing a numpydoc returns section. Also see https://scanpy.readthedocs.io/en/return-formatting/api/scanpy.pp.filter_cells.html, for instance.

This solution is dropping support for them.

Also, do you by chance have another simple solution for having the styling of the return sections similar to the parameters section (what `numpydoc` did 🙂)? Bold font and spacings around colons?

image

falexwolf:
Unfortunately, changing
```
.rst-content dl:not(.docutils) dl dt {
    font-weight: normal;
```
to `bold` changes a whole lot of other stuff. https://scanpy.readthedocs.io/en/return-formatting/api/scanpy.tl.dpt.html

I just see that this how scikit-learn styles it, though
image

Hence, only the spacing after the colon is the remaining inconsistency.

falexwolf:
@ivirshup @flying-sheep I would remove the `, optional` statement from the docstrings, as, what we mean with this is "a parameter has a default value". Hence, it's redundant. However, it's consistently used in all of numpy, scipy, sklearn, pandas, etc.
image

In the auto-generated type annotations, the default values miss completely, and I don't think we'll ever restore the `, optional` descriptor, there.

image

flying-sheep:
> This is the standard way of writing a numpydoc returns section. […] This solution is dropping support for them.

It certainly shouldn’t, since the definition lists *are* rendered in other cases. IDK why not here, this should render as a definition list with one item.

However, I don’t like indenting the whole section except for the first line, so in case it always works once there are multiple definition list items, I don’t worry too much here.

> Also, do you by chance have another simple solution for having the styling of the return sections similar to the parameters section (what numpydoc did :slightly_smiling_face:)? Bold font and spacings around colons?

I’ll figure it out.

> I would remove the `, optional` statement from the docstrings, as, what we mean with this is "a parameter has a default value". Hence, it's redundant. However, it's consistently used in all of numpy, scipy, sklearn, pandas, etc.

We should definitely put the defaults inline, and I also think the “optional” is redundant. What would it even mean to have “a parameter that isn’t optional but has a default value”?

I’m pretty sure people will understand it.

falexwolf:
> IDK why not here, this should render as a definition list with one item.

Hm, maybe because a new line is missing before the start of the list. I don't know now....

> However, I don’t like indenting the whole section except for the first line, so in case it always works once there are multiple definition list items, I don’t worry too much here.

Absolutely agreed! So, let's just forget about the one-item definition list.

flying-sheep:
I figured the one-item-thing out: The emitted code is:

```rst
:param copy: If an :class:`~anndata.AnnData` is passed, determines whether a copy
             is returned.
:type copy: `bool`, optional (default: `False`)

:returns: AnnData, None
              Depending on `copy` returns or updates `adata` with the corrected data matrix.
```

And since `:returns:` is part of a field list, and field lists are defined by the indentation of the *block starting in the second line*, the additional indentation of the second line is ignored.

So yes, only numpydoc-style sections with one item are affected.

flying-sheep:
OK I have something good. I’ve got an appointment now but will fix the remaining docstrings soon.

flying-sheep:
This should work for now. The problem is that this really needs to live in scanpydoc, where we have access to the fancy parsing.

My plan is to merge this now, which has a solution for the simple case of

```rst
parameter : some.type
    Description
```

Later I’ll introduce the same behavior as what scanpydoc does with the parameters:

If the return type is `...) -> Tuple[foo.bar, baz.zab]:`, then I’ll check if there’s a section like

```rst
one_identifier
    Desc
second_identifier
    Desc
```

and replace them with the same code as parameters.

----

This leaves 3 styles:

1. Prose for a single return value
2. The above for returning a tuple
3. “this function adds some AnnData.obs/var fields”

For 3., we have like 3 styles floating around and need to fix one:

1. ``**dpt_pseudotime** : :class:`pandas.Series` (`adata.obs`, dtype `float`)``
2. ``` ``adata.obs['louvain']`` (:class:`pandas.Series`, dtype ``category``) ```
3. `` `adata.uns['leiden']['params']` : `dict` ``

falexwolf:
As mentioned in the commit, fantastic! :smile:

Which style mentioned in https://github.com/theislab/scanpy/pull/610#issuecomment-484124854 did you decide for?

Most manual corrections I saw in your PR pointed to solution (1) [`**dpt_pseudotime** : :class:`pandas.Series` (`adata.obs`, dtype `float`)`]. That's fine, but we should mention it in `CONTRIBUTING.md`. 🙂What do you think, @flying-sheep?

----

Why did you decide against the sklearn-style solution, which would have been simple to get without manual fixes (adding `**name**`)? https://github.com/theislab/scanpy/pull/610#issuecomment-484027492

flying-sheep:
I think you misunderstood a bit.

1. There’s three *types* of return sections – prose, tuple, added anndata.obs/var fields.
2. There’s *one style* for tuple return sections (like parameters). I implemented an automatic way to handle simple `name : type` cases of those and manually formatted the rest. I’ll automate the other cases in scanpydoc, then we can remove the manually formatted ones.
3. There’s *3 styles* for anndata.obs/var field return sections. I left them as they are for now, since we have to decide one.

I only reformatted tuple return sections, I neither formatted nor decided on anything about anndata.obs/var field return sections.

I hope I was more clear this time :smile:

falexwolf:
This is unrelated to your post just now. :smile:

Also, `Parameters` etc. don't render as a heading (`rubric`) like the other sections (`Notes`, `Examples`, `See also` etc.) anymore. They simply render as 
```
<dl class="field-list simple">
<dt class="field-odd">Parameters</dt>
```
whereas
```
<p class="rubric">Notes</p>
```

This is the new styling decision around numpydoc, also visible here: https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html

In order to have it look consistent (as in sklearn), I think we do need
```
.rst-content dl:not(.docutils) dl dt {
    font-weight: bold;
    font-size: 16px;
```

falexwolf:
https://github.com/theislab/scanpy/pull/610#issuecomment-484427693: Great! Let's leave the remaining ones as a task for moving towards more consistency in the future. :smile:

falexwolf:
Adapted the css to enforce the consistent styling of headings on each page: https://github.com/theislab/scanpy/commit/1f579f6f745d01c599a39f38abadb8a32f95c12b

falexwolf:
Tried to meaningfully convey what you described in https://github.com/theislab/scanpy/commit/0dcf6fcd831eadfc64b1c9485d1b7021347db0b9.

flying-sheep:
Good! I fixed some typos and improved the examples a bit. Specifically when there’s something to link, they should do it! `` :class:`pandas.Series` `` instead of ``` ``pd.Series`` ```

falexwolf:
Nice! 🙂

PR #611, opened 2019-04-16
Fixed filter_cells() for inplace=False

LiBuchauer:
For `inplace=False,` the function tried to return `gene_subsets`, which is not defined here, instead of `cell_subsets`.


ivirshup:
Thanks!

PR #614, opened 2019-04-20
Add a batch_key option to HVG function to reduce the batch effects

gokceneraslan:
I added a new `batch_key` option to HVG function. If specified, it runs the HVG selection in every batch separately and then merges the list in order to reduce the batch effects by avoiding the selection of batch-specific genes. This doesn't fully correct the batch effect but reduces it.

Running the function for each batch is trivial but merging is trickier than I thought. How I do it now is as follows:

- hvg is run on each batch and resulting hvg lists are first concatenated into a single dataframe.
- The data frame is grouped by genes. mean, dispersion and normalized dispersion values are aggregated via ~~np.nanmin~~ np.nanmean. Another column which counts "in how many batches a gene is detected as hvg" is also created. ~~I'm not 100% certain about nanmin, but I think it works better than mean or max, because it forces the process to pick genes with high dispersion even in the "worst" batches.~~
- if n_top_genes is given, combined hvg lists are sorted by 1) in how many batches a gene is detected as hvg 2) normalized dispersion. normalized dispersion is used to break the ties. Then top n genes are selected as the final hvg list.
- if n_top_genes is not given, same mean and dispersion thresholds are applied to the combined hvg list.

Here is the code to see the improvement of this approach:

```python
import scanpy as sc
import numpy as np
import pandas as pd

ad = sc.read("pancreas.h5ad", backup_url="https://goo.gl/V29FNk") # adapted from scGen repo
ad.obs["cell_type"] = ad.obs["celltype"].tolist()

ad = sc.AnnData(ad.raw.X, var=ad.raw.var, obs=ad.obs)
sc.pp.normalize_per_cell(ad)
sc.pp.log1p(ad)

sc.pp.highly_variable_genes(ad, batch_key='batch')
sc.pp.pca(ad)
sc.pp.neighbors(ad)
sc.tl.umap(ad)
sc.pl.umap(ad, color=["batch", "cell_type"])
```

image

```python
sc.pp.highly_variable_genes(ad)
sc.pp.pca(ad)
sc.pp.neighbors(ad)
sc.tl.umap(ad)
sc.pl.umap(ad, color=["batch", "cell_type"])
```

image

```python
sc.pp.highly_variable_genes(ad, batch_key='batch', n_top_genes=1000)
sc.pp.pca(ad)
sc.pp.neighbors(ad)
sc.tl.umap(ad)
sc.pl.umap(ad, color=["batch", "cell_type"])
```

image

```python
sc.pp.highly_variable_genes(ad, n_top_genes=1000)
sc.pp.pca(ad)
sc.pp.neighbors(ad)
sc.tl.umap(ad)
sc.pl.umap(ad, color=["batch", "cell_type"])
```

image

Plase review thoroughly :) Tests are missing, I'll add them soon.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -121,7 +121,6 @@ def insert_function_images(app, what, name, obj, options, lines):
     path = Path(__file__).parent / 'api' / f'{name}.png'
     if what != 'function' or not path.is_file(): return
     lines[0:0] = [f'.. image:: {path.name}', '   :width: 200', '   :align: right', '']
-    print(*lines, sep='\n')
 
 
 # -- Test for new scanpydoc functionality --------------------------------------
--- a/scanpy/_settings.py
+++ b/scanpy/_settings.py
@@ -211,7 +211,7 @@ class ScanpyConfig(object):
     @datasetdir.setter
     def datasetdir(self, datasetdir):
         _type_check(datasetdir, "datasetdir", [str, Path])
-        self._datasetdir = Path(datasetdir).absolute()
+        self._datasetdir = Path(datasetdir).resolve()
 
     @property
     def figdir(self):
--- a/scanpy/tests/test_embedding_density.py
+++ b/scanpy/tests/test_embedding_density.py
@@ -18,9 +18,6 @@ def test_embedding_density():
 
 def test_embedding_density_plot():
     # Test that sc.pl.embedding_density() runs without error
-    adata = sc.datasets.blobs()
-    sc.pp.pca(adata)
-    sc.pp.neighbors(adata)
-    sc.tl.umap(adata)
+    adata = sc.datasets.pbmc68k_reduced()
     sc.tl.embedding_density(adata, 'umap')
     sc.pl.embedding_density(adata, 'umap', 'umap_density')


scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -97,90 +103,133 @@ def highly_variable_genes(
     if min_disp is None: min_disp = 0.5
     if min_mean is None: min_mean = 0.0125
     if max_mean is None: max_mean = 3
+    if max_disp is None: max_disp = np.inf
 
-    X = np.expm1(adata.X) if flavor == 'seurat' else adata.X
```
gokceneraslan: These lines are actually not removed, but indented and moved below. It's very unfortunate that we still don't have smarter diff utils.

gokceneraslan:
I just found out that raw was already log transformed. Now I fixed it and nanmin seems a bit too conservative, any ideas?

image

LuckyMD:
Hey! I think it's a great idea to put HVG calculation within batches as a pull request. This is also an important pre-processing step for `sce.mnn_correct`, which I usually run on the intersection of HVGs from each batch. Do you think you could output the intersection as well? Or would it just be a matter of subsetting HVGs that are detected in all batches from the resulting dataframe?

I'm a bit more skeptical for advertising this as a batch-correction approach though. Maybe for time-series data where you can't use other batch correction methods?

gokceneraslan:
Hey. I also thought about the intersection but didn't implement it as the default output for two reasons. 

1) it can be too harsh, especially if there is some biological variation between batches. When we sort the genes based on in how many batches they're detected as HVG and on mean normalized dispersion, there is still a chance for the user to catch such "biological" genes with a high n_top_genes value. 

2) Output of highly_variable_genes should be consistent regardless of batch_key option. So n_top_genes and mean/dispersion cutoff flavors should still work the same way. I feel like using the intersection directly as the output violates that.

However, making `'highly_variable': np.nansum` part available in adata.var is a good idea. Then users can manually make the selection more stringent by selecting genes where this value == nbatches.

LuckyMD:
I agree that using the intersection is too harsh if the only data integration/batch correction you do is this HVG filtering, but for `mnn_correct` for example you only use these HVGs to calculate the technical batch vector. In that case you ideally don't want to capture the biological variation between batch samples. For that I reckon having an option of getting the intersection would be good.

And for point 2... definitely agreed... but again, a different use case for me. The same approach (intersection of HVGs) is suggested for the new CCA in Seurat v3 btw.

gokceneraslan:
@LuckyMD Can you maybe check if the new `highly_variable_nbatches` variable in `adata.var` is useful for finding the intersections? 

`adata.var['highly_variable'] = adata.var['highly_variable_nbatches'] == len(adata.obs.batch.cat.categories)`

LuckyMD:
I'm sure it will be... Just having network issues atm, so server storage is too slow to do anything. Can't really test this atm I'm afraid :/.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -97,90 +105,135 @@ def highly_variable_genes(
...
-        # only a single gene fell in the bin and implicitly set them to have
-        # a normalized disperion of 1
-        one_gene_per_bin = disp_std_bin.isnull()
-        gen_indices = np.where(one_gene_per_bin[df['mean_bin'].values])[0].tolist()
-        if len(gen_indices) > 0:
+    if batch_key is not None:
+        sanitize_anndata(adata)
+        batches = adata.obs[batch_key].cat.categories
+        df = []
+        for batch in batches:
```
ivirshup: Would it be worth it to split calculation and argument handling into separate functions? I think it could clean up the flow control (and reduce indentation). Basically I'm thinking just put everything in the top level `else:` block into it's own function.

gokceneraslan:
Sorry my master branch is messed up, I'll send another PR.

Issue #562, opened 2019-03-26
Dotplot where sizes are proportional to p-value and the color to log2-fold change?


falexwolf:
@fidelram, as discussed today, could we adopt `pl.rank_genes_groups_dotplot` so that it reads this information from `.uns['rank_genes_groups']`?

Maybe just a simple switch? Or having arguments `color` and `size` be a choice from a selection {`pvals`, `pvals_adj`, `log2FC`, `expression`, `frac-genes-expressed`}.

fidelram:
I thing that is not so difficult to achieve this. I submit a PR soon

Fidel Ramírez

gokceneraslan:
> @fidelram, as discussed today, could we adopt `pl.rank_genes_groups_dotplot` so that it reads this information from `.uns['rank_genes_groups']`?
> 
> Maybe just a simple switch? Or having arguments `color` and `size` be a choice from a selection {`pvals`, `pvals_adj`, `log2FC`, `expression`, `frac-genes-expressed`}.

I would also love that actually 😄 

`rank_genes_groups` results (LFC, p-val etc) and things like `mean-expression`, `frac-genes-expressed` are all cluster-specific features, which reminds me of an `obs`-like structure with clusters in rows. Right now, mean expression and fractions are calculated in a private function (`_prepare_dataframe`) in `plotting/_anndata.py` but we can move this to `utils.py` or so, call it in sc.tl.rank_genes_groups() and store the resulting data frame in ad.uns?

fidelram:
@gokceneraslan the idea of a `obs`-like structure for clusters or in general for categorical data is interesting. Currently, in `.uns` we are using numpy structured arrays for ranked genes which are not so easy to navigate. A `obs`-like object using `pandas` could facilitate data manipulation and visualization.

@falexwolf should we give it a try or do you see problem with this?

LuckyMD:
If this is stored in `.uns` we would have to wait for pandas support for `.uns`.

flying-sheep:
Which is tracked in #134 I think.

ivirshup:
Why is it that .obs, .var, and .uns don't have data frames in them? `np.recarray` don't seem like a very popular data structure elsewhere.

Also, I'd like to suggest that storing all differential expression within the anndata object might get complicated, and deserve it's own class. It'd be nice if it could be easy to tell what cells and genes were compared, what exactly was being tested, and which direction is "up". That said, the results should definitely be easily accessible as a data frame.

[...]

PR #615, opened 2019-04-23
Qc metrics update

ivirshup:
Two major, and two minor, updates to qc metric calculation:

## Tests run much faster now

`test_qc_metrics.py` used to take ~30 seconds, now takes ~2.

These tests have been kinda slow for a while. This was mostly due to numba compilation. I was using `numba.njit(parallel=True)`, which cannot be cached so compilation occurred every time the tests ran. However, I expect most use cases only calculate QC metrics once in a session, and only for large datasets (at least 300,000 cells) is parallelization + compilation faster than performing the calculation in a single thread. Now a cached single threaded version is used unless the dataset is large.

## Can now calculate observation and variable metrics separately

Split the calculation of qc metrics into two functions for obs and var. These separate calls are now available as: `describe_obs` and `describe_var` after `pd.DataFrame.describe`. This is mostly to go along with my split-apply-combine experiments. In particular a use case like:

```python
(adata
    .groupby(obs="leiden")
    .apply(sc.pp.describe_var)
    .combine(...)
)
```

Where metrics like number of fraction of cells, mean expression, etc. are calculated within each group (useful for things like #562).

## Minor updates

* User can now choose to use expression from `layers` or `raw` instead of `adata.X`
* Doc updates 🤞 (am I polluting `sc.pp._docs.py` too much?)

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -1,3 +1,4 @@
+.. module:: scanpy
 .. automodule:: scanpy
 
 API
@@ -131,12 +132,11 @@ Simulations
 Plotting: PL
 ------------
 
-The plotting module :class:`scanpy.plotting` largely parallels the ``tl.*`` and a few of the ``pp.*`` functions.
+The plotting module :mod:`scanpy.plotting` largely parallels the ``tl.*`` and a few of the ``pp.*`` functions.
 For most tools and for some preprocessing functions, you'll find a plotting function with the same name.
 
-.. toctree::
-   :hidden:
-   :maxdepth: 1
+.. autosummary::
+   :toctree: .
 
    plotting
 
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -30,6 +30,8 @@ For visual quality control, see :func:`~scanpy.pl.highest_expr_gens` and
    :toctree: .
 
    pp.calculate_qc_metrics
+   pp.describe_obs
+   pp.describe_var
    pp.filter_cells
    pp.filter_genes
    pp.highly_variable_genes
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -30,8 +30,6 @@ For visual quality control, see :func:`~scanpy.pl.highest_expr_gens` and
    :toctree: .
 
    pp.calculate_qc_metrics
-   pp.describe_obs
-   pp.describe_var
    pp.filter_cells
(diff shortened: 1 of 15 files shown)


falexwolf:
Cool! :smile:

But is it possible to stay within the naming scheme? Or maybe even better, to avoid cluttering the namespace with many similar functions, add a parameter to `calculate_qc_metrics` that allows to restrict to `obs` and `var` if wanted?

Updating `docs/release_notes.rst` would also be great!

ivirshup:
I'd have to think about the naming scheme, since its something I've been going back and forth on recently. I'm playing around with these semantic a bit over in my [mantis](https://github.com/ivirshup/mantis) repo, but I don't think I'm ready to make a call on which one I like best. Right now I'm split between following [plyexperiment](https://github.com/sa-lee/plyexperiment) and `xarray`.

I think I'd like to keep them as separate functions, since the `calculate_qc_metrics` function already has about as complicated a relationship as I'd want with parameter values and return type.

How about these keep their current name for now, but we don't export these functions beyond `_qc.py`? This will let me play around with their naming scheme and the usefulness a bit more, while scanpy gets the faster, more thorough tests and improved `calculate_qc_metrics`?

falexwolf:
> How about these keep their current name for now, but we don't export these functions beyond _qc.py?

That's OK for me! I thought about naming these `calculate_qc_metrics_obs` and `..._var`, but that's not great, either... `describe_obs` and `describe_var` are probably better names compared to  `calculate_qc_metrics`... So let's just not add a new bunch of functions for now and think about a lasting solution for Scanpy 2, where we can also break backwards compat.

The `mantis` repo looks fantastic!!!

ivirshup:
@falexwolf, this is good now, yeah?

ivirshup:
@flying-sheep and idea what's up with this build error? Docstrings are failing tests, but look fine to me. Seems related to https://github.com/theislab/scanpy/commit/3cacdc87ab47bae70b415e93f2fea74a018c39e2.

falexwolf:
Yes, it's good to go! Sorry about the trivial conflict in the release notes: I just made 1.4.2 based on the changes of the last two weeks on master, which I had the chance to test in the past week...

ivirshup:
No problem, thanks!

I think conflicts in release note are going to be common. Maybe those commits could be separate from the PR?

PR #616, opened 2019-04-23
Minor fixes grab-bag

ivirshup:
Three minor fixes:

1. Remove noisy print statement from doc builds. Looked like it might have been left over from debugging, and doesn't seem to break anything. @flying-sheep, does that sound right?
2. When setting `sc.settings.datasetdir`, I'd meant to use `Path.resolve` instead of `Path.absolute`.
3. Sped up embedding density test by using a dataset where umap was precomputed. Is this fine @LuckyMD?


flying-sheep:
sounds good from my side

LuckyMD:
great!

ivirshup:
Great, thanks for the quick responses!

Issue #613, opened 2019-04-18
pl.umap/tsne/draw_graph colored by principal component or diffusion component


LisaSikkema:
Dear scanpy developers,

I would like to color my tsnes, umaps and graph layouts by specific principal components or diffusion components. Any chance you could add an option to use columns from obsm objects as colors for these plots? (Or maybe that's already possible and I missed it...?)
Thanks for your help.

Best,
Lisa

LuckyMD:
Hi Lisa,
I reckon this is a sensible request for cases when you give PCs or DCs a biological interpretation. It shouldn't be hard to implement I guess. But @fidelram will have a better overview of this.

A temporary workaround would be to do the following:
```
adata.obs['pc1'] = adata.obsm['X_pca'][:,0]
sc.pl.umap(adata, color='pc1')
```

gokceneraslan:
I think it'll be very cool to make this lookup dynamic somehow e.g. `sc.pl.umap(adata, color='X_PCA:1')`.

ivirshup:
@gokceneraslan, how about a tuple like `({obsm_key}, {obsm_value_column})`? It's more verbose, but I think more flexible, particularly if `obsm` starts to allow more kinds of values.

Here's an example (using some stuff on personal branches):

<img width="703" alt="image" src="https://user-images.githubusercontent.com/8238804/56583661-6c62a680-661d-11e9-8314-ca0b0bf8a309.png">

PR #619, opened 2019-04-25
Add some data access helpers to utils

ivirshup:
This adds two new convenience functions to `utils`.

## `obs_values_df`

Basically does the data access part of the scatter plots (actually copied the core of the code from there). Basically, lets you get a data frame of values from obs, obsm, and expression matrix back as a dataframe. I'd planned on this being the data access part of `ridge_plot` PR, but I've found it generally useful for data access. Also finding a feature-ful KDE that isn't buggy has been an issue for the ridge plots.

This uses the obsm access I had suggested to @gokceneraslan in #613.

I'm also open to adding a `var_values_df` to this PR, I just haven't had a use case yet.

## `rank_genes_groups_df`

Returns a dataframe of differential expression results, because accessing DE results right now is a pain. This was a part of #467, but I can just remove it from there.

## Whats left to do.

Docs, but it's boilerplate. Do we have centralized docstrings for things like `gene_symbols`, `use_raw`, `layers`, and `adata`?


falexwolf:
This looks very good!

No, we don't yet have centralized docs for `gene_symbols`, `adata` etc. but of course, we should have them. Maybe even just in the root `scanpy/_docs.py`. What do you think?

ivirshup:
Thanks!

I've updated the docs, but it turned out not much was actually shared. Where should I put these in the api docs? A new `utils` section, or something under `Further modules`? I'm thinking I'd just include `rank_genes_groups_df` and `obs_values_df` on the site.

scanpy/utils.py
```diff
@@ -563,6 +569,183 @@ def unique_categories(categories):
...
+    adata
+        Object to get results from.
+    group
+        Which group (key from :func:`scanpy.tl.rank_genes_groups` `groupby`) to
+        return results from.
+    key
+        Key differential expression groups were stored under.
+    pval_cutoff
+        Minimum adjusted pval to return.
+    logfc_min
```
falexwolf: After some discussions, we thought that `log2FC` is the most unambiguous and most commonly used name, and we would rename the slot this way in `v2.0` (https://github.com/theislab/scanpy/issues/453#issue-404768505). Do you agree? Would you adapt the parameter names?

scanpy/utils.py
```diff
@@ -563,6 +569,183 @@ def unique_categories(categories):
...
+    if logfc_max is not None:
+        d = d[d["logfoldchanges"] < logfc_max]
+    if gene_symbols is not None:
+        d = d.join(adata.var[gene_symbols], on="names")
+    return d
+
+
+# Would an array be faster?
+@doc_params(raw_layer_params=_docs.doc_raw_layers)
+def obs_values(
```
falexwolf: Is there some duplicatation with https://github.com/theislab/anndata/blob/34f4eb63710628fbc15e7050e5efcac1d7806062/anndata/base.py#L1464?

I think we could have a public function in AnnData for this purpose.

falexwolf: Your implementation in combination with `obs_values_df` is definitely better. But I really think it should go into AnnData, next to `.to_df()` which right now just gives the data matrix, but one could give it the options you have below.

ivirshup: Definitely duplication, I hadn't realized we had that function in AnnData and was just going off how we assign colors for scatter plots. I agree it's better to have one function for this, and AnnData makes sense for where to put it.

Do all the current argument make sense for AnnData? For example, what about `gene_symbols`? If AnnData is meant to be agnostic to datatype, I could handle resolving gene_symbols in `obs_values_df`?

Also, these functions have different conventions for `layers`. Which one should we standardize on?

scanpy/utils.py
ivirshup: Sure! No strong opinion on this, as long as we're definitely calculating a log2 fold change.

falexwolf: Yes, we're definitely calculating log2. I think this is an established convention.

scanpy/utils.py
falexwolf: Right, `gene_symbols` doesn't make sense for AnnData! Your suggestion makes sense.

Maybe Fidel didn't use the original https://github.com/theislab/scanpy/blob/9b522f54e0f839e1a0c9874ca658400bfe79a894/scanpy/plotting/_anndata.py#L311 in his functions? Unfortunate if I didn't notice. I don't have strong opinion on the convention for layers as long as it's there. We could also have a Slack discussion with Fidel, if you think there might be issues.

ivirshup:
Since the `obs_values_df ` will now depend on an updated version of AnnData, I'm thinking I'll move this version of `rank_genes_groups_df` over to #467 so that can get merged.

Edit: Actually, this isn't the case since we'll need backwards compatibility anyways, nvm

scanpy/utils.py
ivirshup: Just checked to make sure I did it right, but turns out I didn't...

Still: `log2fc` or `log2FC`? My pinkies would prefer less caps, but I'd probably also tab complete it most times.

ivirshup:
Does `obs_values_df` need the `_df`, or could it be `obs_values`?

scanpy/utils.py
ivirshup: Ah, I was looking at this function: https://github.com/theislab/scanpy/blob/9e4e5ee02e04cf618872d9b098e24f0542e8b227/scanpy/plotting/_tools/scatterplots.py#L651-L736

I don't think there's an issue per-say, I just think it'd be easier for me to follow/ debug the plotting functions if they were a little more standardized. We pretty frequently want a dataframe (or at least aligned arrays) of values per cell or gene, but this is done in a variety of ways.

ivirshup: On layers, it looks like scanpy uses `layer=None` as the default and anndata does `layer='X'` (via `find . -name "*.py" -exec grep -n "layer=" {} +`).

I prefer the `scanpy` style, since if someone specifies `layer='X'` and they actually have a layer named `'X'` they probably want to use that.

falexwolf: Oh, yes, I also prefer the Scanpy style. @Koncopd built most of the layers for AnnData, any reason to move away from 'X'? Any reason why you chose to do it that way?

falexwolf:
I think `obs_values` is fine. But maybe, `aggregate_obs` is even better, as this describes what it does (aggregating annotations of observations with partial (projections of) observations).

It's no problem at all to make the next Scanpy release depend on the current AnnData release, both in the requirements and the minimal version check upon importing Scanpy.

scanpy/utils.py
falexwolf: > I don't think there's an issue per-say, I just think it'd be easier for me to follow/ debug the plotting functions if they were a little more standardized. We pretty frequently want a dataframe (or at least aligned arrays) of values per cell or gene, but this is done in a variety of ways.

100% agreed.

scanpy/utils.py
falexwolf: I'm fine with `log2fc` but I think diffxpy uses `log2FC` and if you think about an axes label, the capped version might be more appealing. Your call.

falexwolf: But, please update https://github.com/theislab/scanpy/issues/453#issue-404768505. 🙂

ivirshup: It looks like diffxpy uses `"log2fc"`, at least as of https://github.com/theislab/diffxpy/commit/0054f900aa9569ec00da5aa30876d8ee9d1d3092, so I think I'll go with that

scanpy/utils.py
Koncopd: @ivirshup Could you please give an example where we have `layer='X'`, i don't see it.
@falexwolf not sure, maybe i had loom as a model, there is a similar thimg there if i remember correctly.

Koncopd: Ah, i [see](https://github.com/theislab/anndata/blob/34f4eb63710628fbc15e7050e5efcac1d7806062/anndata/base.py#L1464). I'm not sure i wrote these.

ivirshup:
To me, `aggregate_obs` implies a reduction, but that might mostly be because `pd.DataFrame.aggregate`. I could go for `collect_obs`, but if this goes into `sc.get` there's already a verb. I think `sc.get.obs_df` would work (though kind of ambiguous with `adata.obs`) as would `sc.get.obs_values`.

falexwolf:
`sc.get.select_obs` (meaning `selected_obs`)? You're right about `aggregate`. `obs_values` is fine, too (it's really just that it reminds of `.values`).

falexwolf:
Should we merge this? Or is this going to partly end up in AnnData? :)

ivirshup:
Kinda? This is waiting on https://github.com/theislab/anndata/pull/144 and a following AnnData point release.

But I think that PR is ready to go.

ivirshup:
Getting back around to this, I think it's pretty close to ready.

Two last things to consider:
*  Name change of `obs_values_df`. I like`obs_df` since it fits with `obs_vector` and has a nice short name.
* Support for `.raw`, I've dropped it at the moment, but maybe should add it back in.

ivirshup:
Also, should this go in `sc.get`?

ivirshup:
I think this functionality is ready to go. I'm going to try using this a bit before deciding if it needs an argument for `raw`. I'm going to merge this so I can use the `sc.get.rank_genes_groups` for #467. @flying-sheep, let me know if I messed up any docs and I'll fix it separately.

docs/release_notes.rst
```diff
@@ -12,6 +12,7 @@ Post v1.4 :small:`May 13, 2019`
 
 New functionality:
 
+- New module :ref:`sc.get<module-get>` adds helper functions for extracting data in convenient formats :pr:`619` :smaller:`thanks to I Virshup`
```
flying-sheep, line 15: why not `` :mod:`scanpy.get` ``?

flying-sheep:
The docs look great! I just wonder about the above: In the release notes, we refer to everything as `scanpy.*`, not `sc.*`

docs/release_notes.rst
ivirshup: I think it should link to the docs, which ``:mod:`scanpy.get` `` didn't seem to do.

ivirshup: It could definitely get changed to ``:ref:`scanpy.get<module-get>` ``, though I'm not sure if how the styling can be fixed to look like a module.

flying-sheep: It can’t. We should fix that `` :mod:`scanpy.get` `` doesn’t link to the proper location.

I guess that could be done by creating the module entry via [`.. module:: scanpy.get`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py:module) directly before using `.. autosummary::`

flying-sheep: Fixed it! 0485cb8c5d28e7316622a2d8c26e538014506e68

I just put the module link targets at the correct position and immediately reset the “current module” to scanpy each time, so that relative links still work!

LuckyMD:
Do we have some kind of tutorial around the new `sc.get` module?

Issue #620, opened 2019-04-25
t-tests fails when variance of both groups is 0


ivirshup:
So I was writing a test for `rank_genes_groups_df` from #619 when I got some results that seem pretty wrong. Here's an example:

```python
import scanpy as sc

import numpy as np
import pandas as pd
from scipy import stats

from itertools import repeat, chain

# Making data where "gene0" is definitely differentially expressed.

a = np.zeros((20, 3))
a[:10, 0] = 5
adata = sc.AnnData(
    a,
    obs=pd.DataFrame(
        {"celltype": list(chain(repeat("a", 10), repeat("b", 10)))},
        index=[f"cell{i}" for i in range(a.shape[0])]
    ),
    var=pd.DataFrame(index=[f"gene{i}" for i in range(a.shape[1])]),
)

# Running differential expression with t-test:

sc.tl.rank_genes_groups(adata, groupby="celltype")
print(adata.uns["rank_genes_groups"]["pvals"]["a"])
# array([1., 1., 1.])  # This seems wrong

sc.tl.rank_genes_groups(adata, groupby="celltype", method="t-test")
print(adata.uns["rank_genes_groups"]["pvals"]["a"])
# array([1., 1., 1.])  # This also seems wrong

# Checking to make sure I'm not forgetting something obvious
print(stats.ttest_ind([0,0,0,0,0], [5,5,5,5,5]))
# Ttest_indResult(statistic=-inf, pvalue=0.0)   # This seems right

# Wilcoxon seems fine:

sc.tl.rank_genes_groups(adata, groupby="celltype", method="wilcoxon")
print(adata.uns["rank_genes_groups"]["pvals"]["a"])
# array([1.57052284e-04, 1.00000000e+00, 1.00000000e+00])  # This seems right
```

`"logreg"` on the other hand, throws an error:

```python
sc.tl.rank_genes_groups(adata, groupby="celltype", method="logreg")
<ipython-input-7-29e46f287a31> in <module>
----> 1 sc.tl.rank_genes_groups(adata, groupby="celltype", method="logreg")

~/github/scanpy/scanpy/tools/_rank_genes_groups.py in rank_genes_groups(adata, groupby, use_raw, groups, reference, n_genes, rankby_abs, key_added, copy, method, corr_method, log_transformed, **kwds)
    397     adata.uns[key_added]['scores'] = np.rec.fromarrays(
    398         [n for n in rankings_gene_scores],
[...]

PR #621, opened 2019-04-25
Fix t-tests when variance is zero

ivirshup:
Just a quick fix, should probably replace most of this code with `scipy.stats.ttest_ind`/ `scipy.stats.stats.ttest_ind_from_stats` with `equal_var=False`.

### DIFF
--- a/scanpy/tools/_rank_genes_groups.py
+++ b/scanpy/tools/_rank_genes_groups.py
@@ -196,7 +196,8 @@ def rank_genes_groups(
             else: raise ValueError('Method does not exist.')
 
             denominator = np.sqrt(vars[igroup]/ns_group + var_rest/ns_rest)
-            denominator[np.flatnonzero(denominator == 0)] = np.nan
+            with np.errstate(divide='ignore', invalid='ignore'):
+                scores = np.divide((means[igroup] - mean_rest), denominator) #Welch t-test
             scores = (means[igroup] - mean_rest) / denominator #Welch t-test
             scores[np.isnan(scores)] = 0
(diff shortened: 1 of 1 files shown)


ivirshup:
#620

ivirshup:
Someone should review this before it's merged. I think this won't cause any problems, but I'm also not too familiar with the DE code.

LuckyMD:
I wonder why this wasn't done in the first place. Is scipy not already a dependency of scanpy? Or is this slower than the initial implementation?

falexwolf:
Looks great! I wasn't aware of this high-dimensional version of a t-test in Scipy, which seems to be as efficient as the current implementation. I only investigated thoroughly for Wilcoxon rank and found that Scipy doesn't have a scalable version to offer.

But yes, this will get merged after 1.4.1.

falexwolf:
1.4.1 is out, are we sure this is as scalable as it was before and not a backwards breaking change. If yes, we can merge immediately.

ivirshup:
I'm not completely sure this doesn't break anything, but the regression tests pass. The internal code is very similar, so I'm not too worried about these changes.

It does look like it's (very) slightly slower. Running this a thousand times for pbmc68k dataset took ~2.3% longer (about 1.4 ms per run) than the previous version. That said, we're very inefficient about mean and variance calculation, so I think that's a better place to optimize.

Edit: I've force pushed to fix some minor formatting issues (trailing white space, blank line, typo) that I didn't think deserved it's own commit.

falexwolf:
Great! 2.3% is nothing...

PR #623, opened 2019-04-26
Log PCA time at info level, like other steps in the pipeline (recipes…

tomwhite:
…, neighbors).


falexwolf:
OK. I initially thought that PCA is such a fast step that much logging is not needed. But you're right. :)

PR #624, opened 2019-04-26
Take n_bins into account for the cell_ranger flavor of highly_variable_genes

e-sollier:
I've changed one line in the highly_variable_genes function, so that n_bins is taken into account with the cell_ranger flavor (currently only the seurat flavor uses this parameter).

Additionally, I have noticed that, in the current version, the bins are slightly offset: after -INF, it starts at 10, instead of 5, which results in the first bin containing twice as many genes as the other bins. I don't know if this is intentional (for example, to exactly reproduce the results of cell ranger) or not.
In the version that I suggest, I have removed this offset. As a consequence, with the default parameter of n_bins=20, my new version does not exactly reproduce the results of the previous version. In order to exactly reproduce the current results, we would have to keep this offset by doing range(2,n_bins+1) instead of range(1,n_bins).


flying-sheep:
pre-commit.ci autofix

PR #628, opened 2019-05-02
Multi-panel embedded density

fidelram:
This PR adds multiple panels to the plot embedding_density function.

image

### DIFF
--- a/scanpy/plotting/_tools/__init__.py
+++ b/scanpy/plotting/_tools/__init__.py
@@ -1,3 +1,4 @@
+from collections import abc
 import numpy as np
 import pandas as pd
 from scipy.sparse import issparse
@@ -650,7 +651,11 @@ def embedding_density(
     fg_dotsize:  Optional[int] = 180,
     vmax:  Optional[int] = 1,
     vmin:  Optional[int] = 0,
+    ncols: Optional[int] = 4,
+    hspace: Optional[float] = 0.25,
+    wspace: Optional[None] = None,
     save: Union[bool, str, None] = None,
+    show: Optional[bool] = None,
     **kwargs
 ):
     """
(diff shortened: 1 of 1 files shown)


fidelram:
The failing test is un-related to the PR.

@LuckyMD can you take a look at the changes?

scanpy/plotting/_tools/__init__.py
```diff
@@ -721,52 +726,92 @@ def embedding_density(
...
+
+        axs = []
+        for count, group_name in enumerate(group):
+            if group_name not in adata.obs[groupby].cat.categories:
+                raise ValueError('Please specify a group from the `.obs` category over which the density '
+                                 'was calculated. Invalid group name: {}'.format(group_name))
+
+            ax = pl.subplot(gs[count])
+            group_mask = (adata.obs[groupby] == group_name)
+            adata.obs['_density'] = adata.obs[key][group_mask]
```
LuckyMD: Can we assume that we will not overwrite a `'_density'` slot in the original `AnnData` object? I created a new one for that reason. Not sure if `_VarName` can also be assumed to be reserved for `adata.obs` slots.

LuckyMD (approved):

LuckyMD:
I really like this! I was trying to figure out how to easily do this within the current `plot_scatter` framework, but I couldn't figure out how.

One thing though... You could add an option `group='all'`, which should be the default. That would be much more user-friendly than the current setup.

scanpy/plotting/_tools/__init__.py
fidelram: good point. I didn't want to copy the whole AnnData object to save memory but indeed this could rewrite a valid obs. name. I will set up some random name to avoid this.

fidelram:
Now it looks like this:

image

scanpy/plotting/_tools/__init__.py
```diff
@@ -754,6 +767,12 @@ def embedding_density(
     norm = colors.Normalize(vmin=vmin, vmax=vmax)
     cmap.set_over('black')
     cmap.set_under('lightgray')
+    # a name to store the density values is needed. To avoid
+    # overwriting a user name a new random name is created
+    while True:
+        density_col_name = '_tmp_embedding_density_column_{}_'.format(np.random.randint(1000, 10000))
```
LuckyMD, line 778: I love the attention to detail here 😄

LuckyMD:
It feels like I tried to give the density tool wings by putting together lego bricks which just about kept it in the air, and then you just crossed it with a pegasus 😆.

falexwolf:
Very cool!

PR #640, opened 2019-05-11
Add outline to the legend text in scatterplots

gokceneraslan:
This PR adds support to add an outline around the legend text in scatterplots. Its effect is similar to fontweight='bold' but arguably more elegant and readable.

Before:

`sc.pl.umap(adata, color='paul15_clusters', legend_loc='on data') `

image

After:

`sc.pl.umap(adata, color='paul15_clusters', legend_loc='on data', legend_fontoutline=1, legend_fontweight='normal')`

image

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -9,8 +9,8 @@ from scipy.sparse import issparse
 from matplotlib import pyplot as pl
 from matplotlib import rcParams
 from matplotlib import gridspec
+from matplotlib import patheffects
 from matplotlib.colors import is_color_like
-import matplotlib.patheffects as patheffects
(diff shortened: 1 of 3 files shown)


LuckyMD:
Might be a good idea to make font outline default for `legend_loc='on data'`? The first figure isn't particularly nice to look at.

fidelram:
@gokceneraslan Do you know why the test is failing? Did you change some of the defaults? I like the change, is really an improvement.

gokceneraslan:
I didn't change any defaults I think. Test fails in rename_category of pandas in pbmc, not sure it's related.

fidelram:
It seems not to be related indeed. I will merge this.

PR #651, opened 2019-05-16
Ingest class

Koncopd:
As discussed with @falexwolf this PR introduces a new Ingest class to process new small pieces of data.

> sc.pp.neighbors(adata)  # adata is huge with 1M observations
> 
> ingest = sc.Ingest(adata)  # represents the existing data, learned annotations, structure and exposes it to functionality that allows to ingest new data very quickly
> 
> adata_small.obsm['X_model'] = model(adata_small.X)
> 
> ingest.neighbors(adata_small)  # adata_small with just 1000 observations
> 
> now, we have the updated neighbors graph with 1,001,000 observations
> we want to do the same things as always
> 
> by leveraging the neighbors of the new data within the old data, 
> map the new data into the embedding (umap), by just computing a correction to the existing embedding: a new data point gets the mean position of its k nearest neighbors
> 
> ingest.umap(adata_small)
> 
> update the clustering (mapping the 1000 observations into the existing clusters): a new data point maps into a cluster if the majority of its neighbors is a member of the cluster 
> 
> ingest.louvain(adata_small)


Koncopd:
This is just an initial setup.

Koncopd:
@falexwolf 
some usage examples
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/use_Ingest.ipynb

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,136 @@
...
+            self._init_neighbors(adata)
+
+        if 'X_umap' in adata.obsm:
+            self._init_umap(adata)
+
+    def pca(self, adata_small, n_pcs=None, inplace=True):
+        #todo - efficient implementation for sparse matrices
+        rep = adata_small.X
+        rep = rep.toarray() if issparse(rep) else rep.copy()
+        rep -= rep.mean(axis=0)
```
ivirshup: Should this be conditional on whether the original was mean centered as well?

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,136 @@
...
+    def umap(self, adata_small):
+        rep = self.same_rep(adata_small)
+        adata_small.obsm['X_umap'] = self._umap.transform(rep)
+
+    def knn_classify(self, adata_small, classes_key, k=10, **kwargs):
+        #i.e. ingest.knn_classify(adata_small, 'louvain')
+        cat_array = self._adata.obs[classes_key]
+        neighbors, _ = self.neighbors(adata_small, k, **kwargs)
+
+        values = [cat_array[inds].mode()[0] for inds in neighbors]
```
ivirshup: Does this take any meaningful amount of the run time? I suspect this could be made more efficient, but I'm not sure if it'd be worth it.

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,136 @@
+import pandas as pd
+import numpy as np
+from umap import UMAP
+from umap.distances import named_distances
+from umap.nndescent import make_initialisations, make_initialized_nnd_search, initialise_search
```
ivirshup: Should we be relying on pynndescent directly as in https://github.com/theislab/scanpy/pull/659?

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,136 @@
...
+from sklearn.utils import check_random_state
+from scipy.sparse import issparse
+
+from ..preprocessing._simple import N_PCS
+from ..neighbors import _rp_forest_generate
+
+
+class Ingest:
+
+    def _init_umap(self, adata):
```
ivirshup, line 148: Could this be a function (maybe in `sc.get`?) that takes an AnnData and returns its UMAP transformer?

scanpy/tools/_ingest.py
ivirshup: Oh, also we shouldn't import UMAP at the top level. It takes a few seconds to import, so we should defer it.

ivirshup (commented):
I really like that this functionality is making it into scanpy! I've comments on specific implementation details but mostly I think the methods look good. I do have a few questions about the structure, however.

* With the current functionality, could this be a module instead of a class? I'm not sure I see the advantage of this being a separate class, rather than a set of functions.
  * A potential factor here is how long it takes to run the `_init_` methods.
  * I don't think I see the `Ingest` object being updated when new data is processed. Is this correct?
* I think there should be checks to make sure the transformation is valid. Mostly this means checking the shape and var_names are compatible.

scanpy/tools/_ingest.py
ivirshup: Thinking about this again, what's the reason to include KNN classification in here? Given the large low weight component I've seen in UMAP connectivity graphs, I'm worried this will give noisy results.

Additionally [the recent benchmarking paper](https://www.biorxiv.org/content/10.1101/644435v1) found poor relative performance for the KNN classifier they tried.

scanpy/tools/_ingest.py
Koncopd: Yes, you are right, i just forgot about it.

scanpy/tools/_ingest.py
Koncopd: The problem is that we don't have pynndescent as a dependency for now.
Yes, i will change to the deferred import.

Koncopd (commented):
I think `map_labels` should be named `map_labels_knn`, this reflects the meaning of the function better.

ivirshup:
> I think `map_labels` should be named `map_labels_knn`

Why change the name from `knn_classify`? I think it fits, especially since this is so close to sklearn's `KNeighborClassifier`?

Koncopd:
> > I think `map_labels` should be named `map_labels_knn`
> 
> Why change the name from `knn_classify`? I think it fits, especially since this is so close to sklearn's `KNeighborClassifier`?

@ivirshup this was the decision of @falexwolf . Maybe something like `map_labels(..., method='knn')` which allows adding other methods when they are implemented?

falexwolf:
Yes on your last suggestion, @Koncopd ! :)

Koncopd:
@falexwolf @ivirshup 
New Ingest api usage
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/Ingest.ipynb

LuckyMD:
This looks really cool! So when you do `ing.to_adata()` you only get back the mapped datapoints and not the whole dataset? It would be interesting to see a plot of both on top of one another to appreciate the label mapping.

Koncopd:
@LuckyMD `ingest.to_adata` just returns a copy of `adata_new` (or the same object depending on inplace) with all mapped representations (pca, umap) and labels.

LuckyMD:
I see. So you'd have to concatenate the data to plot them together?

Koncopd:
Yes. Also i can add a function which returns a dataset with joint representations and labels only.

Koncopd:
Updated the notebook above with `to_adata_joint` example - this function returns the concatenation of `adata_ref` and `adata_new` with projected representations and mapped labels.

Koncopd:
Test are failing because of the new version of dask.

Koncopd:
I think it is more or less complete.
Here are the tutorials
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/Ingest-realistic.ipynb
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/Ingest-simple.ipynb

ivirshup:
Just starting to try this out, but I hit a bug. Here's a script to reproduce and the traceback:

<details>
<summary>Script </summary>

```python
import scanpy as sc
from scanpy.tools._ingest import ingest
import numpy as np
import pandas as pd
from functools import reduce

def simplify_annot(annot):
    names = annot.columns.str.extract(r"\[(.*)\].*$", expand=False)
    unique_names, idxs = np.unique(names, return_index=True)
    new_annot = annot.iloc[:, idxs].copy()
    new_annot.columns = unique_names
    return new_annot

def process(dset):
    dset.layers["counts"] = dset.X.copy()
    sc.pp.normalize_total(dset)
    sc.pp.log1p(dset)
    sc.pp.highly_variable_genes(dset)
    sc.pp.pca(dset)
    sc.pp.neighbors(dset, n_neighbors=30)
    sc.tl.umap(dset)

dset1 = sc.datasets.ebi_expression_atlas("E-GEOD-81608", filter_boring=True) 
dset2 = sc.datasets.ebi_expression_atlas("E-GEOD-83139", filter_boring=True)
# dset3 = sc.datasets.ebi_expression_atlas("E-ENAD-27", filter_boring=True)

# dsets = [dset1, dset2, dset3]
dsets = [dset1, dset2]
for dset in dsets:
    dset.obs = simplify_annot(dset.obs)
    sc.pp.calculate_qc_metrics(dset, inplace=True)

shared_genes = reduce(np.intersect1d, [dset.var_names for dset in dsets])
dsets = [dset[:, shared_genes].copy() for dset in dsets]

for dset in dsets:
    process(dset)

# dset1, dset2, dset3 = dsets
dset1, dset2 = dsets

dset1.obs["inferred cell type (dset1)"] = dset1.obs["inferred cell type"]

dset12 = ingest(dset2, dset1, obs="inferred cell type (dset1)", return_joint=True)
```

Traceback:

```python
  File "/Users/isaac/github/scanpy/scanpy/tools/_ingest.py", line 33, in ingest
    return ing.to_adata(inplace) if not return_joint else ing.to_adata_joint()
  File "/Users/isaac/github/scanpy/scanpy/tools/_ingest.py", line 222, in to_adata_joint
    adata = AnnData(np.vstack((self._adata_ref.X, self._adata_new.X)))
  File "/Users/isaac/github/anndata/anndata/core/anndata.py", line 566, in __init__
    filename=filename, filemode=filemode)
  File "/Users/isaac/github/anndata/anndata/core/anndata.py", line 735, in _init_as_actual
    X = X.astype(dtype, copy=False)
ValueError: setting an array element with a sequence.
```

</details>

I've reproduced on 2bffd6a1400 and after merging master with that. This occurs whether inplace is true or not. It also occurs on anndata master as well as the current release. As you'd expect, it doesn't happen with `return_joint=False`.

Koncopd:
Thank you. I guess this is because dset{1,2}.X is sparse, i need to fix this.

ivirshup:
Since I can get some labels out from just not getting an object back, I tried a couple things.

Using the labelled half of `E-ENAD-27` to add labels to the other cells:

<details>
<summary> plots </summary>

image

</details>

Using `E-GEOD-81608` (many cells, few labels) to label `E-GEOD-83139` (few cells, many labels) seemed to work fine considering we're moving to a lower resolution label set. Here's the latter projected onto the former:

<details>
<summary> plots </summary>

image

</details>

The reverse (former projected onto the latter) didn't seem to work so well:

<details>
<summary> plots </summary>

image

</details>

If it'd be useful, I could clean up the notebooks I ran these in and get you some confusion matrices tomorrow.

Koncopd:
@ivirshup thanks. 
I fixed return_join, it should work now.

scanpy/tools/_ingest.py
```diff
@@ -26,10 +30,21 @@ def ingest(
...
-    ing.map_embedding(embedding_method)
+    embedding_method = _str_to_list(embedding_method)
+    for method in embedding_method:
+        ing.map_embedding(method)
     if obs is not None:
         ing.neighbors(**kwargs)
-        ing.map_labels(obs, labeling_method)
+
+        obs = _str_to_list(obs)
+        labeling_method = _str_to_list(labeling_method)
```
ivirshup: Just a few style thoughts.

* I think `_str_to_list` may be [simple enough it doesn't need a function](http://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection).
* Could the preparing of `obs` and `labeling_method` (via `_str_to_list`) happen at the top of the function? I don't think it would break anything, and I find code easier to follow when a variables value is consistent through as much of the function as possible.

ivirshup:
Great! I'll check it out when I have a chance. If this is close to ready, could it also start getting some tests?

Just to clarify, would a notebook with the pancreas integration stuff be useful to you?

Koncopd:
Hi, yes, i will start adding tests. 
And i think it can be useful but only if you have time for it.

scanpy/tools/_ingest.py
Koncopd: I'll fix it.

Koncopd:
@falexwolf @ivirshup 
Tests are also ready.

Koncopd:
Some things on pancreas dataset
https://github.com/Koncopd/anndata-scanpy-benchmarks/blob/master/ingest_pancreas.ipynb

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,239 @@
...
+    embedding_method='umap',
+    labeling_method='knn',
+    return_joint = False,
+    **kwargs
+):
+    """
+    Note
+    ----
+    This doesn't update the neighbor graph.
+    """
```
ivirshup, line 100: I'm a little confused about what some of these arguments are doing and what values can be passed to them. Could you update the the docs?

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,239 @@
...
+            return adata.obsm[self._use_rep]
+        return adata.X
+
+    def transform(self, adata_new):
+        self._obs = pd.DataFrame(index=adata_new.obs.index)
+        #not sure if it should be AxisArrays
+        self._obsm = AxisArrays(adata_new, 0)
+
+        self._adata_new = adata_new
+        self._obsm['rep'] = self._same_rep()
```
ivirshup, line 298: What does this method do?

As it only seems to modify itself, should it be called `fit` or even something else? Based on the sklearn conventions, I would expect `transform` to return transformed values.

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,239 @@
...
+
+    if obs is not None:
+        ing.neighbors(**kwargs)
+        for i, col in enumerate(obs):
+            ing.map_labels(col, labeling_method[i])
+
+    return ing.to_adata(inplace) if not return_joint else ing.to_adata_joint()
+
+
+class Ingest:
```
ivirshup, line 134: Do we definitely want this class to be public? At the moment, I'm not sure if there's functionality being hidden if we just expose `ingest`. If we do want to make this public, I'll want to go over it's API in more detail.

An option here is to initially merge this with the `Ingest` class being private, with the possibility of making it public in the future. What do you think?

scanpy/preprocessing/_simple.py
```diff
@@ -515,12 +515,16 @@ def pca(
 
     if data_is_AnnData:
         adata.obsm['X_pca'] = X_pca
+        adata.uns['pca'] = {}
+        adata.uns['pca']['params'] = {
+            'zero_center': zero_center,
+            'use_highly_variable': use_highly_variable
+        }
```
ivirshup, line 527: If you write an AnnData object any values of `uns` that are `None` don't get written. While this is maybe something we should change, are you sure that all the values you rely on in `ingest` wouldn't be set to `None` with typical usage? This could be tested by checking that the results of merging into an object, and merging into a copy which has been round tripped to disk are equivalent.

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,239 @@
...
+            self._obsm['X_umap'] = self._umap_transform()
+        elif method == 'pca':
+            self._obsm['X_pca'] = self._pca()
+        else:
+            raise NotImplementedError('Ingest supports only umap embeddings for now.')
+
+    def _knn_classify(self, labels):
+        cat_array = self._adata_ref.obs[labels]
+
+        values = [cat_array[inds].mode()[0] for inds in self._indices]
```
ivirshup, line 362: Would you be interested in adding weighted knn classification? I think all we would need to do is calculate the connectivities, which I think we could do with [these lines](https://github.com/lmcinnes/umap/blob/424268f6a99a401e3b9eae9c1e8dadebf619db83/umap/umap_.py#L1828-L1837).

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,239 @@
...
+
+        obs_update = self._obs.copy()
+        obs_update.index = adata[adata.obs['batch']=='1'].obs_names
+        adata.obs.update(obs_update)
+
+        for key in self._obsm:
+            if key in self._adata_ref.obsm:
+                adata.obsm[key] = np.vstack((self._adata_ref.obsm[key], self._obsm[key]))
+
+        return adata
```
ivirshup, line 437: Could you document these functions (`to_adata` and `to_adata_joint`)? I think I know what they're supposed to do, but I'm not totally sure.

scanpy/tests/test_ingest.py
```diff
@@ -0,0 +1,145 @@
...
+    assert ing._obsm['rep'].shape == (adata_new.n_obs, 30)
+    assert not ing._pca_centered
+
+    sc.pp.neighbors(adata_ref, use_rep='X')
+
+    ing = sc.tl.Ingest(adata_ref)
+    ing.transform(adata_new)
+
+    assert ing._use_rep == 'X'
+    assert ing._obsm['rep'] is adata_new.X
```
ivirshup, line 65: In general, I think we should test the functionality instead of the implementation. For example, instead of checking `assert not ing._pca_centered`, could we check if the initial PCA wasn't centered, the new data wasn't centered before it was transformed?

scanpy/tools/_ingest.py
```diff
@@ -0,0 +1,239 @@
...
+        if self._use_rep == 'X':
+            return adata.X
+        if self._use_rep in adata.obsm.keys():
+            return adata.obsm[self._use_rep]
+        return adata.X
+
+    def transform(self, adata_new):
+        self._obs = pd.DataFrame(index=adata_new.obs.index)
+        #not sure if it should be AxisArrays
+        self._obsm = AxisArrays(adata_new, 0)
```
ivirshup, line 295: I don't think it does, really that class is mostly to support indexing of AnnData objects, but this class can't be subset. I suppose it could be used for the shape checks, but that's easy enough to do separately.

ivirshup (changes requested):
## Notebook feedback

This looks really good! I wanted to take a closer look so I replicated it on my machine. A few plots I found useful (note I kept the original labels in `obs` as `"celltype_orig"`):

### Just show labels of added data

Here's an example:

```python
with plt.rc_context({"figure.figsize": (5, 5),"figure.dpi": 100}):
    ax = sc.pl.umap(ad_joint, show=False, size=10)
    sc.pl.umap(ad_joint[ad_joint.obs["sample"] == "Wang"], color=["celltype_orig"], ax=ax, size=10)
```

image

### Confusion matrices of changed labels

This would be very useful to see if labels change unexpectedly. Here's some code I used for this:

<details>
<summary> `confusion` function definition </summary>

```python
import numpy as np
import pandas as pd
from sklearn.metrics import confusion_matrix
import scanpy as sc

def confusion(
    orig: "Union[pd.Series, np.array, Sequence]",
    new: "Union[pd.Series, np.array, Sequence]",
    data=None,
    *,
    normalize: bool = True
) -> pd.DataFrame:
    """Given an original and new set of labels, create a labelled confusion matrix.
    
    Params
    ------
    orig
        Original labels
    new
        New labels
    normalize
        Should the confusion matrix be normalized?
        
    Usage
    -----
    >>> import seaborn as sns
    >>> confmtx = confusion(labels["orig"], labels["new"])
    >>> sns.heatmap(confmtx)
    """
    # TODO: Decide whether I want to remove this
    if isinstance(data, sc.AnnData):
        orig = data.obs[orig]
        new = data.obs[new]
    elif data is not None:
        orig = data[orig]
        new = data[new]
        
    # Coercing so I don't have to deal with it later
    orig, new = pd.Series(orig), pd.Series(new)
    assert len(orig) == len(new)

    unique_labels = pd.unique(np.concatenate((orig.values, new.values)))

    # Compute
    mtx = confusion_matrix(orig, new, labels=unique_labels)
    if normalize:
        sums = mtx.sum(axis=1)[:, np.newaxis]
        mtx = np.divide(mtx, sums, where=sums != 0)

    # Label
    if orig.name is None:
        orig_name = "Original labels"
    else:
        orig_name = orig.name
    if new.name is None:
        new_name = "New labels"
    else:
        new_name = new.name
    df = pd.DataFrame(
        mtx,
        index=pd.Index(unique_labels, name=orig_name),
        columns=pd.Index(unique_labels, name=new_name)
    )
    
    # Filter
    df = df.loc[df.index.isin(pd.unique(orig)), df.columns.isin(pd.unique(new))]
    
    return df
```


</details>

Here's an example of usage

```python
import seaborn as sns

new_labels = ad_joint.obs[ad_joint.obs["sample"] == "Wang"][["cell_type", "celltype_orig"]]
sns.heatmap(confusion(new_labels["celltype_orig"], new_labels["cell_type"]))
```

image

### Provenance

Could you include some information on where the datasets that you're integrating here came from, and how they've been processed?

### Merging larger datasets looks good!

I've merged the Segerstolpe dataset into the Baron one to test this out a bit more, and it looks really good!

<details>
<summary> Code </summary>

```python
baron = adata[adata.obs["sample"] == "Baron"].copy()
seger = adata[adata.obs["sample"] == "Segerstolpe"].copy()

sc.pp.pca(baron)
sc.pp.neighbors(baron)
sc.tl.umap(baron)

ad_bs = sc.tl.ingest(
    seger, baron,
    obs='celltype',
    embedding_method=['umap', 'pca'], k=15,
    return_joint=True
)
```

</details>

On thing that is particularly cool is all the "unclassified" cells get labeled as "schwann":

image

## `sc.tl.ingest`

### Arguments

As mentioned in review, could you write up the docs for the arguments? I just used what you have in that notebook, but it's not clear what each one does at the moment.

### `obs_names`

This might be a separate issue, but could we not change the cell labels when we join objects if there is no overlap? I'd like to be able to do: `ad_joint[ad_new.obs_names]`.

## `Ingest`

I like the state the `ingest` function is in. I think the `Ingest` class might need some more work, especially if we want to make it public. Do you definitely want `Ingest` to be made public in this PR? That could happen later.

scanpy/tools/_ingest.py
Koncopd: Yes, i will add all needed descriptions soon.

scanpy/tools/_ingest.py
Koncopd: Do you mean distance weighted knn?

scanpy/preprocessing/_simple.py
Koncopd: For now they wouldn't be None, but in general it should be checked, yes.

scanpy/tools/_ingest.py
Koncopd: Yes, this is the idea, to make only the function `ingest` public for now.

scanpy/tools/_ingest.py
Koncopd: Definitely i will do it.

scanpy/tools/_ingest.py
Koncopd: It does transform `adata_new` to the same representation which was used by `adata_ref` to calculate neighbors. But it doesn't return any values, yes, maybe fit is a better name for this function

scanpy/tests/test_ingest.py
Koncopd: Do we need testing that the initial pca wasn't centered here?
I mean it is specified as `sc.pp.pca(adata_ref, n_comps=30, zero_center=False)` a few lines above, so it should be centered and i don't see any need to test it here.

scanpy/tools/_ingest.py
Koncopd: I found it to be convenient here, maybe i need to get rid of it because it is an internal AnnData thing.

Koncopd:
@ivirshup 
Thank you for your analysis.
I will definitely add all needed arguments' descriptions a bit later.

As for the `Ingest` class, i don't think it should be public, the public thing is the `ingest` function.

Also i added the possibility of specifying `adata.concatenate` arguments in both `ingest` and `to_adata_joint`. So to get the same `obs_names` you can use `sc.tl.ingest(..., index_unique=None)` or `ing.to_adata_joint(index_unique=None)`.

scanpy/tools/_ingest.py
ivirshup: Something like that, I was thinking UMAP’s connectivity.

scanpy/tests/test_ingest.py
ivirshup: When you check the value of `_pca_centered` you’re checking some internal state, not what computation happened. I think this should be tested more like this (this is basically pseudo code, since I’m not at a computer):

```python
refc = sc.pp.pca(adata, zero_center=True, copy=True)
ref = sc.pp.pca(adata, zero_center=False, copy=True)

# Not sure if I have these calls exactly right
# The intent is the `new` objects have been projected into the ref objects pca space
newc = sc.tl.ingest(adata, refc, embedding_method=“pca”, return_joint=False)
new = sc.tl.ingest(adata, ref, embedding_method=“pca”, return_joint=False)

assert np.allclose(newc.obsm[“X_pca”], refc.obsm[“X_pca”])
assert np.allclose(new.obsm[“X_pca”], ref.obsm[“X_pca”])
assert not np.allclose(new.obsm[“X_pca”], newc.obsm[“X_pca”])
```

You could alternatively use an oracle here, like scikit-learn’s PCA transformer.

Checking the value of `_pca_centered` doesn’t tell us whether these cases are true. Additionally, testing the behavior of the method rather than internal construction means we’re free to change the internals without having to rewrite tests.

This is basically the [“test behavior, not implementation”](https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html) mantra.

flying-sheep:
@Koncopd once this is ready from your side, you can re-request @ivirshup’s review.

falexwolf:
I'd be fine with it, but happy to wait for @ivirshup's review!

flying-sheep:
Well, he reviewed it before and none of the conversations is marked as resolved. Let’s let them close those and then we can squash & merge.

scanpy/datasets/__init__.py
```diff
@@ -68,9 +68,9 @@ def burczynski06() -> AnnData:
     blood mononuclear cells"
     J Mol Diagn 8, 51 (2006). PMID:16436634.
     """
-    filename = settings.datasetdir / 'burczynski06/GDS1615_full.soft.gz'
-    url = 'ftp://ftp.ncbi.nlm.nih.gov/geo/datasets/GDS1nnn/GDS1615/soft/GDS1615_full.soft.gz'
-    adata = read(filename, backup_url=url)
+    filename = settings.datasetdir / "burczynski06/GDS1615_full.soft.gz"
+    url = "ftp://ftp.ncbi.nlm.nih.gov/geo/datasets/GDS1nnn/GDS1615/soft/GDS1615_full.soft.gz"
+    adata = sc.read(filename, backup_url=url)
```
flying-sheep:
```suggestion
    adata = read(filename, backup_url=url)
```

flying-sheep (changes requested):
Why all the string quote changes? Black should read its config and leave those alone.

Also there’s a merge error, please pay real good attention that things like this don’t happen.

flying-sheep (approved):

Koncopd:
@flying-sheep Yeah, sorry, missed this. Thank you for correcting.

As for black, i don't understand why it happened. I changed the quotes back and run black on the file, now it doesn't change it...

I will go through the discussion with @ivirshup again, fix some things and then will re-request the review.

flying-sheep:
Great, thank you a lot, this seems like it was quite some work!

I think the resolve feature is pretty effective in keeping track of what one already adressed.

ivirshup:
@Koncopd, I'm happy to go over this again whenever. I haven't looked too closely yet, but it looks like a lot has been addressed.

falexwolf:
Guys, I'll merge this for now so that I can conveniently play around with it in practical settings and potentially improve. We don't need to advertise for now and the next release might be a bit ahead anyway.

I'm reading the latest comments as you being essentially positive. ☺️

PR #659, opened 2019-05-22
Use pynndescent dependency to support threaded nearest neighbors

tomwhite:
Pynndescent 0.3.0 was released yesterday with support for multi-threading. This change allows scanpy to take advantage of multi-threading for computing nearest neighbors.

To use it, wrap the call to scanpy in a `joblib.parallel_backend` context manager:

```python
from joblib import parallel_backend
with parallel_backend('threading', n_jobs=16):
        sc.pp.neighbors(adata)
```

Running on the 130K dataset on a 16 core machine before the change:

```
computing neighbors
    using 'X_pca' with n_pcs = 50
    finished (0:01:31.54)
```
 and with the change:

```
computing neighbors
    using 'X_pca' with n_pcs = 50
    finished (0:00:32.02)
```

A threefold speedup.

(Note that there is a small [bug](https://github.com/lmcinnes/pynndescent/pull/58) in pynndescent 0.3.0, which means that `n_jobs` needs to be set explicitly. When that's fixed you'll be able to leave it out to use all cores on a machine.)

### DIFF
--- a/scanpy/neighbors/__init__.py
+++ b/scanpy/neighbors/__init__.py
@@ -173,6 +173,85 @@ def neighbors_update(adata, adata_new, k=10, queue_size=5, random_state=0):
     return indices[:, :k], dists[:, :k]
 
 
+# Generates a timestamp for use in logging messages when verbose=True
+def ts():
+    import time
+    return time.ctime(time.time())
+
+
+def nearest_neighbors(
+    X, n_neighbors, metric, metric_kwds, angular, random_state, verbose=False
+):
+    """Compute the ``n_neighbors`` nearest points for each data point in ``X``
+    under ``metric``. This may be exact, but more likely is approximated via
+    nearest neighbor descent.
+
+    Parameters
+    ----------
+    X: array of shape (n_samples, n_features)
+        The input data to compute the k-neighbor graph of.
+
+    n_neighbors: int
+        The number of nearest neighbors to compute for each sample in ``X``.
+
+    metric: string or callable
+        The metric to use for the computation.
+
+    metric_kwds: dict
+        Any arguments to pass to the metric computation function.
+
+    angular: bool
+        Whether to use angular rp trees in NN approximation.
+
+    random_state: np.random state
+        The random state to use for approximate NN computations.
+
+    verbose: bool
+        Whether to print status data during the computation.
+
+    Returns
+    -------
+    knn_indices: array of shape (n_samples, n_neighbors)
+        The indices on the ``n_neighbors`` closest points in the dataset.
+
+    knn_dists: array of shape (n_samples, n_neighbors)
+        The distances to the ``n_neighbors`` closest points in the dataset.
+    """
+    if verbose:
+        print(ts(), "Finding Nearest Neighbors")
+
+    if metric == "precomputed":
+        # Note that this does not support sparse distance matrices yet ...
+        # Compute indices of n nearest neighbors
+        knn_indices = np.argsort(X)[:, :n_neighbors]
+        # Compute the nearest neighbor distances
+        #   (equivalent to np.sort(X)[:,:n_neighbors])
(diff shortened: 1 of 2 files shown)


scanpy/neighbors/__init__.py
```diff
@@ -173,6 +173,85 @@ def neighbors_update(adata, adata_new, k=10, queue_size=5, random_state=0):
...
+            n_iters=n_iters,
+            max_candidates=60,
+            verbose=verbose,
+        )
+        knn_indices, knn_dists = nn_descent._neighbor_graph
+        rp_forest = nn_descent._rp_forest
+
+    if verbose:
+        print(ts(), "Finished Nearest Neighbor Search")
+    return knn_indices, knn_dists, rp_forest
```
gokceneraslan, line 253: Why don't we return `nn_descent` object here instead of the rp_forest? If we store it in `adata.uns` then we can build other methods for querying the graph. What do you think?

Koncopd: Hm, but then we need to be able to write it on disk and read it back, also we need to store forests for the Ingest class - https://github.com/theislab/scanpy/pull/651 . What you are suggesting is basicly the functionality of the Ingest.

falexwolf:
Very cool! Thank you, Tom!

Is Leland also going to make the change within UMAP?

So that, in effect `from umap.umap_ import nearest_neighbors` is more or less equivalent to your implementation here?

scanpy/neighbors/__init__.py
falexwolf: Yes, this whole discussion circles around "how we efficiently store things on disk while providing users convenient objects". The `Ingest` class will take care of several aspects relating to the graph (querying the graph for ingesting new data [we can also think about another name here]). You are right that there might be other use cases for querying the graph. For now, I'd remain conservative and simply keep the forest here. It's not even entering AnnData by default...

One could even think about infrastructure within AnnData. But let's have this discussion somewhere else.

tomwhite:
The code in UMAP and pynndescent is very close now, but I don't know when UMAP is going to use the pyyndescent dependency. So I thought it would be useful to have this PR in the meantime.

ivirshup:
This is super cool!

Is the level of parallelism exposed here amenable to multi-machine parallelism (like the MapReduce strategy mentioned in the nn-descent paper)?

tomwhite:
Thanks @ivirshup.

The parallelism is achieved using a MapReduce scheme operating on a single NumPy array, as described here: https://github.com/lmcinnes/pynndescent/pull/12. This would be amenable to multi-machine parallelism, and in fact I have started a [Dask implementation](https://github.com/tomwhite/pynndescent/tree/dask) that should work on a cluster. However, I haven't benchmarked the Dask implementation, so I don't know how competitive it is with the single (multi-core) machine version using threads. I have successfully run pynndescent on 10^7 rows on a single machine (50 columns, 96 cores), and I don't see why it wouldn't go further than that, although the bottleneck is memory for the heap updates.

ivirshup:
Even cooler!

Do you recall whether the memory usage for a million cells was anything prohibitive?

I was actually thinking of utilizing distribution across machines as a way to scale out if memory usage started becoming an issue (or even just using the same batching strategy on one machine). This could be very useful for large scale dataset integration. Though I'm curious if using nn-descent could introduce some bias towards merging data.

gokceneraslan:
We might consider waiting until the sparse matrix support of pynndescent is thoroughly tested (e.g. https://github.com/lmcinnes/pynndescent/issues/65)

ivirshup:
@tomwhite, what's the status of this PR?

tomwhite:
This is no longer needed since UMAP 0.4 (not yet released) will use pynndescent if it is installed, see https://github.com/lmcinnes/umap/pull/278.

Issue #646, opened 2019-05-15
Simpler gene annotations in plotting functions


gokceneraslan:
We typically have some marker information in the form of an Excel sheet, pandas DataFrame, and eventually a Python dictionary. Using these as gene annotations in various plotting functions (not pl.rank_genes_groups_* family but the others) is a very common task and it looks awesome thanks to @fidelram's `var_group_*` parameters.

It would be even more fantastic to be able to pass simple dict (e.g. the ones we already use in [Malte's marker_gene_overlap](https://scanpy.readthedocs.io/en/latest/api/scanpy.tl.marker_gene_overlap.html#scanpy.tl.marker_gene_overlap)) to plotting functions where `var_group_positions` and `var_group_labels` are populated automatically. 

One caveat is that there might be genes covered by multiple keys, but this is similar to supplying overlapping `var_group_position`s in current api, which can exit with an error.

I already have a function for that but it's absolutely super ugly. I can send a PR after tidying it up, but if anyone else wants to do it, it's perfectly fine.

LuckyMD:
I do this all the time as well. Could you enlighten me as to the `var_group_*` you are talking about? That sounds like it could be very helpful for me.

gokceneraslan:
https://scanpy-tutorials.readthedocs.io/en/latest/visualizing-marker-genes.html#Stacked-violins

What I'm talking about is actually literature-driven marker equivalent of rank_genes_groups plots.

LuckyMD:
Cool... I haven't used the new plots yet. Looks very handy. Your wrapper is an automated use of `var_group_X` to label marker genes of particular clusters in these plots?

fidelram:
I should dig on a function that I have somewhere that translates a dict into the proper input for the plotting functions. This is a use case that  I frequently find. 

I think that a good solution would be to check if the var names is a dict, in which case the var_labels and var_positions are automatically set. Or do you have any other idea? I don’t know if this will make the documentation too complicated. 

At the moment I am travelling, but next week I can look into this.

jorvis:
I know this is a bit specific, but the current sc.pl.rank_genes_groups() and sc.pl.rank_genes_groups_violin() have a 'gene_symbols' argument to allow display based on other columns in .var.   For example, if I have this in Var:

```
Var columns:
                   gene_symbol
index                         
ENSMUSG00000000001       Gnai3
ENSMUSG00000000028       Cdc45
ENSMUSG00000000031         H19
[...]

PR #661, opened 2019-05-23
Dict param to plots #646

fidelram:
This PR addresses #646 by adding the option to pass a dict to the plotting functions heatmap, dotplot, matrixplot, tracksplot and stacked_violin. 

Now, when `var_names` is a dictionary the `var_group_labels` and `var_group_positions` are set such that the dictionary key is a label and the group is the dict values. In the following example the 'brackets' plot on top of the image are prepared based on the markers dictionary:

```PYTHON
marker_genes_dict = {'B-cell': ['CD79A', 'MS4A1'], 
                     'T-cell': 'CD3D',
                     'T-cell CD8+': ['CD8A', 'CD8B'],
                     'NK': ['GNLY', 'NKG7'],
                     'Myeloid': ['CST3', 'LYZ'],
                     'Monocytes': ['FCGR3A'],
                     'Dendritic': ['FCER1A']}
# use marker genes as dict to group them
ax = sc.pl.dotplot(pbmc, marker_genes_dict, groupby='bulk_labels')
```
image

This PR also introduces a small change in `sc.pl.stacked_violin` by setting `cut=0` as default parameter for `seaborn.violin`. This produces in my opinion better plots by removing the extension of the violin past extreme points. This is specially useful to avoid the violin plot to extend below zero expression values. 

**Update**: I set the dependencies to `matplotlib==3.0.*` and `scipy==1.2` to solve failing tests. More details in the conversation


fidelram:
Some of the tests fail for reasons unrelated to the PR (`test_preprocessing`). Locally all tests pass for me. Any ideas?

flying-sheep:
I think it’s again some flakiness of some dependency (or our interaction with it)

LuckyMD:
Are we already able to write `pd.DataFrames` from `anndata.uns` to h5ad? I think someone was working on this... @Koncopd?

fidelram:
@flying-sheep I see that the same tests are failing in other PRs. Do you have any idea what and when some change was introduced that broke the tests

fidelram:
Looking at the commits page I can see that the tests has been failing for some days already. 

First was a problem with a notebook test (test_pbmc3k) that seems innocuous but should be addressed. This is related to release 1.4.3 (https://github.com/theislab/scanpy/commit/85acb6c8949d43d08a26437dceab4fa5db79e246). The commits are unrelated to the failing test so I assume that some dependency was updated  

However, after this commit https://github.com/theislab/scanpy/commit/115d635bf950354509053d976b90c1db518bcffe more errors are found. But again, I don't see any relevant changes that will cause the problems.

One of the errors is that statsmodels is using a deprecated module from scipy.misc:

```
>   from scipy.misc import factorial
E   ImportError: cannot import name 'factorial'
../../../virtualenv/python3.6.7/lib/python3.6/site-packages/statsmodels/distributions/edgeworth.py:7: ImportError
```

This was introduced after scipy 1.3 was recently updated (https://github.com/statsmodels/statsmodels/issues/5759). It seems that currently, the only solution is to install statsmodels directly from the master branch. Or downgrade scipy.

fidelram:
setting scipy==1.2  fixes several errors but there is another one maybe related to matplotlib.

fidelram:
The remaining failed test is related to matplotlib 3.1.0 and 3d scatter plots. There is a report of a similar error (https://github.com/matplotlib/matplotlib/issues/14298). My suggestion is to wait for those issues to be solved and then upgrade the dependencies.

flying-sheep:
Could you please add comments to the requirements.txt (need to be on their own lines)?

This way, in case we forget, we’ll have reminders why the `==`s are there.

fidelram:
@flying-sheep good idea. Added in PR #664

PR #676, opened 2019-06-03
Logging overhaul

### LINKED ISSUE #256: Get rid of logging module and use our own
We should replace the whole logging module with

```py
import logging

root_logger = logging.getLogger('scanpy')
root_logger.setLevel('INFO')
root_logger.propagate = False  # Don’t pass log messages on to the root logger and its handler

handler = logging.StreamHandler(sys.stderr)  # Why did we use stdout?
handler.setFormatter(logging.Formatter('%(message)s'))
handler.setLevel('INFO')
root_logger.addHandler(handler)

def get_logger(name):
   return root_logger.getChild(name)
```

and in all submodules just do

```py
from .logging import get_logger

logger = get_logger(__name__)
```

flying-sheep:
Here we go. ~~It’s not done yet~~:

- [x] The timing stuff isn’t yet implemented. I think we just do timing stuff when doing “info” level logging, correct?

    If yes, we probably just need to override `RootLogger.info` so it sets `self.handlers[0].formatter.passed_time = kwargs.get('t', False)`.

- [x] All the places hackily using `_settings_verbosity_greater_or_equal_than(2|3)` have to work differently. I propose that we just add a kwarg `deepinfo: str` or so which only adds the passed string to the message if the active verbosity is higher than the function’s (i.e. calling `logging.warn('foo', deepinfo='bar')` and the loglevel/verbosity is 'INFO' or noisier adds 'bar')

- [x] We now use vanilla log function syntax, so no more using it like `print`. We switched to Python 3.6+ though, so I propose f-strings everywhere! That looks better and works.

Fixes #256

### DIFF
--- a/scanpy/_settings.py
+++ b/scanpy/_settings.py
@@ -2,7 +2,22 @@ import inspect
 import sys
 from pathlib import Path
 from time import time
-from typing import Tuple, Union, Any, List, Iterable, TextIO
+from typing import Tuple, Union, Any, List, Iterable, TextIO, Optional
+
+from . import logging
+from .logging import _set_log_level, _set_log_file, RootLogger
+
+_VERBOSITY_TO_LOGLEVEL = {
+    'error': 'ERROR',
+    'warn': 'WARNING',
+    'info': 'INFO',
+    'hint': 'HINT',
+    'debug': 'DEBUG',
+}
+verbosity_names = list(_VERBOSITY_TO_LOGLEVEL.keys())
+# Python 3.7 ensures iteration order
+for v, level in enumerate(list(_VERBOSITY_TO_LOGLEVEL.values())):
+    _VERBOSITY_TO_LOGLEVEL[v] = level
 
 
 def _type_check(var: Any, varname: str, types: Union[type, Tuple[type, ...]]):
@@ -43,7 +58,11 @@ class ScanpyConfig:
         _vector_friendly: bool = False,
         _low_resolution_warning: bool = True
     ):
+        # logging
+        self._root_logger = RootLogger('INFO')  # level will be replaced
+        self.logfile = logfile
         self.verbosity = verbosity
+        # rest
         self.plot_suffix = plot_suffix
         self.file_format_data = file_format_data
         self.file_format_figs = file_format_figs
@@ -55,7 +74,6 @@ class ScanpyConfig:
         self.figdir = figdir
         self.max_memory = max_memory
         self.n_jobs = n_jobs
-        self.logfile = logfile
         self.categories_to_ignore = categories_to_ignore
         self._frameon = _frameon
         """bool: See set_figure_params."""
@@ -76,6 +94,9 @@ class ScanpyConfig:
         self._previous_memory_usage = -1
         """Stores the previous memory usage."""
 
+        for fn in verbosity_names:
+            setattr(logging, fn, getattr(self._root_logger, fn))
+
     @property
     def verbosity(self) -> int:
         """
@@ -93,7 +114,10 @@ class ScanpyConfig:
 
     @verbosity.setter
     def verbosity(self, verbosity):
-        verbosity_str_options = ["error", "warn", "info", "hint"]
+        verbosity_str_options = [
+            v for v in _VERBOSITY_TO_LOGLEVEL
+            if isinstance(v, str)
+        ]
         if isinstance(verbosity, int):
             self._verbosity = verbosity
         elif isinstance(verbosity, str):
@@ -107,6 +131,8 @@ class ScanpyConfig:
                 self._verbosity = verbosity_str_options.index(verbosity)
         else:
             _type_check(verbosity, "verbosity", (str, int))
+        from .logging import _set_log_level
+        _set_log_level(self, _VERBOSITY_TO_LOGLEVEL[self._verbosity])
 
     @property
     def plot_suffix(self) -> str:
@@ -248,8 +274,22 @@ class ScanpyConfig:
         self._n_jobs = n_jobs
 
     @property
-    def logfile(self) -> Union[Path, TextIO]:
-        """:class:`~pathlib.Path` of logfile or writable object.
+    def logpath(self) -> Optional[Path]:
+        """The file path `logfile` was set to."""
+        return self._logpath
+
+    @logpath.setter
+    def logpath(self, logpath: Union[str, Path, None]):
+        _type_check(logpath, "logfile", (str, Path))
+        self._logpath = Path(logpath)
+        # set via “file object” branch of logfile.setter
+        self.logfile = self._logpath.open('a')
+
+    @property
+    def logfile(self) -> TextIO:
+        """The open file to write logs to.
+
+        Set it to a :class:`~pathlib.Path` or :class:`str: to open a new one.
         The default `None` corresponds to :obj:`sys.stdout` in jupyter notebooks
         and to :obj:`sys.stderr` otherwise.
         
@@ -259,9 +299,15 @@ class ScanpyConfig:
 
     @logfile.setter
     def logfile(self, logfile: Union[str, Path, TextIO, None]):
-        _type_check(logfile, "logfile", (str, Path, type(None)))
-        stream = sys.stdout if self._is_run_from_ipython() else sys.stderr
-        self._logfile = Path(logfile) if logfile else stream
+        if not hasattr(logfile, 'write') and logfile:
+            self.logpath = logfile
+        else:  # file object
+            if not logfile:  # None or ''
+                logfile = sys.stdout if self._is_run_from_ipython() else sys.stderr
+            self._logfile = logfile
+            self._logpath = None
+            from .logging import _set_log_file
+            _set_log_file(self)
 
     @property
     def categories_to_ignore(self) -> List[str]:
--- a/scanpy/_settings.py
+++ b/scanpy/_settings.py
@@ -1,5 +1,6 @@
 import inspect
 import sys
+from enum import IntEnum
 from pathlib import Path
 from time import time
 from typing import Tuple, Union, Any, List, Iterable, TextIO, Optional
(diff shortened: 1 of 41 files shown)


LuckyMD:
>     3.\ We switched to Python 3.7+ though

No more support for python 3.6? That was a quick move from 3.5 support to only 3.7+...

flying-sheep:
Oh, whoops, 3.6 of course. That was the one introducing f-strings. Sorry, I’m not that good in ancient history :wink:

scanpy/_settings.py
```diff
@@ -259,9 +299,15 @@ def logfile(self) -> Union[Path, TextIO]:
...
-        _type_check(logfile, "logfile", (str, Path, type(None)))
-        stream = sys.stdout if self._is_run_from_ipython() else sys.stderr
-        self._logfile = Path(logfile) if logfile else stream
+        if not hasattr(logfile, 'write') and logfile:
+            self.logpath = logfile
+        else:  # file object
+            if not logfile:  # None or ''
+                logfile = sys.stdout if self._is_run_from_ipython() else sys.stderr
+            self._logfile = logfile
+            self._logpath = None
```
ivirshup, line 317: There's a logic error here. If you set the `logpath`, it sets `_logpath` then `logfile`. Setting `logfile` sets `_logpath` to `None`.

On my machine, on commit e5a38fd86c1f

```python
import scanpy as sc
sc.settings.logpath = "./tmp.log" 
sc.settings.logpath is None
# True
```

scanpy/preprocessing/_simple.py
```diff
@@ -138,15 +138,14 @@ def filter_cells(
...
-                   if min_counts is None else str(min_counts) + ' counts', no_indent=True)
+            msg += 'less than '
+            msg += f'{min_genes} genes expressed' if min_counts is None else f'{min_counts} counts'
         if max_genes is not None or max_counts is not None:
-            logg.info('more than ',
-                   str(max_genes) + ' genes expressed'
-                   if max_counts is None else str(max_counts) + ' counts', no_indent=True)
+            msg += 'more than '
+            msg += f'{max_genes} genes expressed' if max_counts is None else f'{max_counts} counts'
+        logg.info(msg)
```
ivirshup, line 148: Also minor: why not `time=True`?

Similar for all functions where there's an `info` at the beginning and end.

scanpy/_settings.py
```diff
@@ -2,7 +2,22 @@
...
+    'error': 'ERROR',
+    'warn': 'WARNING',
+    'info': 'INFO',
+    'hint': 'HINT',
+    'debug': 'DEBUG',
+}
+verbosity_names = list(_VERBOSITY_TO_LOGLEVEL.keys())
+# Python 3.7 ensures iteration order
+for v, level in enumerate(list(_VERBOSITY_TO_LOGLEVEL.values())):
+    _VERBOSITY_TO_LOGLEVEL[v] = level
```
ivirshup, line 21: Minor issue, but I think this would be more clear:

```python
_VERBOSITY_TO_LOGLEVEL = {
    'error': 'ERROR',
    'warn': 'WARNING',
    'info': 'INFO',
    'hint': 'HINT',
    'debug': 'DEBUG',
    0: 'ERROR',
    1: 'WARNING',
    2: 'INFO',
    3: 'HINT',
    4: 'DEBUG'
}

VERBOSITY_NAMES = [
    k for k in _VERBOSITY_TO_LOGLEVEL if isinstance(k, str)
]
```

`VERBOSITY_NAMES` could even be set by a literal, since the duplication is pretty minor. I think it's worth using global variable naming conventions for it as well.

Also, `VERBOSITY_NAMES` could replace `verbosity_str_options` on line 117.

ivirshup (changes requested):
Mostly good! I've got some minor comments, and found one bug (logic error in setting `logpath`, so it's `None` when it shouldn't be). The only change I'm requesting is that one.

I didn't look too closely at all the changes to usage in scanpy, but it could be worth a check.

It could be worth adding some tests for this? Probably just a few would be fine. I'm thinking:

* time should be recorded correctly
* setting the log level works, makes the correct things be logged
* setting the log file works

scanpy/_settings.py
```diff
@@ -2,7 +2,22 @@
...
 from time import time
-from typing import Tuple, Union, Any, List, Iterable, TextIO
+from typing import Tuple, Union, Any, List, Iterable, TextIO, Optional
+
+from . import logging
+from .logging import _set_log_level, _set_log_file, RootLogger
+
+_VERBOSITY_TO_LOGLEVEL = {
+    'error': 'ERROR',
+    'warn': 'WARNING',
```
ivirshup: Found one more issue here:

```python
>>> sc.logging.warn("hi")                                                                        
/usr/local/bin/ipython:1: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
  #!/usr/local/opt/python/bin/python3.7
WARNING: hi
>>> sc.logging.warning
AttributeError                            Traceback (most recent call last)
<ipython-input-9-8d955ccc14ce> in <module>
----> 1 sc.logging.warning

AttributeError: module 'scanpy.logging' has no attribute 'warning'
```

scanpy/external/_tools/_palantir.py
```diff
@@ -133,7 +133,7 @@ def palantir(adata: AnnData):
 
     """
 
-    logg.info('Palantir diffusion maps', r=True)
+    logg.info('Palantir diffusion maps')
```
falexwolf, line 136: If you don't reset (or start) a timer, there is no way at which point the timer should know when to start timing. So, I fear, there needs to be an additional parameter here.

scanpy/logging.py
```diff
@@ -1,136 +1,92 @@
...
 
+class RootLogger(logging.RootLogger):
+    def __init__(self, level):
+        super().__init__(level)
+        self.propagate = False
+        self.last_time = time_.time()
 
-def info(*args, **kwargs):
-    return _msg(*args, v='info', **kwargs)
+    def log(self, level, msg, *, extra=None, deep=None, time_passed=None):
```
falexwolf: Ah, you even used the `time_passed` argument. 🙂

scanpy/neighbors/__init__.py
```diff
@@ -108,11 +107,15 @@ def neighbors(
...
     if neighbors.rp_forest is not None:
         adata.uns['neighbors']['rp_forest'] = neighbors.rp_forest
-    logg.info('    finished', time=True, end=' ' if _settings_verbosity_greater_or_equal_than(3) else '\n')
-    logg.hint(
-        'added to `.uns[\'neighbors\']`\n'
-        '    \'distances\', distances for each pair of neighbors\n'
-        '    \'connectivities\', weighted adjacency matrix')
+    logg.info(
+        '    finished',
+        time=True,
```
falexwolf: OK, you're assuming that any previous `logg.info` call starts the timer. But that means that everything in between has to be logged with `logg.debug`. But often, we also want to logg out info also within the function. Hm. What do you think?

falexwolf:
Very cool, @flying-sheep! Really the only concern I have is whether we need an option to prevent setting the timer (or for setting the timer) when calling `logg.info`...

scanpy/_settings.py
flying-sheep: Good point, I switched everything to `warning`!

scanpy/_settings.py
flying-sheep: I just introduced an IntEnum that gets used instead of integers.

I guess we should simplify the code to use that one!

scanpy/_settings.py
flying-sheep: Thank you, fixed

scanpy/preprocessing/_simple.py
flying-sheep: I think I didn’t change that anywhere. The idea was to change as little as possible while changing the API

scanpy/external/_tools/_palantir.py
flying-sheep: I checked and it seems like automatically resetting at every `info` call works, as there seem to be no `info` loggings in the middle of some long running process.

scanpy/logging.py
flying-sheep: yeah! I heavily modified your code, but based the implementation on it.

scanpy/neighbors/__init__.py
flying-sheep: It doesn’t seem like there’s any info call in the middle of a long running process at the moment.

But theoretically changing that behavior is of course possible.

flying-sheep:
I’m doing all the tests @ivirshup proposed. If my explanations to the questions I left open are sufficient, I’ll merge this! :smile:

scanpy/preprocessing/_simple.py
ivirshup: Fair enough

scanpy/_settings.py
ivirshup: Always good to reduce duplication!

scanpy/_settings.py
ivirshup: Was `warn` used before? If so, maybe we could keep a deprecated alias `warn` for backwards compatibility?

scanpy/neighbors/__init__.py
flying-sheep: I switched to explicit time passing

flying-sheep:
OK, one more change. The log functions now all return the current time and have the optional parameter `time: datetime`. If you pass something there, the time will be logged:

```py
start = log.info('foo')
# do stuff
log.hint('bar', time=start)  # --> bar (00:00:02)
```

You can customize where the time ends up via `log.*('blah {time_passed}: blub', time=...)`

falexwolf:
Cool solution! Really cool new logging module! :)

scanpy/neighbors/__init__.py
falexwolf: Great!

falexwolf:
`logg.warning` vs. `logg.warn`. I liked the short and verbal `.warn` better. I know there is some confusion, because of how python's core warning and logging modules possible, but meanwhile, several other packages have adapted Scanpy's logging module.

All of them now need to change each line from `logg.warn` to `logg.warning` and even I will tend to make a lot of errors being used to `logg.warn` (still most of the time using emacs without autosuggest...). So, as Isaac, I'd also like the equivalent `logg.warn` function but wouldn't even deprecate it.

flying-sheep:
Well, the logging module is undocumented, only `logging.print_versions()` is public. And I changed the signature, it no longer works like `print`, so they’ll have to change it anyway. But I think the new API is pretty sweet, so we can stabilize and document it now.

Python has `warnings.warn` and `logging.warning`. I think we should follow suit. You can of course always do `from warnings import warn` and use that. It doesn’t interact with our logging system (yet), though!

PR #685, opened 2019-06-10
Move adding `--internet-tests` option back to `conftest.py`

ivirshup:
`pytest_addoption` and `pytest_collection_modifyitems` have to be defined where `pytest` gets called from, otherwise they don't work.

@flying-sheep, I saw you had changed this in d979267f48607fd609954c96cd5c586b6135dc30. Was there a functional reason you moved these things, or was it aesthetic?

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,7 @@
 import pytest
 
+# Command line options for pytest must be added from conftest.py from where
+# `pytest` is called.
 def pytest_addoption(parser):
     parser.addoption(
         "--internet-tests",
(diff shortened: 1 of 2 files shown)


flying-sheep:
Yes, sorry, I wasn’t aware of this. Please add a small comment explaining this.

ivirshup:
Added. Do you think we should keep multiple `conftest.py` files, or have it all in one place?

flying-sheep:
multiple are fine, I like having the test setup in the tests directory

PR #703, opened 2019-06-24
Defer import of seaborn

ivirshup:
Cuts around 1 second off import of scanpy.

```sh
isaac@Mimir ~/github/scanpy$ time python3 -c "import scanpy"                        ✭defer-seaborn 
python3 -c "import scanpy"  3.36s user 0.61s system 104% cpu 3.801 total
isaac@Mimir ~/github/scanpy$ git checkout master                                    ✭defer-seaborn 
Switched to branch 'master'
Your branch is up to date with 'upstream/master'.
isaac@Mimir ~/github/scanpy$ time python3 -c "import scanpy"                               ✭master 
python3 -c "import scanpy"  4.23s user 0.48s system 108% cpu 4.324 total
```

I'd like to cut down import time even more, but I figure this is a good place to start.

PR #704, opened 2019-06-24
Defer loading of umap to speed up import

ivirshup:
Follow up to #703, this should take import times down another second. It does add a dependency, but that dependency will be in the stdlib from 3.8+.

After this it looks like the next biggest contributors to import times are caused by `numba` (which would be hard to factor out) and `sklearn.metrics`, but these are on a smaller scale. Evidence of speedup:

```sh
isaac@Mimir ~/github/scanpy$ time python3 -c "import scanpy as sc"                      ✹ ✭master 
python3 -c "import scanpy as sc"  3.10s user 0.41s system 116% cpu 3.010 total
isaac@Mimir ~/github/scanpy$ git checkout defer-umap                                    ✹ ✭master 
Switched to branch 'defer-umap'
Your branch is up to date with 'origin/defer-umap'.
isaac@Mimir ~/github/scanpy$ time python3 -c "import scanpy as sc"                  ✹ ✭defer-umap 
python3 -c "import scanpy as sc"  2.18s user 0.42s system 131% cpu 1.981 total
```

### DIFF
--- a/scanpy/utils.py
+++ b/scanpy/utils.py
@@ -25,29 +25,29 @@ EPS = 1e-15
 
 def check_versions():
     from distutils.version import LooseVersion
+    from importlib_metadata import version
 
     if sys.version_info < (3, 6):
         warnings.warn('Scanpy prefers Python 3.6 or higher. '
                       'Currently, Python 3.5 leads to a bug in `tl.marker_gene_overlap` '
                       'and we might stop supporting it in the future.')
 
-    import anndata, umap
-    # NOTE: pytest does not correctly retrieve anndata's version? why?
-    #       use the following hack...
-    if anndata.__version__ != '0+unknown':
-        if anndata.__version__ < LooseVersion('0.6.10'):
-            from . import __version__
-            raise ImportError('Scanpy {} needs anndata version >=0.6.10, not {}.\n'
-                              'Run `pip install anndata -U --no-deps`.'
-                              .format(__version__, anndata.__version__))
+    anndata_version = version("anndata")
+    umap_version = version("umap-learn")
 
-    if umap.__version__ < LooseVersion('0.3.0'):
+    if anndata_version < LooseVersion('0.6.10'):
+        from . import __version__
+        raise ImportError('Scanpy {} needs anndata version >=0.6.10, not {}.\n'
+                            'Run `pip install anndata -U --no-deps`.'
+                            .format(__version__, anndata_version))
+
+    if umap_version < LooseVersion('0.3.0'):
         from . import __version__
         # make this a warning, not an error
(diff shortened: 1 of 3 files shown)


flying-sheep:
Please make the import conditional using

1. a `python_version < 3.8` [environment marker](https://www.python.org/dev/peps/pep-0345/#environment-markers) and
2. Try importing the 3.8 location (`importlib.metadata`) first before importing the backport module.

ivirshup:
Done.

Just so I can figure out how I did it later, here's the commands for profiling import times:

```python
import cProfile
from pstats import SortKey, Stats

# This imports scanpy to current env, can't be run twice in a session
cProfile.run("import scanpy", "import-scanpy.profile")
p = Stats("import-scanpy.profile")
p.sort_stats(SortKey.CUMULATIVE)
p.print_stats(30)
```

scanpy/utils.py
```diff
@@ -25,7 +25,12 @@
 
 def check_versions():
     from distutils.version import LooseVersion
-    from importlib_metadata import version
+    import importlib
+
+    if importlib.util.find_spec("importlib.metadata") is not None:
+        from importlib.metadata import version
+    else:
+        from importlib_metadata import version
```
flying-sheep, line 32: I think it’s more pythonic to simply do

```py
try:
    from importlib.metadata import version
except ImportError:  # < Python 3.8
    from importlib_metadata import version

flying-sheep:
I started to test also with 3.8, but sadly there’s a bug in scipy: scipy/scipy#10354

ivirshup:
I think the `<` should work, it's part of [PEP-508](https://www.python.org/dev/peps/pep-0508/#motivation). At least that's what [this blog post](https://hynek.me/articles/conditional-python-dependencies/) says.

Didn't those builds just fail before they got to that requirement?

flying-sheep:
You’re right! I wasn’t aware of PEP 508.

The build appears to fail [here](https://travis-ci.org/theislab/scanpy/jobs/550248884#L285), but actually fails much later, due to numpy/numpy#13790.

I filed pypa/pip#6651 to deal with the error being unclear.

ivirshup:
I don't think we need to test against 3.8 until a stable release is out. I'm thinking we can just drop that from travis and merge?

flying-sheep:
Done!

LuckyMD:
I'm getting an error loading scanpy (#739 ), and it points to the line you moved about deferring loading of umap-learn. 

When I revert back to commit abf95c645828e29edf5a7a27b05d9397f3c36f65 (the commit a couple before this), it works.

flying-sheep:
Doesn’t seem like a problem with this. Further discussion in #739.

PR #705, opened 2019-06-24
making Cell ranger's normalized dispersion (for highly variable gene selection) not absolute

LisaSikkema:
Deleted abs(...) from normalized dispersion definition. Taking the absolute dispersion will also result in selection of genes with dispersion lower than the bin's median, whereas we only want to select genes with dispersion higher than the bin's median.

### DIFF
--- a/scanpy/preprocessing/_highly_variable_genes.py
+++ b/scanpy/preprocessing/_highly_variable_genes.py
@@ -128,12 +128,9 @@ def _highly_variable_genes_single_batch(
         with warnings.catch_warnings():
             warnings.simplefilter('ignore')
             disp_mad_bin = disp_grouped.apply(robust.mad)
-        df['dispersions_norm'] = (
-            np.abs(
-                df['dispersions'].values
+            df['dispersions_norm'] = (df['dispersions'].values
                 - disp_median_bin[df['mean_bin'].values].values
-            ) / disp_mad_bin[df['mean_bin'].values].values
-        )
+                ) / disp_mad_bin[df['mean_bin'].values].values
     else:
         raise ValueError('`flavor` needs to be "seurat" or "cell_ranger"')
     dispersion_norm = df['dispersions_norm'].values.astype('float32')


ivirshup:
Thanks for the PR!

@Koncopd does this look right to you? I'm not too familiar with this code, but this looks like [what cellranger does](https://github.com/10XGenomics/cellranger/blob/5f5a6293bbc067e1965e50f0277286914b96c908/lib/python/cellranger/analysis/stats.py#L36-L67) to me.

LuckyMD:
This would be backwards breaking, right? Might need to output a warning for that.

Koncopd:
@ivirshup this looks fine to me, but i just took the cell ranger version for the refactored `highly_variable_genes ` from the last version, so also not an expert. But again, this looks right.

ivirshup:
@Koncopd, ah, this was as far back as I could trace the code initially: https://github.com/theislab/scanpy/commit/af4a82a2540eee65c732cb5e401d2145846e6d97. Now I've found an earlier commit from @falexwolf at https://github.com/theislab/scanpy/commit/43e71fe8577a8b3a51dc2117bd431911001d9869.

@falexwolf, does this change look right to you?

@LuckyMD, I'm not sure if this would count as backward breaking if it's a bug. Should definitely go into  the release notes, maybe as warning.

LisaSikkema:
Hi! Thank you for looking into this. Any progress with the issue?

LuckyMD:
@falexwolf @ivirshup it would be good to clear this up soon. If this is correct, then we choose HVGs incorrectly every time.

LuckyMD:
I wonder if the absolute value was initially used as the dispersions can be negative due to negative mean expression values? That could be remedied by taking the absolute value of only the dispersions though, and not the dispersion offset by the median dispersion.

LisaSikkema:
These figures might be clarifying:
Original code:
image
Corrected code:
image

ivirshup:
I'd vote to make this change. I'd just hoped to figure out if there was a reason it was not this way in the first place.

LuckyMD:
@ivirshup Agreed. It seems using an `np.abs()` call would have a rationale behind it.

LisaSikkema:
We could ask the authors of the paper, they even mention taking the absolute value in their methods if I remember correctly

LuckyMD:
Which paper? The Zheng et al. 2017?

LisaSikkema:
Yes: 
"Normalized dispersion is calculated as the absolute difference between dispersion and median dispersion of the expression mean, normalized by median absolute deviation within each bin."

flying-sheep:
Great, thank you!

/edit: because I became confused later: they mentioned taking the absolute in the paper, but it must be a mistake since they don’t do it in the code, since as the plot above shows, it doesn’t make sense to do it.

PR #716, opened 2019-06-27
Update verbosity call in sc.external.tl.phate

milescsmith:
The phate too makes a call to a function named `_settings_verbosity_greater_or_equal_than()`, which appears to no longer exist, preventing phate from running at all.  I exchanged that call for a simple comparison that yields what I believe the old function returned.  In any case, the change results in phate being able to run.


flying-sheep:
Thank you!

PR #721, opened 2019-07-02
Make External more visible in docs

### LINKED ISSUE #588: Give `external` higher billing in the docs?
At the moment external modules are kind of hidden in the docs. I think it'd be worth making them more visible (at least on the same page as everything else). I've been giving this a shot, but have hit the limit of my sphinx/ rst abilities.

Two ideas for how they could be more discoverable:

* They get their own heading under `api`
* They're mixed in with everything else (so everything stays organized by topic), but their names are prepended with `sce` while scanpy functions are prepended with `sc`.

ivirshup:
Fixes #588

I'm not sure how to deal with references to external under api, so there are two copies of external at the moment. Sphinx gives warnings about this, and it'll probably mess with search. Any ideas @flying-sheep?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 # Docs
 /docs/_build/
 /docs/api/scanpy.*.rst
+/docs/external/scanpy.*.rst
 !/docs/api/scanpy.api.rst
 !/docs/api/scanpy.api.AnnData.rst
 !/docs/api/scanpy.plotting.rstk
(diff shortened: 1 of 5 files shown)


ivirshup:
Fixes #737

flying-sheep:
> I'm not sure how to deal with references to external under api, so there are two copies of external at the moment.

can you elaborate please? I don’t understand this.

Issue #666, opened 2019-05-29
sc.tl.umap numba error when used with init_pos="paga"


jenzopr:
Hi Alex,

UMAP throws an error if I use `scanpy.tl.ump` with initial positions from `sc.tl.paga`. Based on the error (see below) I thought it was a problem of UMAP itself. However, the error is not thrown when called without initial positions from paga.

Here is the output / error:

```pytb
sc.tl.umap(adata, init_pos='paga')

computing UMAP
    using 'X_pca' with n_pcs = 50

---------------------------------------------------------------------------
TypingError                               Traceback (most recent call last)
<ipython-input-35-924452b37e5b> in <module>
----> 1 sc.tl.umap(adata, init_pos='paga')

/opt/conda/lib/python3.7/site-packages/scanpy/tools/_umap.py in umap(adata, min_dist, spread, n_components, maxiter, alpha, gamma, negative_sample_rate, init_pos, random_state, a, b, copy)
    137         neigh_params.get('metric', 'euclidean'),
    138         neigh_params.get('metric_kwds', {}),
--> 139         verbose=max(0, verbosity-3))
    140     adata.obsm['X_umap'] = X_umap  # annotate samples with UMAP coordinates
    141     logg.info('    finished', time=True, end=' ' if _settings_verbosity_greater_or_equal_than(3) else '\n')

/opt/conda/lib/python3.7/site-packages/umap/umap_.py in simplicial_set_embedding(data, graph, n_components, initial_alpha, a, b, gamma, negative_sample_rate, n_epochs, init, random_state, metric, metric_kwds, verbose)
    984         initial_alpha,
    985         negative_sample_rate,
--> 986         verbose=verbose,
    987     )
    988 

/opt/conda/lib/python3.7/site-packages/numba/dispatcher.py in _compile_for_args(self, *args, **kws)
    348                 e.patch_message(msg)
    349 
--> 350             error_rewrite(e, 'typing')
    351         except errors.UnsupportedError as e:
    352             # Something unsupported is present in the user code, add help info

/opt/conda/lib/python3.7/site-packages/numba/dispatcher.py in error_rewrite(e, issue_type)
    315                 raise e
    316             else:
--> 317                 reraise(type(e), e, None)
    318 
    319         argtypes = []

/opt/conda/lib/python3.7/site-packages/numba/six.py in reraise(tp, value, tb)
    656             value = tp()
    657         if value.__traceback__ is not tb:
--> 658             raise value.with_traceback(tb)
[...]

PR #724, opened 2019-07-03
Fix init_pos argument of sc.tl.umap

ivirshup:
* Now actually allows passing an array (previously threw error from `if init_pos in adata.obsm.keys()`)
* Additionally allows providing an array (even through key of obsm or via paga) of dtype other than float32
* Code converting arrays to float32 can be removed if https://github.com/lmcinnes/umap/pull/262 gets merged and released
* Should solve #666 related issues

### DIFF
--- a/scanpy/tests/test_embedding.py
+++ b/scanpy/tests/test_embedding.py
@@ -1,0 +1,13 @@
+import numpy as np
+from sklearn.utils.testing import assert_array_almost_equal
+
+import scanpy as sc
+
+def test_umap_init_dtype():
+    pbmc = sc.datasets.pbmc68k_reduced()
+    sc.tl.umap(pbmc, init_pos=pbmc.obsm["X_pca"][:, :2].astype(np.float32))
+    embed1 = pbmc.obsm["X_umap"].copy()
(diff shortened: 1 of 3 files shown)


ivirshup:
The test takes a while to run. I'm not sure how much of that is this being the first to import UMAP, but I'll look into it before merging.

falexwolf:
Great!

ivirshup:
Reducing the size of the dataset used from 700 to 100 only reduced total time from ~7.5 seconds to 6 seconds, so I think it's mostly the UMAP import. I've done that anyways and fixed a warning in the embedding plots.

flying-sheep:
if we import everything slow in conftest.py, then the test times won’t be misleading.

Thanks for slaying the “issue of the beast” :metal::stuck_out_tongue_closed_eyes:

ivirshup:
Ha, yeah, the issue number definitely encouraged me to fix it.

That's a good idea! I'm not sure I know what the pytest execution model is like, but does it ever start new processes for different tests?

flying-sheep:
> I'm not sure I know what the pytest execution model is like, but does it ever start new processes for different tests?

Apparently [it needs a plugin](https://pypi.org/project/pytest-parallel/) for that.

PR #733, opened 2019-07-09
Check return types of distributed computations

ivirshup:
Add checks to make sure the distributed backend is staying distributed. This currently fails for dask arrays, since they become numpy arrays when subset.

@tomwhite Is this it right to assume that dask arrays should stay as dask arrays, and Zappy arrays should stay as Zappy arrays? Currently dask arrays are becoming numpy arrays when the parent AnnData is subset via the last line here (though this exact code is being replaced):

https://github.com/theislab/anndata/blob/75296bb210d141fa1de24960412bfcab77a9604d/anndata/core/anndata.py#L637-L657

Also, are `ZappyArray`s always read-only like the `DirectZappyArray`s are? I'm wondering about the set of anndata features they should work with.

### DIFF
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -664,8 +664,10 @@ def normalize_per_cell(
         warnings.simplefilter("ignore")
         counts_per_cell += counts_per_cell == 0
         counts_per_cell /= counts_per_cell_after
-        if not issparse(X): X /= materialize_as_ndarray(counts_per_cell[:, np.newaxis])
-        else: sparsefuncs.inplace_row_scale(X, 1/counts_per_cell)
+        if not issparse(X):
+            np.true_divide(X, counts_per_cell[:, np.newaxis], out=X)
+        else:
+            sparsefuncs.inplace_row_scale(X, 1/counts_per_cell)
     return X if copy else None
 
 


tomwhite:
Thanks for opening this @ivirshup. The principle should be that Anndata doesn't change array types to numpy arrays. It may not have done this in the past (and this is one example), but we should fix that. Hopefully with improvements in numpy itself (like https://www.numpy.org/neps/nep-0018-array-function-protocol.html, and `__array_ufunc__`) it should be achievable. (I don't know how much work is needed in Anndata to do that though.)

> Also, are `ZappyArray`s always read-only like the `DirectZappyArrays` are?

Yes, the idea is that they are immutable, and they are changed by transformation (e.g. applying a ufunc) to create a new array.

ivirshup:
> The principle should be that Anndata doesn't change array types to numpy arrays.

I mostly agree with this, but think there would be a fair amount of work needed in AnnData. Most array types have special treatment in at least one place. A lot of this is due to our need to support sparse matrices. I'm hoping this can be reduced with some new stuff I'm adding though. This is definitely a good use case to test with.

There's also the issue of when converting implicitly makes sense. We will do transformations from sparse to dense if the values becomes dense. We will also convert between sparse matrix formats if it will make calculations more efficient.

On the topic of Zappy arrays. Can you take a `view` of a zappy array? This would be useful for some of the expectations around subsetting AnnData objects.

ivirshup:
Looking at the view code again, I see a key pain point for working with generic arrays is how our views work. Right now, we have to make a new subclass for every array type we accept, so that we can take views of arbitrary indices.

ivirshup:
I've managed to mostly fix this with only a slightly hacky change to AnnData (no longer requiring X to have a view type defined when making a copy). However the `normalize_per_cell` test still fails, as it seems like the matrix isn't actually being modified. I think it's because the reference to the dask array on [this line](https://github.com/theislab/scanpy/blob/610a955f025f5f17328865926a9341a55553e081/scanpy/preprocessing/_simple.py#L667) becomes a copy when assigned to.

I've asked about this behaviour, and if it might change or throw a warning in  https://github.com/dask/dask/issues/5199

LuckyMD:
@aopisco This might be relevant to your questions about zarr arrays in backed/distributed mode.

ivirshup:
~~@tomwhite, I'm getting a little lost with dask at the moment. Right now this works on my machine, but doesn't work on travis. Any idea what's going on?~~

Nope, I was dumb. It's obviously that anndata hasn't been updated in this build.

Issue #762, opened 2019-08-01
sc.pl.scatter for var


coh-racng:
I get the following error when trying to use sc.pl.scatter to plot gene expression, with use_raw=False. I am using sc.pl.scatter instead of sc.pl.umap, .tsne, etc.,  because of the need to use custom basis names.
```
  File "/opt/Python/3.6.5/lib/python3.6/site-packages/scanpy/plotting/_anndata.py", line 118, in scatter
    ax=ax)
  File "/opt/Python/3.6.5/lib/python3.6/site-packages/scanpy/plotting/_anndata.py", line 390, in _scatter_obs
    c = adata.raw.obs_vector(key, layer=layers[2])
TypeError: obs_vector() got an unexpected keyword argument 'layer'
```

The following snippet is copied from `_scatter_obs()` in /scanpy/plotting/_anndata.py
``` python
        # coloring according to gene expression
        elif (use_raw
              and adata.raw is not None
              and key in adata.raw.var_names):
            c = adata.raw.obs_vector(key)
        elif key in adata.var_names:
            c = adata.raw.obs_vector(key, layer=layers[2])
```
Should line 390 be c = adata.obs_vector(key, layer=layers[2]) since it is handling the case when use_raw==False and adata.raw.obs_vector does not take layer as argument.

ivirshup:
Thanks for the bug report! That sounds totally right, would you mind submitting a PR to fix that?

fidelram:
@coh-racng I would like to add that for your specific intention the best way is to load the `plot_scatter` function that accepts `basis` as parameter and works well with layers.

The code should be:
```PYTHON
from scanpy._plotting.scatterplots import plot_scatter`
plot_scatter(adata, basis='<name>'....)
```

@ivirshup, @falexwolf  I think we should add `plot_scatter` to the API maybe renaming it `plot_embedding` to help users like @coh-racng. Currently we have two different ways to make scatter plots: One for embeddings (`plot_scatter`) and other more generic for obs and vars (`sc.pl.scatter`) that accepts `x` and `y` parameters.

ivirshup:
Definitely agree with you @fidelram, I think the main thing stopping us before was unfortunately just picking the name. `plot_{...}` doesn't really match any of our other functions. Names that come up for me are: `sc.pl.scatter` (obviously taken), `sc.pl.dimred`, or `sc.pl.scatter_obsm`.

fidelram:
[...]

PR #764, opened 2019-08-02
 Fix bug for plotting var when use_raw=False

coh-racng:
#762


ivirshup:
Thanks for the PR! Could you add a test too?

ivirshup:
@coh-racng, I've merged your fix in #790 with a test added. Thanks!

Issue #348, opened 2018-11-06
sc.pl.violin not making multipanels


cartal:
Hi, 

**Issue**: When I try to use the multipanel option, the command runs without errors but the plot never shows up:

`sc.pl.violin(adata, keys = ['IL10', 'IFNG', 'CD3G'], multi_panel = None, groupby=None)`

**Question**: Is there a way to split the violin plots form the sc.pl.violin() into a specific number of rows? Similar to the `ncols=` in R?
 
Thank you!

fidelram:
Can you help me with a quick example to reproduce the problem? You can use example datasets contained in scanpy.

MichaelPeibo:
Hi @cartal 
I tried to plot violin plot in multiple columns. How did you go through with it?

rboiarsky:
Can we reopen this issue? I still don't see this functionality available. The ask is to be able to specify the number of rows or columns for the arrangement of the output panels from `sc.pl.violin`. Right now if I plot 8 genes, for example, they all show up on one row, yielding tiny plots. It would be nice to be able to pass in something like `ncols=4` so that the 8 panels will be arranged as a 2x4 instead of a 1x8.

ivirshup:
@rpeys I think that might be better as a new feature request. I'd like to keep this closed since the bug that was reported here has been fixed.

PR #776, opened 2019-08-08
Fix sc.pl.violin not making violinplots

gokceneraslan:
When groupby is not None and multipanel is True, nothing is plotted. Now we fallback to non-multipanel if groupby is specified. Related to #348.


fidelram:
@gokceneraslan Thanks for looking at this. Can you add test for this?

PR #784, opened 2019-08-15
Preparations for UMAP 0.4 release

Koncopd:
https://github.com/theislab/scanpy/issues/779

Also i see some changes in the internal nnescent of umap and it is problematic for [Ingest](https://github.com/theislab/scanpy/pull/651).

### DIFF
--- a/scanpy/neighbors/__init__.py
+++ b/scanpy/neighbors/__init__.py
@@ -319,7 +319,8 @@ def compute_connectivities_umap(
                                           set_op_mix_ratio=set_op_mix_ratio,
                                           local_connectivity=local_connectivity)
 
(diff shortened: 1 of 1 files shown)


gokceneraslan:
Do you think we also need to check if the umap version is >= 0.4 or so?

flying-sheep:
Nah, if it returns a tuple, we take the first element, if not, we assume it’s a sparse matrix

gokceneraslan:
Did anyone run scanpy tests with umap 0.4 branch?

flying-sheep:
Not me, but I checked the current and old umap code and found that this PR will only fix things and not cause problems. IDK if umap 0.4 causes further problems

PR #789, opened 2019-08-20
Fix paga when no argument is given

ivirshup:
@flying-sheep 

I've moved the argument handling from the functions to the constructor since `groups` is used later in the function, and I thought changing it's value midway seemed inelegant. This should get builds working again.


flying-sheep:
Thank you! IDK how I missed that it’s used further down.

PR #790, opened 2019-08-20
Fix scatter `use_raw`

### LINKED ISSUE #762: sc.pl.scatter for var

ivirshup:
Fixes #762. Update to #764.

Just adding a test to make sure the `use_raw` argument does something. I'm opening this a new pull request since git didn't me using other PRs branch name.

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -387,7 +387,7 @@ def _scatter_obs(
               and key in adata.raw.var_names):
             c = adata.raw.obs_vector(key)
         elif key in adata.var_names:
-            c = adata.raw.obs_vector(key, layer=layers[2])
+            c = adata.obs_vector(key, layer=layers[2])
         elif is_color_like(key):  # a flat color
             c = key
             colorbar = False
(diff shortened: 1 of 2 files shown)


ivirshup:
Thanks for originally opening this @coh-racng!

PR #791, opened 2019-08-20
Pass on copy and return_info to dca

gamazeps:
The arguments were not passed in preprocessing/_dca.py, which
makes debugging much more complicated  than needed.


gamazeps:
The failing test is here https://travis-ci.org/theislab/scanpy/jobs/574409535#L386

However test_var_df should not be affected by this PR, and this test is also failing on master,

flying-sheep:
Yes, sorry, that was my fault. Rerunning the tests for this PR

Issue #775, opened 2019-08-08
Normalization fun with vmin/vmax


gokceneraslan:
Dear people,

In our plotting functions, vmin and vmax are great arguments, we all love them. First of all, let's all hold our hands, close our eyes and thank matplotlib developers for implementing them.

But when we jointly plot some continuous variables that are not on the exact same scale (e.g. some genes), it's not possible to specify a single vmin/vmax value that fits all variables especially if they have some outliers. This deeply saddens us and forces us to watch a few more episodes of Stranger Things and it certainly doesn't help :(

I would like to hear your thoughts about how to fix that. But before that, as a responsible person, I did some homework and I spent around 34 minutes to understand how color normalization works in matplotlib (https://matplotlib.org/3.1.1/tutorials/colors/colormapnorms.html) and tried to implement a custom normalization class (https://matplotlib.org/3.1.1/tutorials/colors/colormapnorms.html#custom-normalization-manually-implement-two-linear-ranges). 

My idea is simply to specify vmin/vmax in terms of quantiles of the color vector which can be shared between variables instead of a specific value. One way, I thought, might be to pass a `norm` argument with a custom normalization object to our lovely `plot_scatter`. However, as far as I understand, it's not possible because in the quantile function in the custom normalization class requires the entire color vector for each continuous variable which is not super convenient because it's too much preprocessing to find different quantile values for each variable and pass a vmin/vmax vector to the plotting function. Not user-friendly and still requires modifications in the code :(

Instead, I added two ugly arguments named `vmin_quantile` and `vmax_quantile` to the `plot_scatter` function which allows me to specify a single quantile value for vmin/vmax which is then translated into real values separately for each variable:

![image](https://user-images.githubusercontent.com/1140359/62720493-20731c00-b9d8-11e9-9dc9-f91cf052c4e1.png)

This solved my problem but I was wondering if it makes sense to add this to scanpy. What do you think?

Finally, here is the way I use it:

`plot_scatter(adata, basis='umap', color=['louvain', 'NKG7', 'GNLY', 'KIT'], cmap='Reds', vmax_quantile=0.999)`

PS: It needs a few more lines to be accessible from other functions like sc.pl.umap.

LuckyMD:
[...]

PR #794, opened 2019-08-21
Embedding improvements

fidelram:
This PR allows the creation of embedding as follows:

image

sumary PR:
 * Added `sc.pl.embedding` to the API. This makes possible to select any 'basis' for plotting. This name is consistent with other functions that accept 'basis' as parameter: `sc.pl.embedding_density` and `scvelo.velocity_embedding` (#762) 
 * vmax and vmin now could be a list to indicate individual vmin/vmax range when using multiple plots (#775).
 * vmax and vmin can indicate a quantile using the format `qNN` to indicate the NN quantile.  Eg. `vmax=['q90', 'q80']` (#775)
 * vmax and vmin can be a user defined function. (#775)
 * added the ~~`add_contour`~~ `add_outline`  parameter which produces figures like the one on top (I will be happy to know if there is a better name for this type of scatter plot as `contour` may refer to some specific type of plots)

TODO:
- [x] Correct style
- [x] Pass tests
- [x] Add tests for new parameters
-  [x] Find better name for 'contour'


flying-sheep:
Awesome! maybe “outline” would be a good word?

LuckyMD:
This looks very cool! I wonder if there could also be an option to put a contour between cluster that have a fluid boundary (for example what I assume are clusters 6 and 7 above). I like the above plot, but I imagine this might help with colours being difficult to discern in certain colour maps.

fidelram:
@flying-sheep 'outline' could be good. We already have font outline that works similarly. Can you take a look at the new type hints that I added. I am not sure if I did it right?

@LuckyMD you mean to do something like this (top image)?

image

LuckyMD:
Exactly!

fidelram:
@LuckyMD I will think about it. The implementation does not seem trivial at the moment and will require yet a new parameter which I am hesitant to add, unless `add_contour` can accept something line `add_contour='separate categories'`. But at least, using overlapping images and without further modifications the issue is solved.

flying-sheep:
> 'outline' could be good. We already have font outline that works similarly.

Great!

> Can you take a look at the new type hints that I added. I am not sure if I did it right?

The typing is pretty good! There’s one rule I follow, which is to be specific:

- If your function has parameter `a`, does `for elem in a`, and expects `elem`s to be `str`s, you can say `Iterable[str]`. If you use `a[i]`, say `Sequence[str]`. You don’t want to artificially limit the user by saying you need a `List[str]` if a `Tuple[str]` can be passed or even any `Iterator[str]` is sufficient.
- If you say what you *return*, be concrete, e.g. `List[str]`. You know what exact type you return.
- If you accept a callable, specify its signature: `Callable[[ArgType1, ArgType2], RetType]`. There’s nothing more annoying than to dive into the code because the library doesn’t specify what kind of function you can supply.

So you should change

- `callable`→`Callable[[???], ?]`
- `Sequence`→`Sequence[?]`
- `Optional[dict]`→`Optional[Mapping[?, ?]]`

Also stylewise: Once `(` and `)` aren on separate lines, never have anything after `(`, and before `)`:

```py
def _get_vmin_vmax(
    […]
    color_vector: Sequence[float],
):
    '''
    […]
```

```py
logg.error(
    "The parameter […]"
    […]
    "of plots."
)
```

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -262,17 +330,70 @@ def plot_scatter(
...
+    Alternatively, vmin/vmax could be a function that is applied to
+    the list of color values (`color_vector`).  E.g.
+
+    def my_vmax(color_vector): np.percentile(color_vector, p=80)
+
+
+    Parameters
+    ----------
+    index: This index of the plot
+    color_vector: list or values for the plot
```
flying-sheep: You need to say

```rst
color_vector
    list or values for the plot
```

else this breaks the docs. The same applies to `index` and `basis` in the other function.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -44,6 +45,10 @@ def plot_scatter(
     legend_fontweight: str = 'bold',
     legend_loc: str = 'right margin',
     legend_fontoutline: Optional[int] = None,
+    vmax: Union[float, str, callable, Sequence, None] = None,
+    vmin: Union[float, str, callable, Sequence, None] = None,
+    add_contour: Optional[bool] = False,
+    contour_config: Optional[dict] = {'color': ('black', 'white'), 'edge_width': (0.3, 0.05)},
```
flying-sheep: Why not `contour_edgewidth: Tuple[float, float]` and `contour_color: Tuple[str, str]`? This way people can’t misspell their dict keys and have clearer typings.

LuckyMD:
@flying-sheep I quite like your style guidelines. It might be a good idea to designate a particular function that does it well and is complex enough to include all of these options. That way it's easy to follow style when writing a new function. Something like a contributors template.

flying-sheep:
Good idea. Done in 740c4a510ec598ab03ff3de1d9b1c091f0aac292

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -262,17 +330,70 @@ def plot_scatter(
...
+
+    Returns
+    -------
+
+    [vmin, vmax] containing None or float values
+
+    '''
+    out = []
+    for v_name in ['vmin', 'vmax']:
+        if len(eval(v_name)) == 1:
```
flying-sheep: Please remove `eval` and do `for v_name, v in [('vmin', vmin), ('vmax', vmax)]`

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -262,17 +330,70 @@ def plot_scatter(
...
+            v_value = eval(v_name)[0]
+        else:
+            try:
+                v_value = eval(v_name)[index]
+            except IndexError:
+                logg.error(f"The parameter {v_name} is not valid. If setting multiple vmin values,"
+                           "check that the length of the {v_name} list is equal to the number "
```
flying-sheep: forgot an “f” here, and wrote “vmin” instead of using “`{v_name}`” one line above

ivirshup:
This looks great!

A few ideas:

* For having an outline to separate overlapping clusters, I don't think I like that one of the outlines would be plotted over the other cluster. In the plots shown above (https://github.com/theislab/scanpy/pull/794#issuecomment-523515331) I think the upper image is less clear about the extent of the overlap than the lower one, and suggests a greater importance of group `3`. Maybe there could be some indication of ambiguity for the region of overlap?
* For the string based quantile selection, is there another package which allows writing operations like this? My concern is that string based DSLs can get messy. It would be nice to make sure we're choosing a unambiguous spec which we can extend in the future and use in other functions. An example of a spec would be SQL reduction operations (like `PERCENTILE_DISC`), but hopefully there would be something less verbose.
* For the basis argument, could we not require the key in `obsm` start with `X_`? I'm thinking the key would just go through a check like:

```python
if basis in adata.obsm:
    basis_key = basis
elif f"X_{basis}" in adata.obsm:
    basis_key = f"X_{basis}"
else:
    raise KeyError(
        f"Could not find entry in `obsm` for '{basis}'.\n"
        f"Available keys are: {list(adata.obsm.keys())}."
    )
```

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -205,8 +237,44 @@ def plot_scatter(
...
+                # the default contour is a black edge followd by a
+                # thin white edged added around connected clusters.
+                # To add a contour
+                # three overlapping scatter plots are drawn:
+                # First black dots with slightly larger size,
+                # then, white dots a bit smaller, but still larger
+                # than the final dots. Then the final dots are drawn
+                # with some transparency.
+
+                black_width, white_width = contour_config['edge_width']
```
flying-sheep: Please rename `black_*` to `outline_*` and `white_*` to `bg_*` or `gap_*` or something similar. For bright-on-dark plots the non-semantic names you chose don’t make sense :wink:

fidelram:
@flying-sheep regarding the typing I ended up with the following line:

```PYTHON
 vmax: Union[float, str, Callable[[Sequence[float]], float], Sequence[Union[str, float, Callable[[Sequence[float]], float]]], None] = None,
```
Does it looks ok to you?

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -450,10 +581,22 @@ def _get_data_points(adata, basis, projection, components) -> Tuple[List[np.ndar
...
+        raise KeyError(
+            f"Could not find entry in `obsm` for '{basis}'.\n"
+            f"Available keys are: {list(adata.obsm.keys())}."
+        )
+
     n_dims = 2
     if projection == '3d':
         # check if the data has a third dimension
-        if adata.obsm['X_' + basis].shape[1] == 2:
+        if adata.obsm[basis_key].shape[1] == 2:
```
gokceneraslan, line 607: Can we please make sure that basis='umap' work? basis='umap' is a lot more user friendly than basis='X_umap' 😊

gokceneraslan: Oh ok, sorry I didn't see the part that looks up the key in obsm. Please ignore the previous comment.

gokceneraslan:
This looks very cool Fidel. It's not a big deal but if we use numbers in 0-100 range for vmax-vmin, it's a percentile, so p80 makes more sense.

flying-sheep:
>  I ended up with […] Does it looks ok to you?

Yeah! You could simplify though:

```py
VMinMax = Union[str, float, Callable[[Sequence[float]], float]]

def embedding(
    ...
    vmin: Union[VMinMax, Sequence[VMinMax], None] = None,
    vmax: Union[VMinMax, Sequence[VMinMax], None] = None,
    ...
): ...

def _get_vmin_vmax(
    vmin: Sequence[VMinMax],
    vmax: Sequence[VMinMax],
    ...
) -> ...: ...
```

> It's not a big deal but if we use numbers in 0-100 range for vmax-vmin, it's a percentile, so p80 makes more sense.

you can also be more or less precise: `q001`→`0.001`, `q9`→`0.9`

fidelram:
@ivirshup I looked around and could not find any example in other packages that use input strings as in this case. I agree that the current solution is not ideal, but other solutions seem more complicated. For example, we can add a new parameter called `percentile` that then interprets `vmin` and `vmax` as percentiles/quantiles.  However, this limits the option to mix different types of values. 

In seaborn they use the parameter `robust` to set vmin and vmax as the .02 and .98 quantiles but the quantiles can not be changed. 

Unless we have a strong opinion against this solution I suggest we keep it.

flying-sheep:
Looks great! Optimally we’d add a test image like the bottom one in https://github.com/theislab/scanpy/pull/794#issuecomment-523515331, but I’d be up for merging this as-is.

LuckyMD:
> For having an outline to separate overlapping clusters, I don't think I like that one of the outlines would be plotted over the other cluster. In the plots shown above ([#794 (comment)](https://github.com/theislab/scanpy/pull/794#issuecomment-523515331)) I think the upper image is less clear about the extent of the overlap than the lower one, and suggests a greater importance of group `3`. Maybe there could be some indication of ambiguity for the region of overlap?

@ivirshup I see your point regarding the accurate representation of what is shown. My thoughts were more along the lines of making it easy to distinguish clusters. That would be particularly useful when you have >20 clusters that are hard to distinguish by colours alone. And the overlaps may not be as important for a large-scale overview.

flying-sheep:
Great stuff, thank you!

gokceneraslan:
I cannot get the new percentile feature to work, am I doing something wrong?

```python
%pdb
import scanpy as sc

adata = sc.datasets.pbmc3k()
sc.pp.pca(adata, svd_solver='arpack')
sc.pp.neighbors(adata)
sc.tl.umap(adata)

sc.pl.embedding(adata, basis='umap', color='CST3', vmax='p90')
```
image

Even if the format is wrong it should be ignored, no?

gokceneraslan:
`sc.pl.embedding(adata, basis='umap', color='CST3', vmax=['p90'])` works.

gokceneraslan:
Type checking would have caught this quite easily because `vmax` is not a sequence in `sc.pl.embedding(adata, basis='umap', color='CST3', vmax='p90')` call.

@flying-sheep why don't we use mypy et al. for checking type hints?

flying-sheep:
Hmm, there’s this code in `embedding` which probably doesn‘t work as `List` here is `typing.List` instead of `collections.abc.List`.

```r
if not isinstance(vmax, List):
    vmax = [vmax]
```

Don’t know how I missed this. The correct check would of course be:

```r
if isinstance(vmax, str) or not isinstance(vmax, collections.abc.Sequence):
    vmax = [vmax]
```

fidelram:
I am surprised that this is coming out. I thought I had solved the issue as I don't get the error. Thanks @flying-sheep for addressing this.

flying-sheep:
Hmm, the code in this PR isn’t the current code anymore. I’m fixing it in #839

PR #795, opened 2019-08-22
Remove broken RNA velocity code and reference scvelo

### LINKED ISSUE #792: Fix and export or delete rna_velocity
We have unused and untested RNA velocity code in #130.

Our plotting functions accept the `arrows` argument which expects a `Delta_` from velocyto.

I guess everything here is superseded by scvelo, so should we just get rid of the `rna_velocity` code and make `arrows` work with scvelo?

flying-sheep:
Fixes #792


scanpy/plotting/_docs.py
```diff
@@ -31,7 +31,8 @@
 edges_color : matplotlib color(s), optional (default: `'grey'`)
     Color of edges. See :func:`~networkx.drawing.nx_pylab.draw_networkx_edges`.
 arrows
-    Show arrows (requires to run :func:`~scanpy.tl.rna_velocity` before).
+    Show arrows (requires to run :func:`scvelo.tl.velocity_embedding` before).
```
flying-sheep, line 34: @VolkerBergen is `scvelo.tl.velocity_embedding` correct here?

VolkerBergen: Perfectly done. Keeping the plotting functionality while pointing towards the new outsourced modules. 

`scvelo.tl.velocity_embedding` is correct while requiring some preceding steps (such as `tl.velocity` and `tl.velocity_graph`). However, since it is raising a message of required preceding steps, `tl.velocity_embedding` might be the right one to put in here. Alternatively, simply scvelo (instead of specifying velocity_embedding) would probably suffice as well.

flying-sheep: OK, great!

Issue #351, opened 2018-11-08
Interest in HTO/ADT demultiplexing and analysis?


wflynny:
I'm starting to regularly analyze samples multiplexed with HTOs a la [Stoeckius et al](https://www.biorxiv.org/content/early/2017/12/21/237693) and multimodal CITE-seq ADTs.  I have some rough ports of Seurat's `HTODemux` and some other functionality.  

Is anyone else analyzing these kinds of experiments using scanpy and is there interest to bring functionality to do so into the main scanpy branch?

falexwolf:
This is very interesting! It would be awesome if you linked to a small example to learn what you do exactly! I guess, a PR would then be more than welcome! 🙂

wflynny:
@falexwolf Mostly just following along with the Seurat vignettes here using scanpy:

* [Demultiplexing with cell hashing](https://satijalab.org/seurat/hashing_vignette.html)
* [Multimodal analysis with CITE-seq ADTs](https://satijalab.org/seurat/multimodal_vignette.html)

Mostly I'm using [CITE-seq-count](https://github.com/Hoohm/CITE-seq-Count) to generate counts matrices then storing the ADT counts in `adata.obsm["X_adt"]` and HTO counts in `adata.obsm["X_hto"]`.  From there, I generate additional metadata about the cells which I store in either `adata.obs` (e.g, for HTOs, `adata.obs.global_classification` stores `["singlet", "doublet", "negative"]` and `adata.obs.tag_class` stores `["hashtag_1", "hashtag_2", ...]`).

I'd be happy to contribute a PR in the next couple weeks.  A quick question though:  the workflow involves some IO, normalization, classification, and plotting, so where would be the best entry point for this?  I was thinking about stashing all related functionality in a `multimodal.py` module, but then should they be part of the `tools` api or a separate api like `sc.mm`?  Or should I spread out the functionality across the existing `pp`, `tl`, `pl` apis?

Leprechaun777:
Hi, I have lots of interest in adding ADT and HTO functionality as well.

Thank you!

njbernstein:
I've got some code to do this. I'll try to submit a PR soon

gokceneraslan:
I'm also interested in this since I'll be analyzing some HTO data soon. 

[...]

PR #797, opened 2019-08-22
Cell hashing

njbernstein:
Here is a stab at adding cell hashing demultiplexing to scanpy. This will address https://github.com/theislab/scanpy/issues/351.

### DIFF
--- a/scanpy/preprocessing/_cell_hashing.py
+++ b/scanpy/preprocessing/_cell_hashing.py
@@ -1,0 +1,110 @@
+from typing import Iterable
+
+from anndata import AnnData
+
+from scipy.signal import argrelextrema
+from scipy.stats import gaussian_kde
+import numpy as np
+import itertools
+import pandas as pd
+
+
+def _demultiplex_per_barcode(x):
+    """
+    Approximates method in https://www.nature.com/articles/s41592-019-0433-8
+    MULTI-seq: sample multiplexing for single-cell RNA
+    sequencing using lipid-tagged indices
+    """
+    percentiles = np.percentile(x, q=np.arange(.01, 99.9, step=.1))
+    kde = gaussian_kde(x)
+    dist = kde(percentiles)
+    maxima = argrelextrema(dist, np.greater_equal)[0]
+    min_maxima = np.min(maxima)
+    max_maxima = np.max(maxima)
+    dist[min_maxima], dist[max_maxima]
+    intermediate_thresh = (percentiles[min_maxima] + percentiles[max_maxima]) / 2
+    all_low_maxima = maxima[np.where(percentiles[maxima] < intermediate_thresh)[0]]
+    low_maxima = all_low_maxima[np.argmax(dist[all_low_maxima])]
+    info = []
+    for q in range(1, 100):
+        threshold = np.percentile([percentiles[low_maxima],
+                                   percentiles[max_maxima]], q=q)
+        info.append([x >= threshold, threshold])
+    return info, [percentiles, dist]
+
+
+def demultiplex_hashing(adata: AnnData,
+                        percentile_candidates: Iterable[int] = None,
+                        inplace: bool = True):
+    """
+    Demultiplexing cell hashing data
+
+    Params
+    ------
+    adata
+        The annotated data matrix of shape ``n_obs`` × ``n_vars``. Rows correspond
+        to cells and columns to genes.
+    percentile_candidates
+        Choose exact percentiles to evaluate thresholding at
+
+    inplace
+        Whether to update ``adata`` or return dictionary with normalized copies of
+        ``adata.X`` and ``adata.layers``.
+
+    Returns
+    -------
+    Returns `adata.obs` with columns "ID" and "CLASSIFICATION" which identify
+    which cell hashing barcodes a cell has
+    """
+    barcodes_hits = []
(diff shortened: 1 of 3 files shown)


njbernstein:
@flying-sheep Let me know next steps

njbernstein:
The  test is failing because `scipy.stats` is not a wanted import. Let me know how you'd like to deal with that

scanpy/preprocessing/_cell_hashing.py
```diff
@@ -0,0 +1,109 @@
...
+                        inplace: bool = True):
+    """
+    Demultiplexing cell hashing data
+
+    Params
+    ------
+    adata
+        The annotated data matrix of shape ``n_obs`` × ``n_vars``. Rows correspond
+        to cells and columns to genes.
+    percentile_candidates
```
fidelram: Can you add if the adata object needs any particular formatting or pre-processing steps? Maybe here or in the description. For example for 10x, just processing the sample with cell ranger is enough or other steps are required?

scanpy/preprocessing/_cell_hashing.py
```diff
@@ -0,0 +1,109 @@
...
+        Whether to update ``adata`` or return dictionary with normalized copies of
+        ``adata.X`` and ``adata.layers``.
+
+    Returns
+    -------
+    Returns `adata.obs` with columns "ID" and "CLASSIFICATION" which identify
+    which cell hashing barcodes a cell has
+    """
+    barcodes_hits = []
+    number_of_barcodes = adata.X.shape[1]
```
fidelram, line 77: I am not sure about the logic here. Normally, the columns contain genes, I assume that in cell-hashing there will be barcodes, but could the data be mixed ? Maybe if you clarify what is needed/expected in the description of the function would make the code easy to read.

scanpy/preprocessing/_cell_hashing.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    Demultiplexing cell hashing data
+
+    Params
+    ------
+    adata
+        The annotated data matrix of shape ``n_obs`` × ``n_vars``. Rows correspond
+        to cells and columns to genes.
+    percentile_candidates
+        Choose exact percentiles to evaluate thresholding at
```
fidelram: This percentiles refer to what exactly? Can you elaborate a bit?

njbernstein:
@fidelram  lemme know if the new comments make things more clear

fidelram:
@ivirshup @flying-sheep I would like to merge this code but I have some questions:

- Do you know why the test is failing?
- Currently, this is located in `preprocessing`, but I think that the right place is `external.pp` would you agree on that?

ivirshup:
> Do you know why the test is failing?

@flying-sheep and I have put some effort into reducing import times for scanpy (they were getting up to a few seconds). This can most effectively be done by deferring the import of slow-to-import packages until they're actually needed. Phil added a test which blacklists top level import of some of particularly egregious packages (scipy and seaboarn for example).

Here, the imports for `argrelextrema` and `gaussian_kde` are pretty slow (about half a second). These should be moved inside `_demultiplex_per_barcode` function instead of being top level.

> Currently, this is located in preprocessing, but I think that the right place is external.pp would you agree on that?

I think I would be fine with this going in preprocessing, since it isn't tied to another tool/ code base. I'm also not too familiar with the current state of demultiplexing techniques so I'd be fine to defer to you on this. Could you elaborate on why you think it should go in external?

fidelram:
@ivirshup : Thanks for the background explanation. 

@njbernstein Can you move the import statements inside the `_demultiplex_per_barcode` to remove the test errors? 

I think the tool should go to `external` to point out that this is based on a method that we have not tested and thus the responsibility of its accuracy and implementation lies on the external contributor.

flying-sheep:
Hmm, I’m pretty happy with my self-documented test code:

```py
    def test_deferred_imports(imported_modules):
        slow_to_import = {
            'umap',             # neighbors, tl.umap
            'seaborn',          # plotting
            'sklearn.metrics',  # neighbors
            'scipy.stats',      # tools._embedding_density
            'networkx',         # diffmap, paga, plotting._utils
            # TODO: 'matplotlib.pyplot',
            # TODO (maybe): 'numba',
        }
        falsely_imported = slow_to_import & imported_modules
>       assert not falsely_imported
E       AssertionError: assert not {'scipy.stats'}
```

Do you think this could be clearer?

fidelram:
@flying-sheep I think the output is clear once you know what is about. Since this error may happen to future contributions that are not aware of the efforts to reduce import times, I think is better to be explicit. Something like: "Slow import detected (scipy.stats). Please check that slow-to-import packages are not in top level calls but inside the functions that require them".

scanpy/preprocessing/_cell_hashing.py
gokceneraslan: I think these are the columns here represent hashing barcodes, not genes. @njbernstein I think both RNA UMIs and hashing barcode counts should be stored in the same AnnData object. Hashing barcodes can be stored in `adata.obs` or `adata.obsm` but definitely not in `adata.X`. 

I think it makes more sense to first discuss 1) how to store hashing barcodes and 2) how to demultiplex them either in the PR or in the #351 issue.

PR #798, opened 2019-08-23
Migrate to setuptools_scm and get rid of packaging cruft.

flying-sheep:
No vendored versioneer.py anymore, no setup.cfg! This is vastly simpler.

Could you please check if my logic holds?

Package metadata and content will be correctly derived from the git repo’s status.

1. before building, `setuptools_scm` will be installed through `setup_requires=['setuptools_scm']` or `[build-requires]`.
2. during building, due to `use_scm_version=True`, it will populate the package metadata with a proper version from git. (Looks like `scanpy-1.4.5.dev60+g0adf706`)

`scanpy.__version__` will be the version of scanpy you’re currently developing (if applicable) or the installed one:

1. scanpy tries to use `setuptools_scm.get_version`
2. if `setuptools_scm` can’t be imported or we’re not in a git repo, it will just use the version from package metadata (like `check_versions` does for other installed packages)


setup.py
```diff
@@ -1,10 +1,9 @@
 import sys
-if sys.version_info < (3,):
+if sys.version_info < (3, 6):
     sys.exit('scanpy requires Python >= 3.6')
```
flying-sheep, line 3: Unrelated change, but seems to make sense?

flying-sheep:
OK, I checked every part, let’s go!

PR #805, opened 2019-08-29
Show PCA loadings for genes with lowest loadings

gokceneraslan:
Currently, we do not show the genes with the lowest loadings in `sc.pl.pca_loadings`. This PRs add an option for that:

image


gokceneraslan:
image

Maybe we should consider making this True by default.

flying-sheep:
Is it that useful to see it by default? Why would you want to know unimportant genes?

IMHO it would be more interesting to know genes that have high loadings in another early PC but not in this one, right?

LuckyMD:
Hi negative loadings are also important genes as you can see on the example of `LTB` and `HLA-DRA`

flying-sheep:
Ah, sure, they just factor in in the inverse direction. Thank you, it’s been a while.

Sure, then it should be True by default.

gokceneraslan:
Made it True by default and merged. Thanks.

scanpy/plotting/_tools/__init__.py
```diff
@@ -82,12 +85,16 @@ def pca_loadings(
...
     components = np.array(components) - 1
     if np.any(components < 0):
         logg.error("Component indices must be greater than zero.")
         return
-    ranking(adata, 'varm', 'PCs', indices=components)
+    ranking(adata,
+            'varm',
+            'PCs',
+            indices=components,
+            include_lowest=include_lowest)
```
flying-sheep, line 97: Please use [black line wrapping](https://black.readthedocs.io/en/stable/the_black_code_style.html#how-black-wraps-lines) here too

scanpy/plotting/_anndata.py
```diff
@@ -576,15 +577,29 @@ def ranking(
...
                     rotation='vertical', verticalalignment='bottom',
                     horizontalalignment='center', fontsize=8)
+        if include_lowest:
+            pl.text(len(indices), (score[g]+score[neg_indices[0]])/2, '⋮', color=color,
+                    rotation='vertical', verticalalignment='bottom',
+                    horizontalalignment='center', fontsize=8)
+            for ig, g in enumerate(neg_indices):
+                pl.text(ig+len(indices)+2, score[g], labels[g], color=color,
+                        rotation='vertical', verticalalignment='bottom',
+                        horizontalalignment='center', fontsize=8)
```
flying-sheep, line 597: Please use [black line wrapping](https://black.readthedocs.io/en/stable/the_black_code_style.html#how-black-wraps-lines) here and above

scanpy/plotting/_anndata.py
```diff
@@ -576,15 +577,29 @@ def ranking(
                            top=1-(n_rows-1)*bottom-0.1/n_rows)
     for iscore, score in enumerate(scores.T):
         pl.subplot(gs[iscore])
-        indices = np.argsort(score)[::-1][:n_points+1]
+        if not include_lowest:
+            indices = np.argsort(score)[::-1][:n_points+1]
+        else:
+            sorted_scores = np.argsort(score)[::-1]
```
flying-sheep, line 583: Please call this `score_order` or so, move it before the conditional, and use it in both branches.

flying-sheep (commented):
Nice! Just a few stylistic points left.

gokceneraslan:
I didn't see these comments @flying-sheep , sorry about that and thanks for taking care.

I discovered a bug today, "⋮" character is not visible on some machines:

image

Any ideas?

flying-sheep:
I think it’s pretty impossible to know if they’ll render – Depends on the fonts available on the system and the way the font rendering stack falls back to other fonts.

My approach would be to check which systems have the problem, and if it’s only some Linux server or some obsolete stuff like e.g. Windows versions up to Vista, ignore it. If it’s a commonly used and still supported desktop OS / Linux distribution, we have to deal with it.

The reason I excluded Linux servers is that server admins often set up things minimalistically, excluding “GUI stuff” so trying to support those highly heterogenous systems will only bring pain. When people want better fonts, then fontconfig is happy to provide them with the means to do so.

gokceneraslan:
@flying-sheep I hit this vertical ellipsis bug again. I think the problem is that, at least on this machine, setting default sans-serif font to Arial messes up the rendering of "⋮" character. Rendering is fine with DejaVu or Helvetica though. Any ideas why Arial is problematic on Mac OS? Why do we change Matplotlib "font.sans-serif" anyway?

See below:

image

flying-sheep:
> Why do we change Matplotlib "font.sans-serif" anyway?

No idea, I don’t even like Arial. Alex updated the fonts last in 6c68b8ba2821f27bd0b8f499a1d543dff9cc51b2, and setting the fonts happened in the initial commit:

https://github.com/theislab/scanpy/blob/c22e48abe45a6ccca5918bbf689637caa4b31250/scanpy/plotting.py#L605

@falexwolf do you recall why you did that? Can we just remove that line?

gokceneraslan:
There are plots that are very sensitive to font type/size, I think, like rank plots (e.g. `sc.pl.pca_loadings`). Alex did it in order to standardize the style of plots and to prevent unexpected things, I guess. 

If this is only a ⋮ problem, we can switch to "...", it's not a big deal. Interestingly, Arial on my laptop can render other unicode characters well, ⋮ is problematic though :)

flying-sheep:
Nah, why are they sensitive?

gokceneraslan:
Because they get unreadable if fonts overlap.

flying-sheep:
Continued in #897

PR #806, opened 2019-09-01
Minor PAGA plotting improvements related to legend fonts

gokceneraslan:
This PR does a few things:

- Adds `fontoutline` argument to sc.pl.paga as a convenient shortcut. Similar to https://github.com/theislab/scanpy/pull/640. This can also be achieved via text_kwds but it's super verbose and less intuitive for beginners.
- Removes the usage of empty dictionaries as default arguments, because it might lead to [erroneous results in successive calls](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). 
- `sc.pl.paga_compare` now adds `legend_font{size,weight,outline}` arguments to paga_graph_params by default which results in better agreement between the embedding view and paga view.


flying-sheep:
Great! Will this change the test pics? If not LGTM.

gokceneraslan:
There is no plotting test for PAGA, I think. It'd be great to have some :)

gokceneraslan:
Added PAGA plotting tests and merged.

PR #812, opened 2019-09-04
The method for annotating genes with cell types

PrimozGodec:
With this PR we propose our new annotation method to Scannpy:

Annotator marks the data with cell type annotations based on marker genes.
Over-expressed genes are selected with the  Mann-Whitney U tests and cell
types are assigned with the hypergeometric test. This function first selects
genes from gene expression data with the Mann Whitney U test, then annotate
them with the hypergeometric test, and finally filter out cell types that
have zero scores for all cells. The results are scores that tell how
probable each cell type is for each cell.

Hope you like the method and merge it to Scampy.


ivirshup:
Thanks for the PR! Gene-set based annotation would be pretty useful to have here.

Is there any chance there's a preprint we could look at for a little more context on the method?

PrimozGodec:
@ivirshup it is currently in writing. Will write back to you when we will have it ready.

fidelram:
Any update on this? Can you add a test (probably reusing the example already in the method docstring)?

PrimozGodec:
@fidelram, we are still working paper/preprint. I will post it soon. 

I will add tests. So in order for the test to work should I add my library in the requirements.txt? What I observed is that other external packages are not included in project requirements.

fidelram:
Yeah, please add it together with a comment mentioning where is needed (e.g  `external.tl.annotator` )

ivirshup:
@PrimozGodec, probably don't add this to `requirements.txt`, since the requirement should be optional for install. I think instead you should mark it with something like:

```python
from importlib.util import find_spec

@pytest.mark.skipif(find_spec('pointannotator') is None, reason="pointannotator not installed")
```

You can add a requirement for the package to this line in `setup.py`: https://github.com/theislab/scanpy/blob/d8f32c040f3a5f4fc07998b269796ca58de84b40/setup.py#L41

Maybe we should eventually have a second requirements file for CI testing, like we do for anndata.

PrimozGodec:
I added unit tests and reformated the code.

flying-sheep:
Thank you. We’re using pytest though, so please write the tests that way:

1. Remove the class and make all its methods top-level functions
2. Make `setUp` into `fixture`s
3. Just use `assert`

```py
@pytest.fixture
def markers():
    return pd.DataFrame(
        ...
    )


@pytest.fixture
def adata():
    ...
    return AnnData(data.values, var=data.columns.values)


def test_remove_empty_column(adata, markers):
    ...
    annotations = annotator(adata, markers, num_genes=20)
    ...
    assert len(annotations) == len(self.anndata)
    ...
```

flying-sheep:
Only remaining thought: I have slight concerns about the name being too generic, but then again, this does exactly what people expect a “cell type annotator based on marker genes” to do.

scanpy/external/tl/_annotator.py
```diff
@@ -0,0 +1,146 @@
...
+    expressions. For annotation, we use data by Villani et al. (2017)[1] and
+    marker genes by Franzén et al. (2019)[2].
+
+    [1]  Villani, A. C., Satija, ... Jardine, L. (2017). Single-cell
+         RNA-seq reveals new types of human blood dendritic cells, monocytes,
+         and progenitors. Science, 356(6335).
+
+    [2]  Oscar Franzén, Li-Ming Gan, Johan L M Björkegren, PanglaoDB:
+         a web server for exploration of mouse and human single-cell RNA
+         sequencing data, Database, Volume 2019, 2019.
```
flying-sheep: Please put the references into `references.rst` and reference them correctly, e.g. `[Villani17]_`

scanpy/external/tl/_annotator.py
```diff
@@ -0,0 +1,146 @@
...
+            Negative of the logarithm of a p-value
+    normalize : bool, optional (default = False)
+        If this parameter is True data will be normalized during the
+        a process with a log CPM normalization.
+        That method works correctly data needs to be normalized.
+        Set this `normalize` on True if your data are not normalized already.
+
+    Returns
+    -------
+    pd.DataFrame
```
flying-sheep: That’s not right, it returns an `AnnData` object. But you don’t need to mention that here, that’s what a return type annotation is for: `def fun(...) -> ReturnType:`

Also please orient yourself at the way other scanpy functions work: They have an `inplace` argument that controls if `adata`’s `obs`/`var` is updated or a `DataFrame` or similar is returned.

This function currently filters out cells, but it would be better to just mark them as having no marker cell type found.

ivirshup: @flying-sheep, I find this comment is a little confusing. `inplace=False` will often return some other dtype. `copy` determines whether the anndata object is copied, and that copy modified.

flying-sheep: Let’s be perfectly clear: `inplace` and `copy` are both used, sometimes `copy` is being deprecated in favor of `inplace`.

`copy` means “if copy=True, return a copy of the AnnData object, with var/obs modified. if copy=False, return `None` and modify the AnnData object’s var/obs”

`inplace` means “if inplace=True, modify the AnnData object, if inplace=False, return an object with computation results”. There’s two variants for when it’s True: a) the AnnData object’s obs/var get new columns (for almost everything) or b) the AnnData object is filtered (in `pp.filter_cells`).

I updated my comment above

PrimozGodec: I added an in-place argument

scanpy/external/tl/_annotator.py
PrimozGodec: Done

Zethson:
We decided not to add more packages to `external` but you are more than welcome to add your own package to the scverse ecosystem: https://scverse.org/packages/#ecosystem

PR #813, opened 2019-09-04
Make the dca learning rate accessible from scanpy.

gamazeps:
scanpy currently does not allow to select the dca learning rate.

This can create convergence issues (for example dca diverges 60% of the time on Tabula Muris when trained on CPU, but oddly enough converges when trained on GPU)


flying-sheep:
Makes sense, thank you!

PR #816, opened 2019-09-05
fix zheng17

### LINKED ISSUE #511: memory requirements
Hi,
I am trying to run the full 1.3M 10X mouse cell dataset (using the 1M_neurons_filtered_gene_bc_matrices_h5.h5 file from 10X website).
I have 126GB RAM and Intel® Xeon(R) W-2123 CPU @ 3.60GHz × 8 which is above the requirements you mention needed to run the full cluster.py method without subsampling. (https://github.com/theislab/scanpy_usage/tree/master/170522_visualizing_one_million_cells)
I get a memory error at the normalization and filter_genes_dispersion stage, should i modify the code in anyway? (without subsampling)
Thanks,Shobi

### LINKED ISSUE #811: Memory issue running zheng17 on ~1M cells
Running https://github.com/theislab/scanpy_usage/blob/master/170522_visualizing_one_million_cells/cluster.py on the latest released Scanpy (1.4.4.post1) gives a memory error:

```
reading 1M_neurons_filtered_gene_bc_matrices_h5.h5
Variable names are not unique. To make them unique, call `.var_names_make_unique`.
 (0:01:39)
running recipe zheng17
filtered out 3983 genes that are detectedin less than 1 counts
Killed
```

This is running with 60GB of memory (n1-standard-16), but also occurs with 104GB (n1-highmem-16). It looks like there has been a regression somewhere since this used to run OK. I think the error may be happening in anndata.

Koncopd:
This fixes #811 and should fix #511

The analysis of the memory requirments is in the second issue.

### DIFF
--- a/scanpy/preprocessing/_recipes.py
+++ b/scanpy/preprocessing/_recipes.py
@@ -60,7 +60,7 @@ def recipe_seurat(adata, log=True, plot=False, copy=False):
     if copy: adata = adata.copy()
     pp.filter_cells(adata, min_genes=200)
     pp.filter_genes(adata, min_cells=3)
-    pp.normalize_per_cell(adata, counts_per_cell_after=1e4)
+    normalize_total(adata, target_sum=1e4)
     filter_result = filter_genes_dispersion(
         adata.X, min_mean=0.0125, max_mean=3, min_disp=0.5, log=not log)
     if plot:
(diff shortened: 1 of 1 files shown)


flying-sheep:
So what’s the difference between normalize_per_cell and normalize_total?

If they’re replacable, why don’t we just use the more efficient one?

Koncopd:
@flying-sheep `normalize_total` is newer and more efficient. And we should use it instead of `normalize_per_cell`, yes.

flying-sheep:
The tests are failing because there’s some master fixes missing from your branch, so this is fine!

PR #819, opened 2019-09-07
Make partition modularity available in adata.uns

gokceneraslan:
This PR stores modularity metric into adata.uns. I also added `key_added` as a suffix to uns key "louvain", so that we can run louvain with different parameters and store their quality metrics separately.

```python
import scanpy as sc

sc.settings.verbosity = 3
adata = sc.datasets.pbmc3k()
sc.pp.pca(adata, svd_solver='arpack')
sc.pp.neighbors(adata)
sc.tl.louvain(adata, resolution=2.0, key_added='newkey')
adata.uns
```

image

I'm really bad at writing warning messages and naming things, so feel free to edit these :)

### DIFF
--- a/scanpy/tests/test_clustering.py
+++ b/scanpy/tests/test_clustering.py
@@ -56,9 +56,11 @@ def test_louvain_basic(adata_neighbors):
 
 def test_partition_type(adata_neighbors):
     import louvain
+    sc.tl.louvain(adata_neighbors, partition_type=louvain.ModularityVertexPartition)
     sc.tl.louvain(adata_neighbors, partition_type=louvain.RBERVertexPartition)
     sc.tl.louvain(adata_neighbors, partition_type=louvain.SurpriseVertexPartition)
 
     import leidenalg
+    sc.tl.leiden(adata_neighbors, partition_type=leidenalg.ModularityVertexPartition, resolution=None)
     sc.tl.leiden(adata_neighbors, partition_type=leidenalg.RBERVertexPartition)
     sc.tl.leiden(adata_neighbors, partition_type=leidenalg.SurpriseVertexPartition, resolution=None)
--- a/scanpy/tests/test_clustering.py
+++ b/scanpy/tests/test_clustering.py
@@ -8,7 +8,10 @@ def adata_neighbors():
 
 
 def test_leiden_basic(adata_neighbors):
-    sc.tl.leiden(adata_neighbors)
+    sc.tl.leiden(adata_neighbors, use_weights=True, directed=True)
+    sc.tl.leiden(adata_neighbors, use_weights=True, directed=False)
+    sc.tl.leiden(adata_neighbors, use_weights=False, directed=True)
+    sc.tl.leiden(adata_neighbors, use_weights=False, directed=False)
 
 @pytest.mark.parametrize('clustering,key', [
     (sc.tl.louvain, 'louvain'),
@@ -43,7 +46,10 @@ def test_clustering_subset(adata_neighbors, clustering, key):
 
 def test_louvain_basic(adata_neighbors):
     sc.tl.louvain(adata_neighbors)
-    sc.tl.louvain(adata_neighbors, use_weights=True)
+    sc.tl.louvain(adata_neighbors, use_weights=True, directed=True)
+    sc.tl.louvain(adata_neighbors, use_weights=True, directed=False)
+    sc.tl.louvain(adata_neighbors, use_weights=False, directed=True)
+    sc.tl.louvain(adata_neighbors, use_weights=False, directed=False)
     sc.tl.louvain(adata_neighbors, use_weights=True, flavor="igraph")
     sc.tl.louvain(adata_neighbors, flavor="igraph")
 
@@ -52,3 +58,7 @@ def test_partition_type(adata_neighbors):
     import louvain
     sc.tl.louvain(adata_neighbors, partition_type=louvain.RBERVertexPartition)
     sc.tl.louvain(adata_neighbors, partition_type=louvain.SurpriseVertexPartition)
+
+    import leidenalg
+    sc.tl.leiden(adata_neighbors, partition_type=leidenalg.RBERVertexPartition)
+    sc.tl.leiden(adata_neighbors, partition_type=leidenalg.SurpriseVertexPartition, resolution=None)
--- a/scanpy/tools/_leiden.py
+++ b/scanpy/tools/_leiden.py
@@ -170,8 +170,8 @@ def leiden(
             q /= m if directed else m*2
         if isinstance(part, (
             leidenalg.RBConfigurationVertexPartition,
-            leidenalg.ModularityVertexPartition),
-        ):
+            leidenalg.ModularityVertexPartition,
+        )):
             qual_type = ' (scaled modularity)'
         else:
             qual_type = ''
--- a/scanpy/tools/_leiden.py
+++ b/scanpy/tools/_leiden.py
@@ -161,10 +161,13 @@ def leiden(
         n_iterations=n_iterations,
     )
     if 'quality' in dir(part):
-        adata.uns[uns_key]['quality'] = part.quality()
+        q = part.quality()
+        if q > 1.0: # scale by 2*num_edges
+            q /= g.ecount()
+        adata.uns[uns_key]['quality'] = q
         quality_msg = (
             f'\n'
-            f'    quality of the partitioning is {adata.uns[uns_key]["quality"]:.2f}\n'
+            f'    quality of the partitioning (scaled modularity) is {q:.3f}\n'
             f'    added "quality" key to adata.uns["{uns_key}"]'
         )
     else:
--- a/scanpy/tools/_leiden.py
+++ b/scanpy/tools/_leiden.py
@@ -162,8 +162,9 @@ def leiden(
     )
     if 'quality' in dir(part):
         q = part.quality()
-        if q > 1.0: # scale by 2*num_edges
-            q /= g.ecount()
+        if isinstance(part, leidenalg.RBConfigurationVertexPartition) and q > 1.0: #scale quality
+            m = partition_kwargs['weights'].sum() if use_weights else g.ecount()
+            q /= m if directed else m*2
         adata.uns[uns_key]['quality'] = q
         quality_msg = (
             f'\n'
--- a/scanpy/tools/_leiden.py
+++ b/scanpy/tools/_leiden.py
@@ -153,7 +153,7 @@ def leiden(
         categories=natsorted(np.unique(groups).astype('U')),
     )
     # store information on the clustering parameters
-    uns_key = 'leiden' if key_added == 'leiden' else f'leiden_{key_added}'
+    uns_key = key_added if key_added.startswith('leiden') else f'leiden_{key_added}'
     adata.uns[uns_key] = {}
     adata.uns[uns_key]['params'] = dict(
         resolution=resolution,
@@ -165,10 +165,15 @@ def leiden(
         if isinstance(part, leidenalg.RBConfigurationVertexPartition) and q > 1.0: #scale quality
             m = partition_kwargs['weights'].sum() if use_weights else g.ecount()
             q /= m if directed else m*2
+        if isinstance(part, (leidenalg.RBConfigurationVertexPartition,
+                             leidenalg.ModularityVertexPartition)):
+            qual_type = ' (scaled modularity)'
+        else:
+            qual_type = ''
         adata.uns[uns_key]['quality'] = q
         quality_msg = (
             f'\n'
-            f'    quality of the partitioning (scaled modularity) is {q:.3f}\n'
+            f'    quality of the partitioning{qual_type} is {q:.3f}\n'
             f'    added "quality" key to adata.uns["{uns_key}"]'
         )
     else:
--- a/scanpy/tools/_leiden.py
+++ b/scanpy/tools/_leiden.py
@@ -153,9 +153,8 @@ def leiden(
         categories=natsorted(np.unique(groups).astype('U')),
     )
     # store information on the clustering parameters
-    uns_key = key_added if key_added.startswith('leiden') else f'leiden_{key_added}'
-    adata.uns[uns_key] = {}
-    adata.uns[uns_key]['params'] = dict(
+    adata.uns[key_added] = {}
+    adata.uns[key_added]['params'] = dict(
         resolution=resolution,
         random_state=random_state,
         n_iterations=n_iterations,
@@ -169,7 +168,7 @@ def leiden(
         initial_membership=part.membership,
     )
     q = modularity_part.quality()
-    adata.uns[uns_key]['modularity'] = q
+    adata.uns[key_added]['modularity'] = q
     logg.info(
         '    finished',
         time=start,
@@ -177,7 +176,7 @@ def leiden(
             f'found {len(np.unique(groups))} clusters and added\n'
             f'    {key_added!r}, the cluster labels (adata.obs, categorical)\n'
             f'    modularity: {q:.3f}, resolution: {resolution}\n'
-            f'    added "modularity" key to adata.uns["{uns_key}"]'
+            f'    added "modularity" key to adata.uns["{key_added}"]'
         ),
     )
     return adata if copy else None
--- a/scanpy/tools/_leiden.py
+++ b/scanpy/tools/_leiden.py
@@ -163,6 +163,7 @@ def leiden(
     if 'quality' in dir(part):
         adata.uns[uns_key]['quality'] = part.quality()
         quality_msg = (
+            f'\n'
             f'    quality of the partitioning is {adata.uns[uns_key]["quality"]:.2f}\n'
             f'    added "quality" key to adata.uns["{uns_key}"]'
         )
@@ -173,7 +174,7 @@ def leiden(
         time=start,
         deep=(
             f'found {len(np.unique(groups))} clusters and added\n'
-            f'    {key_added!r}, the cluster labels (adata.obs, categorical)\n'
+            f'    {key_added!r}, the cluster labels (adata.obs, categorical)'
             f'{quality_msg}'
         ),
     )
--- a/scanpy/tools/_leiden.py
+++ b/scanpy/tools/_leiden.py
@@ -173,7 +173,7 @@ def leiden(
         time=start,
         deep=(
             f'found {len(np.unique(groups))} clusters and added\n'
-            f'    {key_added!r}, the cluster labels (adata.obs, categorical)'
+            f'    {key_added!r}, the cluster labels (adata.obs, categorical)\n'
             f'{quality_msg}'
         ),
     )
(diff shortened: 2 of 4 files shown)


ivirshup:
I think it's really good to record the key added. I have a couple questions about the quality score.

* Should it still be called a modularity score if we aren't using the modularity quality function?
* Do you have some use cases for recording the modularity score? My impression was that it may not have much interpretable meaning, especially between different graphs.

Also, should this stuff be mirrored to `leiden`?

LuckyMD:
I think a use case for recording modularity would be to assess the quality of a partition on the same network. Louvain or Leiden are both heuristics that are not guaranteed to get the global optimal partition. So if you run with a different random seed, you will get a different result. By outputting the modularity you can compare the quality of the partition from different random seeds.

scanpy/tools/_louvain.py
```diff
@@ -106,6 +106,9 @@ def louvain(
             restrict_categories,
             adjacency,
         )
+    uns_key = 'louvain' if key_added == 'louvain' else f'louvain_{key_added}'
```
LuckyMD: You could also catch the case if someone puts `key_added='louvain_newkey'`.
`uns_key = key_added if key_added.startswith('louvain') else f'louvain_{key_added}'`

gokceneraslan: Sounds good! I didn't want to overengineer it but people use louvain/leiden prefix a lot, so I did it :)

gokceneraslan:
> I think it's really good to record the key added. I have a couple questions about the quality score.
> 
> * Should it still be called a modularity score if we aren't using the modularity quality function?

If another flavor is used, we do not record or print anything about quality, so it's ok. But there is a chance that user supplies a partition type from Louvain/Leiden that is not using modularity optimization (e.g. CPM or Surprize). In this case, we do not print the term `scaled modularity` anymore.

> * Do you have some use cases for recording the modularity score? My impression was that it may not have much interpretable meaning, especially between different graphs.
> 

To me, scaled modularity is like any statistical measure which gives a rough idea about a concept, like correlation or silhouette coef. It's far from conclusive just by itself, but it gives a "feeling" of how "well-clustered" the data is (and how good we are at finding them). Without complementing it with other measures, it's not more than just a "feeling" :)

> Also, should this stuff be mirrored to `leiden`?

It's done.

ivirshup:
> If another flavor is used, we do not record or print anything about quality, so it's ok. But there is a chance that user supplies a partition type from Louvain/Leiden that is not using modularity optimization (e.g. CPM or Surprize). In this case, we do not print the term scaled modularity anymore.

By default we don't use modularity as the quality function, we use the `RBConfigurationVertexPartition`. I believe that we should get a quality score from the leiden and louvain packages regardless of quality function though. Could we use the term `quality_score` and also store `quality_function` used in params like: `.uns[key_added]["params"]["quality_function"] = partition_type.__name__`?

> To me, scaled modularity is like any statistical measure which gives a rough idea about a concept, like correlation or silhouette coef. It's far from conclusive just by itself, but it gives a "feeling" of how "well-clustered" the data is (and how good we are at finding them). Without complementing it with other measures, it's not more than just a "feeling" :)

A couple follow up points on this and @LuckyMD's points

* I don't actually know how different the quality score can be for different solutions. Any chance you have some stats on quality scores from multiple clusterings? I'm mostly wondering if "good" clusterings are associated with high quality scores.
* I think if a user sees a value like "quality" they could ascribe more meaning to it than it deserves. I think we should add some docs about what it means, and how to interpret it.

scanpy/tools/_louvain.py
```diff
@@ -133,7 +137,22 @@ def louvain(
...
                 **partition_kwargs,
             )
-            # adata.uns['louvain_quality'] = part.quality()
+            if 'quality' in dir(part):
+                q = part.quality()
+                if isinstance(part, louvain.RBConfigurationVertexPartition) and q > 1.0: #scale
+                    m = weights.sum() if use_weights else g.ecount()
+                    q /= m if directed else m*2
+                if isinstance(part, (louvain.RBConfigurationVertexPartition,
+                                     louvain.ModularityVertexPartition)):
```
flying-sheep: please use [black line wrapping](https://black.readthedocs.io/en/stable/the_black_code_style.html#how-black-wraps-lines).

scanpy/tools/_louvain.py
```diff
@@ -133,7 +137,22 @@ def louvain(
...
+                    qual_type = ' (scaled modularity)'
+                else:
+                    qual_type = ''
+                adata.uns[uns_key]['quality'] = q
+                quality_msg = (
+                    f'\n'
+                    f'    quality of the partitioning{qual_type} is {q:.3f}\n'
+                    f'    added "quality" key to adata.uns["{uns_key}"]'
+                )
```
flying-sheep: This block of code should be exactly the same as in the _leiden module. Please create a helper function instead of copying and pasting code.

flying-sheep (changes requested):

scanpy/tools/_louvain.py
LuckyMD: I am "people" in this case ;).

LuckyMD:
>A couple follow up points on this and @LuckyMD's points:
>    I don't actually know how different the quality score can be for different solutions. Any chance you have some stats on quality scores from multiple clusterings? I'm mostly wondering if "good" clusterings are associated with high quality scores.
>    I think if a user sees a value like "quality" they could ascribe more meaning to it than it deserves. I think we should add some docs about what it means, and how to interpret it.

I'm not sure I entirely understand your point. The quality score is modularity, which is optimized. Thus a "good" partition is a high quality score by definition. Or what are you referring to as "good"? 

I have looked at some stats for communities in protein-protein interaction networks, and the quality of the communities can change dramatically with louvain output there (could find the link if you think it's relevant). However, modularity as a score is fairly degenerate toward the optimal score and therefore the value often doesn't change that much between the optimized partitions. I'm not sure how this is on the knn graphs though.

gokceneraslan:
> By default we don't use modularity as the quality function, we use the `RBConfigurationVertexPartition`. I believe that we should get a quality score from the leiden and louvain packages regardless of quality function though. Could we use the term `quality_score` and also store `quality_function` used in params like: `.uns[key_added]["params"]["quality_function"] = partition_type.__name__`?
> 

Strictly speaking, you are correct that the quality function of `RBConfigurationVertexPartition` is not exactly the same as modularity, although it is called unscaled modularity in the [code](https://github.com/vtraag/louvain-igraph/blob/master/src/RBConfigurationVertexPartition.cpp#L123). 

There are two main differences between RBConfigurationVertexPartition and ModularityVertexPartition which uses typical modularity optimization. 1) Scaling by the number of edges and 2) the resolution parameter (as it's written [here in the note](https://louvain-igraph.readthedocs.io/en/latest/reference.html#rbconfigurationvertexpartition)). I account for 1) in the code but using a resolution parameter other than 1.0 would lead to values different than modularity due to 2). Right now, for example, you can get a perfect quality (=1.0) by just setting the resolution to 0.0 :D I don't think that'd mislead users though. After all, that's what the algorithm uses for optimization.

I can think of two solutions. We can report typical modularity regardless of the `partition_type`, namely:

```
modularity_part = leidenalg.ModularityVertexPartition(g, initial_membership=part.membership)
q = modularity_part.quality()
```

or we can report the original quality value as "raw quality" (whatever it is) and the modularity together. It's in the "hint" verbosity level anyway.

Regarding the suggestion to record  `partition_type.__name__`, I think it's a good idea. I'd record it in the `uns[uns_key]['partition_type']` though, not in `quality_function`.

> > To me, scaled modularity is like any statistical measure which gives a rough idea about a concept, like correlation or silhouette coef. It's far from conclusive just by itself, but it gives a "feeling" of how "well-clustered" the data is (and how good we are at finding them). Without complementing it with other measures, it's not more than just a "feeling" :)
> 
> A couple follow up points on this and @LuckyMD's points
> 
> * I don't actually know how different the quality score can be for different solutions. Any chance you have some stats on quality scores from multiple clusterings? I'm mostly wondering if "good" clusterings are associated with high quality scores.
> * I think if a user sees a value like "quality" they could ascribe more meaning to it than it deserves. I think we should add some docs about what it means, and how to interpret it.

That makes sense. Maybe calculating typical modularity and using the term modularity is safe enough, since definition of modularity is available everywhere.

scanpy/tools/_louvain.py
gokceneraslan: Done (I hope)

scanpy/tools/_louvain.py
gokceneraslan: leiden and louvain modules are over 90% the same. So making this a helper would be just deferring the inevitable merger of the two. So I'd like to put the same effort to merge them instead of writing a helper function just for the partition quality.

How does that sound?

gokceneraslan: Related issue: https://github.com/theislab/scanpy/issues/570

flying-sheep: sounds great!

scanpy/tools/_louvain.py
```diff
@@ -133,7 +136,24 @@ def louvain(
...
-            # adata.uns['louvain_quality'] = part.quality()
+            if 'quality' in dir(part):
+                q = part.quality()
+                if isinstance(part, louvain.RBConfigurationVertexPartition) and q > 1.0: #scale
+                    m = weights.sum() if use_weights else g.ecount()
+                    q /= m if directed else m*2
+                if isinstance(part, (
+                    louvain.RBConfigurationVertexPartition,
+                    louvain.ModularityVertexPartition),
+                ):
```
flying-sheep: Hmm, not quite. The idea is that you can add or remove single lines without touching other lines (e.g. for nicer git diffs). So you want the lines that have the same semantic meaning (one entry in a tuple) to look exactly the same.

```py
if isinstance(part, (
    louvain.RBConfigurationVertexPartition,
    louvain.ModularityVertexPartition,
)):
```

gokceneraslan: Ah ok, sorry. I liked the `)):` part actually :)

flying-sheep: Thank you!

flying-sheep (approved):

scanpy/tests/test_clustering.py
```diff
@@ -8,7 +8,10 @@ def adata_neighbors():
 
 
 def test_leiden_basic(adata_neighbors):
-    sc.tl.leiden(adata_neighbors)
+    sc.tl.leiden(adata_neighbors, use_weights=True, directed=True)
+    sc.tl.leiden(adata_neighbors, use_weights=True, directed=False)
+    sc.tl.leiden(adata_neighbors, use_weights=False, directed=True)
+    sc.tl.leiden(adata_neighbors, use_weights=False, directed=False)
```
ivirshup, line 14: Would you mind timing these tests vs what they were previously? I think I had previously done something like this, but it turned out the clustering took a while. Ditto for similar tests below.

Since we don't actually validate the values, these could be sped up by reducing the samples (possibly also sample classes) present.

scanpy/tools/_leiden.py
```diff
@@ -141,18 +141,43 @@ def leiden(
         categories=natsorted(np.unique(groups).astype('U')),
     )
     # store information on the clustering parameters
-    adata.uns['leiden'] = {}
-    adata.uns['leiden']['params'] = dict(
+    uns_key = key_added if key_added.startswith('leiden') else f'leiden_{key_added}'
```
ivirshup: Why `f"leiden_{key_added}"` instead of `key_added`?

ivirshup:
On the definition of modularity, I did go back over some literature and saw that modularity sometimes takes multiple meanings within a paper. It can be either the [specific quality function](https://leidenalg.readthedocs.io/en/latest/reference.html#modularityvertexpartition), or when used like "modularity optimization" can refer to the whole class of partition optimizing algorithms (which are generic wrt quality function) like `louvain` ([I like section IV F of this paper for an overview](https://arxiv.org/abs/1608.00163v2)).

@LuckyMD

> The quality score is modularity, which is optimized. Thus a "good" partition is a high quality score by definition. Or what are you referring to as "good"?

I think of the quality function/score as being determined by the `partition_type`. 

To me, a good partition is one that seperates data points into discrete groups which reflect some true underlying structure. I put this in quotes since it’s ill-defined, however we can tell when it’s definitely not true. A high quality score for a partitioning is just a high quality score for a partitioning.

@gokceneraslan 

> we can report the original quality value as "raw quality" (whatever it is) and the modularity together.

Personally, I would just report the quality metric calculated by the quality function used. To me, the point of returning this value would be to know if the optimization went well, which is probably best measured by looking at the optimized value. This would also simplify the code a bunch. I think there's another case for trying to tell if it's a "good" partitioning, but I think that should be handled seperatly.

> Regarding the suggestion to record partition_type.__name__, I think it's a good idea. I'd record it in the uns[uns_key]['partition_type'] though, not in quality_function.

That's reasonable. Just to be sure, we'd keep it in `uns[uns_key]["params"]['partition_type']` like it is now?

scanpy/tools/_leiden.py
```diff
@@ -141,18 +141,43 @@ def leiden(
...
+    adata.uns[uns_key] = {}
+    adata.uns[uns_key]['params'] = dict(
         resolution=resolution,
         random_state=random_state,
         n_iterations=n_iterations,
+        use_weights=use_weights,
+        directed=directed,
+        partition_type=None if partition_type is None else partition_type.__name__,
     )
+    if 'quality' in dir(part):
```
ivirshup: Could this be:

```suggestion
   if isinstance(part, MutableVertexPartition):
```

scanpy/tools/_louvain.py
```diff
@@ -133,7 +136,24 @@ def louvain(
...
+                    louvain.ModularityVertexPartition,
+                )):
+                    qual_type = ' (scaled modularity)'
+                else:
+                    qual_type = ''
+                adata.uns[uns_key]['quality'] = q
+                quality_msg = (
+                    f'\n'
+                    f'    quality of the partitioning{qual_type} is {q:.3f}\n'
+                    f'    added "quality" key to adata.uns["{uns_key}"]'
```
ivirshup: Should the `partition_type`, `random_state`, and `resolution` be logged as well?

ivirshup (changes requested):
A few minor comments

gokceneraslan:
@ivirshup 

> Personally, I would just report the quality metric calculated by the quality function used. To me, the point of returning this value would be to know if the optimization went well, which is probably best measured by looking at the optimized value.

The quality score returned by RBConfigurationVertexPartition is unscaled modularity, so it's something like `41726.23`. So how would one interpret that? I don't think there is any point in reporting raw RBConfigurationVertexPartition quality value.

scanpy/tools/_leiden.py
gokceneraslan: Because people typically use the leiden prefix e.g. key_added="leiden_2.0".

gokceneraslan:
> That's reasonable. Just to be sure, we'd keep it in `uns[uns_key]["params"]['partition_type']` like it is now?

We don't store the partition_type anywhere now, but yes we'll after this PR.

scanpy/tools/_louvain.py
gokceneraslan: I'm adding resolution, too. But I feel like adding all of them would be too verbose.

scanpy/tools/_leiden.py
gokceneraslan: This part is gone now, I simply report standard modularity regardless of the partititon_type.

scanpy/tests/test_clustering.py
gokceneraslan: image

I think higher coverage is always better than increasing the test time unless the increase is really wild.

gokceneraslan: Previously:

image

gokceneraslan: For louvain:

image

gokceneraslan:
After the discussions we had here, I'm reporting the standard modularity now, regardless of the partition_type. Please have a look and let me know if it makes sense.

scanpy/tools/_leiden.py
ivirshup: I guess that’s the point I’m trying to make. People will specify the key they want to use, so I don’t think we should modify it. For example, if I want to use the params from a leiden call in some code I write, I’d prefer:

```python
sc.tl.leiden(adata, key_added=key_added, ...)
adata.uns[key_added][“quality”]
```

over

```python
sc.tl.leiden(adata, key_added=key_added, ...)
uns_key = key_added
if not key_added.startswith(“leiden”):
    uns_key = f“leiden_{uns_key}”
adata.uns[uns_key][“quality”]
```

Basically, if I wanted the key_added to start with “leiden”, I would’ve had the key added start with “leiden”.

scanpy/tests/test_clustering.py
ivirshup: More coverage is always better, but I run the tests more often if it doesn’t take forever. I have been thinking we should have a separate more rigorous set of tests where execution time doesn’t matter, but that’s a separate topic.

These time changes are fine though. I think I had previously set `use_weights` and `directed` via fixtures, which caused the adata_neighbors fixture to get recalculated for each computation, which was the expensive part.

scanpy/tools/_leiden.py
gokceneraslan: OK, now I get it :) Done.

scanpy/tools/_louvain.py
ivirshup: Fair enough. I just like the idea of being able to run the exact same call, but this is stored in the params anyways.

ivirshup:
Sorry for the long delay – I was away on a retreat. By the way, congrats on the spatial letter!

I'm not sure I understand why you'd report the standard modularity if the partitioning was done with multi resolution modularity or some other quality function. To me, this makes the metric being pretty disconnected from the computation that was run. It seems likely that there could be non-proportional relationships between the whatever quality function is used and unscaled modularity. For example there could be a case where: partitioning A has higher unscaled modularity than partitioning B, but B has higher multi resolution modularity quality with resolution .8 than A. If the multi resolution modularity is what was run, I think the logging should reflect that run resulted in a "better" optimization.

Separately, if it's meant to assess the quality of the clustering, why not calculate something like silhouette? From my perspective, it would be because that's separate enough from the process of clustering that it should be run separately.

scanpy/tools/_leiden.py
```diff
@@ -141,18 +141,30 @@ def leiden(
         categories=natsorted(np.unique(groups).astype('U')),
     )
     # store information on the clustering parameters
-    adata.uns['leiden'] = {}
-    adata.uns['leiden']['params'] = dict(
+    adata.uns[key_added] = {}
```
ivirshup, line 156: Since this is a behavior change, be sure to mention it in the release note.

flying-sheep: @gokceneraslan could you please address this and rebase the PR?

gokceneraslan:
I think the PR is now good as is. I will be quite busy for some time, therefore I won't have time for further discussions. Please feel free to close, update or merge it.

flying-sheep:
@gokceneraslan did you address @ivirshup’s questions?

cyrus303 (approved):

PR #820, opened 2019-09-08
Make write_knn_indices option of Neighbors.compute_neighbors() public

gokceneraslan:
`compute_neighbors` method of the Neighbors class has a nice option called `write_knn_indices` which saves kNN indices into the Neighbors object. But this object and its members are not accessible from sc.pp.neighbors since we don't save the reference to it. 

Making the write_knn_indices option available from `sc.pp.neighbors` gives access to knn indices which provide additional information on top of connectivities and distances since these matrices are symmetrized. 

With knn_indices one can do cool things like building a mutual kNN graph, e.g.:

```python
import scipy.sparse as sp

lm = sp.lil_matrix((adata.n_obs, adata.n_obs))
lm.rows = adata.uns['neighbors']['knn_indices']
lm.data = np.ones_like(adata.uns['neighbors']['knn_indices'])
lm = lm.tocsr()
lm.setdiag(0)
lm.eliminate_zeros()
lm = lm.multiply(lm.T) # build mnn mask

adata.uns['neighbors']['distances'] = adata.uns['neighbors']['distances'].multiply(lm)
adata.uns['neighbors']['connectivities'] = adata.uns['neighbors']['connectivities'].multiply(lm)
```

### DIFF
--- a/scanpy/neighbors/__init__.py
+++ b/scanpy/neighbors/__init__.py
@@ -30,6 +30,7 @@ def neighbors(
     method: str = 'umap',
     metric: Union[str, Metric] = 'euclidean',
     metric_kwds: Mapping[str, Any] = {},
+    write_knn_indices: bool = False,
     copy: bool = False,
 ) -> Optional[AnnData]:
     """\
@@ -69,6 +70,8 @@ def neighbors(
         A known metric’s name or a callable that returns a distance.
     metric_kwds
         Options for the metric.
+    write_knn_indices
+        Saves kNN indices into adata.uns['neighbors'].
     copy
         Return a copy instead of writing to adata.
 
@@ -91,7 +94,7 @@ def neighbors(
     neighbors.compute_neighbors(
         n_neighbors=n_neighbors, knn=knn, n_pcs=n_pcs, use_rep=use_rep,
         method=method, metric=metric, metric_kwds=metric_kwds,
-        random_state=random_state,
+        write_knn_indices=write_knn_indices, random_state=random_state,
     )
     adata.uns['neighbors'] = {}
     adata.uns['neighbors']['params'] = {'n_neighbors': n_neighbors, 'method': method}
@@ -106,6 +109,8 @@ def neighbors(
     adata.uns['neighbors']['connectivities'] = neighbors.connectivities
     if neighbors.rp_forest is not None:
         adata.uns['neighbors']['rp_forest'] = neighbors.rp_forest
+    if write_knn_indices and knn:
+        adata.uns['neighbors']['knn_indices'] = neighbors.knn_indices
     logg.info(
         '    finished',
         time=start,


ivirshup:
> since these matrices are symmetrized

This sounds familiar, but I don't remember where it happens. Could you point me towards that?

Never mind, I think I found it in [`fuzzy_simplical_set`](https://github.com/lmcinnes/umap/blob/439db748b9959b53d6678b6fdc6cb18e8f49c6c6/umap/umap_.py#L566-L574), but made me think of another question:

Do we want to save the indices and distances as a pair of arrays, or just save a sparse matrix of the original distances? I think the latter would be easier to work with.

gokceneraslan:
> This sounds familiar, but I don't remember where it happens. Could you point me towards that?
> 
> Never mind, I think I found it in [`fuzzy_simplical_set`](https://github.com/lmcinnes/umap/blob/439db748b9959b53d6678b6fdc6cb18e8f49c6c6/umap/umap_.py#L566-L574), 

Yes, exactly.

> but made me think of another question:
> 
> Do we want to save the indices and distances as a pair of arrays, 

What do you mean? If we merge this, we will have `adata.uns['neighbors']['knn_indices']`  and `adata.uns['neighbors']['distances']` (only if it's requested by the user). Is this what you mean? 

> or just save a sparse matrix of the original distances? I think the latter would be easier to work with.

I'm not following. `adata.uns['neighbors']['distances']` encodes a different type of information since it represents the graph structure and it's symmetrized. `knn_indices` on the other hand represents the "raw" output of kNN method. Just like `adata.uns['neighbors']['rp_forest']`, it's additional information about the kNN. It can be used for other things like kNN classifiers or building other types of graphs like mutual kNN as I mentioned.

Just to clarify, I don't propose using knn_indices as a replacement of distance or connectivitiy matrix. It's just to be able to access more details of the kNN construction. Furthermore, it's optional and it's False by default.

ivirshup:
My main point was, why not store `lm` (from the `lm = lm.tocsr()` line of your example) instead of `knn_indices`? That said, I think we may already have that info.

Taking a step back, I started playing around with this, and noticed my distance matrices aren't symmetric. Are yours?

```python
import scanpy as sc
sc.pp.neighbors(pbmc, n_neighbors=15)

d = pbmc.uns["neighbors"]["distances"]
c = pbmc.uns["neighbors"]["connectivities"]

(c != c.T).data.any() # False
(d != d.T).data.any() # True
# checking just sparsity structure:
d_inds = d != 0
(d_inds != d_inds.T).data.any() # True
```

If it not, I suspect `lm` from your example:

```python
lm = sp.lil_matrix((adata.n_obs, adata.n_obs))
lm.rows = adata.uns['neighbors']['knn_indices']
lm.data = np.ones_like(adata.uns['neighbors']['knn_indices'])
lm = lm.tocsr()
```

may be equivalent to `d != 0`. Is that the case?

gokceneraslan:
Oooooh, that's so true :D I didn't know that distance matrix is asymmetric:

image

I was always using the connectivity matrix, that's why I didn't figure it out.

 Thanks a lot for discovering this, I'm closing the PR then.

PR #822, opened 2019-09-09
Don't allow h5py 2.10.0

ivirshup:
Trying to fix the build

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ matplotlib==3.0.*
 pandas>=0.21
 scipy>=1.3
 seaborn
-h5py
+h5py!=2.10.0
 tables
 tqdm
 importlib_metadata>=0.7; python_version < '3.8'


ivirshup:
@gokceneraslan 

This fixes the builds for now. I'm guessing there's some bug added on their end, but haven't confirmed this yet.

Some more notes:

A reason I think a bug could be on our end is that current master of anndata reads that dataset just fine, so maybe the previous code relied on unintended behavior.

gokceneraslan:
Cool! Thanks a lot.

flying-sheep:
Where’s the bug on their end? You did file one, right?

flying-sheep:
I can’t reproduce this locally and would like to get the dependencies unlocked. Where’s the bug report at h5py?

ivirshup:
I'm pretty sure I fixed this around here: https://github.com/theislab/anndata/commit/38e687c32b87cc9cadde2fbd977b36bfb08f00f3, but I opened the issue here: https://github.com/h5py/h5py/issues/1307

flying-sheep:
Thank you!

PR #823, opened 2019-09-09
WIP Scran Normalization

NataliaVoro:
Implements Normalization like scran's computeSumFactors method.

### DIFF
--- a/Untitled.ipynb
+++ b/Untitled.ipynb
@@ -1,0 +1,135 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 11,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import scanpy as sc\n",
+    "import numpy as np"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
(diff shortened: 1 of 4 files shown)


LuckyMD:
looking forward to this being implemented! Will greatly increase the speed of most of my analysis workflows 😃

jipeifeng:
looking forward to it !

cdedonno:
Hey guys @LuckyMD, @Koncopd, @falexwolf, @flying-sheep. I think this feature would be very useful to have in Scanpy, but this PR has been sort of forgotten. 

I would be up to take care of this, but it would be my first contribution and I'd like some advice on how to move forward on this. I take it the main issue with the PR is the missing test for the scran normalization, is that correct?

LuckyMD:
A student of @mbuttner started this, and yes, I believe it's only missing tests. I would probably make a small test case and record the results of the `R` version on there, and use it as a test for this. You can check out the `testing/` directory for tests. 

To start, just fetch this branch and you should be able to commit to it directly as a member of theislab github.

cdedonno:
Unfortunately the PR was not just missing some tests. There are quite a lot of missing parts and the code is not correct. This will take much more effort than expected. I will try working on it on my own branch and close this PR.

mbuttner:
I'm sorry to hear this and hope that you can still use it as a basis for your own implementation. Thank you so much for your effort!

cdedonno:
Yes, it's definitely better than starting from scratch!

Issue #758, opened 2019-07-30
batch_key of highly_variable_genes not working


danielStrobl:
Hey!

I'm currently having an issue with the `batch_key` functionality of `sc.pp.highly_variable_genes` introduced in https://github.com/theislab/scanpy/pull/622 by @gokceneraslan.

If I try setting batch_key, I get a TypeError:
```pytb
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
 in 
----> 1 sc.pp.highly_variable_genes(adata, batch_key="sample", flavor='cell_ranger', n_top_genes=4000, inplace=False)

~/.conda/envs/sc-tutorial/lib/python3.7/site-packages/scanpy/preprocessing/_highly_variable_genes.py in highly_variable_genes(adata, min_disp, max_disp, min_mean, max_mean, n_top_genes, n_bins, flavor, subset, inplace, batch_key)
    336             dtypes.append([('highly_variable_nbatches', int),
    337                            ('highly_variable_intersection', np.bool_)])
--> 338         return np.rec.fromarrays(arrays, dtype=dtypes)

~/.conda/envs/sc-tutorial/lib/python3.7/site-packages/numpy/core/records.py in fromarrays(arrayList, dtype, shape, formats, names, titles, aligned, byteorder)
    606 
    607     if dtype is not None:
--> 608         descr = sb.dtype(dtype)
    609         _names = descr.names
    610     else:

TypeError: data type not understood
```
I have tried with multiple categorical columns and it worked with none of them.

ivirshup:
It looks like this error occurs whenever `batch_key` is specified and `inplace=False`.

MCVE:

```python
import scanpy as sc
pbmc = sc.datasets.pbmc68k_reduced()
pbmc.X = pbmc.raw.X  # So we have reasonable values to calculate on

# These do not throw an error:
sc.pp.highly_variable_genes(pbmc, batch_key="phase")
sc.pp.highly_variable_genes(pbmc, inplace=False)

# This throws an error
sc.pp.highly_variable_genes(pbmc,  batch_key="phase", inplace=False)
```

This does raise the question of what `inplace=False` should return for the batch case. I'd think a recarray (maybe this should change to a dataframe?) with metrics for each of the batches. You could end up with columns with names like: `means_{batch1}`, `dispersions_{batch1}` etc.

@danielStrobl, what were you expecting this to return?

LuckyMD:
[...]

PR #824, opened 2019-09-09
highly_variable_gene batch_key variant fix

LuckyMD:
This PR addresses part of #758. I just added a gene filtering addition to the `batch_key` variant of `sc.pp.highly_variable_genes()`. This ensures that the function does not fail because a gene is not expressed in a batch.

I would welcome some feedback on setting dispersions to 0 for filtered out genes. I think this is the standard set by Seurat, and also what is implemented by @gokceneraslan in the original function. 

I have tested on my own data for correct implementation. Testing for result is more difficult in this I guess...

### DIFF
--- a/scanpy/preprocessing/_highly_variable_genes.py
+++ b/scanpy/preprocessing/_highly_variable_genes.py
@@ -10,6 +10,7 @@ from ._distributed import materialize_as_ndarray
 from ._utils import _get_mean_var
 from .._utils import sanitize_anndata
 from ._simple import filter_genes
+from .._settings import settings, Verbosity
 
 def _highly_variable_genes_single_batch(
     adata: AnnData,
@@ -262,14 +263,15 @@ def highly_variable_genes(
         sanitize_anndata(adata)
         batches = adata.obs[batch_key].cat.categories
         df = []
+        gene_list = adata.var_names
         for batch in batches:
             adata_subset = adata[adata.obs[batch_key] == batch]
 
             # Filter to genes that are in the dataset
-            genes_before = adata_subset.var_names
-            filter_genes(adata_subset, min_cells=1)
-            genes_after = adata_subset.var_names
-            genes_missing = genes_before.difference(genes_after)
+            with settings.verbosity.override(Verbosity.error):
+                filt = filter_genes(adata_subset, min_cells=1, inplace=False)[0]
+
+            adata_subset = adata_subset[:,filt]
 
             hvg = _highly_variable_genes_single_batch(
                 adata_subset,
@@ -282,21 +284,19 @@ def highly_variable_genes(
 
             # Add 0 values for genes that were filtered out
             missing_hvg = pd.DataFrame(
-                np.zeros((len(genes_missing), len(hvg.columns))),
+                np.zeros((np.sum(~filt), len(hvg.columns))),
                 columns=hvg.columns,
             )
             missing_hvg['highly_variable'] = missing_hvg['highly_variable'].astype(bool)
-            missing_hvg['gene'] = genes_missing
+            missing_hvg['gene'] = gene_list[~filt]
             hvg['gene'] = adata_subset.var_names.values
             hvg = hvg.append(missing_hvg, ignore_index=True)
 
             # Order as before filtering
-            tmp = hvg.set_index('gene')
-            tmp = tmp.loc[genes_before]
-            tmp['gene'] = tmp.index
-            tmp.index = hvg.index
+            idxs = np.concatenate((np.where(filt)[0], np.where(~filt)[0]))
+            hvg = hvg.loc[np.argsort(idxs)]
 
-            df.append(tmp)
+            df.append(hvg)
 
         df = pd.concat(df, axis=0)
         df['highly_variable'] = df['highly_variable'].astype(int)
--- a/scanpy/preprocessing/_highly_variable_genes.py
+++ b/scanpy/preprocessing/_highly_variable_genes.py
@@ -9,7 +9,7 @@ from .. import logging as logg
 from ._distributed import materialize_as_ndarray
 from ._utils import _get_mean_var
 from .._utils import sanitize_anndata
-
+from ._simple import filter_genes
 
 def _highly_variable_genes_single_batch(
     adata: AnnData,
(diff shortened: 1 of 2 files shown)


flying-sheep:
I think this could be done more efficiently by using the index returned from `filter_genes(..., inplace=False)` in `_highly_variable_genes_single_batch` and getting rid of the whole data frame merging in afterwards.

flying-sheep (changes requested):
I think this could be done more efficiently by using the index returned from `filter_genes(..., inplace=False)` in `_highly_variable_genes_single_batch` and instead of the whole data frame merging you add in the current version of your changes.

LuckyMD:
> I think this could be done more efficiently by using the index returned from `filter_genes(..., inplace=False)` in `_highly_variable_genes_single_batch` and instead of the whole data frame merging you add in the current version of your changes.

I guess that would depend if you want to have a `filter_genes` call in the HVG selection function every time, or whether you only want it in there in a case where `filter_genes` normally doesn't work. You typically use it on the whole dataset, but not per batch.

Another issue atm is that if you set the verbosity high, then the `filter_genes()` call gives you an output, which is not really intended as the user can't see the function call.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -250,25 +250,52 @@ def highly_variable_genes(
...
-                                                      min_disp=min_disp, max_disp=max_disp,
-                                                      min_mean=min_mean, max_mean=max_mean,
-                                                      n_top_genes=n_top_genes,
-                                                      n_bins=n_bins,
-                                                      flavor=flavor)
-            hvg['gene'] = adata.var_names.values
+
+            # Filter to genes that are in the dataset
+            genes_before = adata_subset.var_names
+            filter_genes(adata_subset, min_cells=1)
```
ivirshup: This makes a copy of the whole `AnnData`, which isn't needed here. I'd follow @flying-sheep's recommendation to use `inplace=False` or calculate it directly. You can probably hide the logging output with a context manager if you want.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -250,25 +250,52 @@ def highly_variable_genes(
...
+            missing_hvg['highly_variable'] = missing_hvg['highly_variable'].astype(bool)
+            missing_hvg['gene'] = genes_missing
+            hvg['gene'] = adata_subset.var_names.values
+            hvg = hvg.append(missing_hvg, ignore_index=True)
+
+            # Order as before filtering
+            tmp = hvg.set_index('gene')
+            tmp = tmp.loc[genes_before]
+            tmp['gene'] = tmp.index
+            tmp.index = hvg.index
```
ivirshup: Do these lines end up doing anything?

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -250,25 +250,52 @@ def highly_variable_genes(
...
-            hvg['gene'] = adata.var_names.values
+
+            # Filter to genes that are in the dataset
+            genes_before = adata_subset.var_names
+            filter_genes(adata_subset, min_cells=1)
+            genes_after = adata_subset.var_names
+            genes_missing = genes_before.difference(genes_after)
+
+            hvg = _highly_variable_genes_single_batch(
+                adata_subset,
```
ivirshup, line 277: This function should probably take an array/ sparse matrix directly, since it doesn't actually use anything else from the `AnnData`.

ivirshup (changes requested):

ivirshup:
Am I correct in thinking this is a fix for a batch having zero variance for a gene? What was the output in that case?

LuckyMD:
This is mainly a fix for cases when multiple genes have zero variance. The output in that case was that `sc.pp.highly_variable_genes()` failed as the bin boundaries were too close to one another. The error I got is:
```
ValueError: Bin edges must be unique: array([          -inf, 9.99999996e-13, 9.99999996e-13, 3.71624832e-03,
       4.50723944e-03, 5.04237041e-03, 7.96065722e-03, 9.17631686e-03,
       1.15813100e-02, 1.34968273e-02, 1.62843971e-02, 1.89858746e-02,
       2.27864407e-02, 2.76163086e-02, 3.43018658e-02, 4.27573830e-02,
       5.52219763e-02, 7.87758350e-02, 1.42211060e-01, 3.10728383e+00,
                  inf]).
You can drop duplicate edges by setting the 'duplicates' kwarg
```

I figured the best way forward would be to exclude those genes from the function, rather than changing the bins in the `_highly_variable_genes_single_batch()` function with the `duplicates` argument as suggested in the error.

scanpy/preprocessing/_highly_variable_genes.py
LuckyMD: They didn't before... I had forgotten to change the line after (`df.append(hvg)` is now `df.append(tmp)`). Thanks for the heads up. As I mentioned, I have only tested that it runs so far.

ivirshup:
> This is mainly a fix for cases when multiple genes have zero variance.

Could you add that as the test case? When some genes aren't expressed in a batch you won't get an error.

> the best way forward would be to exclude those genes from the function

I think the approach of masking out the non-expressed genes sounds reasonable, since that's what you'd probably do if it were just one dataset. I'd definitely defer to @gokceneraslan on any more about the appropriateness of the approach.

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: There’s only one for filtering warnings, so I quickly built one:

https://github.com/theislab/scanpy/blob/e63c2df73fc324675b5cff52b444101852efe055/scanpy/_settings.py#L39-L46

Usage:

```py
with sc.settings.verbosity.override(sc.Verbosity.error):
    sc.noisy_function(adata)
```

LuckyMD: Thanks! Used that now.

scanpy/preprocessing/_highly_variable_genes.py
LuckyMD: Hmm... I didn't make that function. That was done by @gokceneraslan when he added `batch_key`. I would therefore be inclined to leave it like that in this PR.

LuckyMD:
@ivirshup Is this fine for you as is? I think I've addressed all the points now. I'd be keen for this to be merged asap as we need this for another project where this code is failing.

Note that I have not addressed the issue with `inplace=False` when using the `batch_key` parameter.

LuckyMD:
I had to use the pbmc3k dataset for testing, as the error doesn't occur on blobs or pbmc68k_reduced. To test I need sufficient genes that have 0 variance in a subset of the cells.

LuckyMD:
Cancel that @flying-sheep sheep helped me find a way around to test with `pbmc68k_reduced`. This should speed up Travis again.

flying-sheep (approved):

flying-sheep:
Works for me! I’d say we refactor the helper function in a separate PR

PR #825, opened 2019-09-09
Rotate only xticks in sc.pl.violin

gokceneraslan:
`rotation` argument is supposed to rotate only xticks, not both x and y according to the documentation:

image


flying-sheep:
Yes. Thank you!

PR #830, opened 2019-09-11
Add ability to run nearest neighbors, umap and louvain on a GPU using RAPIDS

tomwhite:
This adds RAPIDS as an (experimental) option for running some of the graph algorithms on a GPU.

Example usage is demonstrated here: https://github.com/tomwhite/scanpy_usage/tree/rapids-gpu/1909_rapids_gpu

In summary:

```python
    sc.pp.neighbors(adata, method='rapids')
    sc.tl.louvain(adata, flavor='rapids')
    sc.tl.umap(adata, method='rapids')
```

Timings (on 130K samples):

| Step      | CPU time (s) | GPU time (s) | Speedup |
| --------- | ------------ | ------------ | ------- |
| Neighbors | 47           | 15           | 3x      |
| Louvain   | 70           | 1            | 70x     |
| UMAP      | 186          | 15           | 12x     |

Comments:
* In general the output figures are quite different for each of the algorithms compared to regular Scanpy. (See the figures in the link above.)
* RAPIDS uses exact nearest neighbors (using https://github.com/facebookresearch/faiss I think), whereas the default NNDescent algorithm in Scanpy is approximate.
* Louvain community detection results in a different number of communities (28 vs 45). I'm not sure if it is possible to change the termination criteria or other parameters that would make the two outputs the same, or at least more similar.
* UMAP is different too, and in particular it is bunched in one corner. There are a few very small clusters in the opposite corner, which looks like it might be a bug.
* UMAP computes the nearest neighbors again from scratch - it can’t reuse the ones computed earlier since RAPIDS only exposes the fit/transform methods, and not `simplicial_set_embedding` that regular UMAP exposes.
* In general it would be useful to understand what random number generators each of the alternatives use, how they can be controlled, and what prospects there are for achieving identical output (if any).
* This is using RAPIDS version 0.7 since it is the version available on the Google Cloud image I used (https://cloud.google.com/deep-learning-vm/docs/images).

### DIFF
--- a/scanpy/neighbors/__init__.py
+++ b/scanpy/neighbors/__init__.py
@@ -62,9 +62,11 @@ def neighbors(
         `n_neighbors` nearest neighbor.
     random_state
         A numpy random seed.
-    method : {{`'umap'`, `'gauss'`, `None`}}  (default: `'umap'`)
+    method : {{`'umap'`, `'gauss'`, `'rapids'`, `None`}}  (default: `'umap'`)
         Use 'umap' [McInnes18]_ or 'gauss' (Gauss kernel following [Coifman05]_
         with adaptive width [Haghverdi16]_) for computing connectivities.
+        Use 'rapids' for the RAPIDS implementation of UMAP (experimental, GPU
+        only).
     metric
         A known metric’s name or a callable that returns a distance.
     metric_kwds
@@ -233,6 +235,18 @@ def compute_neighbors_umap(
     )
 
     return knn_indices, knn_dists, forest
+
+
+def compute_neighbors_rapids(
+    X: np.ndarray,
+    n_neighbors: int
+):
+    from cuml.neighbors import NearestNeighbors
+    nn = NearestNeighbors(n_neighbors=n_neighbors)
+    X_contiguous = np.ascontiguousarray(X, dtype=np.float32)
+    nn.fit(X_contiguous)
+    knn_distsq, knn_indices = nn.kneighbors(X_contiguous)
+    return knn_indices, np.sqrt(knn_distsq) # cuml uses sqeuclidean metric so take sqrt
 
 
 def _get_sparse_matrix_from_indices_distances_umap(knn_indices, knn_dists, n_obs, n_neighbors):
@@ -611,8 +625,8 @@ class Neighbors:
             logg.warning(f'n_obs too small: adjusting to `n_neighbors = {n_neighbors}`')
         if method == 'umap' and not knn:
             raise ValueError('`method = \'umap\' only with `knn = True`.')
-        if method not in {'umap', 'gauss'}:
-            raise ValueError('`method` needs to be \'umap\' or \'gauss\'.')
+        if method not in {'umap', 'gauss', 'rapids'}:
+            raise ValueError('`method` needs to be \'umap\', \'gauss\', or \'rapids\'.')
         if self._adata.shape[0] >= 10000 and not knn:
             logg.warning('Using high n_obs without `knn=True` takes a lot of memory...')
         self.n_neighbors = n_neighbors
@@ -629,6 +643,8 @@ class Neighbors:
                     knn_indices, knn_distances, X.shape[0], n_neighbors)
             else:
                 self._distances = _distances
+        elif method == 'rapids':
+            knn_indices, knn_distances = compute_neighbors_rapids(X, n_neighbors)
         else:
             # non-euclidean case and approx nearest neighbors
             if X.shape[0] < 4096:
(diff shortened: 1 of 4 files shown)


flying-sheep:
Wow, nice, those are some impressive speedups!

flying-sheep:
I think this looks good and simple enough. Could you please fix the CI errors?

Also there’s 3 added optional deps: cuml, cudf, and cugraph. I assume they’re all different CUDA packages. Could you add them into an `extra` in setup.py?

The RAPIDS umap branch doesn’t use a metric argument. Does it support metrics other than euclidean?

tomwhite:
> The RAPIDS umap branch doesn’t use a metric argument. Does it support metrics other than euclidean?

No, it only supports Euclidean distance currently.

tomwhite:
@flying-sheep thanks for taking a look. I've fixed the CI errors now and added the deps as suggested.

scanpy/neighbors/__init__.py
```diff
@@ -642,7 +671,7 @@ def compute_neighbors(
             self.knn_indices = knn_indices
             self.knn_distances = knn_distances
         start_connect = logg.debug('computed neighbors', time=start_neighbors)
-        if not use_dense_distances or method == 'umap':
+        if not use_dense_distances or method == 'umap' or method == 'rapids':
```
flying-sheep:
```suggestion
        if not use_dense_distances or method in {'umap', 'rapids'}:
```

scanpy/neighbors/__init__.py
```diff
@@ -611,8 +638,8 @@ def compute_neighbors(
             logg.warning(f'n_obs too small: adjusting to `n_neighbors = {n_neighbors}`')
         if method == 'umap' and not knn:
             raise ValueError('`method = \'umap\' only with `knn = True`.')
-        if method not in {'umap', 'gauss'}:
-            raise ValueError('`method` needs to be \'umap\' or \'gauss\'.')
+        if method not in {'umap', 'gauss', 'rapids'}:
```
flying-sheep, line 643:
```suggestion
        if method == 'rapids' and metric != 'euclidean':
            raise ValueError("`method` 'rapids' only supports the 'euclidean' `metric`.")
        if method not in {'umap', 'gauss', 'rapids'}:
```

scanpy/tools/_umap.py
```diff
@@ -89,6 +90,8 @@ def umap(
         `spread`.
     copy : `bool` (default: `False`)
         Return a copy instead of writing to adata.
+    method : {{'umap', 'rapids'}}  (default: `'umap'`)
```
flying-sheep: This function doesn’t use `@doc_params`, so normal braces should be sufficient.

```suggestion
    method : {`'umap'`, `'rapids'`}  (default: `'umap'`)
```

scanpy/tools/_umap.py
```diff
@@ -122,28 +125,48 @@ def umap(
...
+            gamma,
+            negative_sample_rate,
+            n_epochs,
+            init_coords,
+            random_state,
+            neigh_params.get('metric', 'euclidean'),
+            neigh_params.get('metric_kwds', {}),
+            verbose=settings.verbosity > 3,
+        )
+    elif method == 'rapids':
```
flying-sheep, line 151: Maybe a warning would be enough here?

```suggestion
    elif method == 'rapids':
        metric = neigh_params.get('metric', 'euclidean')
        if metric != 'euclidean':
            raise ValueError(
                f'`sc.pp.neighbors` was called with `metric` {metric!r}, '
                "but umap `method` 'rapids' only supports the 'euclidean' metric."
            ) 
```

scanpy/neighbors/__init__.py
```diff
@@ -611,8 +638,8 @@ def compute_neighbors(
...
+            raise ValueError('`method` needs to be \'umap\', \'gauss\', or \'rapids\'.')
```
flying-sheep:
```suggestion
            raise ValueError("`method` needs to be 'umap', 'gauss', or 'rapids'.")
```

flying-sheep (changes requested):
OK, great! I addressed the metric problem in some suggestions.

Do you think testing this on Travis is possible (i.e. running this on a CPU) or do the packages *need* a GPU?

I’d also like to ask you to add the dependencies to a `rapids` extra in setup.py:

https://github.com/theislab/scanpy/blob/b9dbe36b3e9c52a3c793af991c9cd7c78db9f92a/setup.py#L30-L33

If tests are possible, we need to add some and add the `rapids` extra to the list of test extras:

https://github.com/theislab/scanpy/blob/b9dbe36b3e9c52a3c793af991c9cd7c78db9f92a/.travis.yml#L20

tomwhite:
The packages do need a GPU, unfortunately. There is no way to fallback to run on a CPU, so I don't think Travis can test them.

flying-sheep (approved):

flying-sheep:
OK, too bad … Thank you very much for the PR, this looks like really impressive improvements!

PR #831, opened 2019-09-11
added cache_compression kwarg to read()

flying-sheep:
Since theislab/anndata#123 doesn’t seem to gain traction …


flying-sheep:
Hey @VolkerBergen, does it work like you want?

VolkerBergen:
Thanks, that option is very useful and works perfectly.

mini-benchmark on my data (compared to non-compressed): 
gzip: 35% memory, 3x runtime
lzf: 66% memory, 2x runtime

PR #839, opened 2019-09-13
Unify all isinstance calls

flying-sheep:
This should take care of everything: All `isinstance` calls are with concrete types or ABCs, never `typing` types. We should maybe make a helper for this common case:

```py
T = TypeVar("T", Any)

def _ensure_sequence(obj: Union[T, Sequence[T]])  -> Sequence[T]:
    if isinstance(obj, str) or not isinstance(obj, cabc.Sequence):
        obj = [obj]
    return obj
```


scanpy/_utils.py
```diff
@@ -126,11 +126,11 @@ def descend_classes_and_funcs(mod: ModuleType, root: str, encountered=None):
     for obj in vars(mod).values():
         if not _one_of_ours(obj, root):
             continue
-        if isinstance(obj, Callable) and not isinstance(obj, MethodType):
+        if callable(obj) and not isinstance(obj, MethodType):
```
gokceneraslan, line 129: 👍

gokceneraslan (approved):
LGTM!

gokceneraslan:
@flying-sheep What do you think about https://pypi.org/project/pytest-mypy/ ?

flying-sheep:
I think separating static analysis from running the tests is the way to go (in #841 I added black checking as a separate step.)

Also mypy is very strict, so we might have to fix *a lot*.

ivirshup:
I'm not sure the benefits from using mypy would outweigh the pain in using unstable numpy type-stubs

ivirshup:
Also, potential unintended consequence of some of those instance checks:

```python
In [3]: isinstance(np.ones(5), collections.abc.Sequence)                                            
Out[3]: False
```

`np.ndarray` don't get classified as sequences.

flying-sheep:
Ugh, good point. Many of them are luckily short string sequences, which will tend to be lists anyway, but uuugh.

The reason is that they don’t implement `__reversed__`…

ivirshup:
Since `[::-1]` is so much more obvious 😉

It's pretty common for me to pass ndarrays or Series to these methods. Mostly this comes from programatic usage where I'm getting values from something like `unique` or `setdiff1d`.

flying-sheep:
It’s both bullshit that that numpy/numpy#2776 is unfixed since 2012 and Python doesn’t have instance checks for collections without testing for all the mixed-in methods.

What we mostly want to test for is if something is iterable and/or indexable. For “sized iterable”, this is possible via `isinstance(x, cabc.Collection)`, but everything more complex has all those mixin methods that are checked for …

At least what we have now is better than `isinstance(x, list)`.

flying-sheep:
Actually, I call bullshit even more:

```py
class seq_array(np.ndarray):
    def __reversed__(self):
        return iter(self[::-1])
    def index(self, value) -> int:
        return np.in1d(self, value).nonzero()[0]
    def count(self, value) -> int:
        return (self == value).sum()


assert issubclass(seq_array, cabc.Collection)
assert issubclass(seq_array, cabc.Reversible)
for meth in "__contains__ __iter__ __reversed__ index count".split():
    assert hasattr(seq_array, meth), meth
print(issubclass(seq_array, cabc.Sequence))
```

prints `False`. wat.jpg

/edit: Hilarious. Sequence doesn’t implement `__subclasscheck__`, so only things that are `Sequence.register`ed are considered sequences.

ivirshup:
Is a multidimensional array still a sequence? Should `iter(s: Sequence[T]) -> Iterable[T]`? That's not what a multidimensional array does. I guess I'm not even sure what the definition of a sequence is meant to be. That thread brings up a good case of 0-dimensional arrays.

On generic types, working with both numpy and pythons type system is a bit of a pain. `isinstance(np.int64(1), int) == False`. Plus numpy's `dtype` system is a bit of a mess in general.

This is the stuff that makes me go write Julia for a while.

flying-sheep:
Of course it is, but it’s a sequence of sequences (… of sequences of sequences …):

```py
>>> list(iter(np.array([[1,2],[3,5]])))                                                                                                                                                                                                  
[array([1, 2]), array([3, 5])]
```

Yes, 0D-arrays aren’t sequences, but I’m OK with runtime errors if you pass one of those here.

PR #844, opened 2019-09-16
Use cacheing and parallelism for qc metrics

ivirshup:
This simplifies `top_segment_proportions_sparse_csr` by using improvements in numba which allow cacheing parallel code.

A downside of this is it takes a really long time to compile on first run, which might be off-putting.

Side note: I accidentally ran formatting before committing, so some other lines got changed too.

### DIFF
--- a/scanpy/preprocessing/_qc.py
+++ b/scanpy/preprocessing/_qc.py
@@ -1,4 +1,5 @@
 from typing import Optional, Tuple, Collection
+from warnings import warn
 
 import numba
 import numpy as np
@@ -69,8 +70,6 @@ def describe_obs(
         Whether to place calculated metrics in `adata.obs`.
     X
         Matrix to calculate values on. Meant for internal usage.
-    parallel
-        Deprecated argument, doesn't do anything.
 
     Returns
     -------
@@ -79,6 +78,11 @@ def describe_obs(
 
     {doc_obs_qc_returns}
     """
+    if parallel is not None:
+        warn(
+            "Argument `parallel` is deprecated, and currently has no effect.",
+            FutureWarning
+        )
     # Handle whether X is passed
     if X is None:
         X = _choose_mtx_rep(adata, use_raw, layer)
@@ -227,8 +231,6 @@ def calculate_qc_metrics(
     {doc_expr_reps}
     inplace
         Whether to place calculated metrics in `adata`'s `.obs` and `.var`.
-    parallel
-        Deprecated argument, doesn't do anything.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


scanpy/preprocessing/_qc.py
```diff
@@ -70,8 +70,7 @@ def describe_obs(
     X
         Matrix to calculate values on. Meant for internal usage.
     parallel
-        Whether values are calculated in parallel. If `None`, heuristic based
-        on numba compilation time is used.
+        Deprecated argument, doesn't do anything.
```
flying-sheep: Just remove it from the docs, also below, and raise a `FutureWarning` or `DeprecationWarning` if specified.

ivirshup:
Just a heads up, I would like to run a few benchmarks on this before merging.

flying-sheep:
changed the title to WIP

ivirshup:
Benchmark results from my laptop:

|  | Parallel | Single |
|-|--------|--------|
| Compilation | ~12s | ~5s |
| Run (3k cells, 37k genes) | ~70ms | ~120ms | 
| Run (50k cells, 35k genes) | ~3.8s | ~8.1s | 
| Run (370k cells, 33k genes) | ~13.7s |  ~17.4 s |

So... probably worth it? I recall the difference being more pronounced with larger dataset sizes, but that was with a different numba version and maybe a different machine.

flying-sheep:
> A downside of this is it takes a really long time to compile on first run, which might be off-putting.

Right, numba only compiles stuff when first run (because otherwise it can’t know the types) so this doesn’t slow down importing scanpy.

> So... probably worth it?

We could wrap it in a function that checks the number of cells and only compiles this to faster code when necessary. If we do this in a generic way we could even defer importing numpy, saving on import duration (although there probably isn’t much functionality without running numpy-driven functions)

ivirshup:
> We could wrap it in a function that checks the number of cells and only compiles this to faster code when necessary.

So that's what this PR would replace. The reason I thought this could be replaced is that `numba` now allows on-disk cacheing of parallelized functions. This means that the function would only have to be compiled once per install. That cache only get's invalidated if function's source code get's modified, so this shouldn't cause too much pain for development testing times.

I've added a note to the documentation mentioning this, so I think it's fine.

flying-sheep:
OK! A global, per-install cache. Where is it stored?

ivirshup:
[Probably the nearby `__pycache__`](https://numba.pydata.org/numba-doc/dev/reference/envvars.html#envvar-NUMBA_CACHE_DIR)

flying-sheep:
The solidness of their fallback strategy brings a tear to my eye. So. diligent.

flying-sheep (approved):

PR #846, opened 2019-09-16
Run static analysis concurrently

ivirshup:
Runs static analysis concurrently with tests, while currently static analysis is run first. This cuts down on total test time, and will always test both correctness and style.


flying-sheep:
Right. This will of course cost travis more money, but will be better for us…

PR #847, opened 2019-09-16
Add cache_compression setting

flying-sheep:
Default: lzf. see theislab/anndata#123

To review without being bogged down with whitespace changes, check: https://github.com/theislab/scanpy/pull/847/files?w=1

### DIFF
--- a/scanpy/_settings.py
+++ b/scanpy/_settings.py
@@ -76,6 +76,7 @@ class ScanpyConfig:
         cachedir: Union[str, Path] = "./cache/",
         datasetdir: Union[str, Path] = "./data/",
         figdir: Union[str, Path] = "./figures/",
+        cache_compression: Union[str, None] = 'lzf',
         max_memory=15,
         n_jobs=1,
         logfile: Union[str, Path, None] = None,
@@ -98,6 +99,7 @@ class ScanpyConfig:
         self.cachedir = cachedir
         self.datasetdir = datasetdir
         self.figdir = figdir
+        self.cache_compression = cache_compression
         self.max_memory = max_memory
         self.n_jobs = n_jobs
         self.categories_to_ignore = categories_to_ignore
@@ -275,6 +277,24 @@ class ScanpyConfig:
     def figdir(self, figdir: Union[str, Path]):
         _type_check(figdir, "figdir", (str, Path))
         self._figdir = Path(figdir)
+
+    @property
+    def cache_compression(self) -> Optional[str]:
+        """\
+        Compression for `sc.read(..., cache=True)` (default `'lzf'`).
+
+        May be `'lzf'`, `'gzip'`, or `None`.
+        """
+        return self._cache_compression
+
+    @cache_compression.setter
+    def cache_compression(self, cache_compression: Optional[str]):
+        if cache_compression not in {'lzf', 'gzip', None}:
+            raise ValueError(
+                f"Invalid cache_compression={cache_compression} "
+                "∉ {{'lzf', 'gzip', None}}"
+            )
+        self._cache_compression = cache_compression
 
     @property
     def max_memory(self) -> Union[int, float]:
--- a/scanpy/_settings.py
+++ b/scanpy/_settings.py
@@ -291,8 +291,8 @@ class ScanpyConfig:
     def cache_compression(self, cache_compression: Optional[str]):
         if cache_compression not in {'lzf', 'gzip', None}:
             raise ValueError(
-                f"Invalid cache_compression={cache_compression} "
-                "∉ {{'lzf', 'gzip', None}}"
+                f"`cache_compression` ({cache_compression}) "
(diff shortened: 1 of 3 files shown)


scanpy/readwrite.py
```diff
@@ -30,6 +31,12 @@
 """Available file formats for reading data. """
 
 
+class Empty(Enum):
+    token = 0
+
+_empty = Empty.token
```
ivirshup, line 37: This is to differentiate between "no option passed" and an explicit "no compression", right?

Would we want to adopt this convention everywhere? Is there a more pythonic way to do this?

scanpy/_settings.py
```diff
@@ -76,6 +76,7 @@ def __init__(
         cachedir: Union[str, Path] = "./cache/",
         datasetdir: Union[str, Path] = "./data/",
         figdir: Union[str, Path] = "./figures/",
+        cache_compression: Union[str, None] = 'lzf',
```
ivirshup, line 79: Will `lzf` as a default be a problem for interoperability? I know `h5py` ships with it, but does the hdf5 library or any of the `R` wrappers?

scanpy/_settings.py
```diff
@@ -276,6 +278,24 @@ def figdir(self, figdir: Union[str, Path]):
...
+        May be `'lzf'`, `'gzip'`, or `None`.
+        """
+        return self._cache_compression
+
+    @cache_compression.setter
+    def cache_compression(self, cache_compression: Optional[str]):
+        if cache_compression not in {'lzf', 'gzip', None}:
+            raise ValueError(
+                f"Invalid cache_compression={cache_compression} "
+                "∉ {{'lzf', 'gzip', None}}"
```
ivirshup: I'd probably go with something like:

```python
"`cache_compression` must be in {{'lzf'. 'gzip', None}}"
```


instead of <math>&#x02209;</math>. It's more ascii and biologist friendly.

ivirshup (commented):
Reading this gave me an idea for an alternative way to achieve the same end.

I'm thinking there should be a setting like `h5ad_compression`. This is what the cache would use, as well as `AnnData.write_h5ad`. This could be changed on an environment by environment basis, or with a scanpy settings context manager. This would remove the need for the `cache_compression` argument, but achieve the same goals. We would probably still need the enum for `AnnData.write.*`

We could also have a `zarr_compression`, potentially even `file_format`.

scanpy/readwrite.py
flying-sheep: No and no. It’s only used if `None` doesn’t cut it, e.g.

- here: `compression=None` already means something, so we can’t use it as default and let it mean “use the setting”, because it would be very confusing if `compression=None` means “no compression” in one API and “use the default compression” in another.
- in `slice`-like interfaces: `slice(1) == slice(None, 1, None)`, but `slice(1, None) == slice(1, None, None)`, so the default for the second argument can’t be `None`.

scanpy/_settings.py
flying-sheep: Very good point for theislab/anndata#123, moot point here: This is about our cache, no need for interoperability here.

scanpy/readwrite.py
ivirshup: Alright, I think that's what I meant by

> differentiate between "no option passed" and an explicit "no compression"

since `None` already has a meaning here. Another approach would be to do some meta programming and figure out whether an argument was passed. A sentinel value is probably better here.

flying-sheep: You’re right of course. I should have said “exactly, no, and no”, but ended up only replying to your last paragraph’s questions: It’s to differentiate, we shouldn’t adopt it elsewhere, and [this is the most pythonic way if we can’t use `None`](https://www.python.org/dev/peps/pep-0484/#support-for-singleton-types-in-unions).

scanpy/_settings.py
ivirshup: This was more relevant to the point in my main review post, which was why not just define a `h5ad_compression` setting instead.

flying-sheep:
@ivirshup I think writing a file for uploading it to the web, for read caches, and for for checkpoints of a pipeline has different requirements. I think a `h5ad_compression` or even `hdf5_compression` setting could have its place, but separately from the `cache_compression`.

We’ll have to think about naming though. Maybe we want to namespace our settings like matplotlib’s rcparams?

brianpenghe:
Can I ask how to use cache_compression?

flying-sheep:
ah, right, we should probably document it better. The attributes you can set on `sc.settings` are documented [here](https://scanpy.readthedocs.io/en/stable/api/scanpy._settings.ScanpyConfig.html), `cache_compression` [here](https://scanpy.readthedocs.io/en/stable/api/scanpy._settings.ScanpyConfig.cache_compression.html) specifically.

PR #857, opened 2019-10-01
Improve accuracy and speed of _get_mean_var

ivirshup:
This is a PR I've been meaning to get around to for a while. Previously discussed with @fidelram and @falexwolf. This calculates mean and variance for sparse matrices much faster than the current implementation. In addition, this accumulates to 64bit floats, which can greatly increase accuracy for large datasets and values.

Some numbers using 50k cells from tabula muris using the call `sc.pp._utils._get_mean_var`.

|   | speed | memory (`%memit`) |
|--|-------|---------|
|current master| ~6.2 seconds | ~4.7 GB |
|this PR | 1.2 seconds | 0.5 MB |

Pretty good improvements.

For accuracy, I'll compare against `sklearn.utils.sparsefuncs.mean_variance_axis`:

```python
import numba
import numpy as np
from scipy import sparse
from sklearn.utils.sparsefuncs import mean_variance_axis

import seaborn as sns

csr64 = sparse.random(10000, 1000, format="csr", dtype=np.float64)

csr32 = csr64.astype(np.float32)

m_my64, v_my64 = sparse_mean_variance_axis(csr64, 0)
m_sk64, v_sk64 = mean_variance_axis(csr64, 0)
m_my32, v_my32 = sparse_mean_variance_axis(csr32, 0)
m_sk32, v_sk32 = mean_variance_axis(csr32, 0)

sns.distplot(m_sk64 - m_sk32, color="blue")
sns.distplot(m_my64 - m_my32, color="red")
plt.show()
```

<details>
<summary> plot </summary>

residual_mean

</details>

```python
sns.distplot(m_sk64 - m_sk32, color="blue")
sns.distplot(m_my64 - m_my32, color="red")
plt.show()
```

<details>
<summary> plot </summary>

residual_var

</details>

### DIFF
--- a/scanpy/preprocessing/_utils.py
+++ b/scanpy/preprocessing/_utils.py
@@ -1,29 +1,134 @@
 import numpy as np
-from scipy.sparse import issparse
-
+from scipy import sparse
+import numba
 
 STANDARD_SCALER_FIXED = False
 
 
 def _get_mean_var(X):
-    # - using sklearn.StandardScaler throws an error related to
-    #   int to long trafo for very large matrices
-    # - using X.multiply is slower
-    if not STANDARD_SCALER_FIXED:
-        mean = X.mean(axis=0)
-        if issparse(X):
-            mean_sq = X.multiply(X).mean(axis=0)
-            mean = mean.A1
-            mean_sq = mean_sq.A1
-        else:
-            mean_sq = np.multiply(X, X).mean(axis=0)
-        # enforece R convention (unbiased estimator) for variance
-        var = (mean_sq - mean ** 2) * (X.shape[0] / (X.shape[0] - 1))
+    if sparse.issparse(X):
+        mean, var = sparse_mean_variance_axis(X, axis=0)
     else:
-        from sklearn.preprocessing import StandardScaler
+        if STANDARD_SCALER_FIXED:
+            from sklearn.preprocessing import StandardScaler
 
-        scaler = StandardScaler(with_mean=False).partial_fit(X)
-        mean = scaler.mean_
-        # enforce R convention (unbiased estimator)
-        var = scaler.var_ * (X.shape[0] / (X.shape[0] - 1))
+            scaler = StandardScaler(with_mean=False).partial_fit(X)
+            mean, var = scaler.mean_, scaler.var_
+        else:
+            mean = X.mean(axis=0, dtype=np.float64)
+            var = X.var(axis=0, dtype=np.float64)
+    # enforce R convention (unbiased estimator) for variance
+    var = var * (X.shape[0] / (X.shape[0] - 1))
     return mean, var
+
+
+def sparse_mean_variance_axis(mtx: sparse.spmatrix, axis: int):
+    """
+    This code and internal functions are based on sklearns
+    `sparsefuncs.mean_variance_axis`.
+
+    Modifications:
+    * allow deciding on the output type, which can increase accuracy when calculating the mean and variance of 32bit floats.
+    * This doesn't currently implement support for null values, but could.
+    * Uses numba not cython
+    """
+    assert axis in (0, 1)
+    if isinstance(mtx, sparse.csr_matrix):
+        if axis == 0:
+            return sparse_mean_var_minor_axis(
+                mtx.data, mtx.indices, mtx.shape[0], mtx.shape[1], np.float64
+            )
+        elif axis == 1:
+            return sparse_mean_var_major_axis(
+                mtx.data,
+                mtx.indices,
+                mtx.indptr,
+                mtx.shape[0],
+                mtx.shape[1],
+                np.float64,
+            )
+    elif isinstance(mtx, sparse.csc_matrix):
+        if axis == 0:
+            return sparse_mean_var_major_axis(
+                mtx.data,
+                mtx.indices,
+                mtx.indptr,
+                mtx.shape[1],
+                mtx.shape[0],
+                np.float64,
+            )
+        elif axis == 1:
+            return sparse_mean_var_minor_axis(
+                mtx.data, mtx.indices, mtx.shape[1], mtx.shape[0], np.float64
+            )
+    else:
+        raise ValueError(
+            "This function only works on sparse csr and csc matrices"
+        )
+
+
+@numba.njit(cache=True)
+def sparse_mean_var_minor_axis(data, indices, major_len, minor_len, dtype):
+    """
+    Computes mean and variance for a sparse matrix for the minor axis.
+
+    Given arrays for a csr matrix, returns the means and variances for each
+    column back.
+    """
+    non_zero = indices.shape[0]
+
+    means = np.zeros(minor_len, dtype=dtype)
+    variances = np.zeros_like(means, dtype=dtype)
+
+    counts = np.zeros(minor_len, dtype=np.int64)
+
+    for i in range(non_zero):
+        col_ind = indices[i]
+        means[col_ind] += data[i]
+
+    for i in range(minor_len):
+        means[i] /= major_len
+
+    for i in range(non_zero):
+        col_ind = indices[i]
+        diff = data[i] - means[col_ind]
+        variances[col_ind] += diff * diff
+        counts[col_ind] += 1
+
+    for i in range(minor_len):
+        variances[i] += (major_len - counts[i]) * means[i] ** 2
+        variances[i] /= major_len
+
+    return means, variances
+
+
+@numba.njit(cache=True)
+def sparse_mean_var_major_axis(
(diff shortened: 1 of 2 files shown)


ivirshup:
Remaining Qs:

* Does this need support for nans?
* This code is based off `sklearn.utils.sparsefuncs.mean_variance_axis`. Do we need a copy of the sklearn license here? Do we already include that since it's a dependency?

flying-sheep:
The original issue seems to be scikit-learn/scikit-learn#2969

It says that issues arise if there’s a 64 bit index, so the logic should look like

```py
if not issparse(X):
    mean = np.mean(X, axis=0, dtype=np.float64)
    mean_sq = np.multiply(X, X).mean(axis=0, dtype=np.float64)
    var = mean_sq - mean ** 2
elif STANDARD_SCALER_FIXED or np.issubdtype(X.indices.dtype, np.int32):
    from sklearn.preprocessing import StandardScaler
    scaler = StandardScaler(with_mean=False).partial_fit(X)
    mean, var = scaler.mean_, scaler.var_
else:
    mean, var = sparse_mean_variance_axis(X, axis=0) 
```

But actually StandardScaler handles dense matrices just fine, so why not

```py
if not issparse(X) or STANDARD_SCALER_FIXED or np.issubdtype(X.indices.dtype, np.int32):
    from sklearn.preprocessing import StandardScaler
    scaler = StandardScaler(with_mean=False).partial_fit(X)
    mean, var = scaler.mean_, scaler.var_
else:
    mean, var = sparse_mean_variance_axis(X, axis=0) 
```

ivirshup:
Oh, I had assumed the issue was with the dense matrices. In this case, I think we could just go with

```python
if issparse(X):
    mean, var = sparse_mean_variance_axis(X, axis=0)
else:
    mean = np.mean(X, axis=0, dtype=np.float64)
    mean_sq = np.multiply(X, X).mean(axis=0, dtype=np.float64)
    var = mean_sq - mean ** 2
```

This way we always get the improved accuracy.

flying-sheep:
so `StandardScaler` only gives 32 bit floats and isn’t faster than your solution?

ivirshup:
A little surprisingly, looks like it is!

```python
import numpy as np
from sklearn.preprocessing import StandardScaler
from scipy import sparse
import scanpy as sc

s = sparse.random(10000, 10000, format="csr", dtype=np.float32)

def old(X):
    scaler = StandardScaler(with_mean=False).partial_fit(X)
    return scaler.mean_, scaler.var_

%timeit old(s)
# 10.6 ms ± 58.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit sc.pp._utils.sparse_mean_variance_axis(s, 0)
# 5.8 ms ± 180 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

I think it'd be worth it to have better accuracy even if it's a bit more expensive though. I've definitely had variances be much less stable than I would like.

flying-sheep:
Great, then let’s get rid of that branch!

scanpy/preprocessing/_utils.py
```diff
@@ -1,29 +1,127 @@
...
+            )
+        elif axis == 1:
+            return sparse_mean_var_minor_axis(
+                mtx.data, mtx.indices, mtx.shape[1], mtx.shape[0], np.float64
+            )
     else:
-        from sklearn.preprocessing import StandardScaler
+        raise ValueError(
+            "This function only works on sparse csr and csc matrices"
+        )
```
flying-sheep: Please condense this to two branches instead of 4. Also the parameter `axis` should be called `minor_axis`, right?

```py
csr = isinstance(mtx, sparse.csr_matrix)
if not csr and not isinstance(mtx, sparse.csc_matrix):
    raise ValueError(
        "This function only works on sparse csr and csc matrices"
    )
shape = mtx.shape if csr else mtx.shape[::-1]
if axis == int(csr):
    return sparse_mean_var_major_axis(
        mtx.data, mtx.indices, mtx.indptr, *shape, np.float64
    )
else:
    return sparse_mean_var_minor_axis(
        mtx.data, mtx.indices, *shape, np.float64
    )
```

flying-sheep (changes requested):
Looks good, but you could eliminate duplication

scanpy/preprocessing/_utils.py
ivirshup: I'm not a fan of `if axis == int(csr)`. For one thing it would have to be rewritten for COO. Maybe this?

```python
@singledispatch
def sparse_mean_variance_axis(mtx, axis):
    """
    This code and internal functions are based on sklearns
    `sparsefuncs.mean_variance_axis`.

    Modifications:
    * allow deciding on the output type, which can increase accuracy when
    calculating the mean and variance of 32bit floats.
    * This doesn't currently implement support for null values, but could.
    * Uses numba not cython
    """
    raise NotImplementedError(
        "This function is only defined for sparse csr and csc matrices"
    )


@sparse_mean_variance_axis.register(sparse.csr_matrix)
def sparse_mean_variance_axis_csr(mtx, axis):
    assert axis in (0, 1)
    if axis == 0:
        return sparse_mean_var_minor_axis(
            mtx.data, mtx.indices, *mtx.shape, np.float64
        )
    elif axis == 1:
        return sparse_mean_var_major_axis(
            mtx.data, mtx.indices, mtx.indptr, *mtx.shape, np.float64
        )


@sparse_mean_variance_axis.register(sparse.csc_matrix)
def sparse_mean_variance_axis_csc(mtx, axis):
    assert axis in (0, 1)
    if axis == 0:
        return sparse_mean_var_major_axis(
            mtx.data, mtx.indices, mtx.indptr, *mtx.shape[::-1], np.float64
        )
    elif axis == 1:
        return sparse_mean_var_minor_axis(
            mtx.data, mtx.indices, *mtx.shape[::-1], np.float64
        )
```

It's also purely internal code, so I'm  not sure it's worth spending too much time on style.

flying-sheep: My code is the complete thing, so nothing has to be rewritten. `int(csr)` means `1 if csr else 0` (the `int` is actually superfluous since `issubclass(bool, int)`).

If you think my version is too cute, you can use better variable names or as above an `if else` expression instead of converting the bool to an int.

I’m also OK with your version if you replace the last function with:

```py
@sparse_mean_variance_axis.register(sparse.csc_matrix)
def sparse_mean_variance_axis_csc(mtx, axis):
    # use csr implementation
    return sparse_mean_variance_axis(mtx.T, 1 - axis)
```

I’m pretty sure the `csc_matrix.T` returns a `csr_matrix` view of the same data.

ivirshup: Oh, I meant to say a boolean switch stops being sufficient if there are 3 options. `i.e` is sparse matrices could be `scipy`'s CSR and CSC or `pydata/sparse`'s COO. If COO is introduced the logic has to be redone, as opposed to just adding another case.

In general I prefer the original, since the logic is pretty simple. I think [this blog post](https://mrocklin.github.com/blog/work/2019/06/23/avoid-indirection) is a good discussion of this idea, as is the one linked at the bottom. As you can probably tell, I generally like fancy code, but the tool has to fit the problem.

I tried `mtx.T` before, but it was a bit slower. `scipy.sparse` can be slow to construct sparse matrices for some reason (not sure why, since it doesn't do much validation).

Also, would it be `abs(axis - 1)` not `1 - axis`?

flying-sheep: `1 - axis` means `0 if axis is 1 else 1` if axis ∈ {0, 1}.

> I tried mtx.T before, but it was a bit slower.

A pity. In this case we should go for my code, as yours would need a 4th function to prevent code duplication and that’s a lot of functions where 1 would be clean enough:

```py
if isinstance(mtx, sparse.csr_matrix):
    ax_minor = 1
    shape = mtx.shape
elif isinstance(mtx, sparse.csc_matrix):
    ax_minor = 0
    shape = mtx.shape[::-1]
else:
    raise ValueError(
       "This function only works on sparse csr and csc matrices"
    )
if axis == ax_minor:
    return sparse_mean_var_major_axis(
        mtx.data, mtx.indices, mtx.indptr, *shape, np.float64
    )
else:
    return sparse_mean_var_minor_axis(
        mtx.data, mtx.indices, *shape, np.float64
    )
```

The only confusing part is `if axis == ax_minor: ..._major_axis()`, but it’s correct: For `csr_matrices`, axis 0 is the major axis and axis 1 is the minor axis.

ivirshup: Yeah, it's always get's me turned around a bit that `axis` is the one being reduced, so you get the means for the other axes. I think referring to major and minor is a better convention than the original sklearn names though.

flying-sheep (approved):

PR #858, opened 2019-10-01
update ValueError message in pca

fbnrst:
replaced the deprecated filter_gene_dispersion with highly_variable_genes


flying-sheep:
Thank you!

PR #861, opened 2019-10-07
update bbknn reference

ktpolanski:
BBKNN came out as an actual paper a couple of months ago, updating the reference to point to it.


flying-sheep:
Right. Thank you!

PR #862, opened 2019-10-07
Add trimap to scanpy

eamid:
TriMap: Large-scale Dimensionality Reduction Using Triplets

TriMap is a dimensionality reduction method that uses triplet constraints to form a low-dimensional embedding of a set of points. The triplet constraints are of the form "point i is closer to point j than point k". The triplets are sampled from the high-dimensional representation of the points and a weighting scheme is used to reflect the importance of each triplet.

TriMap provides a significantly better global view of the data than the other dimensionality reduction methods such t-SNE, LargeVis, and UMAP. The global structure includes relative distances of the clusters, multiple scales in the data, and the existence of possible outliers. We define a global score to quantify the quality of an embedding in reflecting the global structure of the data. Also, TriMap is significantly faster than t-SNE, LargeVis, and UMAP and scales better to large datasets.

paper: https://arxiv.org/abs/1910.00204

code: https://github.com/eamid/trimap


fidelram:
Thanks for this contribution! Can you look at the travis errors? See https://github.com/theislab/scanpy/pull/797#issuecomment-536861482 for background.

To fix formatting errors you can use https://github.com/psf/black

Furthermore, can you add a small example to the docstring? (see for example: https://github.com/theislab/scanpy/blob/master/scanpy/external/tl/_palantir.py)

LuckyMD:
I guess this should probably be added to scanpy external as other external packages are?

fidelram:
@LuckyMD I agree.

eamid:
Thanks for getting back to me. I fixed the formatting errors and moved trimap to scanpy external. 
trimap is no longer imported by default, so the overall import time is unaffected. 
I also added an example to the docstring.

Please let me know if further fixes are required.

eamid:
I fixed the merge  conflicts, it should be good to go!

eamid:
I fixed the merge conflicts. Hope it is all good now!
Please let me know if you need more changes.

scanpy/external/exporting.py
```diff
@@ -450,7 +450,7 @@ def cellbrowser(
         Coordinates missing from `adata` are skipped.
         By default (or when specifying `'all'` or `None`), these keys are tried:
         [`"tsne"`, `"umap"`, `"pagaFa"`, `"pagaFr"`, `"pagaUmap"`, `"phate"`,
-        `"fa"`, `"fr"`, `"kk"`, `"drl"`, `"rt"`].
+        `"fa"`, `"fr"`, `"kk"`, `"drl"`, `"rt"`, `"trimap"`].
```
flying-sheep, line 453: Nice catch. @theislab/scanpy: we should create this string using code, not by requiring people to find and update it.

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,148 @@
...
+    n_outliers=5,
+    n_random=5,
+    metric='euclidean',
+    weight_adj=500.0,
+    lr=1000.0,
+    n_iters=400,
+    init_pos=None,
+    verbose=False,
+    copy=False,
+):
```
flying-sheep: Please add type hints here. Check other methods for how to do `metric` using `compat.Literal`.

Don’t forget the return type!

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,148 @@
...
+     author = {{Amid}, E. and {Warmuth}, M. K.},
+     title = "{TriMap: Large-scale Dimensionality Reduction Using Triplets}",
+     journal = {ArXiv e-prints},
+     archivePrefix = "arXiv",
+     eprint = {1910.00204},
+     year = 2019,
+    }
+
+    Input
+    ------
```
flying-sheep: Please call this section `Parameters`, use numpydoc style and get rid of “default” (it’s added automatically)

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,148 @@
...
+
+    lr: Learning rate (default = 1000.0)
+
+    n_iters: Number of iterations (default = 400)
+
+    knn_tuple: Use the pre-computed nearest-neighbors information in form of a
+    tuple (knn_nbrs, knn_distances) (default = None)
+
+    apply_pca: Apply PCA to reduce the dimensions to 100 if necessary before the
+    nearest-neighbor calculation (default = True)
```
flying-sheep: I think the default should be `None`, which means “True, but throw a warning if we have to run sc.pp.pca”. Also the parameter seems to be missing?

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,148 @@
...
+
+    opt_method: Optimization method ('sd': steepest descent,  'momentum': GD
+    with momentum, 'dbd': GD with momentum delta-bar-delta (default))
```
flying-sheep: I don’t see this in the actual parameters, where is it?

Please sort the parameters like they appear in the function and make sure every documented one is also present and vice versa.

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,148 @@
...
+
+    verbose: Print the progress report (default = True)
```
flying-sheep: Please derive this from `sc.settings.verbosity` by default. (`None` by default)

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,148 @@
...
+
+    weight_adj: Adjusting the weights using a non-linear transformation
+    (default = 500.0)
+
+    return_seq: Return the sequence of maps recorded every 10 iterations
+    (default = False)
```
flying-sheep: This parameter is also absent. If you conditionally want to return a non-AnnData, please replace “copy” with “inplace” (the semantics are “inplace=True” → modify “adata”, “inplace=False” → return a different data structure)

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,148 @@
+from ... import logging as logg
+import numpy as np
+import scipy.sparse as scp
+
+
+def trimap(
+    adata,
+    n_dims=2,
```
flying-sheep: Please add a `*,` after `adata,` or `n_dims: int = 2,`.

This way we’re free to reorganize subsequent parameters without it being a breaking change.

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,148 @@
...
+        if scp.issparse(X):
+            raise ValueError(
+                'trimap currently does not support sparse matrices. please'
+                'use a dense matrix or apply pca first.'
+            )
+    if metric not in ['euclidean', 'manhattan', 'angular', 'hamming']:
+        raise ValueError(
+            'metric \'' + metric + '\' is currently not supported '
+            'by trimap. acceptable values are \'euclidean\', \'manhattan\', '
+            '\'angular\', and \'hamming\'.'
```
flying-sheep: please avoid unnecessary escaping and prefer f-strings to concatenating strings with `+`:

```py
f'Metric {metric!r} is currently not supported by trimap. '
'Acceptable values are {", ".join(map(repr, allowed_metrics))}'
```

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,148 @@
...
+        distance=metric,
+        weight_adj=weight_adj,
+        n_iters=n_iters,
+        verbose=verbose,
+    ).fit_transform(X, init=init_coords)
+    adata.obsm['X_trimap'] = X_trimap
+    logg.info(
+        '    finished',
+        time=start,
+        deep=('added\n' "    'X_trimap', TriMap coordinates (adata.obsm)"),
```
flying-sheep:
```suggestion
        deep="added\n    'X_trimap', TriMap coordinates (adata.obsm)",
```

flying-sheep (changes requested):
Looks good apart from some details, and I liked the method when it saw it being presented a few months ago.

eamid:
Thanks @flying-sheep for the thorough feedback! I made the changes. There is still a Travis CI error about slow_to_import modules. Since trimap is now in external, I am now sure how this test is being affected.

flying-sheep:
Very weird. Importing `scipy.sparse` shouldn’t import `scipy.stats`, and you didn’t add any other imports. Could be that this is a change in some updated version of something.

Does the test fail for you? Then you could find out how it gets imported using `import-profiler`

eamid:
Thanks @flying-sheep! I haven't added any new imports other those for type hints and copying anndata: [https://github.com/eamid/scanpy/compare/9ae6c19...b7ed705](https://github.com/eamid/scanpy/compare/9ae6c19...b7ed705)

My previous commit passed the tests. Could this be because of recent commits on your side?

gokceneraslan:
I have the same issue here https://github.com/theislab/scanpy/pull/945, so it's not specific to this PR,fyi.

flying-sheep:
To help debug it, can one of you do what I suggested? I don’t have the problem locally and if one of you does see it, it would be much easier to find out which dependency started to import that module.

flying-sheep (changes requested):
The PR looks really clean now, congrats!

One thing left I just realized: the bibtex code for the reference will look pitiful when rendered as reStructuredText. Can you please add yourself to `references.rst` and just cite yourself as `[Amid19]_` in the docstring instead of using that “Reference” section?

eamid:
Thank you @flying-sheep! All done.

eamid:
@flying-sheep just following up on the PR. Thanks!

scanpy/external/tl/_trimap.py
```diff
@@ -0,0 +1,129 @@
...
+        If `True`, print the progress report.
+        If `None`, `sc.settings.verbosity` is used.
+    copy
+        Return a copy instead of writing to `adata`.
+
+    Returns
+    -------
+    Depending on `copy`, returns or updates `adata` with the following fields.
+
+    **X_trimap** : `np.ndarray`, (`adata.obs`, shape=[n_samples, n_components], dtype `float`)
```
flying-sheep:
```suggestion
    **X_trimap** : :class:`~numpy.ndarray`, (:attr:`~anndata.AnnData.obsm`, shape=(n_samples, n_components), dtype `float`)
```

One typo I missed: `.obs` → `.obsm`.

flying-sheep:
Great, thank you!

eamid:
Thank you @flying-sheep!

PR #865, opened 2019-10-08
dtype fixes for downsample and normalization

ivirshup:
Addressing https://github.com/theislab/scanpy/issues/435#issuecomment-538776417

This PR does two things:

1. `downsample_counts` will convert the resulting downsampled matrix back to the initial dtype by default.
2. `normalize_total` will now work with integer matrices.

I think 2 should definitely be the case. 1 does have a performance cost, but it's close to @falexwolf's [suggestion](https://github.com/theislab/scanpy/issues/435#issuecomment-475999342) and removes a minor foot-gun.

### DIFF
--- a/scanpy/preprocessing/_normalization.py
+++ b/scanpy/preprocessing/_normalization.py
@@ -11,6 +11,8 @@ from .._compat import Literal
 
 def _normalize_data(X, counts, after=None, copy=False):
     X = X.copy() if copy else X
+    if issubclass(X.dtype.type, (int, np.integer)):
+        X = X.astype(np.float32)  # TODO: Check if float64 should be used
     after = np.median(counts[counts>0]) if after is None else after
     counts += (counts == 0)
     counts = counts / after
@@ -182,10 +184,7 @@ def normalize_total(
         layer = adata.layers[layer_name]
         counts = np.ravel(layer.sum(1))
         if inplace:
-            if hasattr(layer, '__itruediv__'):
-                _normalize_data(layer, counts, after)
-            else:
-                adata.layers[layer_name] = _normalize_data(layer, counts, after, copy=True)
+            adata.layers[layer_name] = _normalize_data(layer, counts, after)
         else:
             dat[layer_name] = _normalize_data(layer, counts, after, copy=True)
 
--- a/scanpy/preprocessing/_normalization.py
+++ b/scanpy/preprocessing/_normalization.py
@@ -13,7 +13,8 @@ def _normalize_data(X, counts, after=None, copy=False):
     X = X.copy() if copy else X
     if issubclass(X.dtype.type, (int, np.integer)):
         X = X.astype(np.float32)  # TODO: Check if float64 should be used
-    after = np.median(counts[counts>0]) if after is None else after
+    counts = np.asarray(counts)  # dask doesn't do medians
+    after = np.median(counts[counts>0], axis=0) if after is None else after
     counts += (counts == 0)
     counts = counts / after
     if issparse(X):
--- a/scanpy/preprocessing/_normalization.py
+++ b/scanpy/preprocessing/_normalization.py
@@ -13,12 +13,12 @@ def _normalize_data(X, counts, after=None, copy=False):
     X = X.copy() if copy else X
     after = np.median(counts[counts>0]) if after is None else after
     counts += (counts == 0)
-    counts /= after
+    counts = counts / after
     if issparse(X):
         sparsefuncs.inplace_row_scale(X, 1/counts)
     else:
-        X /= counts[:, None]
-    return X if copy else None
+        np.divide(X, counts[:, None], out=X)
+    return X
 
 
 def normalize_total(
@@ -170,10 +170,7 @@ def normalize_total(
     if inplace:
         if key_added is not None:
             adata.obs[key_added] = counts_per_cell
-        if hasattr(adata.X, '__itruediv__'):
-            _normalize_data(adata.X, counts_per_cell, target_sum)
-        else:
-            adata.X = _normalize_data(adata.X, counts_per_cell, target_sum, copy=True)
+        adata.X = _normalize_data(adata.X, counts_per_cell, target_sum)
     else:
         # not recarray because need to support sparse
         dat = dict(
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -983,8 +983,10 @@ def downsample_counts(
     adata: AnnData,
     counts_per_cell: Optional[Union[int, Collection[int]]] = None,
     total_counts: Optional[int] = None,
+    *,
     random_state: Optional[Union[int, RandomState]] = 0,
     replace: bool = False,
+    dtype: Optional[np.dtype] = None,
     copy: bool = False,
 ) -> Optional[AnnData]:
     """\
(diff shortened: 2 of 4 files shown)


ivirshup:
I'm thinking about reverting the return dtype part, but just keeping having `normalize_total` to work with more dtypes. Thoughts @flying-sheep?

scanpy/preprocessing/_normalization.py
```diff
@@ -11,14 +11,16 @@
...
     after = np.median(counts[counts>0]) if after is None else after
     counts += (counts == 0)
-    counts /= after
+    counts = counts / after
     if issparse(X):
         sparsefuncs.inplace_row_scale(X, 1/counts)
     else:
-        X /= counts[:, None]
-    return X if copy else None
+        np.divide(X, counts[:, None], out=X)
```
flying-sheep, line 23: Why is this and the above change necessary? I though that if `__itruediv__` is not available, python falls back to `__truediv__` anyway?

flying-sheep (commented):
I need more info: why did you do the changes you made in `_normalize_data` and `normalize_total`?

And what are the arguments for/against maintaining the dtype?

scanpy/preprocessing/_normalization.py
ivirshup: Sorry for the late response on this, totally forgot about it.

The conversion of `X` to float happens because you can't do an inplace dtype conversion. For example:

```
>>> a = np.ones(3, dtype=int)
>>> np.divide(a, 2, out=a)
TypeError: No loop matching the specified signature and casting was found for ufunc true_divide
```

This is also the reason behind changing `count /= after` with `counts = counts / after`. If `counts` was generated by `X.sum`, it could be an integer array. Which would do:

```
>>> a /= 2
TypeError: No loop matching the specified signature and casting was found for ufunc true_divide
```

Finally, I replaced `X /= counts[:, None]` with `np.divide(X, counts[:, None], out=X)` because if X was dask array `__itruediv__` doesn't behave like it does for numpy arrays https://github.com/dask/dask/issues/5199.

ivirshup:
Arguments for and against converting values in `downsample_counts`:

If we don't convert dtypes back to what they originally were, there's a slight performance boost since we don't have to have two copies.

I we return an array of integers we run into trouble downstream with functions that aren't tested with integer arrays. Issues from this have been opened a few times, so when I wrote this I thought it might be worth just maintaining the input type. I'm not sure I agree with that now.

scanpy/preprocessing/_normalization.py
flying-sheep: Thank you! Would be cool to have this documented in a comment

LuckyMD:
> If we return an array of integers we run into trouble downstream with functions that aren't tested with integer arrays. Issues from this have been opened a few times, so when I wrote this I thought it might be worth just maintaining the input type. I'm not sure I agree with that now.

This is quite a compelling argument for me (as I was one of the people who reported an issue like this). If an integer matrix is generally returned, then one would have to ensure all other functions will work with this data type as intended (sc.pp.log1p for example). Otherwise this would be backward-breaking.

GMaciag:
I would just like to add that the issue with `normalize_total` can arise not only from the `downsample_counts` function. 

In my case, I am working on `.loom` files generated with *velocyto* - I want to be able to estimate RNA velocity in the end. That means I have 'spliced' and 'unspliced' layers in my anndata object. I wanted to use `normalize_total` on all the layers, which should be possible by setting parameter `layers='all'`. However, I was getting a TypeError, as in  #435 

The workaround described at the end of that issue solved it for me. My point is just that fixing only `downsample_counts` is not enough and functions that work on layers should accept integer data. I think that my case is not that uncommon and will happen more often as people use scVelo with velocyto pipelines. 

Alternatively, you should warn people about it in the tutorials and make them convert everything to float.

flying-sheep:
We should definitely maintain the type in layers, and that means maintaining the type in .X makes sense too.

We should also take care not to downcast more incompatible types: int32 can be expressed as float64, but not in float32. int64 has to stay int64.

ivirshup:
Hah, I've gotten much better at numba since I wrote this function. I figured out I can just get the core part to work on floats and don't have to worry about casting between types. Makes this a much easier decision. Now floats aren't converted to integers in the first place.

> We should also take care not to downcast more incompatible types: int32 can be expressed as float64, but not in float32. int64 has to stay int64.

I think we can be a little flexible on this, and just generally follow numpy promotion rules (except for when they're bad).

flying-sheep:
[these ones](https://docs.scipy.org/doc/numpy/reference/generated/numpy.result_type.html). But you still convert ints to floats? And when are they bad?

ivirshup:
> But you still convert ints to floats?

`downsample_counts` now doesn't convert anything, or even copy if you don't want it to. It just modifies the passed matrix (or copy, if `copy=True`) whether it has integer or floating point values.

> And when are they bad?

This one is pretty weird:

```python
np.result_type(np.int64, np.uint64)
# dtype('float64')
```

But also numpy does promotion (or some other conversion, not quite sure for this one) when it probably shouldn't:

```python
type(np.uint8(255) << 1)
# numpy.int64
```

flying-sheep:
> This one is pretty weird:
> 
> ```py
> np.result_type(np.int64, np.uint64)
> # dtype('float64')
> ```

That’s also a bug, the following code changes the first integer (+=1) and completely destroys the second one, without need.

```py
>>> infos = [np.iinfo(dt).max for dt in [np.int64, np.uint64]]
>>> infos
[9223372036854775807, 18446744073709551615]
>>> np.array(infos)
array([9.22337204e+18, 1.84467441e+19]
>>> np.array(infos).astype(np.uint64)
array([9223372036854775808, 0], dtype=uint64)
```

scanpy/preprocessing/_normalization.py
```diff
@@ -11,14 +11,17 @@
 
 def _normalize_data(X, counts, after=None, copy=False):
     X = X.copy() if copy else X
-    after = np.median(counts[counts>0]) if after is None else after
+    if issubclass(X.dtype.type, (int, np.integer)):
+        X = X.astype(np.float32)  # TODO: Check if float64 should be used
```
flying-sheep, line 15: yeah, we should at least implement that TODO, leaking precision shouldn’t be done willy-nilly.

Unless that’s a huge performance cost of course.

flying-sheep (changes requested):

scanpy/preprocessing/_normalization.py
ivirshup: Oh, by check I meant do a benchmarking and seeing what range of values would be required for imprecision. Looking at the functions we're calling, it's all numpy and nothing is iterative on our end, so I'm not too worried about it. I think it'll be no worse than our default behavior.

What were you thinking for implementation? int64 -> float64 or something like that?

flying-sheep: Everything is simplified by the fact that 1) scRNASeq data is pretty robust to rounding [citation needed] and 2) neither counts nor any normalization we do go far up. it still feels weird:

- float32 can represent ints <= 16777216 precisely
- int64 can represent ints <= 9223372036854775807

I’d sleep more easily if I knew what order of magnitude the highest number has that exists at any point in scanpy (e.g. is there some point where we take count data to the power of something and save it)

flying-sheep (approved):

PR #869, opened 2019-10-10
Fix pl paga when adata.uns color list was not set

fidelram:
`sc.pl.paga` was throwing and error when the color list in `.uns` was not previously set. While fixing the error I realized that some functionality was duplicated between legacy scatter plots and the embedding plots and removed the code duplication.

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -1982,7 +1982,7 @@ def tracksplot(
 
     # get categories colors:
     if groupby + "_colors" not in adata.uns:
-        from ._tools.scatterplots import _set_default_colors_for_categorical_obs
+        from ._utils import _set_default_colors_for_categorical_obs
         _set_default_colors_for_categorical_obs(adata, groupby)
 
     groupby_colors = adata.uns[groupby + "_colors"]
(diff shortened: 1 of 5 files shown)


flying-sheep:
Looks good, but I can’t really see your changes in the diff because you moved the functions. Can you paste a diff of just the function code as a comment?

fidelram:
Here is the summary:
 * `_set_default_colors_for_categorical_obs()` moved from scatterplots to _utils. No modifications
 * `_set_colors_for_categorical_obs()` moved from scatterplots to _utils. No modifications
 * `_validate_palette()` function added, extracting the relevant code from `scatterplots.py:_get_color_values()`
 * `adjust_palette()` was removed as this functionality is replicated to some extent by   `_set_default_colors_for_categorical_obs()`
 * `add_colors_for_categorical_sample_annotation()` was simplified by using the above functions.

FYI: the error from `sc.pl.paga` that I had was caused because the expected output of `adjust_palette()` was `Cycler` but the function actually returned the original type of `palette` which could be `ListedColorMap`, `cabc.Sequence` or `Cycler`.

flying-sheep:
OK, sounds good!

The problems codacy points out are probably in old, moved code and therefore not yours.

Issue #891, opened 2019-10-24
Plotting order in embedding plots with groups parameter 


gokceneraslan:
When we use the `groups` parameter of `sc.pl.umap` (and other embedding plots), plotting order of groups is random. This means that gray background group can be plotted over the groups of interest and obscure dots of other colors.

Here the top 2 plots are plotted with `groups=` and bottom ones are plotted manually via calling sc.pl.umap twice with subsetted anndata (e.g. adata[adata.obs.x != 'y'] and adata[adata.obs.x == 'y']):

![image](https://user-images.githubusercontent.com/1140359/67532602-da535d80-f694-11e9-9ed2-3d544013ff78.png)

LuckyMD:
Wasn’t there recently a PR to introduce ordered categoricals? Or does that only affect legend order and not order of plotting?

fidelram:
@LuckyMD The ordered categoricals with not solve the problem. The issue is with the order of the dots.  I will take a look.

LuckyMD:
I see. Would it be a solution to link the order of plotting to the order of categories when categoricals are ordered?

fidelram:
The solution that I proposed is that if 'groups' is set, then the dots corresponding to the selected groups are put on top. But also, sorting the cells according to the categories order should work.

elo073:
On similar issue: is there a way to plot the cells in random order? So that no specific group would be on top, but randomly mixed.

ivirshup:
You can pass `sort_order=False`, but then it should just be the order in AnnData object.

elo073:
Thank you, that could be helpful. Then I'd have to shuffle the order of the cells in the AnnData object first. Is there any easy way to do this?

ivirshup:
I can think of a couple ways:

```python
view = adata[np.random.choice(adata.n_obs, adata.n_obs, replace=False)]

new = sc.pp.subsample(adata, fraction=1., copy=True)
```

elo073:
Thank you! This works
```
new = sc.pp.subsample(adata, fraction=1., copy=True)
sc.pl.umap(new, color=['scsn'], legend_loc='on data',  size=10, sort_order=False, legend_fontsize=15)
```

leezx:
We can sort the adata itself. Like:
```adata = adata[adata.obs.group.sort_values().index,:]```
This can manually change the order of groups in a graph.

Issue #890, opened 2019-10-24
Reorder Stacked_Violin Subplots


MediciPrime:
Hello @fidelram,

This is not an issue but I would really like some help with using the 'order' parameter in the stacked_violin plot function.  If I have cell type as the y-axis and the genes in the x-axis, how would I rearrange the order to the cell types in the y-axis.  

Providing the 'order' parameter with a list of cell types being used by the function doesn't seem to be working.  I am just really confused how else to interpret these instructions for the stacked_violin function:  order : list of str, optional (default: True)  

Thank you,
Behram

fidelram:
I just tried and worked for me:

Default order:
![image](https://user-images.githubusercontent.com/4964309/67553457-7d47be00-f70d-11e9-9381-b323cc25a739.png)

Order changed: 
![image](https://user-images.githubusercontent.com/4964309/67553505-a0726d80-f70d-11e9-8a35-fe3ded5780ea.png)

If this is not working for you, please check that the old order and the new order has the same elements as this may be the problem.

FYI: The order of the cell type is given by the categorical order in `adata.obs['<column name>'] `. Thus, another way to change the order is to change the pandas categorical order was you want it. 

See this link on how to do it: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.CategoricalIndex.reorder_categories.html

MediciPrime:
Thank you @fidelram!  I guess for me it wasn't working because I wasn't using the 'swap_axes=True' command so maybe the function didn't know what I wanted it to reorder.  However after using 'swap_axes' it started working.  Now I still wanted the cell types to be along the y-axis so I tried your second suggestion for reordering the pandas categorical order like so:

`g0_g1_ctrl_ifn.obs['cluster_id'].cat.reorder_categories(['Glomerular Epithelial Cells', 'Early GEC', 'Proximal and Distal Tubule',
                                                         'Endothelial Cells', 'Cycling Cells', 'Mesenchyme', 'Neuron'], inplace=True)`

And it worked!  Thank you again for your very helpful suggestions!

- Behram

fidelram:
@MediciPrime I am also updating the underlying plotting code to fix the problem with `swap_axes=True`. After this is merged to master, you would not need to do the reordering of the categories.

MediciPrime:
That's awesome thank you for fixing it so quickly @fidelram!

PR #893, opened 2019-10-25
Fix plot issues

fidelram:
Fix issue related to #890 and the reordering of stacked_violin_plots when `swap_axes=False` and for the case when `dendrogram` and `order` are given, in which case dendrogram order takes precedence. 

Fix issue #891, now if `groups` is set for an embedding, the cells belonging to those groups are plotted on top.

I also added missing documentation, pointing that the marker size could be a list of per-cell size. 
Added a new plotting test for the groups option.


flying-sheep:
Great! Can’t say I understand the #890-related fix though: What went wrong before? Is there a test for it?

Please use 4 space indentation, not visual indentation. Basically, running `black` on the any newly changed code should yield minimal changes. Feel free to remove a file where you changed a lot from here:

https://github.com/theislab/scanpy/blob/b3933ac185f9af3908261e939fc5df2336f1932e/pyproject.toml#L9-L13

Then everything will be done automatically. You’ll just have to go through the changes and fix ugly ones like black making `some = code[:]  # comment` into `some = (\n    code\n)  # comment` instead of `#comment\nsome = code`.

fidelram:
@flying-sheep For #890 probably we had never tested that combination of parameters because the output was a broking image. 

If I understand you correctly, black can by applied to only some lines? Apparently PyCharm can be used with black, do you have any experience?

gokceneraslan:
One question related to #891, is there any plotting order for continuous values like higher expression plotted on top? That's more controversial but sometimes dropout et al might obscure cells expressing a gene. What do you think?

fidelram:
@gokceneraslan by default, higher values of the gene/obs column you are plotting are put on top. You can change this behaviour by setting `sort_order=False`. 

Note: The default behaviour can be misleading and give you the impression that for example one cluster is expressing certain gene, when in reality only a small fraction of cells do express the gene, but because they are all plot on top, they seem to be more.

flying-sheep:
> probably we had never tested that combination of parameters because the output was a broking image.

got it!

> If I understand you correctly, black can by applied to only some lines? Apparently PyCharm can be used with black, do you have any experience?

AFAIK black can’t be applied to some lines only. Thus my suggestion to just run it on files where you changed a lot. [about pycharm](https://black.readthedocs.io/en/stable/editor_integration.html)

flying-sheep:
That’s a lot of files reformatted! I didn’t blacken ` scanpy/plotting/_anndata.py` yet because people were working on it. If currently nobody is working on it, we can go ahead.

scanpy/tests/blackdiff.py
```diff
@@ -45,8 +45,9 @@ def black_diff(src: Path, differ=Differ()) -> Tuple[int, int]:
     ]
     for file, (added, removed) in sorted(excluded, key=lambda sd: -sum(sd[1])):
         print(f'{file}: +{added} -{removed}')
-        if added + removed < thresh:
-            print('File has < 10 changes to black formatting. Do it!')
+        if 0 < added + removed < thresh:
+            print(f'File: {file}: has < {threshold} changes to '
+                  f'black formatting. Do it!')
```
flying-sheep: No, if it’s 0, it should be removed from `tool.black.exclude` in `pyproject.toml`.

scanpy/tests/blackdiff.py
```diff
@@ -45,8 +45,11 @@ def black_diff(src: Path, differ=Differ()) -> Tuple[int, int]:
     ]
     for file, (added, removed) in sorted(excluded, key=lambda sd: -sum(sd[1])):
         print(f'{file}: +{added} -{removed}')
-        if added + removed < thresh:
-            print('File has < 10 changes to black formatting. Do it!')
+        if 0 < added + removed < thresh:
+            print(
+                f'File: {file}: has < {threshold} changes to '
+                f'black formatting. Do it!'
+            )
```
flying-sheep: No, if it’s 0, it should be removed from `tool.black.exclude` in `pyproject.toml`. Good idea to change the message in that case though. It should be:

```py
if added + removed == thresh:
    print('This file has 0 changes to black formatting, good job! Please remove it from “tool.black.exclude”')
elif added + removed < thresh:
    print('This file has < {thresh} changes to black formatting. Do it!')
```

Also no need to repeat the file name, it’s the one from the line above.

flying-sheep (changes requested):

scanpy/tests/blackdiff.py
fidelram: Added it because it was not obvious. I didn't know what was happening until I dig in the code and realized that the file before was the one throwing the error. At first I thought the code checks all files and then gives a report.

scanpy/tests/blackdiff.py
```diff
@@ -45,8 +45,17 @@ def black_diff(src: Path, differ=Differ()) -> Tuple[int, int]:
     ]
     for file, (added, removed) in sorted(excluded, key=lambda sd: -sum(sd[1])):
         print(f'{file}: +{added} -{removed}')
-        if added + removed < thresh:
-            print('File has < 10 changes to black formatting. Do it!')
+        if added == removed == 0:
+            print(
+                'This above file has 0 changes to black formatting, good job! '
+                'Please remove it from “tool.black.exclude" in pyproject.toml'
+            )
```
flying-sheep, line 52:
```suggestion
            )
            sys.exit(1)
```

flying-sheep (approved):

PR #896, opened 2019-10-29
Update MAGIC API and docs

scottgigante:
The MAGIC docs and API are a little out of date.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -24,3 +24,4 @@ importlib_metadata; python_version < '3.8'
 typing_extensions; python_version < '3.8'
 setuptools_scm
 umap-learn
+legacy-api-wrap
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ natsort
 joblib
 numba>=0.41.0
 umap-learn>=0.3.0
+legacy-api-wrap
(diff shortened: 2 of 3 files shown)


flying-sheep:
Thank you! As it stands, this destroys backwards compatibility. But fear not:

1. Add [`legacy-api-wrap`](https://pypi.org/project/legacy-api-wrap/) to requirements.txt and docs/requirements.txt
2. Use it to change the API to 1-2 positional parameters (i.e. either `def magic(adata: ..., *, name_list: ... = None, ...)` or with the `*` one parameter to the right)
3. While you’re at it, change `name_list`’s type to `Union[Literal['all_genes', 'pca_only'], Sequence[str], None]`. You can do `from .._compat import Literal` or so (maybe a dot less or more)

scottgigante:
Wow, I learned something new! Thanks!

flying-sheep:
Looks good, just one style thing remaining! You can just do `python -m black path/to/file.py` instead of doing things manually

flying-sheep:
Thank you!

scottgigante:
Thanks @flying-sheep ! I've also started using `black` in my own code now. You might be interested in this - I've added a git hook to automatically format any committed `.py` files with `black`.

https://github.com/KrishnaswamyLab/scprep/blob/master/autoblack.sh

flying-sheep:
Thanks, I use it [this way](https://black.readthedocs.io/en/stable/version_control_integration.html).

scottgigante:
That's much nicer, thanks!

PR #898, opened 2019-10-29
Inherit main requirements

flying-sheep:
To not repeat ourselves @ivirshup (I think) suggested this. Let’s see if readthedocs supports this. If so, this should soon be visible: https://icb-scanpy.readthedocs-hosted.com/en/inherit-requirements/


flying-sheep:
Seems to work!

PR #899, opened 2019-10-30
add 'layer' option to sc.pl.violin

fidelram:
I just discovered that this was missing


flying-sheep:
Thank you, judging from the test pics, those look like very useful changes!

PR #902, opened 2019-11-05
Allow specification of mitochondrial chromosome when retrieving mitochondrial genes

ghuls:
Allow specification of mitochondrial chromosome when retrieving mitochondrial
genes as some organisms use a different chromosome name than "MT".


flying-sheep:
Thank you for your contribution!

The function is already called `mitochondrial_genes`, so I think the parameter is a bit much to type. Why not just `chromosome_name` or even `chromosome`?

ghuls:
It is using `chromosome` now.

flying-sheep:
Great, thank you!

PR #903, opened 2019-11-06
Adding the Self-Assembling Manifold algorithm to scanpy

atarashansky:
This is a pull request to integrate the Self-Assembling Manifold (SAM) algorithm into scanpy.

A brief summary of the method:
SAM iteratively rescales the expressions of genes based on their spatial variability along the intrinsic manifold of the data. Extensive benchmarking has shown this approach to improve dimensionality reduction and feature selection for both 'easy' and 'challenging' datasets. SAM is especially powerful when analyzing datasets with only subtle differences in gene expression between cell types. More information can be found in the eLife publication: https://elifesciences.org/articles/48994

I still need to write the test script as well as ensure that the added code follows the BLACK coding style. Please let me know if there are any other issues I should fix prior to merging.


flying-sheep:
Hi! We didn't look at the method yet, but superficially:

- uppercase letters are only allowed in CONSTANTS and ClassNames, not module names, parameter names or function names
- preprocessing goes into sce.pp. This way there's way less keyword arguments and no need to prefix them. If the preprocessed data isn't useful for other applications, you can put it into .layers
- Scanpy functions should either mutate or return something, depending on the value of the `inplace` parameter

atarashansky:
Thanks for taking a look! I'll fix those issues and push the changes soon.

atarashansky:
Hi @flying-sheep , I fixed the issues you pointed out.

Please let me know if there are any other issues I should address. Thanks!

atarashansky:
Whoops, I had forgotten to format with BLACK so the checks didn't pass. Should be good now, @flying-sheep

atarashansky:
Any chance I could get someone to take a look at this pull request? Thanks!

flying-sheep:
Hi, sorry for the silence, I’m doing my PhD thesis right now. Can you please format it so it passes CI? `black .` should be sufficient.

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+
+    Parameters
+    ----------
+
+    k - int, optional, default 20
+        The number of nearest neighbors to identify for each cell.
+
+    distance : string, optional, default 'correlation'
+        The distance metric to use when identifying nearest neighbors.
+        Can be any of the distance metrics supported by sklearn's 'pdist'.
```
flying-sheep:
```suggestion
        Can be any of the distance metrics supported by
        :func:`~scipy.spatial.distance.pdist`.
```

It’s actually scipy, not sklearn.

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+
+    max_iter - int, optional, default 10
```
flying-sheep:
```suggestion
    max_iter
```

Please remove all the types and defaults from the docstring, they’re added automatically

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+from ... import logging as logg
+
+
+def sam(
+    adata: AnnData,
+    max_iter: int = 10,
+    num_norm_avg: int = 50,
+    k: int = 20,
+    distance: str = 'correlation',
+    standardization: Optional[str] = 'Normalizer',
```
flying-sheep:
```suggestion
    standardization: Literal['Normalizer', 'StandardScaler'] = 'Normalizer',
```

Please do `from ..._compat import Literal` or so to get this.

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+    weight_pcs: bool = True,
+    npcs: Optional[int] = None,
+    n_genes: Optional[int] = None,
+    projection: Optional[str] = 'umap',
```
flying-sheep:
```suggestion
    projection: Literal['umap', 'tsne'] = 'umap',
```

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+    inplace: bool = True,
+    verbose: bool = True,
+) -> Optional[AnnData]:
+    """Self-Assembling Manifolds single-cell RNA sequencing analysis tool.
```
flying-sheep:
```suggestion
    """\
    Self-Assembling Manifolds single-cell RNA sequencing analysis tool.
```

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+        genes with large spatial dispersions from skewing the distribution
+        of weights.
+
+    weight_pcs - bool, optional, default True
+        If True, scale the principal components by their eigenvalues. In
+        datasets with many expected cell types, setting this to False might
+        improve the resolution as these cell types might be encoded by low-
+        variance principal components.
+
+    npcs - int, optional, default None,
```
flying-sheep: Please call this `n_pcs`. You use `n_genes` below and [NPCs](https://en.wikipedia.org/wiki/Non-player_character) are something else :wink:

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+        distance=distance,
+        preprocessing=standardization,
+        weight_PCs=weight_pcs,
+        npcs=npcs,
+        n_genes=n_genes,
+        projection=projection,
+        verbose=verbose,
+    )
+
+    return (s, adata) if inplace else (s, s.adata)
```
flying-sheep: The return type (`Optional[AnnData]`) doesn’t match. I think the best would probably be a return type of `Union[SAM, AnnData]`:

```suggestion
    return s if inplace else s.adata
```

People expect `new_adata = sc.*tl.*(adata, inplace=False)` to work.

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+        `.var['spatial_dispersions']`
+            Spatial dispersions for each gene (these are used to compute the
+            SAM weights)
+        `.var['mask_genes']`
+            If preprocessed with SAM, this boolean vector indicates which genes
+            were filtered out (=False).
+        `.uns['preprocess_args']`
+            Dictionary of parameters used for preprocessing.
+        `.uns['run_args']`
+            Dictionary of parameters used for running SAM.
```
flying-sheep: These (and quite a few others) are pretty generic. SAM-specific stuff stuff should go into `uns['sam']` instead or you should name it in a more SAM-specific way.

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+        `.uns['pca_obj']`
+            The sklearn.decomposition.PCA object.
```
flying-sheep: Hmm, is this serializable to disk? Will this work?

```py
sce.tl.sam(adata)
adata.write('somefile.h5ad')
```

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+        `.uns['X_processed']`
+            The standardized and SAM-weighted data fed into PCA.
```
flying-sheep: Why `.uns` and not `obsm` or `layers`?

scanpy/external/tl/_sam.py
```diff
@@ -0,0 +1,216 @@
...
+        `.uns['neighbors']`
+            A dictionary with key 'connectivities' containing the kNN adjacency
+            matrix output by SAM. If built-in scanpy dimensionality reduction
+            methods are to be used using the SAM-output AnnData, users
+            should recompute the neighbors using `.obs['X_pca']` with
+            `scanpy.pp.neighbors`.
+        `.uns['ranked_genes']`
+            Gene IDs ranked in descending order by their SAM weights.
```
flying-sheep: Shouldn’t this be `.var['sam_rank']` or so?

flying-sheep (changes requested):

flying-sheep:
Oh, and I forgot: Why not `sc.external.pl.sam(adata)` instead of `sam_object.scatter()`?

scanpy/external/tl/_sam.py
atarashansky: Done!

scanpy/external/tl/_sam.py
atarashansky: Possibly a stupid question: `SAM` is defined within the scope of the function. If I want to set the return type to be `Union[SAM, AnnData]`, should I import SAM in the global frame (at the top of `_sam.py`)?

scanpy/external/tl/_sam.py
atarashansky: Good catch. It used to be serializable to disk, but with the latest updates to AnnData, it no longer is. I've since removed it from the `uns` dictionary.

scanpy/external/tl/_sam.py
atarashansky: Good catch! I had since changed this to be stored in the SAM object itself as opposed to the AnnData. It's a large, dense array and ended up not wanting to make AnnData too bloated

scanpy/external/tl/_sam.py
atarashansky: It's the genes themselves sorted in decreasing order of the SAM weights, but it's not really useful since you can just sort `.var['weights']` to get the same information. I removed it!

This is kind of hacky, but to avoid having to refactor all my pipelines/scripts by changing what gets output into `uns` by `SAM`, I incorporated your suggestions by manually changing `uns` after `s.run(...)`:

```
    s.adata.uns['sam'] = {'run_args' : s.adata.uns['run_args'],
     'preprocess_args' : s.adata.uns['preprocess_args']}

    del s.adata.uns['nnm']
    del s.adata.uns['preprocess_args']
    del s.adata.uns['run_args']
    del s.adata.uns['ranked_genes']
```
Hopefully, this is an acceptable compromise!

atarashansky: This is probably a more elegant way of writing the above:
```
    s.adata.uns['sam'] = {}
    for attr in ['nnm','preprocess_args','run_args','ranked_genes']:
        s.adata.uns['sam'][attr] = s.adata.uns.pop(attr,None)
```

atarashansky:
I added the scatter function to `sc.external.pl.sam` as per your suggestion!

scanpy/external/tl/_sam.py
atarashansky: Stupid of me, bumping the import statement to the global frame throws an ImportError since SAM is not a required package. Not sure how to set the return type to `SAM`, then. I replaced `SAM` as `Any` for the time being.

scanpy/external/pl.py
```diff
@@ -135,3 +137,103 @@ def harmony_timeseries(
         return fig
     elif not show:
         return axes
+
+
+def sam(
+    adata: AnnData,
+    projection: Union[str, np.ndarray] = 'X_umap',
+    c: Optional[Union[str, np.ndarray]] = None,
+    cmap: str = 'rainbow',
```
flying-sheep: Please never use rainbow: https://eagereyes.org/basics/rainbow-color-map

```suggestion
    cmap: str = 'viridis',
```

scanpy/external/__init__.py
```diff
@@ -58,6 +58,7 @@
    tl.phate
    tl.palantir
    tl.trimap
+   tl.sam
```
flying-sheep, line 61: You should also add `pl.sam` to the **Plotting** section

flying-sheep (changes requested):
Nice, almost there!

scanpy/external/tl/_sam.py
flying-sheep: Not a stupid question at all! That’s a complex thing we don’t have an easy solution for. What you can do is something like this:

```py
try:
    from sam import SAM
except ImportError:
    SAM = typing.Any
```

flying-sheep:
Great, thank you!

atarashansky:
Thanks so much for your help!! Good luck with your thesis :)

PR #917, opened 2019-11-13
Allow control of parallel in simplicial_set_embedding

tomwhite:
This should not be merged until Scanpy uses UMAP 0.4 (not yet released).

It allows UMAP to take advantage of multiple cores by setting the random state to `None`:

```python
sc.tl.umap(adata, random_state=None)
```

### DIFF
--- a/scanpy/tools/_umap.py
+++ b/scanpy/tools/_umap.py
@@ -131,6 +131,7 @@ def umap(
     if hasattr(init_coords, "dtype"):
         init_coords = check_array(init_coords, dtype=np.float32, accept_sparse=False)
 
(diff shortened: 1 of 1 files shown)


gokceneraslan:
Although I understand technically the link between the seed and parallelization, from a user experience point of view, this is very unintuitive. How about another parameter about parallelization like `parallel` or `multicore` or so? Scanpy can then print a warning saying that results will not be reproducible.

Zethson:
I agree with @gokceneraslan . Since this PR hasn't seen any movement in 3 years I'll close it for now. Feel free to reopen if you want to discuss this further.

PR #927, opened 2019-11-20
score_genes: use_raw=None instead of False, efficiency updates, prettier logging

falexwolf:

PR #928, opened 2019-11-20
Fixup #927

flying-sheep:


scanpy/tools/_score_genes.py
```diff
@@ -79,7 +79,7 @@ def score_genes(
             genes_to_ignore.append(gene)
     if len(genes_to_ignore) > 0:
         logg.warning(f'genes are not in var_names and ignored: {genes_to_ignore}')
-    gene_list = set(gene_list_in_var[:])
```
flying-sheep, line 112: @fidelram https://github.com/theislab/scanpy/commit/d8350a82a9764cf7b2f5180da594407507855889#diff-c617eda21024eb4afae17f6e3c2660ffR73

Why slice this? `set` isn’t a view class.

fidelram: I think that I was trying to get rid of duplicate gene names.

flying-sheep: No, the question is: why `set(x[:])` and not `set(x)`?

Zethson:
@flying-sheep do you still know the context of this PR? Seems small enough to be able to rebase and merge it eventually. Else, I'd close it.

PR #931, opened 2019-11-23
Add base option to log1p and store it in adata.uns

### LINKED ISSUE #929: log1p with different bases
`log2(TP10K+1)` values are more interpretable than `log(TP10K+1)`, which uses natural logarithm, therefore it'd be great to have an option on the base of the log. It's one of the requests also here #45 .

Since neither of np.log or np.log1p accepts any base arguments(isn't this unbelievable), I did it with simple numba functions here: 

https://nbviewer.ipython.org/gist/gokceneraslan/2744cfeda702fb9b9e48d0216427372c?flush_cache=true

I won't have time to send a PR, but feel free to pursue further, adapt the notebook and merge (if you have time).

PS: Also see https://github.com/numpy/numpy/issues/14969.

gokceneraslan:
WIP.

Fixes #929

### DIFF
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -283,7 +283,7 @@ def log1p(
     -------
     Returns or updates `data`, depending on `copy`.
     """
-    if 'log1p' in adata.uns_keys():
+    if 'log1p' in data.uns_keys():
         logg.warning('adata.X seems to be already log-transformed.')
 
     if copy:
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -301,11 +301,11 @@ def log1p(
         if issparse(X):
             np.log1p(X.data, out=X.data)
             if base is not None:
-                X.data /= np.log(base)
+                np.divide(X.data, np.log(base), out=X.data)
         else:
             np.log1p(X, out=X)
             if base is not None:
-                X /= np.log(base)
+                np.divide(X, np.log(base), out=X)
         return X
 
     if isinstance(data, AnnData):
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -283,6 +283,9 @@ def log1p(
     -------
     Returns or updates `data`, depending on `copy`.
     """
+    if 'log1p' in adata.uns_keys():
+        logg.warning('adata.X seems to be already log-transformed.')
+
     if copy:
         if not isinstance(data, AnnData):
             data = data.astype(np.floating)
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -255,12 +255,13 @@ def log1p(
     copy: bool = False,
     chunked: bool = False,
     chunk_size: Optional[int] = None,
+    base: Optional[float] = None,
 ) -> Optional[AnnData]:
     """\
     Logarithmize the data matrix.
 
     Computes :math:`X = \\log(X + 1)`,
-    where :math:`log` denotes the natural logarithm.
+    where :math:`log` denotes the natural logarithm unless a different base is given.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


scanpy/preprocessing/_simple.py
```diff
@@ -275,11 +276,16 @@ def log1p(
...
         `n_obs` of the chunks to process the data in.
+    base
+        Base of the logarithm. Natural logarithm is used by default.
 
     Returns
     -------
     Returns or updates `data`, depending on `copy`.
     """
+    if 'log1p' in data.uns_keys():
+        logg.warning('adata.X seems to be already log-transformed.')
```
flying-sheep, line 287: nice!

flying-sheep:
My questions from #929 don’t apply since you don’t use numba here.

Except for “What's the performance difference here”:

It’s not too bad, but we should use base 2 for everything that isn’t the natural logarithm: log2 can be calculated much faster on regular hardware due to binary storage.

scanpy/preprocessing/_simple.py
```diff
@@ -294,8 +300,12 @@ def log1p(
     def _log1p(X):
         if issparse(X):
             np.log1p(X.data, out=X.data)
+            if base is not None:
+                np.divide(X.data, np.log(base), out=X.data)
         else:
             np.log1p(X, out=X)
+            if base is not None:
+                np.divide(X, np.log(base), out=X)
         return X
```
flying-sheep, line 309: This is most likely 2×–3× as fast when using log2(X)÷log2(base). We should

1. investigate if that’s the case
2. still use log1p if base is None because it’s probably more accurate

flying-sheep (changes requested):

scanpy/preprocessing/_simple.py
ivirshup: As a heads up, there's a good chance this will differ depending on how numpy was built.

scanpy/preprocessing/_simple.py
ivirshup: Do we want to allow something like: `sc.pp.log1p(adata, layer="logcounts")`? We do this with `normalize_total`, which you'd probably want to use before this.

That functionality wouldn't need to be implemented here, I'm thinking the element that was log transformed should be recorded and the check should be for that. Though there would need to be a solution for the case where multiple elements were transformed. This is starting to sound a little complicated though, and no-check would also be fine.

scanpy/preprocessing/_simple.py
flying-sheep: Which ways of building it do you mean? Blas or CPU extensions or what?

log2 is probably always faster because you simply chop off the exponent part of the float, subtract 1 and return it.

ivirshup: Both I guess. MKL vs. OpenBLAS vs. whatever your HPC does can make a large difference. Ability to vectorize seems like a large factor.

This function is also not a bottleneck for typical usage, so I'm not too worried about performance. If this were an implementation for numpy, then it'd be more of an issue.

flying-sheep: SIMD and AVX2 are in basically everything today, but you’re right. Let’s not do premature optimization.

scanpy/preprocessing/_simple.py
flying-sheep: That’s for another feature request or PR, I’d say.

flying-sheep: I’d say that’s for another feature request or PR

gokceneraslan:
Let's please test this thoroughly, I'm not sure about how stable fold change estimates are in base 2.

gokceneraslan:
Also, I think when we save the log1p base info here `data.uns['log1p'] = {'base': base}`, we should always make the base more explicit even if it's `None`. Because if we change the default to base 2 at some point (which I'd really like), then anndatas saved before and after this change would refer to different defaults so exponentiations will be incorrect in LFC and HVG operations.

flying-sheep:
Currently it’s only an option, and nowhere the default, so I think having it in can’t hurt, right?

PR #939, opened 2019-11-29
Restrict networkx to 1.X

gokceneraslan:
We use Graph.node attributes (see https://github.com/theislab/scanpy/blob/master/scanpy/plotting/_tools/paga.py#L776) which are removed in networkx 2.X. More details here: https://networkx.github.io/documentation/stable/release/migration_guide_from_1.x_to_2.0.html


gokceneraslan:
Oops, please ignore.

PR #943, opened 2019-12-02
Add replace option to subsample and rename function to sample

### LINKED ISSUE #2854: sc.pp.subsample, allow for sampling with replacement

Currently sc.pp.subsample does not allow for sampling with replacement.

When n_obs is provided, and it is larger than the size of the adata object, an error message from numpy.random.choice is given.
"obs_indices = np.random.choice(old_n_obs, size=new_n_obs, replace=False)"

It seems like replace is automatically set to False.

It would be great if sc.pp.subsample provided a paramater to change the np.random.choice's 'replace' parameter to True.

gokceneraslan:
so that we can do sampling with replacement. It can be useful for bootstrap samples i.e. sc.pp.subsample(adata, n_obs=adata.n_obs, replace=True). We can also sample fractions > 1, but then it's not really subsampling.


gokceneraslan:
Maybe we should make a new function called `sc.pp.sample`?

cyrus303 (approved):

ivirshup:
I'd like to merge this.

I think we could just rename the function, and deprecate `subsample`. I'd like the name `sample` more for this function if we could add a `dim` argument so users can subsample on the variable as well.

scanpy/preprocessing/_simple.py
```diff
@@ -743,59 +743,127 @@ def scale(
...
+    n_obs: Optional[int] = None,
+    random_state: AnyRandom = 0,
+    copy: bool = False,
+) -> Optional[AnnData]:
+    """\
+    Subsample to a fraction of the number of observations.
+
+    .. warning::
+        .. deprecated:: 1.4.7
+            Use :func:`~scanpy.pp.sample` instead.
```
ivirshup: It looks like this is making the docs fail to build:

```
sphinx.errors.SphinxWarning: /home/docs/checkouts/readthedocs.org/user_builds/icb-scanpy/checkouts/943/scanpy/preprocessing/_simple.py:docstring of scanpy.pp.subsample:5:py:func reference target not found: scanpy.pp.sample
```

scanpy/tests/test_preprocessing.py
```diff
@@ -91,18 +91,37 @@ def test_normalize_per_cell():
...
+    sc.pp.sample(adata, n=10, axis=1)
+    assert adata.n_vars == 10
+    sc.pp.sample(adata, n=11, axis=1, replace=True)
+    assert adata.n_vars == 11
+    sc.pp.sample(adata, fraction=2.0, axis=1, replace=True)
+    assert adata.n_vars == 22
 
+    adata = AnnData(sp.csr_matrix(np.ones((200, 10))))
+    sc.pp.sample(adata, fraction=2.0, axis=1, replace=True)
+    assert adata.n_vars == 20
```
ivirshup: Could there also be tests for the argument handling? I'm thinking things like:

```python
    with pytest.raises(ValueError):
        sc.pp.sample(adata, fraction=2.0, n=100, replace=True)
    with pytest.raises(ValueError):
        sc.pp.sample(adata, fraction=2.0, replace=False)
    with pytest.raises(ValueError):
        sc.pp.sample(adata, n=adata.shape[0] + 1, replace=False)
```

scanpy/tests/test_preprocessing.py
```diff
@@ -91,18 +91,37 @@ def test_normalize_per_cell():
         axis=1).A1.tolist()
 
 
-def test_subsample():
+def test_sample():
```
ivirshup: Could there be tests that the values with the object are correct, not just the shape?

I'm thinking something like using an identity matrix as `X` and making sure each row or column on the sampled axis still has only one value.

scanpy/preprocessing/_simple.py
```diff
@@ -743,59 +743,127 @@ def scale(
...
+    fraction: Optional[float] = None,
+    n_obs: Optional[int] = None,
+    random_state: AnyRandom = 0,
+    copy: bool = False,
+) -> Optional[AnnData]:
+    """\
+    Subsample to a fraction of the number of observations.
+
+    .. warning::
+        .. deprecated:: 1.4.7
```
ivirshup: Should this function also throw a deprecation function on use?

ivirshup (commented):
Looks good! Just a few minor comments.

scanpy/preprocessing/_simple.py
gokceneraslan: Makes sense. Can you point me to an existing deprecation case in scanpy, I searched a bit but couldn't find :/

ivirshup: There are a few if you do a full code search for `depreca`.

One example would be at the top of `calculate_qc_metrics`. In anndata we also have a whole file for testing that deprecated functions work and throw warnings, which you could look at for examples of tests.

src/scanpy/preprocessing/_simple.py
```diff
@@ -825,18 +828,144 @@ def _regress_out_chunk(
...
+                msg = f"If `replace=False`, `{fraction=}` needs to be within [0, 1]."
+                raise ValueError(msg)
+            n = int(fraction * old_n)
+            logg.debug(f"... sampled to {n} {axis_name}")
+        case _:
+            msg = "Providing both `fraction` and `n` is not allowed."
+            raise TypeError(msg)
+    del fraction
+
+    np.random.seed(random_state)
```
eroell: Since this is new code, could move the[ np.random.choice](https://numpy.org/doc/stable/reference/random/generated/numpy.random.choice.html) to [Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator)?

tests/test_preprocessing.py
```diff
@@ -134,34 +143,122 @@ def test_normalize_per_cell():
...
-def test_subsample_copy_backed(tmp_path):
-    A = np.random.rand(200, 10).astype(np.float32)
-    adata_m = AnnData(A.copy())
-    adata_d = AnnData(A.copy())
-    filename = tmp_path / "test.h5ad"
-    adata_d.filename = filename
-    # This should not throw an error
-    assert sc.pp.subsample(adata_d, n_obs=40, copy=True).shape == (40, 10)
+def test_sample_copy_backed(tmp_path):
+    adata_m = AnnData(np.random.rand(200, 10).astype(np.float32))
```
eroell, line 253: np.random.rand wraps [np.random.random_sample](https://numpy.org/doc/stable/reference/random/generated/numpy.random.random_sample.html#numpy.random.random_sample), which also recommends to use [Generator](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator) in new code.

eroell: I think seeding is not too relevant here, but just double checking if there is a policy whether to generally not or to generally do seed tests?

src/scanpy/preprocessing/_simple.py
flying-sheep: Well, the old deprecated one should still work the same for a while, but maybe I can make that happen!

tests/test_preprocessing.py
flying-sheep: I think the only thing that matters is that two calls with the same seed will produce the same results. And I’m testing that!

scanpy/preprocessing/_simple.py
flying-sheep: handled in #3380

scanpy/tests/test_preprocessing.py
flying-sheep: If you feel like contributing this, feel free, otherwise I’ll just rely on numpy working as expected.

src/scanpy/preprocessing/_simple.py
```diff
@@ -825,67 +826,126 @@ def _regress_out_chunk(
...
+        assert not isinstance(subset, AnnData)
+        if copy:
+            subset = subset.copy()
+        return subset, indices
+    assert isinstance(subset, AnnData)
+    if copy:
+        return subset.to_memory() if data.isbacked else subset.copy()
+
+    # in-place
+    if data.isbacked:
```
Intron7: This should be done earlier

src/scanpy/preprocessing/_simple.py
```diff
@@ -825,67 +826,126 @@ def _regress_out_chunk(
...
+            n = int(fraction * old_n)
+            logg.debug(f"... sampled to {n} {axis_name}")
+        case _:
+            msg = "Providing both `fraction` and `n` is not allowed."
+            raise TypeError(msg)
+    del fraction
+
+    rng = np.random.default_rng(rng)
+    indices = rng.choice(old_n, size=n, replace=replace)
+    subset = data[indices] if axis_name == "obs" else data[:, indices]
```
Intron7, line 946: Why do we do this if its only needed for `copy=True` if `isinstance(data, AnnData)`

Intron7 (approved):

PR #945, opened 2019-12-04
Make the optimal_ordering accessible from sc.tl.dendrogram.

gokceneraslan:


gokceneraslan:
No idea about the error in the performance test. @flying-sheep ?

Issue #706, opened 2019-06-25
Wilcoxon rank_genes_groups math domain error (Scanpy 1.4.3)


HKanenew:
Hi guys,

Sorry I am new to github, I just posted this a comment on a closed thread (#530) and wasn't sure if that would be seen or not so I am just posting it up in as a new issue, just in case.

I am having trouble running the rank_gene_groups test using wilcoxon, every time I try it throws a math domain error. I think this might be a duplicate of #566.

As with the OP in #566 I have no trouble running the rank_genes_groups using t-test or logreg, the problem only arises when using method='wilcoxon'. For the example provided below `adata` is using one of our real data sets and I am using custom clusters but I have the same problem when replicating the tutorial workflow. My colleague has had the same experience when trying to work through the tutorial on his system, again using the most recent release of Scanpy.

I've made sure 'log_transformed' was being applied to include the #519 fix provided by **a-munoz-rojas** in the hope that this might help but no such luck, I get the same error either way. Any ideas on how to solve this?

Here is what I ran:

```
sc.tl.rank_genes_groups(adata, groupby='Custom_clusters_Leiden', groups=['NKT1','NKT2','NKT17','IL2+ aNKT1','aNKT2 & aNKT17','TNF- aNKT1'], reference='rest', method='wilcoxon', corr_method='benjamini-hochberg', log_transformed=True)
sc.pl.rank_genes_groups(adata, n_genes=25, sharey=False)
```

And here is the resulting error:

```
ValueError                                Traceback (most recent call last)
<ipython-input-117-a5ba74ea872c> in <module>
----> 1 sc.tl.rank_genes_groups(adata, groupby='Custom_clusters_Leiden', groups=['NKT1','NKT2','NKT17','IL2+ aNKT1','aNKT2 & aNKT17','TNF- aNKT1'], reference='rest', method='wilcoxon', corr_method='benjamini-hochberg', log_transformed=True)
      2 sc.pl.rank_genes_groups(adata, n_genes=25, sharey=False)

~\Anaconda3\lib\site-packages\scanpy\tools\_rank_genes_groups.py in rank_genes_groups(adata, groupby, use_raw, groups, reference, n_genes, rankby_abs, key_added, copy, method, corr_method, log_transformed, **kwds)
    367 
    368                 scores[imask, :] = (scores[imask, :] - (ns[imask] * (n_cells + 1) / 2)) / sqrt(
[...]

Issue #566, opened 2019-03-28
Math domain error when using the Wilcoxon rank-sum


davidepisu:
Another error I get and have no idea how to solve is when using the Wilcoxon rank-sum for testing for differential gene expression:

`sc.tl.rank_genes_groups(adata, 'louvain', method='wilcoxon')
sc.pl.rank_genes_groups(adata, n_genes=25, sharey=False)`

```
ranking genes

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-385-c2fa7bb8ea8d> in <module>
----> 1 sc.tl.rank_genes_groups(adata, 'louvain', method='wilcoxon')
      2 sc.pl.rank_genes_groups(adata, n_genes=25, sharey=False)

~\AppData\Local\conda\conda\envs\Scanpy\lib\site-packages\scanpy\tools\_rank_genes_groups.py in rank_genes_groups(adata, groupby, use_raw, groups, reference, n_genes, rankby_abs, key_added, copy, method, corr_method, **kwds)
    352 
    353                 scores[imask, :] = (scores[imask, :] - (ns[imask] * (n_cells + 1) / 2)) / sqrt(
--> 354                     (ns[imask] * (n_cells - ns[imask]) * (n_cells + 1) / 12))
    355                 scores[np.isnan(scores)] = 0
    356                 pvals = 2 * stats.distributions.norm.sf(np.abs(scores[imask,:]))

ValueError: math domain error


```

The logistic regression and t-test work fine.
I guess it is related to my data....

flying-sheep:
```py
>>> from math import sqrt                                                                                                                                                                                                                
>>> sqrt(-1)
ValueError: math domain error
```

I assume it’s the square root throwing this. Assuming that it only happens when you pass a negative argument, the term inside can only become negative if `ns[imask] < 0` or `ns[imask] > n_cells`

amitbin1:
Hey,
So I don't understand how I can get around this issue with the wilcoxon test. I'm following the scanpy tutorial and getting this 'ValueError: math domain error'.

bioguy2018:
Hi, I am also still receiving this error!

> ---------------------------------------------------------------------------
> ValueError                                Traceback (most recent call last)
> <ipython-input-141-e471c5e20fbd> in <module>
> ----> 1 sc.tl.rank_genes_groups(adata, 'louvain_05',n_genes=100,method="wilcoxon",use_raw=False)
> 
[...]

Issue #530, opened 2019-03-14
Math domain error in rank_genes_groups function


outlace:
I get the error below when trying to run the following:

`>>> sc.tl.rank_genes_groups(adata, 'louvain', groups=['5','16','19','30'], reference='0', method='wilcoxon')`

```bash
C:\Users\myuser\Anaconda3\lib\site-packages\scanpy\tools\_rank_genes_groups.py:298: RuntimeWarning: overflow encountered in long_scalars
  (n_active * m_active * (n_active + m_active + 1) / 12))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-160-dd19114ff660> in <module>
      1 #adata.obs['groups'] = ['group 1'= ['0'], 'group 2'= ['5','16','19','30']]
----> 2 sc.tl.rank_genes_groups(adata, 'louvain', groups=['5','16','19','30'], reference='0', method='wilcoxon') # wilcoxon-rank-sum/mann-whitney u test, the default of Seurat

~\Anaconda3\lib\site-packages\scanpy\tools\_rank_genes_groups.py in rank_genes_groups(adata, groupby, use_raw, groups, reference, n_genes, rankby_abs, key_added, copy, method, corr_method, **kwds)
    296 
    297                 scores = (scores - (n_active * (n_active + m_active + 1) / 2)) / sqrt(
--> 298                     (n_active * m_active * (n_active + m_active + 1) / 12))
    299                 scores[np.isnan(scores)] = 0
    300                 pvals = 2 * stats.distributions.norm.sf(np.abs(scores))

ValueError: math domain error
```

Here `adata` is real data from our lab, not the tutorial data. Have been trying to replicate the cluster analysis tutorial. All previous steps work fine. Interestingly, if I remove group '5' from the list of groups it works. Also, this error only happens with the `wilcoxon` method, not with `t-test`.

ivirshup:
This certainly looks strange. Would you mind making a minimal complete example? If you need, [here's a useful guide](http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) to making one.

falexwolf:
Sorry, there is a small bug in the wilcoxon method, that might hit sometimes. @a-munoz-rojas, it should be resolved after merging your fix, don't you think so? I'd be happy to move forward as soon as the code-overhead issue around double logarithmization is fixed. Should be very simple. Thank you!

a-munoz-rojas:
[...]

PR #952, opened 2019-12-12
Fixup #566 & #706

QuarkFang:
I think the reason of issues(#566 & #706) is that the value of x*y*z in sqrt(x*y*z) is too large and overflow encountered in long_scalars.
After updating these codes, the ValueError disappeared in my PC.

### DIFF
--- a/scanpy/tools/_rank_genes_groups.py
+++ b/scanpy/tools/_rank_genes_groups.py
@@ -397,8 +397,8 @@ def rank_genes_groups(
                 means[imask], vars[imask] = _get_mean_var(X[mask])
                 mean_rest, var_rest = _get_mean_var(X[mask_rest])
 
-                scores[imask, :] = (scores[imask, :] - (ns[imask] * (n_cells + 1) / 2)) / sqrt(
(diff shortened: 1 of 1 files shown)


chmreid:
#530 also related. Are you including the `log_transformed=True` kwarg in your call to `rank_genes_groups()`?

QuarkFang:
> 
> 
> #530 also related. Are you including the `log_transformed=True` kwarg in your call to `rank_genes_groups()`?

oh, I'm not. But when I included it, the ’ValueError: math domain‘ error still appeared.

chmreid:
In #706 @LuckyMD gives a minimal working example in [this comment](https://github.com/theislab/scanpy/issues/706#issuecomment-505335006) that may give you a useful starting point. Their advice to create/post a way to reproduce the error is good too, as it may help you identify the source of the problem in the process (and make it easier for others to help troubleshoot)!

bioguy2018:
Unfortunately I am still facing this problem! 
Is there a way to avoid it?

flying-sheep:
I think this is fixed since https://github.com/scverse/scanpy/pull/1310, please create a reproducer if not

PR #959, opened 2019-12-25
Add layer support to sc.tl.score_genes

gokceneraslan:
... so that we can do crazy things like scoring velocities or scoring "unspliced expression".


ivirshup:
@gokceneraslan I've been thinking we should have options like `layer` and `obsm` in many more places. I've started trying to implement this in a systemic way with an internal API and some test helpers in: #1173.

What would you think of using that here?

gokceneraslan:
@ivirshup is there an elegant way to get subset of genes with  `_get_obs_rep`? because in the current code there are things like:

```python
    _adata = adata.raw if use_raw else adata
    _adata_subset = _adata[:, gene_pool]
```

ivirshup:
Not much beyond `_get_obs_rep(adata[:, genes], ...)` or something like `_get_obs_rep(adata.raw.to_adata()[:, genes], ...)`.

PR #962, opened 2019-12-26
Find gene symbols exist only in .raw attribute

mw201608:
This change fixed an issue of retrieving data for gene symbols which exist only in the .raw attribute of an AnnData object.

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -2985,22 +2985,38 @@ def _prepare_dataframe(
                 f'Given {groupby}, valid observations: {adata.obs_keys()}'
             )
 
-    if gene_symbols is not None and gene_symbols in adata.var.columns:
-        # translate gene_symbols to var_names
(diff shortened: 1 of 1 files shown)


scanpy/plotting/_anndata.py
```diff
@@ -2985,22 +2985,38 @@ def _prepare_dataframe(
...
+                    logg.error(
+                        f"Gene symbol {symbol!r} not found in given "
+                        f"gene_symbols column: {gene_symbols!r}"
+                    )
+                    return
+                translated_var_names.append(
+                    adata.var[adata.var[gene_symbols] == symbol].index[0]
+                )
+            symbols = var_names
+            var_names = translated_var_names
```
ivirshup, line 3019: How about using `sc.get.obs_df` for this? I believe it already has the logic you're implementing.

ivirshup (changes requested):
Thanks for the PR!

I think you could use code we already have in the codebase for this method.

PR #963, opened 2019-12-27
Fix test_preprocessing_distributed

Koncopd:
I don't know what changes caused this, but now there are 2 problems with test_preprocessing_distributed.py.

When `adata_dist.X` is a dask array, `adata_dist.X.chunks` is `((2000, 2000, 2000, 2000, 2000), (1000,))`. It leads to an error in `adata.write_zarr(temp_store, chunks)` because zarr chunks should be a tuple with an integer entry per dimension, not a tuple of tuples.

The second problem is that  `adata_dist.X.to_zarr(temp_store.dir_path("X"))` causes an error because there is already `'X'` in `temp_store`, it needs to be overwritten.

This pr removes these problems but maybe logic of the function should be changed somehow instead of the test.

### DIFF
--- a/scanpy/tests/test_preprocessing_distributed.py
+++ b/scanpy/tests/test_preprocessing_distributed.py
@@ -90,7 +90,7 @@ class TestPreprocessingDistributed:
         temp_store = zarr.TempStore()
         chunks = adata_dist.X.chunks
         if isinstance(chunks[0], tuple):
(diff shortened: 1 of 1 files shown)


falexwolf:
Thanks for the fixes. I'll merge this for now. We'll go through the zarr-based implementation in depth with @ivirshup soon, again.

PR #975, opened 2020-01-08
Harmony_timeseries - framework for connecting scRNA-seq data from discrete time points

awnimo:
Harmony time series framework for connecting scRNA-seq data from discrete time points

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -20,7 +20,7 @@
 .. _Genome Biology (2018): https://doi.org/10.1186/s13059-017-1382-0
 
 
-.. topic:: News - CZI grant :small:`2012-11-14`
+.. topic:: News - CZI grant :small:`2019-11-14`
 
    Scanpy has been selected an `essential open source software for science`_ by
    CZI among `32 projects`_, along with giants such as Scipy, Numpy, Pandas,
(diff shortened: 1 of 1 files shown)


scanpy/external/tl/_harmony_timeseries.py
```diff
@@ -0,0 +1,432 @@
...
+    basis for generated a force directed layout for visualization and also serves as input
+    for computing the diffusion operator which can be used for trajectory detection using
+    `Palantir <https://github.com/dpeerlab/Palantir>`__.
+
+    .. note::
+        More information and bug reports `here <https://github.com/dpeerlab/Harmony>`__.
+
+    Parameters
+    ----------
+    adata :
```
falexwolf: There shouldn't be colons here.

scanpy/external/tl/_harmony_timeseries.py
```diff
@@ -0,0 +1,432 @@
...
+
+from scanpy import logging as logg
+import pandas as pd
+from collections import OrderedDict
+from itertools import chain
+import numpy as np
+import os
+
+
+def harmony_timeseries(adata: list, **kargs):
```
falexwolf: An adata is not of type list. :)

falexwolf (changes requested):
The test data you added has 1 MB altogether. 
image

We can't scale the repository if we allow merging data. Even Scanpy itself auto-downloads almost all of its builtin datasets.

I'd suggest to run the test based on builtin datasets of Scanpy such as the ones listed right here: https://scanpy-tutorials.readthedocs.io/en/latest/integrating-data-using-ingest.html#PBMCs

Please revert the commit or make a new PR without the data. Everything else looks good except for the two minor comments below.

Thank you!

  Alex

PR #988, opened 2020-01-13
Update MAGIC API to 2.x

scottgigante:
Changeset:
* added new MAGIC parameter `knn_max`
* updated MAGIC default parameters to match KrishnaswamyLab/MAGIC
* added tests to check MAGIC implementation interacts with AnnData object correctly
* increased required version of `magic-impute` to 2.0


scanpy/external/pp/_magic.py
```diff
@@ -124,11 +134,15 @@ def magic(
...
-        if not __version__ >= (1, 5):
+        version = tuple([int(v) for v in __version__.split(".")[:2]])
+        min_version = (2, 0)
+        if not version >= min_version:
             raise ImportError(
+                'scanpy requires magic-impute >= v{}.{} (detected: v{}). '
                 'Please update magic package via `pip install --user '
-                '--upgrade magic-impute`'
+                '--upgrade magic-impute`'.format(
+                    min_version[0], min_version[1], __version__
```
flying-sheep: Please use f-strings, not `format`.

.travis.yml
```diff
@@ -17,7 +17,7 @@ cache:
   - directories: data
 install:
   - pip install docutils sphinx
-  - pip install -e .[test,louvain,leiden]
+  - pip install -e .[test,louvain,leiden,magic]
```
flying-sheep, line 20: @falexwolf @ivirshup how should we go about testing external stuff?

I’d think that we should probably test it in a separate job so our jobs will be:

- python current - 1
- python current
- static analysis only
- extras only

scanpy/external/pp/_magic.py
```diff
@@ -17,10 +17,12 @@ def magic(
...
     *,
-    knn: int = 10,
-    decay: int = 15,
-    t: str = 'auto',
-    n_pca: int = 100,
+    knn: int = 5,
+    decay: Optional[float] = 1,
+    knn_max: Optional[int] = None,
+    t: Union[Literal['auto'], int] = 3,
+    n_pca: Optional[int] = 100,
```
flying-sheep, line 24: Please explain in the param docs what `None` means for `n_pca`. I think for all others you’re already explaining but please check.

scanpy/external/pp/_magic.py
```diff
@@ -68,6 +73,11 @@ def magic(
         neighborhoods. For extremely large datasets, using
         n_pca < 20 allows neighborhoods to be calculated in
         roughly log(n_samples) time.
+    solver
+        Which solver to use. "exact" uses the implementation described
+        in van Dijk et al. (2018) [1]_. "approximate" uses a faster implementation
```
flying-sheep: `[1]_`? Is that a footnote you forgot to define?

flying-sheep (changes requested):
Thank you for adding tests! There’s no precedent for testing externals, so let us think about that. I’m not sure if our suite should fail if an external tool breaks, but it would be good to see them tested in some way. Other than that there’s a few small changes I’d like to see.

.travis.yml
ivirshup: External job sounds like a good option. Only downside I immediately see (other than complexity) is that it will make CI take longer to run, but I don't think that's so bad.

I'm thinking we would control this with a new pytest tag, like we do the "internet test" ones?

scottgigante:
Re: testing externals, I've tried my best to just test the way it interfaces with scanpy. i.e., if MAGIC silently fails to return the correct output, scanpy tests would pass so long as the output is the right type / shape. If MAGIC throws an error when run from scanpy, this might be something you would like to address (i.e. by contacting the relevant external developer) regardless.

.travis.yml
flying-sheep: Yeah. Also not super urgent, only once there’s enough tested externals that it makes a difference.

scanpy/external/pp/_magic.py
```diff
@@ -124,9 +135,12 @@ def magic(
...
         )
     else:
-        __version__ = tuple([int(v) for v in __version__.split(".")[:2]])
-        if not __version__ >= (1, 5):
+        version = tuple([int(v) for v in __version__.split(".")[:2]])
+        min_version = (2, 0)
+        if not version >= min_version:
             raise ImportError(
+                'scanpy requires magic-impute >= '
+                f'v{min_version[0]}.{min_version[1]} (detected: v{__version__}). '
```
flying-sheep, line 143:
```suggestion
                f'v{min_version[0]}.{min_version[1]} (detected: v{version}). '
```
Please try this out.

scottgigante: `__version__` is the string version, `version` is the parsed tuple. I could do `v{version[0]}.{version[1]}` if you prefer.

flying-sheep: `__version__` doesn’t exist, you changed it.

flying-sheep: Ah, whoops, I just saw you renaming it, not that `__version__` still exists. Apologies

flying-sheep (approved):

PR #989, opened 2020-01-13
Fix 10x mtx reading functions for anndata 0.7

ivirshup:
Example of this kind of bug in https://github.com/theislab/anndata/issues/293

* Now `read_10x_mtx` does not set an integer name for obs_name/ var_name indices
* Additionally improved code re-use and things tested in 10x reading tests

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,6 @@ exclude = '''
         test_get
         |test_pca
         |test_combat
-        |test_read_10x
         |test_neighbors
         |test_readwrite
         |test_clustering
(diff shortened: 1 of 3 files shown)


ivirshup:
@flying-sheep, any idea what's up with the black error here? Am I at fault, are all the other listed files at fault, some combination of the two?

flying-sheep:
Didn’t you rephrase the message?

> scanpy/tests/test_read_10x.py: +3 -1
> 
> This above file has < {thresh} changes to black formatting. Please black format it and afterwards remove it from “tool.black.exclude" in pyproject.toml

Anyway, it should be “remove it from ‘tool.black.exclude’ *and then* black-format it”, as black won’t run on it if it’s excluded.

PR #994, opened 2020-01-14
Harmony_timeseries - connects data from discrete time points

awnimo:
Harmony_timeseries, a framework for connecting scRNA-seq data from discrete time points

### DIFF
--- a/scanpy/external/tl/_harmony_timeseries.py
+++ b/scanpy/external/tl/_harmony_timeseries.py
@@ -1,0 +1,462 @@
+"""Harmony time series for data visualization with augmented affinity matrix at discrete timepoints
+"""
+
+from scanpy import logging as logg
+import pandas as pd
+from collections import OrderedDict
+from itertools import chain
+import numpy as np
+import os
+
+
+def harmony_timeseries(data: list, **kargs):
+    """
+    Harmony time series for data visualization with augmented affinity matrix at discrete timepoints [Nowotschin18i]_.
+
+    Harmony time series is a framework for data visualization, trajectory detection and interpretation
+    for scRNA-seq data measured at discrete timepoints. Harmony constructs an augmented
+    affinity matrix by augmenting the kNN graph affinity matrix with mutually nearest
+    neighbors between successive time points. This augmented affinity matrix forms the
+    basis for generated a force directed layout for visualization and also serves as input
+    for computing the diffusion operator which can be used for trajectory detection using
+    `Palantir <https://github.com/dpeerlab/Palantir>`__.
+
+    .. note::
+        More information and bug reports `here <https://github.com/dpeerlab/Harmony>`__.
+
+    Parameters
+    ----------
+    data
+        list of :class:`~anndata.AnnData`, or cell x gene `.csv` filenames of count matrices from two time
+        points. Please ensure that replicates of the same time point are consecutive in order
+
+    timepoints
+        defaults to None, a list of timepoints at which each cell was measured. Important for Harmony augmented affinity matrix
+
+    sample_names
+        defaults to None, a list sample names in the form of timepoint_rep# of length `data`
+
+    log_transform
+        defaults to False, property setter passed to harmony
+
+
+    Returns
+    -------
+    `harmony_adata` of :class:`~anndata.AnnData` of normalized data:
+
+    - `.obsm['layout']` force directed layout
+
+    - `.uns['aff']` affinity matrix
+
+    - `.uns['aug_aff']` augmented affinity matrix
+
+    - `.uns['timepoint_connections']` timepoint connections
+
+    Example
+    -------
+
+    >>> import scanpy.external as sce
+    >>> import scanpy as sc
+    >>> import os
+
+    There are two ways to load data used with harmony_timeseries. If you have `csv`
+    files, follow these steps. If you have `AnnData` sets, follow the next snippet.
+
+    *Load sample data from csv files*
+
+    A sample data is available `here <https://github.com/dpeerlab/Harmony/tree/master/data>`_.
+
+    To view the plots, it is recommended to run jupyter notebook
+
+    >>> # harmony_dir = os.path.expanduser('~/Harmony/data/')
+    >>> csv_files = [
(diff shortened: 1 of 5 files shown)


scanpy/external/tl/_harmony_timeseries.py
```diff
@@ -0,0 +1,213 @@
...
+"""
+
+import pandas as pd
+from anndata import AnnData
+
+from ... import logging as logg
+
+
+def harmony_timeseries(data: AnnData, tp: str):
+    """Harmony time series for data visualization with augmented affinity matrix
```
flying-sheep:
```suggestion
    """\
    Harmony time series for data visualization with augmented affinity matrix
```

scanpy/external/tl/_harmony_timeseries.py
```diff
@@ -0,0 +1,213 @@
...
+    at discrete time points [Nowotschin18i]_.
+
+    Harmony time series is a framework for data visualization, trajectory
+    detection and interpretation for scRNA-seq data measured at discrete
+    time points. Harmony constructs an augmented affinity matrix by augmenting
+    the kNN graph affinity matrix with mutually nearest neighbors between
+    successive time points. This augmented affinity matrix forms the basis for
+    generated a force directed layout for visualization and also serves as input
+    for computing the diffusion operator which can be used for trajectory
+    detection using **Palantir**.
```
flying-sheep: As said: Please don’t use anonymous links. Use links like this instead:

```suggestion
    detection using Palantir_.

    .. _Palantir: https://github.com/dpeerlab/Palantir
```

scanpy/external/tl/_harmony_timeseries.py
```diff
@@ -0,0 +1,213 @@
...
+            data_df=adata.to_df(),
+            timepoints=adata.obs["time_points"],
+            timepoint_connections=adata.uns["timepoint_connections"],
+        )
+
+    **Visualization using force directed layouts**
+
+    >>> layout = d.harmony_timeseries.plot.force_directed_layout(
+            affinity_matrix=aug_aff, cell_names=adata.obs.index
+        )
```
flying-sheep: Is there any reason you need to return a class instance? Why not

1. Do both of those steps within `sce.tl.harmony_timeseries`
2. store the resulting matrices within `adata`

It’s the way scanpy works, and unless there’s reasons not to, users will appreciate tools to follow that convention.

scanpy/external/tl/_harmony_timeseries.py
```diff
@@ -0,0 +1,213 @@
...
+
+    Use any of Scanpy or Harmony methods for plotting. Example:
+
+    >>> d.harmony_timeseries.plot.plot_timepoints(
+            layout=layout, timepoints=adata.obs.time_points
+        )
```
flying-sheep: And if it’s possible to eliminate the class as I said above, the plotting function belongs into `sce.pl`

flying-sheep (changes requested):
Thank you for addressing some of my concerns. There’s some issues remaining that you didn’t address, though, while marking the corresponding conversations as resolved.

Did you forget to push some commits or did you simply not get around to do them yet? In the second case, it would be nice to not mark conversations as resolved that you didn’t address yet.

scanpy/external/tl/_harmony_timeseries.py
awnimo: Eliminated the class and reduced it into a simple function . Plotting can be done using `sc.pl.scatter` as illustrated in the documentation. Our harmony package has convenience plotting methods that perhaps we will add to `see.pl` in the future. In the meanwhile `sc.pl.scatter` does the same.

scanpy/external/tl/_harmony_timeseries.py
awnimo: Done!

scanpy/external/tl/_harmony_timeseries.py
flying-sheep: Great! Looks much simpler now. Thank you!

scanpy/external/tl/_harmony_timeseries.py
```diff
@@ -0,0 +1,161 @@
...
+        adata.to_df(),
+        adata.obs[tp],
+        adata.uns['timepoint_connections'],
+        pc_components=n_components,
+    )
+
+    # Force directed layouts
+    layout = harmony.plot.force_directed_layout(aug_aff, adata.obs.index)
+
+    adata.obs[['x', 'y']] = layout
```
flying-sheep: We’re storing embeddings like your layout in `.obsm["X_<embedding>"]`

```suggestion
    adata.obsm["X_harmony"] = layout
```

(Please adapt the example accordingly)

scanpy/external/tl/_harmony_timeseries.py
```diff
@@ -0,0 +1,161 @@
...
+    adata.uns['aff'] = aff
+    adata.uns['aug_aff'] = aug_aff
```
flying-sheep: Which of those are specific to harmony, which ones are only dependent on the data? E.g. if `aff` is *the* affinity matrix for this object, and `aug_aff` is harmony specific, `aug_aff` should be stored as `aff_aug_harmony` or `aff_harmony`.

Heads-up: Starting with AnnData 0.7, which will be released soon, we’re storing n_obs×n_obs matrices in anndata’s `.obsp`. I’ll switch this over ASAP once AnnData 0.7 is out.

flying-sheep (changes requested):
Great, this looks much simpler now, and very scanpy-like.

scanpy/external/tl/_harmony_timeseries.py
awnimo: Actually they are both Harmony specific. I will change the names as suggested. When the release of AnnData 0.7 is out I will make the change to adapt the new data structure.

awnimo: @flying-sheep 
Please let me know if there is any other issues that need to be addressed.

flying-sheep (approved):

PR #1003, opened 2020-01-21
Make external tests for MAGIC optional

### LINKED ISSUE #1001: MAGIC in external causes test failures if its not installed
@scottgigante @flying-sheep, on current master the tests fail for me with a MAGIC not installed error. External tests should have some mark to make sure they don't get run if the required package isn't installed.

scottgigante:
Closes #1001

### DIFF
--- a/scanpy/external/pp/_magic.py
+++ b/scanpy/external/pp/_magic.py
@@ -6,13 +6,14 @@ from typing import Union, Sequence, Optional
 from anndata import AnnData
 from numpy.random.mtrand import RandomState
 from legacy_api_wrap import legacy_api
+from packaging import version
 
 from ... import logging as logg
 from ..._settings import settings
 from ..._compat import Literal
 
 
-__min_version__ = "2.0"
+MIN_VERSION = "2.0"
 
 
 @legacy_api('k', 'a')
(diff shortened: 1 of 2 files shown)


scottgigante:
Realised there's a better way of doing this with pytest.

scanpy/external/pp/_magic.py
```diff
@@ -12,6 +12,9 @@
 from ..._compat import Literal
 
 
+__min_version__ = "2.0"
```
flying-sheep: Don’t use `__dunder__` naming for things that aren’t standardized python names. You can just use `UPPER_WITH_UNDERSCORES` for constants.

scanpy/external/pp/_magic.py
```diff
@@ -136,7 +139,7 @@ def magic(
         )
     else:
         version = tuple([int(v) for v in __version__.split(".")[:2]])
-        min_version = (2, 0)
+        min_version = tuple([int(v) for v in __min_version__.split(".")[:2]])
         if not version >= min_version:
```
flying-sheep: Please compare versions using `packaging.version`

https://github.com/theislab/scanpy/blob/0d44d1d50138e76bdc120cfacfd40c222e47b79e/scanpy/external/tl/_pypairs.py#L150-L152

flying-sheep (changes requested):
Thank you for your quick response. The `tuple` stuff is a bit flaky though, let’s just use standard version parsing and comparing

scottgigante:
Thanks, I hadn't seen `packaging.version` before. Much better!

flying-sheep:
It’s a recent change, driven by the complex stuff people did to compare versions :wink:

flying-sheep (approved):

scottgigante:
It's not clear to me why these `test_10x` tests are failing here and not on master -- there shouldn't be anything in this diff that affects those tests.

flying-sheep:
You’re right: They would be failing on master if master had been tested since anndata 0.7 was released earlier today. We have to merge #989 to get tests to pass again before we can merge this.

/edit: done, should work now

PR #1004, opened 2020-01-22
Fix plotting and harmony, depend on anndata 0.7 for obsp (#1004, #1007)

flying-sheep:
This

- moves all the external plotting routines to `scanpy/external/pl.py`
- adds one for harmony
- Fixes a plotting bug this triggered
- Makes harmony the first tool using obsp/varp

@awnimo can you please test this? Does the plot look like you want it to?


awnimo:
> @awnimo can you please test this? Does the plot look like you want it to?

@flying-sheep The plots look the same as expected. Compared plots using methods from Scanpy and Harmony

PR #1005, opened 2020-01-23
Add a CI build which uses anndata master

ivirshup:
It doesn't look like I can make the install process a matrix expansion, so this only tests against one version of python for now.


flying-sheep (approved):
It’s better that way. Compatibility problems with Python versions are usually not very subtle, and we only test with older versions so we don’t accidentally break stuff by depending on some new API.

flying-sheep:
Huh, it’s commit 3064268, this is the first time I saw a commit ID prefix without letters.

PR #1007, opened 2020-01-23
Ext plotting

awnimo:
Suggested changes to the test function, and harmony_timeseries

### DIFF
--- a/scanpy/external/tl/_harmony_timeseries.py
+++ b/scanpy/external/tl/_harmony_timeseries.py
@@ -61,6 +61,7 @@ def harmony_timeseries(adata: AnnData, tp: str, n_components: Optional[int] = 10
     Example
     -------
 
+    >>> from itertools import product
     >>> from anndata import AnnData
     >>> import scanpy as sc
     >>> import scanpy.external as sce
--- a/scanpy/external/tl/_harmony_timeseries.py
+++ b/scanpy/external/tl/_harmony_timeseries.py
@@ -55,6 +55,8 @@ def harmony_timeseries(adata: AnnData, tp: str, n_components: Optional[int] = 10
         augmented affinity matrix
     `.uns['harmony_timepoint_var']`
         The name of the variable passed as `tp`
+    `.uns['harmony_timepoint_connections']`
+        The links between time points
 
     Example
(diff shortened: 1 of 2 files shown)


scanpy/external/tl/_harmony_timeseries.py
```diff
@@ -128,3 +131,4 @@ def harmony_timeseries(adata: AnnData, tp: str, n_components: Optional[int] = 10
     adata.obsp["harmony_aff"] = aff
     adata.obsp["harmony_aff_aug"] = aug_aff
     adata.uns["harmony_timepoint_var"] = tp
+    adata.uns["harmony_timepoint_connections"] = timepoint_connections
```
flying-sheep, line 134: This won’t work, anndata can’t write a dataframe with numerical column names to disk

Since the connections are currently tivial anyway, (just each one connected to the next), I decided to not store this.

scanpy/tests/test_harmony_timeseries_ext.py
```diff
@@ -1,15 +1,20 @@
...
+adata = AnnData.concatenate(
+    *(adata_ref[i : i + 1000] for i in start),
+    join="outer",
+    batch_key="sample",
+    batch_categories=[f"sa{i}_Rep{j}" for i, j in product((1, 2, 3), (1, 2))],
+)
+adata.obs["time_points"] = adata.obs["sample"].str.split("_", expand=True)[0]
 sc.pp.normalize_total(adata, target_sum=10000)
 sc.pp.log1p(adata)
 sc.pp.highly_variable_genes(adata, n_top_genes=1000, subset=True)
```
flying-sheep, line 20: can you do all this within the test? It takes time to execute and that time shouldn’t be spent on top level, but within the actual test.

flying-sheep (changes requested):

flying-sheep:
Thank you. Did you also check if the plotting function works as intended?

awnimo:
> Thank you. Did you also check if the plotting function works as intended?

Yes I did, it works perfectly fine (see response [here](https://github.com/theislab/scanpy/pull/1004#issuecomment-577771644) too.

PR #1012, opened 2020-01-27
Plotting function for spatial data

giovp:
Support for 10x Visium images and coordinate system

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,6 @@ exclude = '''
     |plotting/(
         __init__
         |palettes
-        |_utils
         |_preprocessing
         |_tools/(__init__|paga)
     )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ exclude = '''
         |palettes
         |_utils
         |_preprocessing
-        |_tools/(scatterplots|__init__|paga)
+        |_tools/(__init__|paga)
     )
     |tests/(
         test_get
--- a/scanpy/plotting/_docs.py
+++ b/scanpy/plotting/_docs.py
@@ -66,7 +66,8 @@ legend_fontoutline
 size
     Point size. If `None`, is automatically computed as 120000 / n_cells.
     Can be a sequence containing the size for each cell. The order should be
-    the same as in adata.obs.
+    the same as in adata.obs. If `img_key` not `None`, size is the scaling factor
+    for the spot size.
 color_map
     Color map to use for continous variables. Can be a name or a
     :class:`~matplotlib.colors.Colormap` instance (e.g. `"magma`", `"viridis"`
--- a/scanpy/plotting/_docs.py
+++ b/scanpy/plotting/_docs.py
@@ -91,7 +91,7 @@ title
 img_key
     Key for image data, stored in `adata.uns`.
 crop_coord
-    Coordinates to use for cropping the image.
+    Coordinates to use for cropping the image (left, right, top, bottom).
 alpha_img
     Alpha value for image.
 scale_spot
--- a/scanpy/plotting/_docs.py
+++ b/scanpy/plotting/_docs.py
@@ -95,9 +95,6 @@ crop_coord
     Coordinates to use for cropping the image (left, right, top, bottom).
 alpha_img
     Alpha value for image.
-scale_spot
-    Scaling factor for scatter points overlaying image (spots), 
-    only for Visium, otherwise use `size`.
 bw
     Plot Image in gray scale.\
 """
--- a/scanpy/plotting/_docs.py
+++ b/scanpy/plotting/_docs.py
@@ -87,7 +87,18 @@ frameon
     :func:`~scanpy.set_figure_params`, defaults to `True`.
 title
     Provide title for panels either as string or list of strings,
-    e.g. `['title1', 'title2', ...]`.\
+    e.g. `['title1', 'title2', ...]`.
+img_key
+    Key for image data, stored in `adata.uns`.
+crop_coord
+    Coordinates to use for cropping the image.
+alpha_img
+    Alpha value for image.
+scale_spot
+    Scaling factor for scatter points overlaying image (spots), 
+    only for Visium, otherwise use `size`.
+bw
+    Plot Image in gray scale.\
 """
 
 doc_vminmax = """\
--- a/scanpy/plotting/_tools/scatterplots.py
+++ b/scanpy/plotting/_tools/scatterplots.py
@@ -13,7 +13,7 @@ from matplotlib import patheffects
 from matplotlib.colors import Colormap
 
 from .. import _utils
-from .._utils import _IGraphLayout, _FontWeight, _FontSize
+from .._utils import _IGraphLayout, _FontWeight, _FontSize, circles
 from .._docs import (
     doc_adata_color_etc,
     doc_edges_arrows,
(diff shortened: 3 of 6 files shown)


flying-sheep:
Looks good! Remaining questions:

- The plan is to add the Visium reading function to `anndata`, right?
- You’re repeating yourself with the docs: `doc_scatter_basic` (and therefore `doc_scatter_embedding`) and the docstring of `pl.spatial` both contain similar text for the same parameters. If you want to reorder them, you could do something fancy (like slicing doc_scatter_embedding) or just mention the parameter names in the free text, something like:

  ```restructuredtext
  Scatter plot in spatial coordinates.

  Use the parameter `img_key` to see the microscopy image in the background.
  Use `crop_coord`, `alpha_img`, and `bw` to control how it is displayed,
  and `scale_spot` to control the size of the Visium spots plotted on top.
  ```

- Is it possible to derive the amount of cropping? Then we could extend the `crop_coord` parameter to this:

  ```py
  Union[
      Iterable[Literal['left', 'l', 'right', 'r', 'top', 't', 'bottom', 'b']],
      Tuple[int, int, int, int],  # l, r, t, b
  ]
  ```

- Maybe it makes sense to add some test data and a test plot? (very low res of course)

ivirshup:
Would you mind posting some example usage?

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -1146,7 +1132,7 @@ def _process_image(adata, data_points, img_key, crop_coord, scale_spot, bw=False
     spot_size = adata.uns[scalef_key] * data_points[0].max() * scale_spot
 
     if crop_coord is not None:
-        img_coord = crop_coord
+        img_coord = [crop_coord[0], crop_coord[1], np.ceil(img.shape[0]-crop_coord[2]), np.ceil(img.shape[0]-crop_coord[3])]
```
flying-sheep:
```suggestion
        crop_coord = np.asarray(crop_coord)
        if len(crop_coord) != 4:
            raise ValueError("Invalid crop_coord of length {len(crop_coord)}(!=4)")
        img_coord = (*crop_coord[:2], *np.ceil(img.shape[0] - crop_coord[2:4]).astype(int))
```

it’s weird that np.ceil doesn’t output ints, but that’s the way it is.

flying-sheep (changes requested):
Looks good! Why did you intoduce scale_spot? My assumption is that scale_spot is dependent on the size of the spots in the real data, while `size` is just for visualizing infinitely small points, am I right?

Also you mentioned example data in chat. Are you going to add some?

flying-sheep:
@davidsebfischer this is a good convention for storing spatial coordinates in AnnData I think. You have spatial code that might want to use it.

giovp:
scale_spot is a float to scale the size of the spots from users. We have original radius dimension but it can be handy to modify it according to cropping/zooming, or simply for visualization purposes.

Regarding examples, I've made a tutorial that I will push to scanpy-tutorial. I'm waiting for @Mirkazemi read function that I need to complete the notebook

flying-sheep:
> We have original radius dimension but it can be handy to modify it according to cropping/zooming, or simply for visualization purposes.

Cropping/zooming won’t make a difference if you plot circles in data space. So there’s our problem: We have the original radius in data space, but you’re plotting markers, whose size is in figure space (i.e. their center position in the final figure is determined and then they’re plotted as circles right into the graphic).

So you need to switch from `ax.scatter` to a `circles` function that does what we need: https://stackoverflow.com/questions/9081553/python-scatter-plot-size-and-style-of-the-marker/24567352#24567352

We can just adapt that one (throw out what we don’t need), make it so the `scatter(...)` calls in “embedding” work with it, and do `scatter = ax.scatter if img_key is None else partial(circles, ax=ax)`.

This means that we don’t have to do difficult math when cropping/zooming, as the spots will always just be the correct size. We can also get rid of `spot_size` and make `size` a scale factor in the image case (1=normal size, 0.8=slightly smaller than in the data, 1.2=slightly larger than in the data)

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -55,7 +55,7 @@ def embedding(
     img_key: Optional[str] = None,
     crop_coord: Tuple[int, int, int, int] = None,
     alpha_img: float = 1.0,
-    scale_spot: float = 0.5,
+    scale_spot: float = 1.0,
```
flying-sheep: Just reuse size for that and update its docs like “if `img_key` is specified, `size` scales …”

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -369,26 +378,36 @@ def embedding(
...
+                else:
+                    cax = ax.scatter(
+                        _data_points[:, 0],
+                        _data_points[:, 1],
+                        s=size,
+                        marker=".",
+                        c=color_vector,
+                        rasterized=settings._vector_friendly,
+                        **kwargs,
+                    )
```
flying-sheep: Please just do once at the top:

```py
scatter = ax.scatter if img_key is None else partial(circles, ax=ax)
```

Make sure `circles` accepts the same parameters and order as `scatter`.

Then you can just replace `cax = ax.scatter(` with `cax = scatter(` everywhere instead of building even more huge if blocks.

scanpy/plotting/_tools/scatterplots.py
flying-sheep: Just reuse size for that and update its docs like “if `img_key` is specified, `size` scales …”

Then update `circles` to handle `size=None` as if it’s `size=1`.

flying-sheep (changes requested):

giovp:
Thanks Phil for the suggestion, should be better now. Reason to keep `size_spot` for the image is because size of the spots are set inside the for loop for plotting.

scanpy/plotting/_docs.py
```diff
@@ -87,7 +88,18 @@
...
+    e.g. `['title1', 'title2', ...]`.
+img_key
+    Key for image data, stored in `adata.uns`.
+crop_coord
+    Coordinates to use for cropping the image (left, right, top, bottom).
+alpha_img
+    Alpha value for image.
+scale_spot
+    Scaling factor for scatter points overlaying image (spots), 
+    only for Visium, otherwise use `size`.
```
flying-sheep: This no longer exists

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -304,30 +356,40 @@ def embedding(
...
                     **kwargs,
                 )
 
             else:
-                cax = ax.scatter(
+                scatter = (
+                    partial(ax.scatter, s=size)
+                    if img_key is None
+                    else partial(circles, s=size_spot)
+                )
```
flying-sheep: Why do you define it in both branches? Just define it once before the `if`.

giovp: Right, modified

scanpy/plotting/_docs.py
giovp: Right, thank you

pyproject.toml
```diff
@@ -47,7 +47,7 @@ exclude = '''
         |palettes
         |_utils
```
flying-sheep: Seems like you formatted `plotting/_utils.py`, so you need to remove this line.

flying-sheep (changes requested):
Great, now only the tests are left (the unicode/numpy.str_ is probably not this PR, but e.g. [this](https://travis-ci.org/theislab/scanpy/jobs/644367920#L854) is)

Also seems like you need to black-format `scatterplots.py` again.

pyproject.toml
giovp: Done

giovp:
I think that specific fail should be fixed with last commit

PR #1013, opened 2020-01-30
Add visium_sge dataset function

Mirkazemi:
10x Spatial Gene Expression data includes tar files. The files in tar format are checked by `check_presence_download_untar()`.

### DIFF
--- a/scanpy/_utils.py
+++ b/scanpy/_utils.py
@@ -546,6 +546,14 @@ def check_presence_download(filename: Path, backup_url):
         from .readwrite import _download
         _download(backup_url, filename)
 
+def check_presence_download_untar(filename: Path, tarfilename: Path, backup_url):
+    """Check if file is present otherwise download and untar."""
+    if not filename.is_file():
+        if not tarfilename.is_file():
+            from .readwrite import _download
+            _download(backup_url, tarfilename)
+        import tarfile
+        tarfile.open(tarfilename).extractall(tarfilename.parents[0])
 
 def lazy_import(full_name):
     """Imports a module in a way that it’s only executed on member access"""
--- a/scanpy/_utils.py
+++ b/scanpy/_utils.py
@@ -546,14 +546,16 @@ def check_presence_download(filename: Path, backup_url):
         from .readwrite import _download
         _download(backup_url, filename)
 
-def check_presence_download_untar(filename: Path, tarfilename: Path, backup_url):
+def check_presence_download_untar(filename: Path, tarfilename: Path, backup_url: str):
     """Check if file is present otherwise download and untar."""
     if not filename.is_file():
-        if not tarfilename.is_file():
-            from .readwrite import _download
-            _download(backup_url, tarfilename)
+        check_presence_download(tarfilename, backup_url)
+        
+        # The following extract only the intended file within some folders, and not all other files.
         import tarfile
-        tarfile.open(tarfilename).extractall(tarfilename.parents[0])
+        tar = tarfile.open(tarfilename)
+        _file = [_ for _ in tar.getnames() if Path(_).name == filename.name][0]
+        tar.extract(_file , tarfilename.parent)
 
 def lazy_import(full_name):
     """Imports a module in a way that it’s only executed on member access"""
--- a/scanpy/datasets/__init__.py
+++ b/scanpy/datasets/__init__.py
@@ -296,8 +296,8 @@ def pbmc3k_processed() -> AnnData:
     )
     return adata
 
-def SGE10x(sample_ID = 'V1_Breast_Cancer_Block_A_Section_1') -> AnnData:
-    """Processed Spatial Gene Expression data from 10x Genomics.
+def visium_sge(sample_ID = 'V1_Breast_Cancer_Block_A_Section_1') -> AnnData:
+    """Processed Visium Spatial Gene Expression data from 10x Genomics.
     Database: https://support.10xgenomics.com/spatial-gene-expression/datasets
     Parameters
     ----------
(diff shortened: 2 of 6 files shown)


giovp:
For me it's good, I would just merge so we have a prototype version for the vignette, then we can extend for other data type/modify it.

scanpy/_utils.py
```diff
@@ -546,6 +546,14 @@ def check_presence_download(filename: Path, backup_url):
         from .readwrite import _download
         _download(backup_url, filename)
 
+def check_presence_download_untar(filename: Path, tarfilename: Path, backup_url):
```
flying-sheep:
```suggestion
def check_presence_download_untar(filename: Path, tarfilename: Path, backup_url: str):
```

scanpy/_utils.py
```diff
@@ -546,6 +546,14 @@ def check_presence_download(filename: Path, backup_url):
...
+    """Check if file is present otherwise download and untar."""
+    if not filename.is_file():
+        if not tarfilename.is_file():
+            from .readwrite import _download
+            _download(backup_url, tarfilename)
```
flying-sheep: Please avoid duplicating code. You can replace those three lines with `check_presence_download(tarfilename, backup_url)`

scanpy/_utils.py
```diff
@@ -546,6 +546,14 @@ def check_presence_download(filename: Path, backup_url):
...
+        import tarfile
+        tarfile.open(tarfilename).extractall(tarfilename.parents[0])
```
flying-sheep:
```suggestion
        tarfile.open(tarfilename).extract(filename, tarfilename.parent)
```

Does this make sense? This way not everything is extracted if there’s more. Please check if it works as intended. Maybe you need to add `str(...)` around `filename`.

scanpy/datasets/__init__.py
```diff
@@ -295,3 +295,107 @@ def pbmc3k_processed() -> AnnData:
...
+    Returns
+    -------
+    Annotated data matrix.
+    """    
+    
+    
+    # setting filenames, tarfilenames, backup_rurls
+    # tarfilenames and backup_urls will be used for downloading data base
+    from urllib.parse import urljoin
+    files = {
```
flying-sheep: please use `dict(...)` syntax for dicts with names that are valid python identifiers (like those ones)

> Is that just the file structure of this specific dataset or a generic format? In the latter case you could also do a `read_visium` function in `scanpy.readwrite` that is used from your dataset function.

This structure has some overlaps with other visium datasets like single cell gene exp but they are not completely identical. Here is a tar file (folder) including spatial data. On the other hand, single cell has some Genome-aligned BAM files.

scanpy/datasets/__init__.py
```diff
@@ -295,3 +295,107 @@ def pbmc3k_processed() -> AnnData:
...
+        'counts'                 : settings.datasetdir / (sample_ID + '_raw_feature_bc_matrix.h5'),
```
flying-sheep: please use `f'{sample_id}_raw_feature_bc_matrix.h5'` here and in similar cases

scanpy/datasets/__init__.py
```diff
@@ -295,3 +295,107 @@ def pbmc3k_processed() -> AnnData:
         backup_url='https://raw.githubusercontent.com/chanzuckerberg/cellxgene/master/example-dataset/pbmc3k.h5ad',
     )
     return adata
+
+def SGE10x(sample_ID = 'V1_Breast_Cancer_Block_A_Section_1') -> AnnData:
```
flying-sheep: please don’t use uppercase letters in function names and parameter names.

Only `TypeNames` and `CONSTANTS` have uppercase letters.

scanpy/datasets/__init__.py
```diff
@@ -295,3 +295,107 @@ def pbmc3k_processed() -> AnnData:
...
+    import matplotlib.image as img
+    adata.uns['hires_image'] = img.imread(str(files['hires_image'])) 
+    adata.uns['lowres_image'] = img.imread(str(files['hires_image']))
+    
+    
+
+    # reading json scalefactors
+    import json
+    with open(files['scalefactors_json_file']) as json_file:
+        adata.uns['scalefactors'] = json.load(json_file)
```
flying-sheep: If this is not too big, just do

```py
adata.uns['scalefactors'] = json.loads(files['scalefactors_json_file'].read_bytes())
```

flying-sheep (changes requested):
Is that just the file structure of this specific dataset or a generic format? In the latter case you could also do a `read_visium` function in `scanpy.readwrite` that is used from your dataset function.

Otherwise great! Good and clean logic, only some code style concerns.

To get everything formatted, just install black and do `black .` in the scanpy directory. Tip: black has instructions on how to run it automatically on changed files from VS Code or PyCharm.

scanpy/datasets/__init__.py
giovp: Maybe something like `visium` and then add in the docs the 10x info, similar to rest of [datasets](https://scanpy.readthedocs.io/en/stable/api/index.html#module-scanpy.datasets)

scanpy/datasets/__init__.py
```diff
@@ -295,3 +295,107 @@ def pbmc3k_processed() -> AnnData:
...
+    # reading h5 file
+    from ..readwrite import read_10x_h5
+    adata = read_10x_h5(files['counts'])
+    
+    adata.var_names_make_unique()
+
+    # reading images
+    # imread can only get string and not a Path object
+    import matplotlib.image as img
+    adata.uns['hires_image'] = img.imread(str(files['hires_image']))
```
giovp:
```suggestion
    adata.uns['image']['hires'] = img.imread(str(files['hires_image'])) 
```

scanpy/datasets/__init__.py
```diff
@@ -295,3 +295,107 @@ def pbmc3k_processed() -> AnnData:
...
+    
+    # reading h5 file
+    from ..readwrite import read_10x_h5
+    adata = read_10x_h5(files['counts'])
+    
+    adata.var_names_make_unique()
+
+    # reading images
+    # imread can only get string and not a Path object
+    import matplotlib.image as img
```
giovp, line 381:
```suggestion
    import matplotlib.image as img
    adata.uns['image']={}
```

scanpy/datasets/__init__.py
```diff
@@ -295,3 +295,107 @@ def pbmc3k_processed() -> AnnData:
...
+    adata.uns['hires_image'] = img.imread(str(files['hires_image'])) 
+    adata.uns['lowres_image'] = img.imread(str(files['hires_image']))
```
giovp:
```suggestion
    adata.uns['image']['lowres'] = img.imread(str(files['lowres_image']))
```

scanpy/datasets/__init__.py
giovp: I think it's actually better to save images in a specific dictionary, cleaner and we can extend it later

scanpy/datasets/__init__.py
Mirkazemi: Done.

scanpy/datasets/__init__.py
Mirkazemi: Done.

scanpy/datasets/__init__.py
Mirkazemi: I changed the function name to visium_sge().

scanpy/datasets/__init__.py
Mirkazemi: Yes! I totally agree.

scanpy/datasets/__init__.py
Mirkazemi: Done.

scanpy/datasets/__init__.py
Mirkazemi: Done.

flying-sheep:
Great! Can you push the changes please so I can have a look?

scanpy/_utils.py
Mirkazemi: Done.

scanpy/_utils.py
Mirkazemi: Done.

scanpy/_utils.py
Mirkazemi: Nice idea. But the code line does not work because tar file is a folder called 'spatial' and intended files (and other files) are located in that folder. I change in a way that only targeted files are extracted (within /spatial). The following only extract targeted files within any structure of folders:
```suggestion
        tar = tarfile.open(tarfilename)
        _file = [_ for _ in tar.getnames() if Path(_).name == filename.name][0]
        tar.extract(_file , tarfilename.parent)```

Mirkazemi: Nice idea. But the code line does not work because tar file is a folder called 'spatial' and intended files (and other files) are located in that folder. I change in a way that only targeted files are extracted (within /spatial). The following only extract targeted files within any structure of folders:
```suggestion
        tar = tarfile.open(tarfilename)
        _file = [_ for _ in tar.getnames() if Path(_).name == filename.name][0]
        tar.extract(_file , tarfilename.parent)```

scanpy/datasets/__init__.py
Mirkazemi: > To get everything formatted, just install black and do `black .` in the scanpy directory. Tip: black has instructions on how to run it automatically on changed files from VS Code or PyCharm.

Thanks for suggesting black. It's quite useful and I reformat the files. But to be honest I do not like it 100% :)
For a set of similar lines of codes I prefer to left spaces to make them more clear, like this:
aaaa = 
a  , , , =
aaa , =
which black prevents.

giovp (commented):
Ready to go for me, looks really good. Thank you Mo!

PR #1015, opened 2020-01-31
Fix test breakages

flying-sheep:
- [x] leiden/louvain tests are fixed by 71fbae4 circumventing pandas-dev/pandas#31499
- [x] `test_read_10x[mtx_path1-h5_path1]` fixed by theislab/anndata#313
- [x] `test_plotting.test_clustermap` is caused by seaborn mwaskom/seaborn#1953

### DIFF
--- a/scanpy/readwrite.py
+++ b/scanpy/readwrite.py
@@ -168,7 +168,8 @@ def read_10x_h5(
             adata = adata[:, list(map(lambda x: x == str(genome), adata.var['genome']))]
         if gex_only:
             adata = adata[:, list(map(lambda x: x == 'Gene Expression', adata.var['feature_types']))]
-        return adata
+        if adata.is_view:
+            return adata.copy()
     else:
         return _read_legacy_10x_h5(filename, genome=genome, start=start)
 
@@ -307,7 +308,7 @@ def read_10x_mtx(
         return adata
     else:
         gex_rows = list(map(lambda x: x == 'Gene Expression', adata.var['feature_types']))
-        return adata[:, gex_rows]
+        return adata[:, gex_rows].copy()
 
 
 def _read_legacy_10x_mtx(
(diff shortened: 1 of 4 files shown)


ivirshup:
@flying-sheep, I made two small changes:

* The 10x readers should no longer return views, fixing `test_read_10x`
* Slightly cleaner providing of categories for leiden/ louvain code.

For the clustermap test, it's not clear to me that the problems are even related to pandas, though the cause might be: https://github.com/pandas-dev/pandas/issues/18720.

There are two images which are compared in this test. I'll post the comparisons here:

# `master_clustermap.png`

I believe the difference is just the margin, so we should be good to just change the test image.

## Expected

master_clustermap

## Actual

master_clustermap

# `master_clustermap_withcolor.png`

This one looks worse, but I'm not sure how to fix it. @fidelram might know better?

## Expected

master_clustermap_withcolor

## Actual

master_clustermap_withcolor

flying-sheep:
Sorry yeah, the branch name might not be accurate when we’re done.

What dependency causes those mismatches? Locally the image comparisons succeed unless I create a venv …

> master_clustermap.png
> 
> I believe the difference is just the margin, so we should be good to just change the test image.

nah, it also has an ugly white gap between tree and heatmap, probably caused by the same reason as the other one …

/edit: the space seems exactly enough to hold the color bar’s x tick labels, that might be the cause.

/edit2: also appears without scanpy’s style or anything: mwaskom/seaborn#1953

jenzopr:
Whats the timeline here? When will there be a release that includes this fix? Or should I downgrade pandas in the meantime?

flying-sheep:
You can just use scanpy master if you need this fix ASAP.

Otherwise I think we should release a new version pretty soon after merging #1038.

jenzopr:
Great! Thanks a lot, Philipp!

PR #1020, opened 2020-02-02
Use matplotlib 3.1 and adapt tests

flying-sheep:
Now only mpl versions start to break more important stuff than 3d plotting, so let’s update.

<details><summary>Outdated</summary>

The umap_with_edges went 

| from | to |
| ---- | -- |
| grafik | grafik |

Is the new one correct? If not, what did I do wrong?

</details>


fidelram:
seems to me like the default palette changed?

flying-sheep:
Yeah, but why does the test not fail on master then? The following works:

```py
import scanpy as sc
assert pbmc.uns['louvain_colors'].tolist() == sc.pl.palettes.zeileis_28[:11]
```

and the Zeileis palette is:

PR #1024, opened 2020-02-03
Spatial

flying-sheep:
Draft PR (@ivirshup look, Github has that now!)

Once @giovp and @Mirkazemi say it’s ready from their side, we can review and merge this.

Docs:

- [read_visium](https://icb-scanpy.readthedocs-hosted.com/en/latest/api/scanpy.read_visium.html) (new)
- [datasets.visium_sge](https://icb-scanpy.readthedocs-hosted.com/en/latest/api/scanpy.datasets.visium_sge.html) (new)
- [pl.spatial](https://icb-scanpy.readthedocs-hosted.com/en/latest/api/scanpy.pl.spatial.html) (new)
- [pl.embedding](https://icb-scanpy.readthedocs-hosted.com/en/latest/api/scanpy.pl.embedding.html) (new parameters)


flying-sheep:
Hi @giovp still up for adding a test dataset and tests? If so, this would be a good moment in time, as we should merge this quickly before more changes to master cause conflicts

giovp:
Yes, I am. Shall I just add the data in `_data` ?

giovp:
Also, how's the multiple slices support for `anndata` @Mirkazemi ? That would also be very useful before merging this PR.

flying-sheep:
> Yes, I am. Shall I just add the data in `_data` ?

Sure, as long as it’s not noticably bigger than any other dataset in there.

flying-sheep:
Hi @giovp! The test data is too large, it’ll take scanpy a long time to clone once this is in `master`. The way we fix it is that we replace the data and then merge our changes into commit bb70446 (creating a new commit from the two and eliminating any trace of the big dataset)

For reference, the test data `filtered_feature_bc_matrix.h5` is <100kb. I’d say you find the smallest of the 10x example datasets, reduce it so the (non-image) data is <100kb all in all, and delete the hires pic. The code should work if there’s only the lores pic anyway, right?

An alternative would be to mark our tests as “internet” tests and dynamically download the data, but I think it’s better to always run the spatial tests.

giovp:
I agree, I've been trying to reduce the dataset and save it as 10x_h5 format again but it seems to be more complicated than I thought. I think it would be best to have an 10x_h5 format so to also use it for the read function test. Also, yes will only keep the lowres image

scanpy/readwrite.py
```diff
@@ -260,6 +276,109 @@ def _read_v3_10x_h5(filename, *, start=None):
...
+            if genome not in adata.var['genome'].values:
+                raise ValueError(
+                    f"Could not find data corresponding to genome '{genome}' in '{filename}'. "
+                    f'Available genomes are: {list(adata.var["genome"].unique())}.'
+                )
+            adata = adata[:, list(map(lambda x: x == str(genome), adata.var['genome']))]
+        # removed gex_only as there are no other data types
+        # the if statement below could also be removed since it's supposed to be v3 chemistry
+        if adata.is_view:
+            adata = _read_legacy_10x_h5(filename, genome=genome, start=start)
```
flying-sheep: @giovp This code is copied from the 10x reading function. Why copy code instead of calling the other function?

flying-sheep:
Congrats @giovp, you landed a big new thing!

PR #1026, opened 2020-02-04
Test that the fallback fr layout works

flying-sheep:
Breaks for me locally…


scanpy/plotting/_tools/paga.py
```diff
@@ -226,15 +226,15 @@ def _compute_pos(
...
+                # this is a super-weird hack that is necessary as igraph’s
+                # layout function seems to do some strange stuff here
                 init_pos[:, 1] *= -1
                 init_coords = init_pos.tolist()
             try:
                 pos_list = g.layout(
                     layout, seed=init_coords,
                     weights='weight', **layout_kwds).coords
-            except:  # hack for excepting attribute error for empty graphs...
+            except AttributeError:  # hack for empty graphs...
```
flying-sheep, line 237: @falexwolf is that what would happen? Please don’t use plain except…

PR #1029, opened 2020-02-05
Add reading from .uns colors for sc.pl.violin

michalk8:
Currently, `sc.pl.violin` does not support support reading colors from `adata['uns']`, this PR fixes it.
Prior to this PR, the below "works":
```python
sc.pl.violin(adata, keys=['initial_size'], stripplot=True, groupby='initial_size')
```
i.e. takes about ~30 seconds to plot, because `groupby` is not categorical (there's no check), now
it raises `AttributeError: Can only use .cat accessor with a 'category' dtype` because of the color addition.


flying-sheep:
Thank you! I’d prefer a nicer, more explicit error that describes the problem. You could use sth like:

```py
if not is_categorical_dtype(adata.obs[groupby].dtype):
    raise ValueError(
        f"The column `adata.obs[groupby]` needs to be categorical, "
        f"but the {groupby!r} column is of dtype {adata.obs[groupby].dtype}."
    )
```

flying-sheep:
Does it work to do `is_categorical_dtype(a_series)` or do you have to do `is_categorical_dtype(a_series.dtype)`?

flying-sheep:
Thanks!

PR #1038, opened 2020-02-11
Fix ingest for the new umap

Koncopd:
This should fix the problems with the new release candidate of umap.
https://github.com/theislab/scanpy/issues/1036

### DIFF
--- a/scanpy/tools/_ingest.py
+++ b/scanpy/tools/_ingest.py
@@ -209,13 +209,11 @@ class Ingest:
     def _init_search(self, dist_func, dist_args):
         from functools import partial
         from umap.nndescent import initialise_search
-        from umap.utils import deheap_sort
-        from umap.umap_ import INT32_MAX, INT32_MIN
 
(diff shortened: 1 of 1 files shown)


flying-sheep:
Nice! But may I ask why you’re still importing everything from umap instead of from pynndescent?

I’d assume if we’d do that we’d be more robust to further umap updates, no?

Koncopd:
Hm, it is still required to follow the internals of umap to reconstruct UMAP object from anndata info. Also this adds dependency on pynndescent.

flying-sheep:
OK. @lmcinnes, is there a better way to construct fitted umap objects from primitives?

lmcinnes:
This looks pretty reasonable for now -- the newer version tried to clean up a lot of t5hese sorts of issues, but backwards compatibility is now an issue I guess.

flying-sheep:
Great, if the new way is going to continue to work, everything’s fine!

PR #1041, opened 2020-02-13
Tests for spatial functions

giovp:


scanpy/readwrite.py
```diff
@@ -339,15 +335,23 @@ def read_visium(
...
 
-        adata.uns['images'] = dict(
-            hires=imread(str(files['hires_image'])),
-            lowres=imread(str(files['lowres_image'])),
-        )
+        adata.uns['images'] = dict()
+        try:
+            adata.uns['images']['hires'] = imread(str(files['hires_image']))
+            adata.uns['images']['lowres'] = imread(str(files['lowres_image']))
+        except:
```
flying-sheep: Don’t use plain `except:`, as it will also try to catch unrecoverable errors:

```py
>>> [
...     e for e in __builtins__.__dict__.values()
...     if type(e) is type
...     and issubclass(e, BaseException)
...     and not issubclass(e, Exception)
... ]
[BaseException, GeneratorExit, SystemExit, KeyboardInterrupt]
```

If you really want to catch everything, do `except Exception`.

flying-sheep (changes requested):

scanpy/readwrite.py
giovp: Fixed

flying-sheep (approved):

PR #1048, opened 2020-02-14
Fixed tests and read function

giovp:


giovp:
Ok, I found a workaround by subsetting the dataset to 100 obs and 100 vars and writing it back to file with this R package 😅 
https://bioconductor.org/packages/release/bioc/html/DropletUtils.html . 

This dataset now works for both `read_visium` and `pl.spatial` tests.

scanpy/readwrite.py
```diff
@@ -349,6 +349,9 @@ def read_visium(
         adata.uns['images'] = dict()
         try:
             adata.uns['images']['hires'] = imread(str(files['hires_image']))
+        except Exception:
+            pass
+        try:
             adata.uns['images']['lowres'] = imread(str(files['lowres_image']))
         except Exception:
             pass
```
flying-sheep, line 357: Ha, good catch

flying-sheep (approved):

flying-sheep:
Nice one!

PR #1063, opened 2020-02-19
Wishbone integration

awnimo:
Wishbone is an algorithm for positioning single cells along bifurcating 
developmental trajectories with high resolution.

### DIFF
--- a/scanpy/external/pl.py
+++ b/scanpy/external/pl.py
@@ -1,4 +1,4 @@
-from typing import Union, List, Optional, Any, Tuple
+from typing import Union, List, Optional, Any, Tuple, Collection
 
 import numpy as np
 import matplotlib.pyplot as plt
@@ -280,11 +280,11 @@ def wishbone_marker_trajectory(
     -------
     Updates `adata` with the following fields:
 
-    `Trunk_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
+    `trunk_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
         Computed values before branching
-    `Branch1_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
+    `branch1_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
         Computed values for the first branch
-    `Branch2_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
+    `branch2_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
         Computed values for the second branch.
     """
 
--- a/scanpy/external/pl.py
+++ b/scanpy/external/pl.py
@@ -241,7 +241,7 @@ def sam(
 @_doc_params(show_save_ax=doc_show_save_ax)
 def wishbone_marker_trajectory(
     adata: AnnData,
-    markers: List[str],
+    markers: Collection[str],
     no_bins: int = 150,
     smoothing_factor: int = 1,
     min_delta: float = 0.1,
--- a/scanpy/external/pl.py
+++ b/scanpy/external/pl.py
@@ -1,9 +1,13 @@
-from typing import Union, List, Optional, Any
+from typing import Union, List, Optional, Any, Tuple, Collection
 
+import matplotlib
 import numpy as np
-from anndata import AnnData
+import pandas as pd
+import seaborn as sns
 import matplotlib.pyplot as plt
+from anndata import AnnData
 from matplotlib.axes import Axes
+from numpy.core.umath_tests import inner1d
 
 from .._utils import _doc_params
 from ..plotting import embedding
@@ -14,6 +18,7 @@ from ..plotting._docs import (
     doc_show_save_ax,
 )
 from ..plotting._tools.scatterplots import _wraps_plot_scatter
+from ..plotting import _utils
 
 
 @_wraps_plot_scatter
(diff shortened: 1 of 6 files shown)


scanpy/external/pl.py
```diff
@@ -234,3 +239,443 @@ def sam(
...
+    smoothing_factor
+        Parameter controlling the degree of smoothing
+
+    Returns
+    -------
+    Updates `adata` with the following fields:
+
+    `weights_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
+        Computed gaussian weights for points at each location
+    `branch_point_bin` : 'int' (`adata.uns`)
```
flying-sheep:
```suggestion
    `branch_point_bin` : :class:`int` (`adata.uns`)
```

scanpy/external/pl.py
```diff
@@ -234,3 +239,443 @@ def sam(
...
+    adata
+        Annotated data matrix.
+    no_bins
+        Number of bins for calculating marker density
+    smoothing_factor
+        Parameter controlling the degree of smoothing
+
+    Returns
+    -------
+    Updates `adata` with the following fields:
```
flying-sheep: Hmm, why is something that updates adata in `pl` and not in `tl`?

scanpy/external/pl.py
```diff
@@ -234,3 +239,443 @@ def sam(
...
+                        adata.obs['branch_wishbone'][
+                            trajectory.index[
+                                (trajectory > bins[i - 1]) & (trajectory < bins[i])
+                            ]
+                        ]
+                        == j
+                    ).sum()
+                    for i in range(1, len(bins))
+                ]
+            )
```
flying-sheep: This looks weird due to the black formatting! I think doing it the other way round should be computationally more efficient and prettier. Something like:

```py
branches = np.array([
    adata.obs['branch_wishbone'].values[trajectory.index[(trajectory > a) & (trajectory < b)]]
    for a, b in pairwise(bins)
])
# or axis=0, test what’s correct
branch_counts = np.array([(branches == i).sum(axis=1) for i in [1, 2, 3]])
```

It might require too much RAM though, let me hear what you think.

scanpy/external/pl.py
```diff
@@ -234,3 +239,443 @@ def sam(
...
+        bp_bin = no_bins
+
+    # return weight to adata
+    adata.uns['weights_wishbone'] = weights
+    adata.uns['branch_point_bin'] = bp_bin
+    adata.uns['bins_wishbone'] = bins
+
+
+@_doc_params(show_save_ax=doc_show_save_ax)
+def wishbone(
```
flying-sheep: This is really huge. Did you copy that from somewhere? If this is copied from the wishbone package, just call the one that’s there instead and write a little wrapper at most!

If you wrote all this code just for Scanpy: Thank you! Wouldn’t it be useful in the `wishbone` package though?

flying-sheep (changes requested):
Hey this is looking great! On first glance all the little things are taken care of.

I’d like to know about the plotting function as described below, then I’ll take a closer look :smile:

scanpy/external/pl.py
awnimo: This code is adapted from Wishbone package and it is part of the plotting scheme within the class. I can definitely move it over to `tl._wishbone` and have it as a separate supporting code for the plotting. This code returns data that could be useful for plotting heatmaps down the road.

scanpy/external/pl.py
awnimo: Unlike other packages we have, Wishbone is actually built with 2 major class objects, one builds a data object to be used, as `adata` for Scanpy, and the second class has the core methods for wishbone, and plotting functions, that accept a Wishbone data object as input.
This plotting code is embedded and adapted from Wishbone, and requires the internal built in data object to be passed. To make it easier for Scanpy community, we decided, to adapt the code from the class in wishbone and make the proper changes such that it works with `AnnData` as input, instead. What you are suggesting to do, is actually wrapping and generating a new class object data using Wishbone and then passing this object to the internal plotting method. The internal plotting method will then generate plot, and additional data for plotting, useful for heatmaps, and stores them internally in the Wishbone data object. This is doable, but could create some confusion.

scanpy/external/pl.py
awnimo: Good catch! again this is the original code adapted from Wishbone class. But, I will definitely test your suggestion and if it returns the expected result, I will change the code. Thank you for the tip!

scanpy/external/pl.py
flying-sheep: Well, that’s easy to solve: You can just refactor the wishbone function to run with plain data instead, and make both your existing plotting function and the upcoming scanpy one a wrapper around the plain-data one. Having two diverging but partially identical functions helps nobody.

Out of curiosity: Why do you use your own data object instead of AnnData?

LuckyMD: I think wishbone may be older than `AnnData` ;)

hisplan: Hi guys,

I work with Awni, and just wanted to drop a line here to have everyone on the same page.

None of the algorithms in Pe'er Lab was designed and built with Scanpy/AnnData in mind. For most, Scanpy/AnnData simply didn't exist back then :-) On top of that, both of us wasn't part of Pe'er Lab when these algorithms were developed (thus, we are not the original authors). Awni is simply trying to help integrate our algorithms into Scanpy. If you are asking why things were done this way or that way, I'm afraid that we have no good answer. It's like we have a USB 3.0, but you want USB Type-C. I really wish we could just go to Staples and buy an adapter :-)

Anyway, thanks for all your help!

flying-sheep: It not existing is the only good reason to not use AnnData :stuck_out_tongue_winking_eye: 

In all seriousness: I think it’s better for code health to use the same code. You are already adapting the original function, so what’s keeping you from doing that inside of the original package? The only references that method has to `self` are:

- `self.branch`
- `self.scdata.data.ix[trajectory.index, marker]`

So you could simply extract the whole method into a function that takes those two as parameters, we can call that one from both `sc.external.pl.wishbone_marker_trajectory` and from `Wishbone.plot_marker_trajectory`.

That would mean 3 lines of code in scanpy and 3 additional lines of code in wishbone instead of 241 copied lines of code in scanpy.

---

All that being said: wishbone isn’t actively being maintained, so the chance of the code changing is low. So if I can trust you to come here and fix it here, too, if that’s not the case, I’m OK with you copying the dead code over here.

awnimo: I took your heed and made a simpler way to wrap the plotting function, without the need to change the original package, yet the new wrapper calls the original method for plotting, and maintains the same functionality. inside `sc.external.tl._wishbone`, I wrapped a data converter that generates a minimal dummy `wishbone` data object, that then can be passed easily to Wishbone's original plotting method in `sc.external.pl.wishbone_marker_trajectory`. It now looks much more cleaner, simpler, and most importantly `AnnData` and Scnapy friendly.

scanpy/external/pl.py
awnimo: With the new change in the latest commit, this comment no longer relevant and can be resolved.

scanpy/external/pl.py
awnimo: One more thing, the plotting of marker trajectory generates data internally that is useful for further visualization. these data are pushed back to adata, as documented in the method.

flying-sheep: Perfect! that’s awesome, thank you.

scanpy/external/pl.py
```diff
@@ -234,3 +236,90 @@ def sam(
...
+    no_bins
+        Number of bins for calculating marker density.
+    smoothing_factor
+        Parameter controlling the degree of smoothing.
+    min_delta
+        Minimum difference in marker expression after normalization to show
+        separate trends for the two branches.
+    figsize
+        width, height
+    return_fig:
```
flying-sheep:
```suggestion
    return_fig
```

scanpy/external/pl.py
```diff
@@ -234,3 +236,90 @@ def sam(
         if colorbar:
             plt.colorbar(cax, ax=axes)
     return axes
+
+
+@_doc_params(show_save_ax=doc_show_save_ax)
+def wishbone_marker_trajectory(
+    adata: AnnData,
+    markers: List[str],
```
flying-sheep:
```suggestion
    markers: Collection[str],
```

Markers needs to be `Iterable` and `Sized` (i.e. have `len()`). You can import `Collection` from `typing`

scanpy/external/pl.py
```diff
@@ -234,3 +236,90 @@ def sam(
...
+    Returns
+    -------
+    Updates `adata` with the following fields:
+
+    `Trunk_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
+        Computed values before branching
+    `Branch1_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
+        Computed values for the first branch
+    `Branch2_wishbone` : :class:`pandas.DataFrame` (`adata.uns`)
+        Computed values for the second branch.
```
flying-sheep: Farther down, you use lowercase-only names, and I think that’s the convention in the rest of scanpy too.

For easier autocompletion, those three should be lowercase as well.

scanpy/external/tl/_wishbone.py
```diff
@@ -0,0 +1,158 @@
...
+import pandas as pd
+from anndata import AnnData
+
+
+def wishbone(
+    adata: AnnData,
+    start_cell: str,
+    branch: bool = True,
+    k: int = 15,
+    components_list: list = [1, 2, 3],
```
flying-sheep:
```suggestion
    components: Iterable[int] = (1, 2, 3),
```

and later do

```py
components = list(components)
```

That way people can just pass an ndarray or tuple or whatever.

scanpy/external/tl/_wishbone.py
```diff
@@ -0,0 +1,158 @@
...
+
+    For further demonstration of Wishbone methods and visualization please follow the
+    notebooks in the package `Wishbone_for_single_cell_RNAseq.ipynb
+    <https://github.com/dpeerlab/wishbone/tree/master/notebooks>`_.\
+    """
+    try:
+        from wishbone.core import wishbone as c_wishbone
+    except ImportError:
+        raise ImportError(
+            "\nplease install wishbone:\n\n" "\thttps://github.com/dpeerlab/wishbone"
```
flying-sheep:
```suggestion
            "\nplease install wishbone:\n\n\thttps://github.com/dpeerlab/wishbone"
```

flying-sheep (changes requested):
That’s awesome, good idea with the dummy class!

Just some cosmetic changes left.

scanpy/external/pl.py
awnimo: fixed and changed to lowercase.

flying-sheep (approved):

scanpy/external/pl.py
```diff
@@ -14,6 +14,8 @@
     doc_show_save_ax,
 )
 from ..plotting._tools.scatterplots import _wraps_plot_scatter
+from ..plotting import _utils
+from .tl._wishbone import load_plot_helper
```
flying-sheep:
```suggestion
from .tl._wishbone import load_plot_helper as _load_plot_helper
```

You just need to do this (and adapt the places using it) to fix the tests. Thank you!

flying-sheep:
Actually, why is there a decorator at all?

let me do this real quick: 37e7782

flying-sheep:
hmm, you didn’t test with the new changes. please do so for the next PR, I’m trusting you with these! Check out ce06987

awnimo:
> hmm, you didn’t test with the new changes. please do so for the next PR, I’m trusting you with these! Check out [ce06987](https://github.com/theislab/scanpy/commit/ce06987e6824471d0fe22c5cc7f9faf8840bf5da)

sorry about that forgot to do this with the last changes I made!

flying-sheep:
Looks good now, thank you!

awnimo:
> Looks good now, thank you!

Thank you for the support and help!

flying-sheep:
Sure, always!

PR #1064, opened 2020-02-20
Make inner keyword argument of violin plot accessible from outside

gokceneraslan:
I noticed that there is no way to plot the "box" inside the violins using inner='box' option. This is the main idea of this PR.

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -719,12 +719,7 @@ def violin(
         g = sns.FacetGrid(obs_tidy, col=x, col_order=keys, sharey=False)
         # don't really know why this gives a warning without passing `order`
         g = g.map(
-            sns.violinplot,
-            y,
-            orient='vertical',
-            scale=scale,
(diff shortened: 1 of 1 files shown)


scanpy/plotting/_anndata.py
```diff
@@ -711,6 +711,8 @@ def violin(
     # of the violin plot (see stacked_violin code) for more info.
     if 'cut' not in kwds:
         kwds['cut'] = 0
+    if 'inner' not in kwds:
+        kwds['inner'] = None
```
ivirshup: How about: `kwds.setdefault("inner")`, to skip the `if` statement?

gokceneraslan: Do you mean `kwds.setdefault("inner", None)`?

gokceneraslan: Oh default is None already, I didn't know that. ok, makes sense :)

fidelram:
Thanks for taking care of this. Looks good.

Issue #979, opened 2020-01-09
editing output of scanpy dotplot and other figures


dylkot:
In the scanpy documentation for sc.pl.dotplot, it indicates that it returns a list of matplotlib.axes.Axes

However, this is not true, it returns a gridspec object instead.

I noticed this because I wanted to make some subtle edits to the results to enhance the figure such as changing axis labels, adding overlapping lines to delineate the marker genes for each cell type, etc. But I am struggling to do so.

I am wondering how the API intends for us to interact with the gridspec object returned to modify the figure. If I edit the code to return the figure object as well as the gridspec, I can access the axis like so

```
ax = fig.add_subplot(gs[1,0])
```

but I can't seem to overwrite the default axis labels or add new lines as commands like
```
ax.set_ylabel('new label')
```

Don't change the figure at all.

This is all Scanpy 1.4.5.post1 but it was the same for 1.4.4.post1

Thanks!

#### Versions:

scanpy==1.4.5.post1 anndata==0.6.22.post1 umap==0.4.0 numpy==1.17.2 scipy==1.3.1 pandas==0.25.1 scikit-learn==0.21.3 statsmodels==0.10.1 python-igraph==0.7.1 louvain==0.6.1

but the same things happened with scanpy==1.4.4.post1
> ...

atarashansky:
Just opened a PR to fix this. Quoting from the PR (#1069):

> Note that if you wish to modify the figure in the same jupyter notebook cell in which the plotting function is called, you should set show=False:

```
fig,ax = sc.pl.dotplot(adata,var_names,show=False)
ax.set_xlabel('test')
```

PR #1069, opened 2020-02-21
Changing scanpy.pl.dotplot to return the matplotlib figure instead of GridSpec

atarashansky:
Fixes issue #979 

Note that if you wish to modify the figure in the same jupyter notebook cell in which the plotting function is called, you should set `show=False`:

```
fig,ax = sc.pl.dotplot(adata,var_names,show=False)
ax.set_xlabel('test')
```

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -1775,7 +1775,7 @@ def dotplot(
 
     Returns
     -------
-    List of :class:`~matplotlib.axes.Axes`
+    :class:`~matplotlib.figure.Figure`
 
     Examples
     -------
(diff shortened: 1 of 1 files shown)


flying-sheep:
Nothing here returns tuples of `(Figure, Axes)`.

Plotting functions usually either return a single axes object (if `show=False`) or a figure if `return_fig=True`.

flying-sheep (changes requested):
I think it would be better if this behaved the way all plotting functions behave:

They usually either return a single Axes object (if `show=False`) or a Figure object if `return_fig=True`.

z5ouyang:
Any updates on this? scanpy=1.4.6 is still not to return matplotlib figure but still GridSpec even with "show=False". Or is there any other way to modify the figure/ax and save it after?
Thanks!

fidelram:
I am trying to address this and other issues in https://github.com/theislab/scanpy/pull/1127

In the new implementation, a dictionary of axes is returned.

fidelram:
After PR #1210 this is now possible for `dotplot`, `matrixplot` and `stacked_viloin` in master.  See [here](https://github.com/theislab/scanpy/pull/1210#issuecomment-641129949) for an example.

Using `show=False` returns a dictionary of axes and an axis can be passed to the plots.

Issue #1033, opened 2020-02-07
highly_variable_genes fails with `batch_key` and `inplace=False`


ivirshup:
Came across this while investigating #1032. 

```python
import scanpy as sc
pbmc = sc.datasets.pbmc3k_processed()
sc.pp.highly_variable_genes(pbmc, inplace=False)
```

```pytb
/Users/isaac/github/scanpy/scanpy/preprocessing/_highly_variable_genes.py:58: RuntimeWarning: invalid value encountered in log
  dispersion = np.log(dispersion)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-36b7cb4d9c0f> in <module>
      1 import scanpy as sc
      2 pbmc = sc.datasets.pbmc3k_processed()
----> 3 sc.pp.highly_variable_genes(pbmc, batch_key="louvain", inplace=False)

~/github/scanpy/scanpy/preprocessing/_highly_variable_genes.py in highly_variable_genes(adata, min_disp, max_disp, min_mean, max_mean, n_top_genes, n_bins, flavor, subset, inplace, batch_key)
    347                 ('highly_variable_intersection', np.bool_),
    348             ])
--> 349         return np.rec.fromarrays(arrays, dtype=dtypes)

/usr/local/lib/python3.7/site-packages/numpy/core/records.py in fromarrays(arrayList, dtype, shape, formats, names, titles, aligned, byteorder)
    615 
    616     if dtype is not None:
--> 617         descr = sb.dtype(dtype)
    618         _names = descr.names
    619     else:

TypeError: data type not understood
````

Separately, could we return a dataframe here?

**Comment by atarashansky** [comment]
In lines 345 - 348, a list of dtypes was getting appended instead of extended. Fixed in PR #1070 
```
dtypes.append([
    ('highly_variable_nbatches', int),
    ('highly_variable_intersection', np.bool_),
])
```

gokceneraslan:
Yeah, if you add a test, something very simple like `sc.pp.highly_variable_genes(pbmc, batch_key='louvain', inplace=False)` we can merge the PR @atarashansky.

> Separately, could we return a dataframe here?

Sure, I can do after @atarashansky's PR is merged.

ivirshup:
This seems to be fixed now.

PR #1070, opened 2020-02-21
Fixes highly_variable_genes failing with `batch_key` and `inplace=False`

atarashansky:
Fixes issue #1033

### DIFF
--- a/scanpy/preprocessing/_highly_variable_genes.py
+++ b/scanpy/preprocessing/_highly_variable_genes.py
@@ -342,7 +342,7 @@ def highly_variable_genes(
                 df['highly_variable_nbatches'].values,
                 df['highly_variable_intersection'].values,
             ])
-            dtypes.append([
+            dtypes.extend([
                 ('highly_variable_nbatches', int),
                 ('highly_variable_intersection', np.bool_),
             ])


ivirshup:
Thanks for the PR! Could you add a test to make sure it works?

atarashansky:
Do you mean to add a test for TravisCI (sorry, I'm new to this)?

ivirshup:
Yeah, there are a bunch of methods which run under CI in under the `scanpy/tests` directory. These should go in the `scanpy/tests/test_highly_variable_genes.py` file. Basically just make sure the results are what they should be. In this case, the returned object should have the same values as those added to the anndata when `inplace=True`.

We use [pytest](https://docs.pytest.org/en/latest/) to run the tests. You can run the tests locally by calling `pytest` at the root of the repo, and you can specify which tests to run with `pytest -k pattern` where `pattern` matches against the test names.

Feel free to ask if you come across any problems!

ivirshup:
@atarashansky, the performance is looking very very good:

```python
import scanpy as sc
import numpy as np
from sklearn.datasets import fetch_20newsgroups_vectorized

X = fetch_20newsgroups_vectorized("all").data
# 18846 x 130107 csr_matrix

a = sc.AnnData(X, dtype=np.float64)
%time implicit = sc.pp.pca(a, pca_sparse=True, dtype=np.float64, copy=True)
# CPU times: user 34.8 s, sys: 5.52 s, total: 40.3 s
# Wall time: 2.93 s
# Peak memory (including dataset) is about 770 MB
%time explicit = sc.pp.pca(a, pca_sparse=False, dtype=np.float64, copy=True)
# CPU times: user 55min 37s, sys: 1min 50s, total: 57min 28s
# Wall time: 7min 43s
# Peak memory is about 36 GB

assert np.allclose(implicit.obsm["X_pca"], explicit.obsm["X_pca"])
assert np.allclose(implicit.varm["PCs"], explicit.varm["PCs"])
```

But the variance and explained variance ratio are still off. Why not calculate them the same way sklearn does?

Also, any thoughts on making a PR there?

ivirshup:
Closing as the underlying issue seems to have been fixed.

PR #1073, opened 2020-02-26
multi-tissue functionality

giovp:
This is an implementation for the support of anndatas with multiple tissues (slides). It depends on the anndata PR theislab/anndata#329
I made a short notebook to explain how it should work https://github.com/giovp/spatial-scripts/blob/master/multiple_slices_functionality.ipynb

There is a lot of room for improvement in terms of code, but in terms of functionality it should cover most of the use cases


giovp:
Old PR, same as #1138

PR #1078, opened 2020-02-29
Fixed inverted axis with img_key=None

### LINKED ISSUE #1059: Flipped y-axis in sc.pl.spatial without img_key

```python
import scanpy as sc

```

giovp:
Fixes #1059, when `img_key=None` and `basis="spatial"` y axis is flipped and plotted correctly.


flying-sheep:
I think the best would be if we did this in `sc.pl.spatial`, not `sc.pl.embedding`.

giovp:
Added a quick test

flying-sheep:
It’s good to have one, but it should of course also make sure that the flipping works!

giovp:
The point of making it in `sc.pl.embedding` is that if there is no image, then there is no scaling factor and it really becomes just a scatterplot. I could use another point size, but then I think it becomes yet another different behavior. Does it make sense?

flying-sheep:
Sure, only problem is that there’s now so much spatial-specific code in `embedding`. But we can fix that after this PR.

I’d still like to see this change though:

> It’s good to have [a test], but it should of course also make sure that the flipping works!

giovp:
Replaced by #1149

PR #1079, opened 2020-03-01
Fix deprecated louvain rapids method

LouisFaure:
Hello,

GPU accelerated louvain community method call nvLouvain has been deprecated.
Now it should be run as follow: df, mod = cugraph.louvain(G)

### DIFF
--- a/scanpy/tools/_louvain.py
+++ b/scanpy/tools/_louvain.py
@@ -162,7 +162,7 @@ def louvain(
         g = cugraph.Graph()
         g.add_adj_list(offsets, indices, weights)
         logg.info('    using the "louvain" package of rapids')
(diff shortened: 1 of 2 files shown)


ivirshup (changes requested):
Thanks for the PR! Could you also add the relevant version restrictions [here](https://github.com/theislab/scanpy/blob/28498953092dc7cbecd0bd67380b1b060367d639/setup.py#L33)?

LouisFaure:
> Thanks for the PR! Could you also add the relevant version restrictions [here](https://github.com/theislab/scanpy/blob/28498953092dc7cbecd0bd67380b1b060367d639/setup.py#L33)?

Sure! I have added version restriction for cuml and cudf as well for consistency

ivirshup (approved):
Great, thanks!

PR #1081, opened 2020-03-03
Split rank_genes_groups

Koncopd:
Splits rank_genes_groups into helper functions.
Related to 1), 2) of [this](https://github.com/theislab/scanpy/issues/723#issuecomment-526079225).
I don't see any point in using dataframe internally (the second point), dict is more convenient here.


falexwolf:
Thank you! This looks great!

What do you think, @ivirshup?

Now, @Koncopd, can you also add a `logg.warn("default of method has been changed to 't-test' from 't-test_overestim_var'")`

And actually change the value?

Finally, can you test this on the Scanpy default tutorial and a a test based on the numerical values in addition to the image based tests? https://github.com/theislab/scanpy/blob/7e058a1a6a082e34a101d65fc7ac5e9cb6563220/scanpy/tests/notebooks/test_pbmc3k.py#L109-L111

Thank you very much!

Otherwise, this is good to be merged, IMO.

Koncopd:
@falexwolf , i'll do.

Koncopd:
Also there are 2 numerical tests here
https://github.com/theislab/scanpy/blob/master/scanpy/tests/test_rank_genes_groups.py
https://github.com/theislab/scanpy/blob/master/scanpy/tests/test_rank_genes_groups_logreg.py

gokceneraslan:
Guys, now that you're dissecting rank_genes_groups, do you mind adding  additional info to the results which is the fraction of cells expressing the genes (similar to what we have in dotplots) People calculate it manually ever time and it can be painful for those who are not familiar with pandas.

Koncopd:
@gokceneraslan could you point me to code or an example?

Koncopd:
Test fails seem unrelated to this PR.

gokceneraslan:
From [Seurat tutorial](https://satijalab.org/seurat/v3.0/de_vignette.html):

image

pct.1 and pct.2 are the ones I mentioned.

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,7 +18,308 @@
...
+
+        # I think it's only nan when means are the same and vars are 0
+        scores[np.isnan(scores)] = 0
+        # This also has to happen for Benjamini Hochberg
+        pvals[np.isnan(pvals)] = 1
+
+        if corr_method == 'benjamini-hochberg':
+            _, pvals_adj, _, _ = multipletests(pvals, alpha=0.05, method='fdr_bh')
+        elif corr_method == 'bonferroni':
+            pvals_adj = np.minimum(pvals * n_genes, 1.0)
```
ivirshup: Instead of doing this in the testing function (i.e. `_t_test`, `_wilcoxon`, etc.), could this be done afterwards since it's the same regardless of test type?

This would also let us remove the `corr_method` argument from these functions.

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,7 +18,308 @@
...
+                mean2=mean_rest,
+                std2=np.sqrt(var_rest),
+                nobs2=ns_rest,
+                equal_var=False,  # Welch's
+            )
+
+        # Fold change
+        foldchanges = (expm1_func(mean_group) + 1e-9) / (
+            expm1_func(mean_rest) + 1e-9
+        )  # add small value to remove 0's
```
ivirshup: Does this change for each testing function? If not, could it be separated?

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,7 +18,308 @@
...
+        # This also has to happen for Benjamini Hochberg
+        pvals[np.isnan(pvals)] = 1
+
+        if corr_method == 'benjamini-hochberg':
+            _, pvals_adj, _, _ = multipletests(pvals, alpha=0.05, method='fdr_bh')
+        elif corr_method == 'bonferroni':
+            pvals_adj = np.minimum(pvals * n_genes, 1.0)
+
+        scores_sort = np.abs(scores) if rankby_abs else scores
+        global_indices = _select_top_n(scores_sort, n_genes_user, n_genes)
```
ivirshup: Similarly, can these be split out from these functions?

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,7 +18,308 @@
...
+
+
+def _ranks_generator(X, mask=None, mask_rest=None):
+    CONST_MAX_SIZE = 10000000
+
+    n_genes = X.shape[1]
+
+    if issparse(X):
+        merge = lambda tpl: vstack(tpl).todense()
+        adapt = lambda X: X.todense()
```
ivirshup: Ideally we avoid using `np.matrix`. Could these be `X.toarray()`?

ivirshup (changes requested):
This really needed doing, thanks for diving into it!

@Koncopd, how clean would you like to get this? I've noted a few initial things which I think would make this a bit more simple, but I'd like to hear your goals here. Are you interested in doing more with the differential expression tooling?

Koncopd:
@ivirshup thanks for catching these things, i'll fix them. 
In addition to some splitting and cleaning i also want to improve wilcoxon implementation (avoid densification at least, i have some code for it). But this should be another step, i think.

Koncopd:
@gokceneraslan i'll add these things.

scanpy/tools/_rank_genes_groups.py
```diff
@@ -177,275 +387,84 @@ def rank_genes_groups(
...
-                for imask, mask in enumerate(groups_masks):
-                    scores[imask, left:right] = np.sum(ranks.loc[mask, :])
-                left = right
+        if mean_group is not None and mean_rest is not None:
+            foldchanges = (expm1_func(mean_group) + 1e-9) / (
+                expm1_func(mean_rest) + 1e-9
+            )  # add small value to remove 0's
+            if 'logfoldchanges' not in d:
+                d['logfoldchanges'] = {}
+            d['logfoldchanges'][group_name] = np.log2(foldchanges[global_indices])
```
gokceneraslan, line 434: Sweet. Does this also produce logfoldchanges for logreg? sc.get.rank_genes_groups_df throws an exception now:

image

Koncopd: Thank you for catching this. I'll check why it happens.

gokceneraslan: Oh sorry for misunderstanding. Error I posted is from stable scanpy, not from the PR. I was asking if it's fixed now.

Koncopd: This is the problem with `sc.get`. logreg only has  `names` and `scores` keys, but `sc.get` tries to query nonexistent  'logfoldchanges' and so on.

Koncopd: Should i add `logfoldchanges` for logreg?

gokceneraslan: I don't want to interrupt the PR, but I think logfoldchange and percentages should be available for every method.

Koncopd: Should the percentage for the reference group be also calculated and stored?

gokceneraslan: I think so. It might look unnecessary for one-vs-rest kind of tests but for clusterA-vs-clusterB kind of tests it'd be useful.

Koncopd:
This is still very messy and very inefficient with this separate calculation of pts. I should work on it more to restructure completely...

PR #1084, opened 2020-03-04
stacked_violin as subplot

flying-sheep:
I think it looks great!

```py
fig, axs = plt.subplots(1, 3, figsize=(15, 5))
sc.pl.rank_genes_groups_violin(adata, groups=adata.obs['leiden'].cat.categories[0], n_genes=8, ax=axs[0], show=False)
sc.pl.stacked_violin(adata, marker_genes, groupby='leiden', rotation=90, ax=axs[1], show=False)
sc.pl.umap(adata, color='leiden', title='', frameon=False, ax=axs[2], show=False)
fig.tight_layout()
````

grafik

(The huge gap is a problem of `plt.subplots`, not of this)


fidelram:
This is very nice! We should do that for all plots.

PR #915, opened 2019-11-13
`sc.metrics` module (add confusion matrix & Geary's C methods)

ivirshup:
This PR adds a module `sc.metrics` for functions which wouldn't modify an anndata object, but are useful calculations. I'm basing this on `sklearn.metrics`, namely, how `sklearn` has separated transformers (`sc.tl`) from measurements. I've started it with two functions, `confusion_matrix` and `gearys_c` but think there are more use cases (e.g. `modularity`). I'm open to this not being a module, but I think these methods should be available and I'm not sure where they'd fit within the current api.

My vision for this module is to make it easier to calculate values based on values you'd get using the scanpy ecosystem. Methods that would be included would be either *a)* not available in other libraries (`gearys_c`) or *b)* are available, but have difficult interfaces (`confusion_matrix`).

## `sc.metrics.confusion_matrix`

Creates a confusion matrix for comparing categorical labels. This is based on `sklearn.metrics.confusion_matrix` but is easier to use, and returns a object with labels. I think this is mostly done, though I'm considering changing the calling convention. Here's an example of usage:

```python
import scanpy as sc
import seaborn as sns

pbmc = sc.datasets.pbmc68k_reduced()
sc.tl.leiden(pbmc)
sns.heatmap(sc.metrics.confusion_matrix("bulk_labels", "leiden", pbmc.obs))
```

image

I've copied `seaborn`s calling convention here, but I think that could change. Right now the above call is equivalent to:

```python
sc.metrics.confusion_matrix(pbmc.obs["bulk_labels"], pbmc.obs["louvain"])
```

But I wonder if it would make more sense to have the DataFrame go first if it's provided. I've also based the API around my usage of confusion matrices, so I'm very open to more general feedback on this. My reason for including it here was the amount of code it took wrapping `sklearn.metrics.confusion_matrix` to get useful output.

## `sc.metrics.gearys_c` ([Wiki page](https://en.wikipedia.org/wiki/Geary%27s_C))

Calculates autocorrelation on a measure on a network. Used in [VISION](https://doi.org/10.1038/s41467-019-12235-0) for ranking gene sets. This is useful for finding out whether some per-cell measure is correlated with the structure of a connectivity graph. In practice, I've found it useful for identifying features that look good on a UMAP:

```python
import numpy as np
pbmc.layers["logcounts"] = pbmc.raw.X

%time gearys_c = sc.metrics.gearys_c(pbmc, layer="logcounts")
[...]

PR #1088, opened 2020-03-08
Fixes for visium reading

ivirshup:
Ping @flying-sheep @giovp @Mirkazemi

Turns out that the `sc.datasets.visium_sge` would just read in whatever images were most recently added, not the one that fit the dataset. This fixes that.

Additionally, `sc.read_visium` now takes a directory as the first argument. If a reading function assumes a directory structure, that directory should be passed, not a specific path inside of it. This follows our other functions like: `sc.read_10x_mtx`.

Similarly, I've rearranged how the example datasets are stored and how the test data is stored to better match the outputs from 10x pipelines.

A few more changes I'd like to make:

* Restructure how elements are added to `uns`, as mentioned in https://github.com/theislab/anndata/issues/295#issuecomment-596164456
* Rename `obsm["X_spatial"]` to `obsm["coords"]` or `obsm["spatial"]`.
* There is a natural connectivity for the observations from the adjacency of wells. This should be easy to add to obsp, or should just be added to obsp when `read_visum` is called. I'm thinking `"spatial_connectivity"` for the default key name.


giovp:
That looks great Isaac, thanks a lot!

> * Restructure how elements are added to `uns`, as mentioned in [theislab/anndata#295 (comment)](https://github.com/theislab/anndata/issues/295#issuecomment-596164456). 

agree, I also think that it could be best in the case of different data types (not visium). It's also best if the tree structure does not change in case of concatenation.

> * Rename `obsm["X_spatial"]` to `obsm["coords"]` or `obsm["spatial"]`.  

what about `X_coords` ?

> * There is a natural connectivity for the observations from the adjacency of wells. This should be easy to add to obsp, or should just be added to obsp when `read_visum` is called. I'm thinking `"spatial_connectivity"` for the default key name.  

This was also in the plans but further down the line (support for multiple slices should be first). I could have a look at this soon.

What about re-open the `theislab/spatial` branch and merge this PR there? I could then work on how to handle the new `uns` structure in the plotting functions and have a definitive version of multiple slices support in anndata.

ivirshup:
> what about `X_coords` ?

Ha, I was mostly just trying to get rid of the `X_`!

> What about re-open the theislab/spatial branch and merge this PR there? I could then work on how to handle the new uns structure in the plotting functions and have a definitive version of multiple slices support in anndata.

I'd like to merge the changes currently in this PR to master since it fixes a bug with dataset reading. The changes to uns structure could go in another PR, but I'm waiting for an email back from 10x to make sure using the `library_id` as a key makes sense. Either way, the logic of getting the transformed coordinates etc. should be abstracted into a function so it's easy to change.

Update: heard back, the `library_id` should be fine, at least for this version.

> support for multiple slices should be first

I'm not sure I'm convinced of this. I've also already got some code ready to go for the connectivities and some examples of what can be done with it.

I'd like to hear what kind of stuff you want to be able to do with multiple slices. Are you interested in stitching together slides or holding arbitrary slides in an AnnData? I think I'd like to see a more fleshed out idea of what kinds of analysis could be done here before deciding on what kind of an API this should have, and cases we should be ready to handle.

Also, I think spatial plotting code should get moved out of `sc.pl.embedding` before we allow plotting multiple slides at a time.

giovp:
> > what about `X_coords` ?
> 
> Ha, I was mostly just trying to get rid of the `X_`!  

ah right, anyway good for me!
> 
> > What about re-open the theislab/spatial branch and merge this PR there? I could then work on how to handle the new uns structure in the plotting functions and have a definitive version of multiple slices support in anndata.
> 
> I'd like to merge the changes currently in this PR to master since it fixes a bug with dataset reading. The changes to uns structure could go in another PR, but I'm waiting for an email back from 10x to make sure using the `library_id` as a key makes sense. Either way, the logic of getting the transformed coordinates etc. should be abstracted into a function so it's easy to change.
>   

What do you mean by transformed coordinates? Also, to understand the inputs for anndata (output of spaceranger) you might have a look at this, if you are not already familiar with https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/output/overview. 
Also, ok for having `uns` changes in another PR, I can work on that as soon as this is merged.
> Update: heard back, the `library_id` should be fine, at least for this version.
>   

good !

> > support for multiple slices should be first
> 
> I'm not sure I'm convinced of this. I've also already got some code ready to go for the connectivities and some examples of what can be done with it.
> 
> I'd like to hear what kind of stuff you want to be able to do with multiple slices. Are you interested in stitching together slides or holding arbitrary slides in an AnnData? I think I'd like to see a more fleshed out idea of what kinds of analysis could be done here before deciding on what kind of an API this should have, and cases we should be ready to handle.
>   

support for multiple slices and concatenation of anndata objects is by far the priority to me. It's a really useful functionality since:
* most people don't work with one slide
* having the same anndata object containing scRNA-seq as well as matched visium tissue would allow for a very straightforward approach to integration and label propagation (with ingest/bbknn). This would also be extremely useful for the tutorial (which I can't update until anndata supports multiple tissues).

I am very interested to see the applications of spatial connectivities you think can be useful. I see the potential but I don't think it's straightforward to make use of that info (especially because in essence the spatial graph derived from visium is completely homogeneous, hence lack of structure).
 
> Also, I think spatial plotting code should get moved out of `sc.pl.embedding` before we allow plotting multiple slides at a time.

Why is that? `sc.pl.spatial` is essentially a scatterplot that calls `sc.pl.embedding` yet using another method (circles instead of scatter, but inherits all the arguments)

ivirshup:
> What do you mean by transformed coordinates?

Don't you have to use a scaling factor for the coordinates to map them onto the image? And the maybe flip either the image or the coordinates?

> I am very interested to see the applications of spatial connectivities you think can be useful.

A few uses I can think of:

* Select all wells which are in cluster 1 and border cluster 0

<details>
<summary> Example code </summary>

```python
g = sc._utils.get_igraph_from_adjacency(adata.obsp["spatial_connectivity"])
g.vs["obs_names"] = adata.obs_names
g.vs["leiden"] = adata.obs["leiden"]

(
    g.es
    .select(
        _between=(
            g.vs.select(leiden_eq="0"),
            g.vs.select(leiden_eq="1")
        )
    )
    .subgraph().vs
    .select(leiden_eq="1")
    ["obs_names"]
)
# Alternative, but equivalent:

adata.obs_names[
    (
        adata.obsp["spatial_connectivity"] @
        (adata.obs["leiden"] == "0")
    ) & 
    (adata.obs["leiden"] == "1")
]
```

</details>

* Feature selection for genes which have more similar values in their neighbouring wells than would be expected by random (this works pretty well for NMF components too)

<details>
<summary> Example code </summary>

From #915

```python
gearys_c(adata.obsp["spatial_connectivity"], adata.X.T)
```

</details>

* Select wells where some feature is expressed in a gradient (not sure what the elegant way to do this one is)

--------------

For multiple spatial objects, I think that makes sense. Having the ability to concatenate makes sense as a priority, I was thinking more of functions that are aware of multiple tissues and having special behaviour for that w.r.t. their images. That's the stuff where I'd like have more confidence in the API before it's put in.

giovp:
> Don't you have to use a scaling factor for the coordinates to map them onto the image? And the maybe flip either the image or the coordinates?  

Yes, this is currently done in `sc.pl.spatial` because it requires first to select the type of image to be plotted.

> A few uses I can think of:
> 
> * Select all wells which are in cluster 1 and border cluster 0
> 
>  Example code
> * Feature selection for genes which have more similar values in their neighbouring wells than would be expected by random (this works pretty well for NMF components too)
> 
>  Example code
> * Select wells where some feature is expressed in a gradient (not sure what the elegant way to do this one is)
> 

Thanks, all really interesting applications indeed. I guess I was more thinking of analysis on the spatial graph (e.g. clustering).

Ok so just to double-check, I will wait for this to get merge and then go on with `uns` and `sc.pl.spatial` fixes, makes sense?

ivirshup:
@giovp sounds good.

Would you want any of the `uns` in a PR to your spatial branch?

PR #1090, opened 2020-03-09
Matplotlib v3.2 fixes

ivirshup:
Matplotlib 3.2 seems to have broken a number of our tests. This is me trying to update them, though I've had trouble reproducing figures compared with travis in the past.

# Heatmaps

For heatmaps, the update was real bad:

blurry_heatmap

# Violin plots

Violin plots have changed as well, but I think it's for the better. They seem less likely to show density where there are no data points. In practice, where `n_genes` < 1000 was used as a cutoff, this looks like:

new_violin

Axis ticks also changed due to this.

# Everything else

For everything else that changed, I couldn't tell the difference. I think plots just moved up or down a bit, and that made them report as changed.


ivirshup:
I'd appreciate a quick review from @fidelram or @flying-sheep on this. It'd be nice to get this soon since the tests failing are blocking the merging of other PRs.

ivirshup:
Since this is blocking me from merging other PRs I'm going to go ahead and merge this.

PR #1102, opened 2020-03-15
Fix ebi downloads

ivirshup:
Should actually fix downloading EBI datasets. URLs for tables and count matrices had changed, and this PR fixes that.

This is a bit of a band-aid, since we should probably be accessing these files through the FTP and we should be checking that already downloaded datasets are current.

I've also only done a spot check on this, and tested it with a couple datasets. It's probably worth testing with a larger set.


ivirshup:
Did some more spot checks, seems to work.

PR #1105, opened 2020-03-15
change uns structure for spatial

giovp:
New version of `uns` as discussed in https://github.com/theislab/anndata/issues/295#issuecomment-596164456

The reason for setting a dummy `library_id` (as "0" for instance) is:
* The `library_id` information is in the molecule_info file [explained here](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/output/molecule_info). Recover it would require to input an additional file path
* All the 10x data I had a look at, both from the same tissue (so 2 slides from same tissue slice) and from different tissues, have `library_id` entry set to "0". So I am not really sure how space ranger set that value but it does not appear to be unique (and therefore not a natural `batch_key` value).
* I think it would be more useful if this value is set according to user choice. Only in the context of `adata.concatenate` it should be modified according to `batch_key`. This is also the only point in the analysis where the `library_id` entry matter.
* I agree with respect to maintaining the tree structure before and after concatenation, so the reason for keeping the `library_id` entry and setting it to a dummy variable by default.

Looking forward to hear what you think @ivirshup and if agree I'll go on with PRs for `anndata.concatenate`. 

Also, let's keep this `spatial` branch open until we really have (almost) everything up and running for spatial analysis.


scanpy/readwrite.py
```diff
@@ -351,15 +357,17 @@ def read_visium(
...
+                    str(files[f'{res}_image'])
+                )
             except Exception:
                 pass
 
         # read json scalefactors
-        adata.uns['scalefactors'] = json.loads(
+        adata.uns["spatial"][library_id]['scalefactors'] = json.loads(
             files['scalefactors_json_file'].read_bytes()
         )
```
ivirshup, line 385:
```suggestion
    adata.uns["spatial"][library_id]["metadata"] = {
        k: attrs[k] for k in ("chemistry_description", "software_version")
    }

```

For adding some metadata

scanpy/readwrite.py
```diff
@@ -332,6 +335,9 @@ def read_visium(
     path = Path(path)
     adata = read_10x_h5(path / count_file, genome=genome)
 
+    adata.uns["spatial"] = dict()
+    adata.uns["spatial"][library_id] = dict()
```
ivirshup, line 351: To get library id (if it holds that it's always present)

```suggestion
    with h5py.File(path / count_file) as f:
        attrs = dict(f.attrs)
    if library_id is None:  # if None was the default value
        library_id = str(attrs.pop("library_id")[0])

    adata.uns["spatial"][library_id] = dict()
```

ivirshup (commented):
> The library_id information is in the molecule_info file

Two points on this:

1. I think we are essentially requiring the entire `outs` directory to read in 10x data, so I don't think requiring another file would be so bad.
2. For every `filtered_feature_bc_matrix.h5` file I have (even normal cellranger ones), the library id has been available in the root group's attributes. Could you check your files to see if this works?

```python
import h5py

with h5py.File(your_feature_mtx.h5, "r") as f:
    print(f.attrs["library_ids"])
```

While `f.attrs["library_ids"]` is an array, I was told by a 10x rep that for spaceranger outputs there should only be one entry. This is because there is no `aggr` command for space ranger.

The `attrs` should also hold information about the chemistry used, and the software used to process the data. This is what I was thinking could go under the `"metadata"` key.

> I think it would be more useful if this value is set according to user choice

We could allow overwriting it, but I think this value was kinda already set by the user. It should be the value of the `--id` argument to `spaceranger`.

> Only in the context of adata.concatenate it should be modified according to batch_key

I'm not sure I'm following. Why would it be modified? I don't think that's logic that should go in anndata.

If I'm thinking about the same problem, I was thinking that the `library_id` would be added as a column in `obs` to the AnnData created by `read_visum`.

>  I'll go on with PRs for anndata.concatenate.

I'm interested to see what your plan is for that. I was thinking this could be as simple as allowing uns to be updated an option like `xarray`'s `no_conflicts` or `override`, since the set of keys should be unique.

There shouldn't be any special code for dealing with spatial data in the anndata concatenation, so feel free to open that whenever.

giovp:
Thanks for the H/T @ivirshup ! I was able to find the library ids and I also added the metadata as suggested.

This should do for now. I have also changed the plotting function to make it work with the new `uns` structure.

The idea for uns concatenation is exactly that one yes. Basically, if the keys are unique, then concatenate, if they are the same, override and throw a warning.

With respect to mixed anndata objects (e.g. one visium adata concatenated with one scRNA-seq), I will just concatenate the obsm and add empty entries to the one missing (like zeros) or something along the lines of masked arrays (although I don't think it's particularly useful in this case).

ivirshup:
No problem!

> The idea for uns concatenation is exactly that one yes. Basically, if the keys are unique, then concatenate, if they are the same, override and throw a warning.

I was thinking that this should have multiple modes, chosen by an argument like `merge_uns`. I'm thinking options would be:

* `None`: the default. Maintain current behaviour of just not merging.
* `"unique"`: Only keep values which are uniquely specified
* `"identical"`: Only keep values which are the same in all objects
* `"override"`: Just take the first value from each

You wouldn't have to implement all of these, just one that makes spatial concatenation work for now.

> With respect to mixed anndata objects (e.g. one visium adata concatenated with one scRNA-seq), I will just concatenate the obsm and add empty entries to the one missing (like zeros) or something along the lines of masked arrays (although I don't think it's particularly useful in this case).

Could there be a `fill_value ` argument here? A way for people to specify what the fill value should be?

I'm mainly against masked arrays since I don't think they're going to work with sparse matrices, and I'm not sure about other array subtypes.

giovp:
Some tests are still failing, but not because of `uns/spatial`. They all throw errors along these lines:
```
assert 'Error: Image files did not match.\n  RMS Value: 15.114361035293829\n
```

ivirshup:
@giovp or @flying-sheep, any idea how to get travis to run this? I'd like to see what's happening there.

Also @giovp, were you going to change the coordinates `obsm` entry key `"coords"`?

giovp:
I did not change the basis to `coords` because a lot of the time the basis is specified as
```python
elif f"X_{basis}" in adata.obsm.keys():
        basis_key = f"X_{basis}"
```
hence maintaining the `X` makes it consistent with the others.

ivirshup:
Is that in embedding? I had thought we'd removed all that. There shouldn't be any special meaning to strings starting with `X_` anyways, it's just a convention.

I'd like to give this a review, but I'd need to see that tests are passing. Do you know why travis isn't running on this?

giovp:
Just re run it, [tests passed](https://travis-ci.org/github/theislab/scanpy/builds/663089197?utm_medium=notification&utm_source=github_status)
No sure why it does not appear on github

giovp:
> Is that in embedding? I had thought we'd removed all that. There shouldn't be any special meaning to strings starting with `X_` anyways, it's just a convention.
> 
Well but until it has not changed I would leave it as is no?

scanpy/readwrite.py
```diff
@@ -379,10 +379,22 @@ def read_visium(
...
+                    )
+                # this statement is necessary for the subsampled dataset in tests
+                elif isinstance(attrs[k], np.ndarray):
+                    adata.uns["spatial"][library_id]["metadata"][k] = str(
+                        attrs[k][0], "utf-8"
+                    )
+                else:
+                    adata.uns["spatial"][library_id]["metadata"][k] = attrs[k]
+            except Exception:
+                pass
```
ivirshup: Could this just be:

```python
{
    k: (str(attrs[k], "utf-8") if isinstance(attrs[k], bytes) else attrs[k])
    for k in ("chemistry_description", "software_version") if k in attrs
}
```

Also, probably better to just add the attributes to the test data rather than change the code in the library.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -1066,15 +1078,16 @@ def _basis2name(basis):
 def _process_image(adata, data_points, img_key, crop_coord, scale_spot, bw=False):
     offset = 100
     cmap_img = None
-    img = adata.uns['images'][img_key]
+    library_id = list(adata.uns['spatial'].keys())[0]
```
ivirshup: Why is this always the first one?

Same with all the other places

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -1066,15 +1078,16 @@ def _basis2name(basis):
...
+    img = adata.uns['spatial'][library_id]['images'][img_key]
```
ivirshup:
```suggestion
    spatial_data = adata.uns["spatial"][library_id]
    img = spatial_data['images'][img_key]
```

And then the same for the other long paths below too.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -756,9 +756,16 @@ def spatial(
...
+
     if img_key is _empty:
         img_key = next(
-            (k for k in ['hires', 'lowres'] if k in adata.uns['images']), None
+            (
+                k
+                for k in ['hires', 'lowres']
+                if k in adata.uns['spatial'][library_id]['images']
+            ),
+            None,
```
ivirshup: What's the intended behavior here? There's a priority for `hires`, `lowres`, and then what happens?

Also, what's the case where `"hires"` doesn't exist?

scanpy/readwrite.py
```diff
@@ -351,18 +365,37 @@ def read_visium(
...
 
-        adata.uns['images'] = dict()
+        adata.uns["spatial"][library_id]['images'] = dict()
         for res in ['hires', 'lowres']:
             try:
-                adata.uns['images'][res] = imread(str(files[f'{res}_image']))
+                adata.uns["spatial"][library_id]['images'][res] = imread(
+                    str(files[f'{res}_image'])
+                )
             except Exception:
```
ivirshup, line 378: This should be more specific. Also, it's pretty bad if you tried to read in a spatial dataset, and it failed right? Why shouldn't this just error?

ivirshup (changes requested):
That travis check wasn't for the latest commit. It was for https://github.com/theislab/scanpy/commit/dcb8e751c9125ce44f2d30152ae2ebab1fb456b0, but the latest commit is https://github.com/theislab/scanpy/pull/1105/commits/24d754ecb52fa7bd1b49d889a9986298b93b3d03.

If you change `X_coords` to `coords`, where do errors occur? Could you also point me to where you're seeing this code? I've definitely been using embeddings in `obsm` whose keys don't start with `"X_"`.

scanpy/readwrite.py
giovp: Done

scanpy/readwrite.py
giovp: Technically no, it could still be the case that there is no image but spatial coordinates. It would still make sense to read it in as visium data, excetp the image entry would be empty.

scanpy/plotting/_tools/scatterplots.py
giovp: Yeah, priority completely arbitrary, this behaviour kicks in only if there is no `img_key` specified, so that at least it tried to plot an image below. Otherwise it's just None and it does not plot any image, only the points in spatial coordinates.

Also, these additions were done only to make the tests pass.

scanpy/plotting/_tools/scatterplots.py
giovp: Because is the only one, the dict is structured like this:
```python
{"spatial":{"library_id":{#everything else}}}
```

giovp: Btw, this is also related to the proposed version of how concatenation should work in theislab/anndata#295

scanpy/readwrite.py
giovp: Thank you for the suggestion

giovp:
> If you change `X_coords` to `coords`, where do errors occur? Could you also point me to where you're seeing this code? I've definitely been using embeddings in `obsm` whose keys don't start with `"X_"`.

Right, my bad, switched to that.

scanpy/plotting/_tools/scatterplots.py
ivirshup: But isn't the point of concatenation that there might two or more?

scanpy/plotting/_tools/scatterplots.py
ivirshup: What test was failing?

scanpy/readwrite.py
ivirshup: So it's possible to run the pipeline without images? How do the images get to the pipeline?

I'm wondering if the number and names of the image files might be variable.

ivirshup: It looks to me like these images [have to exist and have to have these names](https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/output/images).

giovp: Yeah that's true, I thought it's just better to give such flexibility for whatever reason (e.g. shared data with missing stuff) but can change to raise an error

scanpy/plotting/_tools/scatterplots.py
giovp: Done

scanpy/plotting/_tools/scatterplots.py
giovp: test_embedding.py

scanpy/plotting/_tools/scatterplots.py
giovp: Right yes. 
I definitely will have to rethink how this is handled for plotting. For now I've decided to just take the first one if no library_id is given. Had to add an additional argument in embedding for that. Should work but I'm also not very happy on the default behaviour, but for now it's fine I think.

scanpy/readwrite.py
ivirshup: I think this should raise an error. Since this function is specifically for reading visium experiments it reduces complexity a lot to just assume they ran the visium pipeline. Anything more homebrew will also take a more homebrew solution for reading in.

giovp: Done

giovp: One problem is that I had to add the `hires` image to the `_data` folder for test. That image is ~5 Mb. This is potentially a problem considering that now it accounts for half of the `_data` size

ivirshup: This example data has been subset, right? Could you share the script used to generate it?

Could the size of the image be reduced by removing unimportant data? I'm thinking setting unused areas of the image to a solid color, so they'll compress away.

Alternatively, does the image have to be real?

giovp: I thought about it and in fact it doesn't really, I just pushed an image with same resolution but not colors. Should be fine for testing.

yes, the dataset is a subset, and it's been created with a combination of [dropletutils](https://github.com/MarioniLab/DropletUtils) functions and h5py. This was the easiest way I found to subset the dataset and keep attributes in place.

giovp:
Hi @ivirshup this is ready to be reviewed again.
- Used lightweight figure for testing
- Changed docstrings

Tests failing are not related to anything addressed by this PR.

giovp:
Hi @ivirshup this would be ready for review. 
Travis test are failing for some figure in diffusion maps, not sure why though :(

Thank you!

ivirshup (changes requested):
I think this looks good!

The test failures should go away if you rebase onto master. Could you also rebase so that all of the commits where the data and images change are squashed into one? It'll reduce the size of the git history.

PR #1109, opened 2020-03-17
[WIP] Simplify plotting

ivirshup:
@flying-sheep This is generally the kind of simplification I was hoping we could do with plotting. 
It's not much, and is more about the dotplot, heatmap, etc. plotting methods. Also, the test errors I was running into are still happening.

Another example would using a function to choose representations of X the same way for each function. Something like:

```python
def _choose_obs_rep(adata, *, use_raw=False, layer=None, obsm=None, obsp=None):
     """
     Choose array aligned with obs annotation.
     """
     is_layer = layer is not None
     is_raw = use_raw is not False
     is_obsm = obsm is not None
     is_obsp = obsp is not None
     choices_made = sum((is_layer, is_raw, is_obsm, is_obsp))
     assert choices_made <= 1
     if choices_made == 0:
         return adata.X
     elif is_layer:
         return adata.layers[layer]
     elif use_raw:
         return adata.raw.X
     elif is_obsm:
         return adata.obsm[obsm]
     elif is_obsp:
         return adata.obsp[obsp]
     else:
         assert False, (
             "That was unexpected. Please report this bug at:\n\n\t"
             " https://github.com/theislab/scanpy/issues"
         )
```


flying-sheep:
Closing in favor of #1116, where I integrated the commit

PR #1111, opened 2020-03-17
Change the default number of nearest neighbors search in Ingest

Koncopd:
It was configurable with the default `k=10` before, now it uses n_neighbors from `sc.tl.neighbors`.
As discussed with @falexwolf .


Koncopd:
@falexwolf 
Should i merge it?

ivirshup:
Could this get a test?

Also, should this be documented somewhere?

falexwolf (approved):

PR #1112, opened 2020-03-17
Docs 1.4.6

falexwolf:


docs/conf.py
```diff
@@ -145,5 +145,6 @@ def setup(app):
     # Currently undocumented: https://github.com/mwaskom/seaborn/issues/1810
     ('py:class', 'seaborn.ClusterGrid'),
     # Won’t be documented
+    ('py:class', 'scanpy.plotting._utils._AxesSubplot'),
```
falexwolf, line 148: hope this will fix the docs!

flying-sheep: yeah! I really didn’t find a good solution for this, since there’s no class for this in matplotlib, it’s basically an intersection type (anonymous subclass) of both `Axes` and `SubplotBase`.

PR #1113, opened 2020-03-17
try fixing h5py error

flying-sheep:
https://stackoverflow.com/questions/51593527/oserror-unable-to-open-file-unable-to-open-file

/edit: that was not it. I’ll set up a Python 3.7 venv and look if I can reproduce it locally

/edit: nope, cleaning caches …


flying-sheep:
cleaning caches did the trick. let’s see if the absolute path is also needed (rebuilding master): https://travis-ci.org/github/theislab/scanpy/builds/663442852

/edit: it did

PR #1118, opened 2020-03-18
pp.neighbors saves to obsp and uses key_added argument

Koncopd:
It is now possible to have several neighbors graphs in adata.

For example,

```py
sc.pp.neighbors(adata)
sc.pp.neighbors(adata, use_rep='some', key_added='test')
sc.pp.neighbors(adata, use_rep='some1', key_added='test1')
sc.pp.neighbors(adata, key_added='test2')

sc.tl.umap(adata, neighbors_key='test')
sc.tl.diffmap(adata, neighbors_key='test1')
```

and so on.

### DIFF
--- a/scanpy/_utils.py
+++ b/scanpy/_utils.py
@@ -701,16 +701,16 @@ class NeighborsView:
             return key in self._neighbors_dict
 
 
-def _choose_graph(adata, obsp_key, neighbors_key):
+def _choose_graph(adata, obsp, neighbors_key):
     """Choose connectivities from neighbbors or another obsp column"""
-    if obsp_key is not None and neighbors_key is not None:
+    if obsp is not None and neighbors_key is not None:
         raise ValueError(
-            'You can\'t specify both obsp_key, neighbors_key. '
+            'You can\'t specify both obsp, neighbors_key. '
             'Please select only one.'
         )
 
-    if obsp_key is not None:
-        return adata.obsp[obsp_key]
+    if obsp is not None:
+        return adata.obsp[obsp]
     else:
         neighbors = NeighborsView(adata, neighbors_key)
         if 'connectivities' not in neighbors:
--- a/scanpy/_utils.py
+++ b/scanpy/_utils.py
@@ -570,3 +570,42 @@ def lazy_import(full_name):
         # Make module with proper locking and get it inserted into sys.modules.
         loader.exec_module(module)
         return module
+
+
+# --------------------------------------------------------------------------------
+# Neighbors
+# --------------------------------------------------------------------------------
+
+class NeighborsView:
+    def __init__(self, adata, key=None):
+        self._connectivities = None
+        self._distances = None
+
+        if key is None or key == 'neighbors':
+            self._neighbors_dict = adata.uns['neighbors']
+            if 'connectivities' in adata.obsp and 'distances' in adata.obsp:
+                self._connectivities = adata.obsp['connectivities']
+                self._distances = adata.obsp['distances']
+            else:
+                self._connectivities = self._neighbors_dict['connectivities']
+                self._distances = self._neighbors_dict['distances']
+        else:
+            self._neighbors_dict = adata.uns[key]
+            self._connectivities = adata.obsp[self._neighbors_dict['connectivities_key']]
+            self._distances = adata.obsp[self._neighbors_dict['distances_key']]
+
+    def __getitem__(self, key):
+        if key == 'distances':
+            return self._distances
+        elif key == 'connectivities':
+            return self._connectivities
+        else:
+            return self._neighbors_dict[key]
+
+    def __contains__(self, key):
+        if key == 'distances':
+            return self._distances is not None
+        elif key == 'connectivities':
+            return self._connectivities is not None
+        else:
+            return key in self._neighbors_dict
--- a/scanpy/_utils.py
+++ b/scanpy/_utils.py
@@ -583,16 +583,25 @@ class NeighborsView:
 
         if key is None or key == 'neighbors':
             self._neighbors_dict = adata.uns['neighbors']
-            if 'connectivities' in adata.obsp and 'distances' in adata.obsp:
-                self._connectivities = adata.obsp['connectivities']
-                self._distances = adata.obsp['distances']
-            else:
-                self._connectivities = self._neighbors_dict['connectivities']
-                self._distances = self._neighbors_dict['distances']
+            conns_key = 'connectivities'
+            dists_key = 'distances'
+
+            if conns_key in adata.obsp:
+                self._connectivities = adata.obsp[conns_key]
+            elif conns_key in self._neighbors_dict:
+                self._connectivities = self._neighbors_dict[conns_key]
+
+            if dists_key in adata.obsp:
+                self._distances = adata.obsp[dists_key]
+            elif dists_key in self._neighbors_dict:
+                self._distances = self._neighbors_dict[dists_key]
         else:
             self._neighbors_dict = adata.uns[key]
-            self._connectivities = adata.obsp[self._neighbors_dict['connectivities_key']]
-            self._distances = adata.obsp[self._neighbors_dict['distances_key']]
+            conns_key = self._neighbors_dict['connectivities_key']
+            dists_key = self._neighbors_dict['distances_key']
+            
+            self._connectivities = adata.obsp[conns_key]
+            self._distances = adata.obsp[dists_key]
 
     def __getitem__(self, key):
         if key == 'distances':
--- a/scanpy/_utils.py
+++ b/scanpy/_utils.py
@@ -681,3 +681,23 @@ class NeighborsView:
             return self._connectivities is not None
         else:
             return key in self._neighbors_dict
+
+
+def _choose_graph(adata, obsp_key, neighbors_key):
+    """Choose connectivities from neighbbors or another obsp column"""
+    if obsp_key is not None and neighbors_key is not None:
+        raise ValueError(
+            'You can\'t specify both obsp_key, neighbors_key. '
+            'Please select only one.'
+        )
+
+    if obsp_key is not None:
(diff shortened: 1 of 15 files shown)


scanpy/_utils.py
```diff
@@ -570,3 +570,51 @@ def lazy_import(full_name):
         # Make module with proper locking and get it inserted into sys.modules.
         loader.exec_module(module)
         return module
+
+
+# --------------------------------------------------------------------------------
+# Neighbors
+# --------------------------------------------------------------------------------
+
+class NeighborsView:
```
ivirshup, line 621: Could you explain what this class is meant to be used for?

scanpy/_utils.py
```diff
@@ -570,3 +570,51 @@ def lazy_import(full_name):
...
+
+            if conns_key in adata.obsp:
+                self._connectivities = adata.obsp[conns_key]
+            elif conns_key in self._neighbors_dict:
+                self._connectivities = self._neighbors_dict[conns_key]
+
+            if dists_key in adata.obsp:
+                self._distances = adata.obsp[dists_key]
+            elif dists_key in self._neighbors_dict:
+                self._distances = self._neighbors_dict[dists_key]
```
ivirshup: I was thinking we could just pin the new version of scanpy to the new version of anndata, so we would be assuming everything is always in `obsp` and `anndata` is enforcing that. Otherwise, the stuff I'm adding could break pretty trivially.

ivirshup (changes requested):
Thanks for getting this together so quickly!

It's pretty late for me, so I'll give a brief review now, and try to give a more thorough one in the morning.

Main change: tests please! Mainly things like calling `sc.pp.neighbors` with and without `key_added` and making sure the results are exactly the same.

scanpy/_utils.py
Koncopd: This is a helper class to avoid retrieving different keys for connectivities, distances, dictionary keys and so on. It is not meant to prevent problems with deprecated things in `uns['neighbors']`. 
I would say it is just easier to write
```
neighbors = NeighborsView(adata, neighbors_key)

neighbors['distances']
neighbors['connectivities']
neighbors['params']
'connectivities' in neighbors
```
than something like 
```
neighbors_dict = neighbors[neighbors_key]
conns_key = neighbors_dict['connectivities_key']
dists_key = neighbors_dict['distances_key']

adata.obsp[conns_key]
adata.obsp[dists_key]
neighbors_dict['params']
conns_key in adata.obsp
```
every time.

I can remove the checks if you think they are problematic.

scanpy/_utils.py
Koncopd: This checks are added to prevent problems from external modules that change `.uns['neighbors']` like bbknn. This is not meant to catch all problems with deprecated things certainly.

Koncopd:
@ivirshup 
Thanks, i'll add tests ofc.

scanpy/_utils.py
ivirshup: So, generally it's a convenience class for getting accessing graph representations? Could you add this description to a doc-string for the class?

I think it's okay to check if the keys exist, but it shouldn't fall back to checking in `uns`.

scanpy/_utils.py
ivirshup: Where does bbknn store it's values?

If it's as `connectivities` or `distances` in `adata.uns["neighbors"]`, those values will just be placed in `uns` by anndata.

Koncopd:
Without fallback to .uns this fails on every test which uses `sc.datasets.` with neighbors precomputed.

scanpy/_utils.py
```diff
@@ -98,11 +99,11 @@ def deprecated_arg_names(arg_mapping: Mapping[str, str]):
...
         Mapping from deprecated argument name to current argument name.
     """
+
     def decorator(func):
         @wraps(func)
         def func_wrapper(*args, **kwargs):
-            warnings.simplefilter(
-                'always', DeprecationWarning)  # turn off filter
+            warnings.simplefilter('always', DeprecationWarning)  # turn off filter
             for old, new in arg_mapping.items():
```
ivirshup, line 107: Why is this being reformatted? We use lines of length 90 now, right?

ivirshup: Also, could you point me to the code you changed in this file? It's hard to tell with all the formatting changes.

scanpy/tools/_diffmap.py
```diff
@@ -25,6 +31,14 @@ def diffmap(adata: AnnData, n_comps: int = 15, copy: bool = False):
         Annotated data matrix.
     n_comps
         The number of dimensions of the representation.
+    neighbors_key
```
ivirshup, line 34: Could this become one of our documentation variables? Like what we do with plotting arguments, and add to the doc-string with `@_doc_params`?

scanpy/_utils.py
Koncopd: Yes, sorry for that. The formating changes are only in _utils.py, the other changed files have actual code changes. The only new thing in _utils.py is [the helper class](https://github.com/theislab/scanpy/blob/9d399ba8705142faa889750bfb0e955ae31159d7/scanpy/_utils.py#L612).

scanpy/tools/_leiden.py
```diff
@@ -30,6 +30,7 @@ def leiden(
     use_weights: bool = True,
     n_iterations: int = -1,
     partition_type: Optional[Type[MutableVertexPartition]] = None,
+    neighbors_key: Optional[str] = None,
```
ivirshup, line 33: What do you think about this being `connectivities`, with a default value of `"connectivity"` instead? The distances graph isn't used, and really the graph could come from anywhere.

This would also remove the need for `NeighborsView` here, and fits more with what I'm doing for `leiden_multiplex`.

ivirshup (changes requested):
Added a few more comments

I was thinking about how to test this: every function that got changed here should get a test that if the storage location of neighbors changes, the results are exactly the same.

scanpy/_utils.py
ivirshup: Cool, good to know.

I feel like we should probably start splitting this up into multiple files, but that'll be in another PR.

scanpy/neighbors/__init__.py
```diff
@@ -85,6 +86,13 @@ def neighbors(
         A known metric’s name or a callable that returns a distance.
     metric_kwds
         Options for the metric.
+    key_added
+        If not specified, the neighbors data is stored in .uns['neighbors'],
+        distances and connectivities are stored in .obsp['distances'] and
+        .obsp['connectivities'] respectively.
+        If specified, the neighbors data is added to .uns[key_added],
+        distances are stored in .obsp[key_added+'_distances'] and
+        connectivities in .obsp[key_added+'_connectivities'].
```
ivirshup, line 95: Could there be an example of what this looks like in the usage section?

scanpy/plotting/_utils.py
```diff
@@ -482,14 +483,18 @@ def add_colors_for_categorical_sample_annotation(
         _set_default_colors_for_categorical_obs(adata, key)
 
 
-def plot_edges(axs, adata, basis, edges_width, edges_color):
+def plot_edges(axs, adata, basis, edges_width, edges_color, neighbors_key=None):
```
ivirshup, line 486: Like with the leiden function, I think that if only one of the graphs is needed, we could just pass the `obsp` key to that graph.

Koncopd: Added obs_key to leiden, louvain, draw_graph as discussed.

scanpy/tools/_diffmap.py
Koncopd: This is a bit different from function to function.

scanpy/plotting/_utils.py
ivirshup: Could it just be `obsp`? I think it's a little more consistent with the `layer` argument we use in a few places.

Koncopd: Ok, you are right.

ivirshup: Great, thanks!

scanpy/external/exporting.py
```diff
@@ -211,11 +215,12 @@ def spring_project(
 # --------------------------------------------------------------------------------
 
 
-def _get_edges(adata):
+def _get_edges(adata, neighbors_key=None):
+    neighbors = NeighborsView(adata, neighbors_key)
     if 'distances' in adata.uns['neighbors']:  # these are sparse matrices
```
ivirshup: Just ran into this. This shouldn't check for distances in `uns["neighbors"]`. Could this check if `"distances"` and `"connectivities"` are in `NeighborsView` instead?

Koncopd: Yes, thank you for catching this.

ivirshup:
This is great! Thanks @Koncopd!

ivirshup (approved):

gokceneraslan:
image

shall we maybe change 

`sparse matrix (.uns['neighbors'], dtype float32)`

to

`sparse matrix (.uns[key_added], dtype float32) where key_added is "neighbors" if not supplied.` 

or something?

Koncopd:
@gokceneraslan , thanks, forgot to change this, i'll fix.

gokceneraslan:
@Koncopd sorry for the late feedback, but I don't see the "neighbors_key" in the scanpy.tl.paga function. It'd be great to make sure that everything that uses the neighbor graph is covered :)

gokceneraslan:
Also sc.tl.dpt and  sc.tl.diffmap functions are not tested in tests, AFAICS.

Koncopd:
@gokceneraslan 
paga and docs for returns of neighbors are addressed here https://github.com/theislab/scanpy/pull/1179

gokceneraslan:
Thanks! 🙂

gokceneraslan:
Related issue: https://github.com/theislab/anndata/issues/380

PR #1123, opened 2020-03-20
Fix pl.paga with pie charts

michalk8:
Hi everyone,
at the moment, pie charts for paga are a bit brittle, see https://github.com/theislab/cellrank/issues/25.
This pull request is an attempt to fix it. Rather than using `ax.pie`, it's just a bunch of scatterplots with custom markers.
I've tested the performance
```python
foo = {i: {c.to_hex(cm.viridis(_)): 0.001 for _ in range(255)} for i in range(8)}
sc.pl.paga(adata, color=foo, colorbar=False)
```
currect

Performancewise, it takes about ~14 seconds to produce the plot with the proposed changes,
~4s, but I consider that the worst-case scenario.
More importantly, current version doesn't produce a correct plot, see below:
buggy


michalk8:
Hey @ivirshup , could you please give me a feedback on this if there should be any improvements?

ivirshup:
I'm not super familiar with this code, I had no idea a pie chart could even be used here. @falexwolf or @fidelram may be able to say more here.

A few points:

* This should have a regression test, similar to [these](https://github.com/theislab/scanpy/blob/c255fa10fb75f607780ed7d9afc6683cbcecc38e/scanpy/tests/test_plotting.py#L742-L774)
* Could you give some more details about the benchmark? 14 seconds seems far too slow for that plot. Also, is that plotting right? The pie charts all look the same.

michalk8:
Thanks for the feedback. I will add a regression test soon. The plot should be right - there's the color gradient fron blue to green and is a bit more than 1/4 (0.255 to be exact).
As for time, the problem is, that each pie charts are being plotted one by one - that is - there are 2 loops:
```
for node in nodes:
    for pie_fraction in fractions[node]:
        ...
```
I did it because this is the general case of the following matplotlib [example](https://matplotlib.org/3.2.0/gallery/lines_bars_and_markers/scatter_piecharts.html), where they in essence do only `for pie_fraction in fractions`. However, this approach would fail if in the above example
```
foo = {i: {c.to_hex(cm.viridis(_)): 0.001 for _ in range(255)} for i in range(8)}
foo[0] = {'black': 0.5}
```
the the nodes don't contain the same colors, which user could (although not sure why) specify.
I will test out how much speedup can be gained by using the matplotlib approach (assuming the colors for every node are the same) and get back to you.

michalk8:
Ok, while trying to implement what I've suggested, I realized I made a mistake and it won't work - I can't specify different markers per 1 call of `ax.scatter`.

I don't think the speed is a major issue, since the above example is an extreme case (255 categories). Currently, I don't have any trick up my sleeve on how to speed it up.

I've also tried including the regression test, but I can't seem to produce an expected figure. I have the default parameters + 40 dpi as it's in `test_plotting.py`, but the plots that I save are always larger for some reason (tried running it from CLI as well, saving the result from the test case [both options failed]). I've tried whether this is related specifically to the pie chart - it isn't - plotting it without still produces larger plots. @ivirshup any idea what I'm doing wrong?

ivirshup:
The time to plot seems like an issue to me because it's such a simple plot that ends up being generated. It's not obvious to me what part of making that plot would take a long time to calculate, so maybe something unexpected is happening.

I really think @falexwolf or @fidelram are in a better position to give advice on how to implement this plot, and troubleshoot matplotlib.

It would be useful to see examples of the output you're getting, along with the code that generated them.

Marius1311:
Hi all, is there any progress on this? @michalk8 and myself rely on these pie charts for a package we are currently developing and it would be really important to have a fix for this in scanpy.

fidelram:
Hi. This looks good to me. I had used the same trick before. 

I tried to search how to trigger the 'pie' option of paga but I could not find it in the documentation. Do you mind adding some documentation and an example maybe using `sc.datasets.pbmc68k_reduced()`.

As a suggestion, I think that the plots may benefit from adding and outline around the legend as is done for the embeddings. See the parameter legend_fontoutline) https://github.com/theislab/scanpy/blob/668b67765534643bc4257357ef639b2087df0716/scanpy/plotting/_tools/scatterplots.py#L435

michalk8:
@fidelram again sorry for being late. I hope it doesn't matter that I've blackified one of the files (can revert this).
Not sure how relevant this PR will be, since @VolkerBergen has faster and more robust implementation, but as a triage version, it works.

Marius1311:
Thanks! @michalk8 , does this still matter to us, now that we use scvelo's implementation?

PR #1124, opened 2020-03-21
Fix `n_pcs` usage if `use_rep=='X_pca'`

VolkerBergen:
Currently
```
sc.pp.pca(adata, n_comps=50)
neighbors = sc.Neighbors(adata)
neighbors.compute_neighbors(n_pcs=30, use_rep='X_pca')
```
and 
```
sc.pp.pca(adata, n_comps=50)
neighbors = sc.Neighbors(adata)
neighbors.compute_neighbors(n_pcs=30, use_rep=None)
```
yield different results, because the former uses all 50 PCs while the latter uses the user-defined 30 PCs.

Now, both ways, it accounts for the user-defined 30 PCs.

### DIFF
--- a/scanpy/tests/test_neighbors.py
+++ b/scanpy/tests/test_neighbors.py
@@ -3,6 +3,7 @@ import pytest
 from anndata import AnnData
 from scipy.sparse import csr_matrix
 
+import scanpy as sc
 from scanpy import Neighbors
 
 # the input data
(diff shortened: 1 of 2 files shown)


ivirshup (changes requested):
Could this get a test?

ivirshup (approved):
Great, thanks!

PR #1116, opened 2020-03-17
Plotting overhaul

flying-sheep:

Issue #1103, opened 2020-03-15
Legend alignment in dotplot


LuckyMD:
Hey!
The color bar and legend don't seem to align well... might be because i'm subsetting the object.

Here is what happens (reproducible example and my plot, which is even stranger with legend overlap):

<img width="623" alt="Screenshot 2020-03-15 at 13 09 33" src="https://user-images.githubusercontent.com/13019956/76701118-73bff200-66be-11ea-9c99-ee4e0427bbfa.png">
<img width="580" alt="Screenshot 2020-03-15 at 13 02 52" src="https://user-images.githubusercontent.com/13019956/76701119-76224c00-66be-11ea-9b45-b94c1d9a57bd.png">

```python
import scanpy as sc
adata = sc.datasets.pbmc68k_reduced()
sc.pl.dotplot(adata[adata.obs.bulk_labels.isin(['Dendritic', 'CD14+ Monocyte', 'CD4+/CD25 T Reg']),], var_names=['HES4', 'TNFRSF4', 'SSU72'], groupby='bulk_labels', figsize=(8,8))

```

#### Versions:

> scanpy==1.4.5.1 anndata==0.7.1 umap==0.3.10 numpy==1.18.1 scipy==1.4.1 pandas==1.0.1 scikit-learn==0.22.2 statsmodels==0.11.1 python-igraph==0.8.0 louvain==0.6.1

Also, matplotlib is 3.1.3

ivirshup:
Huh. As far as I can tell, this seems to happen only if there are 3 or less groups.

ivirshup:
Probably has something to do with this:

https://github.com/theislab/scanpy/blob/51dc038cb73fc1798373e33d990090439ec57e9e/scanpy/plotting/_anndata.py#L1896-L1905

LuckyMD:
This seems like a bad idea, no? It also seems to throw off the legend other than the colorbar.

fidelram:
In #1127 I changed the way that the colorbar and size legend are plotted and should solve this problem.

gabriellayi:
Hi, I have the same problem, sorry that I just started learning python, so I don't really understand some of the improvements and I also tried some of them, it didn't work. 
Could you please let me know what kind of solution will be good for this issue?

Thanks! 
<img width="765" alt="Screen Shot 2020-05-24 at 17 43 08" src="https://user-images.githubusercontent.com/50899584/82768801-15ae3a00-9de6-11ea-9552-88eb59b19405.png">
[...]

Issue #956, opened 2019-12-17
Avoid argument explosion in plotting functions


fidelram:
- [X] Additional function parameters / changed functionality / changed defaults?
- [ ] New analysis tool: A simple analysis tool you have been using and are missing in `sc.tools`?
- [ ] New plotting function: A kind of plot you would like to seein `sc.pl`?
- [ ] External tools: Do you know an existing package that should go into `sc.external.*`?
- [ ] Other?

To reduce the number of arguments that are passed to plotting functions and to agrupate them by type I was considering the following example syntax:
```PYTHON
sc.pl.umap(adata, color='clusters').scatter_outline(width=0.1)
                                   .legend(loc='on data',  outline=1)
                                   .add_edges(color='black', width=0.1)
```
or 

```
sc.pl.dotplot(adata, ['gene1', 'gene2'], groupby='clusters')
                     .add_dendrogram(width=0.4,color='grey')
                     .swap_axes()
                     .dot_size_legend(title='fraction', location='left')
```

Any comments? 
(I am not sure how to implement something like this)

LuckyMD:
I love this! Question from my side would be how to best make docs for this. Might fragment the documentation too much if we're not careful with this.

fidelram:
With respect to documentation: each plotting function becomes a module with well described functions.   An example is the styling of pandas DataFrames (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.io.formats.style.Styler.html)

LuckyMD:
That looks really good!

gokceneraslan:
I support the idea of tidying up plotting arguments. I think there are mainly two problems. 1) the high number of plotting arguments 2) lack of reusability of plotting "styles". 

Chaining looks really cool and improves 1). Also, it logically partitions the plotting arguments. However, it doesn't solve 2). In other words, if we plot two figures, we'll need to copy the entire thing, and it'll be  very verbose:

```python
sc.pl.umap(adata, color='clusters').scatter_outline(width=0.1)
                                   .legend(loc='on data',  outline=1)
                                   .add_edges(color='black', width=0.1)

sc.pl.umap(adata2, color='fluffy').scatter_outline(width=0.1)
                                  .legend(loc='on data',  outline=1)
                                  .add_edges(color='black', width=0.1)
```

One thing that comes to mind for reusability is to store the result of the chain somewhere and, well, reuse it:

```python
style = sc.pl.styles.scatter_outline(width=0.1)
[...]

PR #1127, opened 2020-03-25
[WIP] Dotplot (aka bubble plot) improvements

fidelram:
This PR addresses issues from #562 #1103

Here is a list of changes for `sc.pl.dotplot` and also for `sc.pl.rank_genes_groups_dotplot`:
 * Colorbar and dot size legends had been improved. Now is also possible to set a title for them.
 * The dotplot functionality has been separated from the main function. This allows to take any two pandas dataFrames, one containing the color values and other containing the size to be plotted.
 * A new dotplot style is added, in which a background square is colored according to the colormap instead of the dot. The style can be chosen as parameter.
 * Now p-values, score and in general any parameter from `sc.rank_genes_groups` can be plotted.
 * `swap_axes` has been added.
 * **update**: method chaining was added to avoid explosion of parameters as suggested in #956. The caveat is that a different name is needed to keep the current functionality of `sc.pl.dotplot`. Now is possible to do `sc.pl.DotPlot(adata, ....).add_dendrogram(size=1.2).swap_axes().show()`.

image

image

Open issues:
 * we need a better way to filter the output from rank_genes_groups (and I would also consider a better name). One option is to filter while plotting.
 * to correctly plot the results from `sc.tl.rank_genes_groups` is better to report the results for all genes (`n_genes`). The default is to keep the top 100 genes per group. Thus, for many genes, the pvalue, logfoldchange etc in other groups is not saved.
* added new parameters, for some I would be happy to have some feedback. One is `style` that has two options  `square color` and `dot color`.

 **Update**
The following is now possible (notice the method chaining):

```PYTHON
markers = {'T-cell': ['CD3D', 'CD3E', 'IL32'], 'B-cell': ['CD79A', 'CD79B', 'MS4A1'], 'myeloid': ['CST3', 'LYZ']}
sc.pl.DotPlot(adata, markers, groupby='bulk_labels') \
               .add_totals(sort='descending', size=1.2) \
               .legend(color_title='log(UMI count+1)', width=1.6)\
               .style(color_map='RdBu_r', dot_min=0.3)\
               .show()
```
image

TODO:
 - [x] Remove small secondary ticks
 - [x] Adapt matrixplot to new object model
 - [ ] Adapt heatmap to object model
 - [x] Adapt stackedviolin to object model
 - [ ] Incorporate modifications from #1116 
 - [x] Fix call from rank_genes_groups
 - [x] 'black' code
 - [x] Update docstrings
 - [ ] Update tests
 - [ ] Update tutorial and readthedocs


scanpy/plotting/_anndata.py
```diff
@@ -1824,94 +1829,37 @@ def dotplot(
...
+        if standard_scale == 'group':
+            mean_obs = mean_obs.sub(mean_obs.min(1), axis=0)
+            mean_obs = mean_obs.div(mean_obs.max(1), axis=0).fillna(0)
+        elif standard_scale == 'var':
+            mean_obs -= mean_obs.min(0)
+            mean_obs = (mean_obs / mean_obs.max(0)).fillna(0)
+        elif standard_scale is None:
+            pass
+        else:
+            logg.warning('Unknown type for standard_scale, ignored')
```
ivirshup: Could this be an `ValueError` instead?

scanpy/plotting/_anndata.py
```diff
@@ -2315,8 +2347,60 @@ def matrixplot(
...
+                    xytext=(0, 10),
+                    fontsize="x-small",
+                    textcoords="offset points",
+                )
+            # for k in total_barplot_ax.spines.keys():
+            #     total_barplot_ax.spines[k].set_visible(False)
+            total_barplot_ax.set_ylim(0, max_y * 2)
+            total_barplot_ax.grid(False)
+            total_barplot_ax.axis("off")
+            #total_barplot_ax.set_xlim(0, p.get_x() + p.get_width())
```
ivirshup: Could this be wrapped up in a function, similar to `_plot_dendrogram`?

scanpy/plotting/_anndata.py
```diff
@@ -1939,103 +1887,170 @@ def dotplot(
...
-        y_ticks = range(mean_obs.shape[0])
-        dendro_ax = fig.add_subplot(axs[1, 1], sharey=dot_ax)
-        _plot_dendrogram(
-            dendro_ax, adata, groupby, dendrogram_key=dendrogram, ticks=y_ticks
+        dendro_ticks = np.arange(mean_obs.shape[0]) + 0.5
+
+    category_height = category_width = 0.35
+    legends_width = 1.2
+    dendro_width = 0.8 if dendrogram else 0
+    if figsize is None:
```
ivirshup: This might explain some issues I've had with writing out dot plots.

What happens if I don't pass `figsize`, but specify it with:

```python
with plt.rc_context({"figure.figsize": (5, 5)}):
    sc.pl.dotplot(adata, ...)
```

scanpy/plotting/_anndata.py
```diff
@@ -1709,18 +1709,23 @@ def dotplot(
     num_categories: int = 7,
     expression_cutoff: float = 0.0,
     mean_only_expressed: bool = False,
-    color_map: str = 'Reds',
+    color_map: str = 'RdBu_r',
```
ivirshup: I don't think this should be diverging by default.

Why do you want to change this?

scanpy/plotting/_anndata.py
```diff
@@ -1709,18 +1709,23 @@ def dotplot(
...
+    style: Optional[str] = 'square color',
```
ivirshup: I don't think the default should change until 2.0. I'm also not sure I like the square version better.

ivirshup (commented):
Gave this a quick once over.

One other point, I don't like that you can't see the size of the circle when the square plot background is white. Would you mind explaining this new style?

> to correctly plot the results from sc.tl.rank_genes_groups is better to report the results for all genes (n_genes). The default is to keep the top 100 genes per group. Thus, for many genes, the pvalue, logfoldchange etc in other groups is not saved.

I completely agree. I personally think we should remove the top genes parameter, or at least default to all, from rank genes groups. The user can very easily filter on their own later. Thoughts @Koncopd?

scanpy/plotting/_anndata.py
fidelram: Dotplot (and matrixplot, heatmap) by default override the ` rcParams['figure.figsize']`, otherwise the output figure will look very bad. I know that this is a different behavior compared to `sc.pl.scatter`.

The solution is to pass `figsize` as argument in case you want to use custom values. 

Do you have any other idea?

scanpy/plotting/_anndata.py
fidelram: Thanks, indeed I wanted to set that back as in the original case.

scanpy/plotting/_anndata.py
fidelram: done

LuckyMD:
I really like the plots above, and really cool that you solved the issue with the legends. Would it also be possible to add some kind of grouping brackets on the cluster level in the same way you have for the gene level? I have cases where I would like to say that club, goblet, basal, and ciliated cells are all airway epithelial cells.

fidelram:
@LuckyMD  should be easy to extend the `Plot` class to add this. I will think about it.

ivirshup (commented):
Hey! This looks good! The code seems a lot cleaner.

I didn't have as much time as I hoped to look through this today. I think I need to take more time to understand the whole structure before I can give useful feedback.

A few questions/ points for now:

* Could you show some examples of the new additions/ let me know where you are on tutorials?
* Could you add tests for functionality where the underlying code is changing, but doesn't have coverage yet? I would mostly just like to see more tests of the plotting code.
* What do you think about the idea of splitting up `_anndata.py` into a few more files? I think it's getting a bit too big, which can make reviewing difficult.
* Could you run `black` over this?

flying-sheep:
Looks like a really cool idea, and undoubtedly inspired by some existing class-based plotting solution.

Scanpy's plots are of course more domain-specific than the generic APIs in Plotnine, Plotly or Altair. Nevertheless, there seem to be quite some generic APIs in your classes like `add_legend` and so on.

Before I review this proper: Did you think about using and extending a generic class-based plotting solution? If not: Could you investigate if any of them can be extended so we don't have to maintain the non-domain-specific parts of the API? If yes: What makes them unsuited for our needs?

fidelram:
@flying-sheep To answer your question. Honestly, I am not so familiar with Plotnine, Plotly or Altair. However, after a quick revision I would say that Altair seems quite interesting and possibly were I could had reused/extended some code.

Yet, at the moment in scanpy we use matplotlib extensively and I didn't even think about the other APIs. Looking closely at Altair I realized that I have a lot to catch up regarding Vega, Vega-lite and the idiosyncrasies specific to Altair before I could start using it. 

Thus, the current effort only integrates the idea of 'chaining' seen in Altair (or in other context in Pandas). In Plotly or Plotnine the 'chaining' is achieved differently but I don't find it as nice or straightforward:

**Plotly:**
```PYTHON
import plotly.graph_objects as go
fig = go.Figure(
    data=[go.Bar(x=[1, 2, 3], y=[1, 3, 2])],
    layout=go.Layout(
        title=go.layout.Title(text="A Bar Chart")
    )
)
fig.show()
```
**Plotnine:**
```PYTHON

from plotnine import ggplot, geom_point, aes, stat_smooth, facet_wrap
from plotnine.data import mtcars

(ggplot(mtcars, aes('wt', 'mpg', color='factor(gear)'))
 + geom_point()
 + stat_smooth(method='lm')
 + facet_wrap('~gear'))
```
**Altair:**
```PYTHON
import altair as alt
from vega_datasets import data

source = data.cars()

alt.Chart(source).mark_circle(size=60).encode(
    x='Horsepower',
    y='Miles_per_Gallon',
    color='Origin',
    tooltip=['Name', 'Origin', 'Horsepower', 'Miles_per_Gallon']
).interactive()
```
The current solution, although using method chaining, is very  *ad hoc* for a specific type of graphs that have predetermined features, like 'dendrogram' or totals for categories or 'brackets' to highlight features.

fidelram:
@ivirshup 
> * Could you show some examples of the new additions/ let me know where you are on tutorials?

I will do that once we are happy with the current code and naming conventions used. My goal was to update this tutorial https://scanpy-tutorials.readthedocs.io/en/latest/visualizing-marker-genes.html but suggestions are welcome. 

> * Could you add tests for functionality where the underlying code is changing, but doesn't have coverage yet? I would mostly just like to see more tests of the plotting code.

I will try to do that.

> * What do you think about the idea of splitting up `_anndata.py` into a few more files? I think it's getting a bit too big, which can make reviewing difficult.

We should to that. Currently, as I see it we have two types of plots: 
 * embedding scatter plots which are separated already
 * the type of plots in this PR that I would describe as `grouping` plots, because they visualize the AnnData matrix  subdivided based on a .`obs` column.  Any better name for this?

> * Could you run `black` over this?

Will do it at the end. Do we have some style policies for black or the defaults are fine?

scanpy/plotting/_tools/__init__.py
```diff
@@ -382,11 +384,40 @@ def _rank_genes_groups_plot(
...
+        values_df = pd.concat(df_list)
+        values_df = pd.pivot(values_df, index='names', columns='group',
+                             values=values_to_plot).fillna(1)
+
+        if values_to_plot in ['pvals', 'pvals_adj']:
+            values_df = -1 * np.log10(values_df)
+            kwds.setdefault('color_title', "-log10(pvalue")
+
+        # TODO. n_genes for tl.rank_genes needs to be larger to get good results
+        values_df = values_df.loc[gene_names].T
```
ivirshup: Why run this here, and not in `dotplot`? Especially since that's the only function that get's passed this dataframe?

I would also subset the dataframes to the genes prior to concatenating and pivoting them. This will mean there's less allocation to do in those steps.

scanpy/plotting/_anndata.py
```diff
@@ -3507,3 +3057,1400 @@ def _check_var_names_type(var_names, var_group_labels, var_group_positions):
...
+
+        # set default values for style
+        self.style()
+
+        self.mean_obs = mean_obs
+        self.fraction_obs = fraction_obs
+        self.kwds = kwds
+
+    def style(self,
+              color_map: str = 'RdBu_r',
```
ivirshup: The default color for dot plots (but really most cases we have) should not default to diverging. Diverging requires a meaningful middle value, which we can't assume in many cases. Log fold change can be diverging, expression level can't unless it's been scaled.

scanpy/plotting/_anndata.py
```diff
@@ -3507,3 +3057,1400 @@ def _check_var_names_type(var_names, var_group_labels, var_group_positions):
...
+        size_legend_ax.spines['bottom'].set_visible(False)
+        size_legend_ax.grid(False)
+
+        ymin, ymax = size_legend_ax.get_ylim()
+        size_legend_ax.set_ylim(-0.9, 4)
+        size_legend_ax.set_title(self.size_title, y=ymax + 0.25,
+                                 size='small')
+
+        xmin, xmax = size_legend_ax.get_xlim()
+        size_legend_ax.set_xlim(xmin,  xmax + 0.5)
```
ivirshup: I like this function.

It takes all of the information it needs as arguments, and its a very small set of arguments. It does one small well defined thing?

Why is it a method?

fidelram:
I just added the class `ViolinPlot` to the group plots. Now is possible to do this:

```PYTHON
from scanpy.plotting._anndata import StackedViolin, DotPlot, MatrixPlot
markers = {'T-cell': ['CD3D', 'CD3E', 'IL32'], 'B-cell': ['CD79A', 'CD79B', 'MS4A1'], 'myeloid': ['CST3', 'LYZ']}

fig, (ax1, ax2, ax3) = plt.subplots(1, 3, figsize=(19,5), gridspec_kw={'width_ratios': [.8, 1, 1]})
StackedViolin(adata, markers, groupby='bulk_labels', ax=ax1).add_dendrogram(True).swap_axes(True).show(False)
DotPlot(adata, markers, groupby='bulk_labels', ax=ax2).add_dendrogram(True).swap_axes(True).show(False)
_ = MatrixPlot(adata, markers, groupby='bulk_labels', ax=ax3).add_dendrogram(True).swap_axes(True).show(False)
```

image

scanpy/plotting/_utils.py
```diff
@@ -1159,7 +1159,41 @@ def make_grid_spec(
...
+    ...     return fix_kwds(kwds, key1="value1", key2="value2")
+    >>> example(key1="value10", key3="value3")
+        {'key1': 'value10, 'key2': 'value2', 'key3': 'value3'}
+
+    """
+
+    for key, value in kwargs.items():
+        kwds_dict[key] = kwds_dict.get(key, value)
+
+    return kwds_dict
```
ivirshup: Couldn't this just be `kwargs.update(kwds_dict)`?

scanpy/plotting/_tools/__init__.py
```diff
@@ -383,10 +392,23 @@ def _rank_genes_groups_plot(
         start = end
 
     group_names = group_names_valid
-    if plot_type == 'dotplot':
-        from .._anndata import dotplot
-        dotplot(adata, gene_names, groupby, var_group_labels=group_names,
-                var_group_positions=group_positions, show=show, save=save, **kwds)
+
+    if plot_type in 'dotplot':
```
ivirshup: `in`?

scanpy/plotting/_anndata.py
```diff
@@ -3507,3 +2696,1868 @@ def _check_var_names_type(var_names, var_group_labels, var_group_positions):
...
+        row_palette = self.kwds.get('color', self.row_palette)
+        if 'color' in self.kwds:
+            del self.kwds['color']
+
+        # All columns should have a unique name, yet, frequently
+        # gene names are repeated in self.var_names.  otherwise the
+        # pd.melt object that is passed to seaborn will merge non-unique columns.
+        # Here, I simply rename the columns using a count from 1..n using the
+        # mapping function `rename_cols_to_int` to solve the problem.
+        _matrix.rename(rename_cols_to_int, axis='columns', inplace=True)
```
ivirshup: Defining a global variable is not a good solution to this problem.

Could just do `df.rename(columns=dict(zip(df.columns, range(len(df.columns)))))`?

ivirshup (changes requested):
Sorry for taking so long to get back to you on this! Overall, I like the idea behind this PR. We have a bunch of similar kinds of plots (those with categorical axes) and it would be great if there was an easy way to consistently add marginal plots to them.

This PR takes a lot of effort to review. This PR adds a lot of code, and that code isn't all the easiest to follow. I think this is largely due to matplotlib generally resulting in hard to follow code, but I think the structure of the new code also needs more consideration. Because of that, I'm going to start with a more broad review, before diving into the code specifics that much.

Could you describe your mental model for these new classes? They seem to do a lot of things and store a lot of data. I'd like to see this cut down, but would want to hear the way you're thinking about it first. To me, these seem very similar to the ideas in `seaborns` `ClusterGrid` (https://github.com/mwaskom/seaborn/blob/5498cdb2f145b15a85adac0c8ff233656f52a943/seaborn/matrix.py#L779) and `jointplot` (https://github.com/mwaskom/seaborn/blob/7216552848c83b678481d5bca01a180120b0301e/seaborn/axisgrid.py#L2130) type plots. I think that code could be a very good template for how this could be structured.

I'm currently not sure I'd be up for making the object based plotting API public. One of the things that the libraries @flying-sheep mention have in common is very well defined models for the objects. In particular, I think altair does a great job with this. I would like to make sure this model is sound before making it something we promise to support. It could definitely be used internally before that.

### Organization

Could this code start going into multiple files now? Github reviewing tools were not made for 5000 line files.

My main issue with the plotting code is more maintainability than number of features. I would like there to be much more clear and consistently used abstractions here. One example of that would be separating data processing from matplotlib code as much as possible. This is an area I'll address more in further review, but it's a heads up about what my priorities are here.

These classes have a lot of attributes, which make the mental overhead of figuring out what's going on in the code pretty high. Can this be reduced?

> Do we have some style policies for black or the defaults are fine?

Black settings are defined in our `pyproject.toml` / `.editorconfig`.

### API

I don't think results for any plots should noticeably change from this PR. Is that happening? I get the impression you want to use a diverging palette by default?

I'm not sure I like the `swap_axes` function. It assumes the user knows what the default axes are. I'd prefer if the x and y axes were passed as arguments to the initial function. Similar to `sns.joinplot`. Maybe this could end up looking like: 

```python
(
    sc.pl.DotPlot(adata, x="leiden", y=gene_list, color="mean", size="frac")
    .add_dendrogram(axis="y")
)
```

### Tests

This needs a lot of tests. Some ideas to start out:

* Should dendrograms look any different for different main plots? What should be the same?
  * Ditto for legends.
* Does making one of these plots change any global state (i.e. `matplotlib`s `rc_params`)?
* Does the order which methods are called matter? Should it?

scanpy/plotting/_tools/__init__.py
fidelram: This is only relevant when `rank_genes_groups` has been called. Thus is only relevant on that context.

scanpy/plotting/_anndata.py
fidelram: The default continues to be a sequential color range in the wrap function `sc.pl.dotplot`. However, a disadvantage of a sequential colormap in this type of plot is that the smallest dots usually can not be differentiated from the background as they tend to have the lighter color. But I can try with some of the sequential colormaps like 'autumn' which do not have very light colors.

scanpy/plotting/_utils.py
fidelram: Humm, probably it may work. How did I miss it?

scanpy/plotting/_tools/__init__.py
fidelram: should be '=='

scanpy/plotting/_anndata.py
fidelram: I remember I had some issue in the past (this code is quite old btw)  which I could not resolve cleanly and had to resort to these weird function. I will try your method.

scanpy/plotting/_anndata.py
fidelram: In contrast to just being a function?

ivirshup: Yes

scanpy/plotting/_anndata.py
ivirshup: Sounds good. I think `viridis` can be fine for this as well, or a version of `reds` where the lowest value is light grey.

fidelram:
This PR is replaced by #1210

PR #1132, opened 2020-03-27
More flexible umap

ivirshup:
This PR lets user choose the connectivity matrix passed to UMAP for layout.

Two arguments have been added to the umap function for this:

* `obsp` for specifying the connectivity matrix in `obsp`
* `key_added` for choosing the key name in `obsm`

Still need to reorganize the arguments for `sc.tl.umap` so the more used ones are documented first.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,6 @@ exclude = '''
         _dpt
         |_sim
         |_paga
-        |_umap
         |_utils
         |_leiden
         |_louvain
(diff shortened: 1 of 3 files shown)


Koncopd:
And what about metric parameter?

ivirshup:
Could you remind me what you mean by that?

Koncopd:
https://github.com/theislab/scanpy/blob/2a16436e9ac3b3ba039a99b9a3674885fe6a4590/scanpy/tools/_umap.py#L171

Is it conceptually right to pass custom connectivities with `metric='euclidean'` by default?

ivirshup:
The metric is only used in the case of multiple connected components, which is also the case where `data` is used. It's for calculating distances between the mediods of the components. I'm not sure there is a great default for either of these values.

This does feel kinda hack-y to me. I'm not really sure what the right thing to do in this case is, and maybe we should not be using UMAPs layout code with graphs that UMAP didn't generate. Unfortunately, it's just much faster than other graph layout algorithms I've tried, and provides good looking results.

@falexwolf, any thoughts?

PR #1135, opened 2020-03-28
Store mean and std in annotations when calling `pp.scale`, only accept AnnData

falexwolf:
This should allow people to reconstruct count data from scaled values.

Moving away from accepting numpy arrrays and sparse matrices in the preprocessing functions has always been on my mind to simplify the code. Nobody is passing anything different from an AnnData.

In the very early days of Scanpy, I thought it'd be nice to also accept other formats of data matrices.


ivirshup (commented):
I think the functionality should get deprecated before it get's removed. A clean way to do this would be `functools.singledispatch`.

falexwolf:
I think that literally nobody is using this function for arrays and sparse matrices, everyone only uses it for AnnDatas. It's complicated to reintroduce the functionality and there is no way the purpose of this PR can be accomplished.

Hence, I'd really just merge it. Quite a few things will break towards 1.5. Now is the time.

I'd make the same change for all other preprocessing functions.

ivirshup:
> I think that literally nobody is using this function for arrays and sparse matrices

I'm not sure about that. I got the impression from https://github.com/theislab/scanpy/issues/1030#issuecomment-607952458 and (IIRC) a conversation with @scottgigante that people would like these functions to work on arrays as well as AnnData objects.

> In the very early days of Scanpy, I thought it'd be nice to also accept other formats of data matrices.

I still think it would be nice to support that, it just requires factoring the code better. I agree recursively calling the same function for argument handling gets very confusing. However, I think we could do something more like this (note, it's not tested yet, and could be cleaner...  it's my ten minute version):

<details>
<summary> Alternative implementation of scale </summary>

```python
@singledispatch
def scale(X, *args, **kwargs):
    """\
    Scale data to unit variance and zero mean.
    .. note::
        Variables (genes) that do not display any variation (are constant across
        all observations) are retained and set to 0 during this operation. In
        the future, they might be set to NaNs.
    Parameters
    ----------
    data
        The (annotated) data matrix of shape `n_obs` × `n_vars`.
        Rows correspond to cells and columns to genes.
    zero_center
        If `False`, omit zero-centering variables, which allows to handle sparse
        input efficiently.
    max_value
        Clip (truncate) to this value after scaling. If `None`, do not clip.
    copy
        If an :class:`~anndata.AnnData` is passed,
        determines whether a copy is returned.
    Returns
    -------
    Depending on `copy` returns or updates `adata` with a scaled `adata.X`,
    annotated with `'mean'` and `'std'` in `adata.var`.
    """
    return scale_array(X, *args, **kwargs)

@scale.register(np.ndarray)
def scale_array(
    X,
    zero_center: bool = True,
    max_value: Optional[float] = None,
    copy: bool = False,
    return_mean_var=False,
):
    if copy:
        X = X.copy()
    if not zero_center and max_value is not None:
        logg.info(  # Be careful of what? This should be more specific
            '... be careful when using `max_value` '
            'without `zero_center`.'
        )
    if max_value is not None:
        logg.debug(f'... clipping at max_value {max_value}')
    mean, std = _scale(X, zero_center)  # the code from here could probably just be 
    # do the clipping
    if max_value is not None:
        X[X > max_value] = max_value
    if return_mean_var:
        return X, mean, var
    else:
        return X


@scale.register(AnnData)
def scale_anndata(
    adata: AnnData,
    *,
    zero_center: bool = True,
    max_value: Optional[float] = None,
    copy: bool = False,
) -> Optional[AnnData]:
    adata = adata.copy() if copy else adata
    view_to_actual(adata)
    adata.X, adata.var["mean"], adata.var["std"] = scale(
        X, 
        zero_center=zero_center, 
        max_value=max_value, 
        copy=False,  # because a copy has already been made, if it were to be made
        return_mean_var=True
    )
    if copy:
        return adata

@scale.register(sparse.spmatrix)
def scale_sparse(
    X, 
    *, 
    zero_center: bool = True,
    copy=False,
    **kwargs
):
    # need to add the following here to make inplace logic work
    if zero_center:
        logg.info(
            '... as `zero_center=True`, sparse input is '
            'densified and may lead to large memory consumption'
        )
        X = X.toarray()
        copy = False  # Since the data has been copied
    return scale_array(X, zero_center=zero_center, copy=copy, **kwargs)
```

</details>

I actually really like this pattern of having an underlying function which has all the logic, but then dispatching through wrappers for the argument handling. It splits out the cases quite nicely, and makes the code flexible. This pattern is very common in Julia, and fairly common in Bioconductor packages.

scottgigante:
I'd like to +1 @ivirshup 's prototype there -- perhaps `scale` is not the best example of a function that people would use on data other than anndata, but this would be a good pattern to follow throughout `scanpy`. Not infrequently I run the following workflow:
```
adata = scanpy.AnnData(data)
dpt = scanpy.tl.dpt(adata)
del adata
```
and it would be awfully nice if `dpt` just accepted my data matrix without me having to wrap it in an AnnData object. This could be true for any scanpy function that doesn't require row/column data.

ivirshup:
Closing as this is was superceded by #1173.

@scottgigante I would be open to that, especially for functions which don't have complex requirements for data input. I'm not sure exactly what this would look like. I'd be interested in seeing a prototype – and PRs 😉.

I think the main issue with supporting this is figuring out the right way to document it. In general, this fits a pattern of multiple dispatch. However, I'm not sure how to make that play nice with sphinx.

PR #1137, opened 2020-03-31
change uns structure for spatial (#1105)

giovp:
Fix uns structure in read visium and tests


giovp:
Made a mistake with squash and rebasing so this should not be merged, and the spatial branch should be deleted. All the changes that were to be added here are in #1138

PR #1138, opened 2020-03-31
fix uns structure in read_visium

giovp:
Hi @ivirshup, so I made a mistake with squash and rebase and basically had all the history in a pretty bad state. 
I managed to revert everything to the first squash, and this should be it.

This PR is essentially #1105, but with correct history.

Also, this should close the scanpy/spatial branch, which should NOT be merged since that branch contains the messed up history after the rebase.

Sorry again for that...


ivirshup (approved):
No worries, managing git history can be a pain. I probably would have just done the same thing here.

Looks good!

PR #1149, opened 2020-04-05
Fix inverted coords when img_key=None

### LINKED ISSUE #1059: Flipped y-axis in sc.pl.spatial without img_key

### LINKED ISSUE #1148: sc.pl.spatial plots vertically flipped spots when  `img_key=None`
sc.pl.spatial plots vertically flipped spots when  `sc.pl.spatial(..., img_key=None)`

Try comparing these in your lymph node demo notebook.
```python
# Plot orientation correct
sc.pl.spatial(adata, img_key = "hires", cmap='magma',
              color=['total_counts', 'n_genes_by_counts'])

# Now spots are flipped
sc.pl.spatial(adata, img_key = None, cmap='magma',
              color=['total_counts', 'n_genes_by_counts'])
```

#### Versions:
scanpy==1.4.6 anndata==0.7.1 umap==0.3.10 numpy==1.17.3 scipy==1.4.1 pandas==0.25.3 scikit-learn==0.22.1 statsmodels==0.10.2 python-igraph==0.7.1 louvain==0.6.1

giovp:
Fix #1059 and #1148 and replace #1078


vitkl:
An alternative solution to this would be `pyplot.imshow(..., origin='lower')` 
https://github.com/theislab/scanpy/blob/master/scanpy/plotting/_tools/scatterplots.py#L301
by default `pyplot.imshow(..., origin='upper')` which makes y-axis flipped compared to the scatterplot

flying-sheep:
@giovp conditionally doing that would probably be cleaner and wouldn’t involve flipping the data right?

giovp:
> An alternative solution to this would be `pyplot.imshow(..., origin='lower')`
> https://github.com/theislab/scanpy/blob/master/scanpy/plotting/_tools/scatterplots.py#L301
> by default `pyplot.imshow(..., origin='upper')` which makes y-axis flipped compared to the scatterplot

Good point, but if `img_key=None` shouldn't we assume that there is no image? So nothing to `pl.imshow`. 
Now the point would be:
* passing a dummy array as "image" and then plot the spots as `circles`, so keeping the normal behavior
* plot the spatial data as a simple scatterplot.  

My reasoning is that, since there is no image to show:
* we probably should assume that also `scalefactors` are empty.
* the correct size of the spots (given by `scalefactors`, and plotted with `circles`) does not matter anymore, since their size is only important in the presence of an image in the background, so a normal scatterplot would do.
* If the user does not want an image in the background, but wants to retain the size of the spots (because `scalefactors` are not empty), then `alpha_img=0` should do the job. 

Because of these reasons, I would go for the second option above, passing the `coords` as basis to the scatterplot fun.

Probably an even better option would be to plot the spots as hexagons, as originally suggested by @flying-sheep , but would wait for that after the plotting module is refactored.

> @giovp conditionally doing that would probably be cleaner and wouldn’t involve flipping the data right?

The reason for flipping is that the coords from space ranger are given with upper origin

```python
sc.pl.embedding(adata_spatial, basis = "coords")
```
image

Happy to discuss and change the behavior, I could have missed a crucial point.
And thank you @vitkl for feedback, we are in dire need of spatial transcriptomics scanpy users !

ivirshup:
@giovp, I've thought about this a bit more and think it should be treated as an image rather than a scatterplot. This is so future updates to the spatial plotting function (like plotting hexagons) won't require a special case here.

If you can't find a scale factor, could we just say it's 1?

giovp:
I think it makes sense. I will open a new PR cause I'll have to change the test as well.

giovp:
Closing, see #1217

PR #1156, opened 2020-04-08
rank_genes_groups refactoring 2nd try

Koncopd:
Related to 1), 2) of [this](https://github.com/theislab/scanpy/issues/723#issuecomment-526079225).

Replaces https://github.com/theislab/scanpy/pull/1081

### DIFF
--- a/scanpy/tools/_rank_genes_groups.py
+++ b/scanpy/tools/_rank_genes_groups.py
@@ -76,7 +76,7 @@ def _ranks(X, mask=None, mask_rest=None):
 
 
 class _RankGenes:
-    def __init__(self, adata, groups, groupby, reference, use_raw, layer):
+    def __init__(self, adata, groups, groupby, reference, use_raw, layer, comp_pts):
 
         if 'log1p' in adata.uns_keys() and adata.uns['log1p']['base'] is not None:
             self.expm1_func = lambda x: np.expm1(x * np.log(adata.uns['log1p']['base']))
@@ -110,6 +110,7 @@ class _RankGenes:
         self.means_rest = None
         self.vars_rest = None
 
+        self.comp_pts = comp_pts
         self.pts = None
         self.pts_rest = None
 
@@ -125,12 +126,12 @@ class _RankGenes:
 
         self.means = np.zeros((n_groups, n_genes))
         self.vars = np.zeros((n_groups, n_genes))
-        self.pts = np.zeros((n_groups, n_genes))
+        self.pts = np.zeros((n_groups, n_genes)) if self.comp_pts else None
 
         if self.ireference is None:
             self.means_rest = np.zeros((n_groups, n_genes))
             self.vars_rest = np.zeros((n_groups, n_genes))
-            self.pts_rest = np.zeros((n_groups, n_genes))
+            self.pts_rest = np.zeros((n_groups, n_genes)) if self.comp_pts else None
         else:
             mask_rest = self.groups_masks[self.ireference]
             X_rest = self.X[mask_rest]
@@ -148,7 +149,8 @@ class _RankGenes:
         for imask, mask in enumerate(self.groups_masks):
             X_mask = self.X[mask]
 
-            self.pts[imask] = get_nonzeros(X_mask) / X_mask.shape[0]
+            if self.comp_pts:
+                self.pts[imask] = get_nonzeros(X_mask) / X_mask.shape[0]
 
             if self.ireference is not None and imask == self.ireference:
                 continue
@@ -160,7 +162,8 @@ class _RankGenes:
                 X_rest = self.X[mask_rest]
                 self.means_rest[imask], self.vars_rest[imask] = _get_mean_var(X_rest)
                 # this can be costly for sparse data
-                self.pts_rest[imask] = get_nonzeros(X_rest) / X_rest.shape[0]
+                if self.comp_pts:
+                    self.pts_rest[imask] = get_nonzeros(X_rest) / X_rest.shape[0]
                 # deleting the next line causes a memory leak for some reason
                 del X_rest
 
@@ -356,6 +359,7 @@ def rank_genes_groups(
     reference: str = 'rest',
     n_genes: Optional[int] = None,
     rankby_abs: bool = False,
+    pts: bool = False,
     key_added: Optional[str] = None,
     copy: bool = False,
     method: _Method = 't-test',
@@ -487,7 +491,7 @@ def rank_genes_groups(
         corr_method=corr_method,
     )
 
-    test_obj = _RankGenes(adata, groups_order, groupby, reference, use_raw, layer)
+    test_obj = _RankGenes(adata, groups_order, groupby, reference, use_raw, layer, pts)
 
     # for clarity, rename variable
     n_genes_user = n_genes
--- a/scanpy/tools/_rank_genes_groups.py
+++ b/scanpy/tools/_rank_genes_groups.py
@@ -66,7 +66,7 @@ def _ranks(X, mask=None, mask_rest=None):
 
     left = 0
     # Calculate rank sums for each chunk for the current mask
-    for chunk_index, right in enumerate(chunk):
+    for right in chunk:
 
         df = pd.DataFrame(data=get_chunk(X, left, right))
         ranks = df.rank()
@@ -223,8 +223,8 @@ class _RankGenes:
         if self.ireference is not None:
 
             for group_index, mask, mask_rest in self._basic_stats():
-                n_active = X_mask.shape[0]
-                m_active = X_rest.shape[0]
+                n_active = np.count_nonzero(mask)
+                m_active = np.count_nonzero(mask_rest)
 
                 if n_active <= 25 or m_active <= 25:
                     logg.hint(
--- a/scanpy/tools/_rank_genes_groups.py
+++ b/scanpy/tools/_rank_genes_groups.py
@@ -167,26 +167,30 @@ class _RankGenes:
                 # deleting the next line causes a memory leak for some reason
                 del X_rest
 
-            yield imask, mask, mask_rest
-
     def t_test(self, method):
         from scipy import stats
 
-        for group_index, mask, mask_rest in self._basic_stats():
+        self._basic_stats()
+
+        for group_index, mask in enumerate(self.groups_masks):
+            if self.ireference is not None and group_index == self.ireference:
+                continue
 
             mean_group = self.means[group_index]
             var_group = self.vars[group_index]
+            ns_group = np.count_nonzero(mask)
 
-            if self.ireference is None:
-                mean_rest = self.means_rest[group_index]
-                var_rest = self.vars_rest[group_index]
-            else:
+            if self.ireference is not None:
                 mean_rest = self.means[self.ireference]
                 var_rest = self.vars[self.ireference]
+                ns_other = np.count_nonzero(self.groups_masks[self.ireference])
+            else:
+                mean_rest = self.means_rest[group_index]
+                var_rest = self.vars_rest[group_index]
+                ns_other = self.X.shape[0] - ns_group
 
-            ns_group = np.count_nonzero(mask)
             if method == 't-test':
-                ns_rest = np.count_nonzero(mask_rest)
+                ns_rest = ns_other
             elif method == 't-test_overestim_var':
                 # hack for overestimating the variance for small groups
                 ns_rest = ns_group
@@ -216,13 +220,20 @@ class _RankGenes:
     def wilcoxon(self):
         from scipy import stats
 
+        self._basic_stats()
+
         n_genes = self.X.shape[1]
         # initialize space for z-scores
         scores = np.zeros(n_genes)
         # First loop: Loop over all genes
         if self.ireference is not None:
 
-            for group_index, mask, mask_rest in self._basic_stats():
+            for group_index, mask in enumerate(self.groups_masks):
+                if group_index == self.ireference:
+                    continue
+
+                mask_rest = self.groups_masks[self.ireference]
+
                 n_active = np.count_nonzero(mask)
                 m_active = np.count_nonzero(mask_rest)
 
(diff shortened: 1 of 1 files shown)


scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,6 +18,329 @@
...
+    for chunk_index, right in enumerate(chunk):
+
+        df = pd.DataFrame(data=get_chunk(X, left, right))
+        ranks = df.rank()
+        yield ranks, left, right
+
+        left = right
+
+
+class _RankGenesGroups:
```
ivirshup: As long as it isn't in a public namespace, it's fine to use public looking names for internal stuff.

Why does this have an `_` in front of it?

scanpy/tools/_rank_genes_groups.py
```diff
@@ -135,317 +462,63 @@ def rank_genes_groups(
             groups_order = [str(n) for n in groups_order]
         if reference != 'rest' and reference not in set(groups_order):
             groups_order += [reference]
-    if (
-        reference != 'rest'
-        and reference not in set(adata.obs[groupby].cat.categories)
-    ):
+    if reference != 'rest' and reference not in set(adata.obs[groupby].cat.categories):
```
ivirshup:
```suggestion
    if reference != 'rest' and reference not in adata.obs[groupby]:
```

I realize this wasn't changed in this PR, but this should already be efficient for categorical dtypes.

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,6 +18,329 @@
...
+            self.means[imask], self.vars[imask] = _get_mean_var(X_mask)
+
+            if self.ireference is None:
+                mask_rest = ~mask
+                X_rest = self.X[mask_rest]
+                self.means_rest[imask], self.vars_rest[imask] = _get_mean_var(X_rest)
+                # this can be costly for sparse data
+                self.pts_rest[imask] = get_nonzeros(X_rest) / X_rest.shape[0]
+                # deleting the next line causes a memory leak for some reason
+                del X_rest
```
ivirshup, line 172: Could you make a small example showing this memory leak? I think I've run into something similar recently.

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,6 +18,329 @@
...
+
+            yield imask, mask, mask_rest
```
ivirshup: Why make this an iterator if you're also saving the values to the object? I think it would be more simple to exclude the yield statement.

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,6 +18,329 @@
...
+            if use_raw and adata.raw is not None:
+                adata_comp = adata.raw
+            X = adata_comp.X
+
+        self.X = X
+        self.var_names = adata_comp.var_names
+
+        self.ireference = None
+        if reference != 'rest':
+            self.ireference = np.where(self.groups_order == reference)[0][0]
```
ivirshup, line 109: Instead of having all these conditional blocks checking for `ireference`, why not just fill the have the  `*_rest` values just be the reference?

I would probably also rename those attributes to `*_other` too.

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,6 +18,329 @@
...
+                scores = scores_all[0]
+            else:
+                scores = scores_all[igroup]
+
+            yield igroup, scores, None
+
+            if len(self.groups_order) <= 2:
+                break
+
+    def compute_statistics(self, method, corr_method, n_genes_user, rankby_abs, **kwds):
```
ivirshup, line 309: Could these arguments get some short documentation?

I was confused for a bit about when you would be using a correlation method 😄

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,6 +18,329 @@
...
+    def logreg(self, **kwds):
+        # if reference is not set, then the groups listed will be compared to the rest
+        # if reference is set, then the groups listed will be compared only to the other groups listed
+        from sklearn.linear_model import LogisticRegression
+
+        # Indexing with a series causes issues, possibly segfault
+        X = self.X[self.grouping_mask.values, :]
+
+        if len(self.groups_order) == 1:
+            raise Exception('Cannot perform logistic regression on a single cluster.')
```
ivirshup:
```suggestion
            raise ValueError('Cannot perform logistic regression on a single cluster.')
```

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,6 +18,329 @@
...
+            generate_test_results = self.wilcoxon()
+        elif method == 'logreg':
+            generate_test_results = self.logreg(**kwds)
+
+        for group_index, scores, pvals in generate_test_results:
+            group_name = str(self.groups_order[group_index])
+            self.groups_names.append(group_name)
+
+            scores_sort = np.abs(scores) if rankby_abs else scores
+            global_indices = _select_top_n(scores_sort, n_genes_user, self.X.shape[1])
```
ivirshup: What do you think about deprecating this feature? I don't think it makes sense to (by default) filter differentially expressed genes to any fixed `n`.

ivirshup (commented):
Huge improvement! Thanks for going thought this! I've added a few questions and thoughts on how this would be more simple. Most of the points are nitpicky, and I think I would be happy with the restructuring as is.

One important point, did you address the correctness issues for the rank test?

-------------------

For the point about treating the `rest` and `reference` cases equivalently, I thought the `_basic_stats` method could have a block like:

```python
        if self.ireference:
            mask_rest = self.groups_masks[self.ireference]
            X_rest = self.X[mask_rest]
            m, v = _get_mean_var(X_rest)
            self.means_rest = np.broadcast_to(m, (n_groups, n_genes))
            self.vars_rest = np.broadcast_to(v, (n_groups, n_genes))
```

So you wouldn't need to do checks in downstream functions.

---------------------

On the structure of the code, I like to minimize the states an object can be in when possible. This just makes it easier to reason about what you can do with an instance. For example, I probably would have tried to write this so the test was like a config for how the test should be run. Then there would be a function which applies that to each pair of arrays.

-------------------------

Also, a personal request for this:

Could we change the default number of top genes to be the number of genes present?

Koncopd:
Hi, @ivirshup . Thanks for the review. I'll address the comments soon.
No, this currently doesn't deal with correctness, tie correction and the other things in the issues, just general structure. But i'll definitely turn to them after this.

About the reference thing. Yes, you are right, i can use this approach of course. However, i didn't want to store the same value multiple times, it doesn't make much difference in performance, but still gives uneasy feelings, i would say.

upd: oh, i see that `numpy.broadcast_to` creates a view, so it should avoid the problem of storing the same value multiple times.

scanpy/tools/_rank_genes_groups.py
fidelram: I second this. All genes should be reported by default.

gokceneraslan:
Thanks so much for the amazing work @Koncopd 🎊

I think it'd be more convenient to have the same type of "keys" regardless of the provided `reference` option in sc.tl.rank_genes_groups. Right now I get 

`['params', 'pts', 'scores', 'names', 'pvals', 'pvals_adj', 'logfoldchanges']` 

keys in .uns['rank_genes_groups'], if I set a reference and 

`['params', 'pts', 'pts_rest', 'scores', 'names', 'pvals', 'pvals_adj', 'logfoldchanges']`

if I don't.

Wouldn't it be better to replace all *_rest in the code with *_reference, and calculate them based on whatever the reference is, rest or a specific group? 

Then we can provide 

`['params', 'pts', 'pts_reference', 'scores', 'names', 'pvals', 'pvals_adj', 'logfoldchanges']` 

without thinking about what the reference is. Seurat reports pct1 for the first group and pct2 for the second group, for example, which is nice and simple, IMO.

Also `sc.get.rank_genes_groups_df` should return pct and pct_reference. It'd be also great to have `pct` cutoffs in the `sc.get.rank_genes_groups_df` function like `pct_min`.  

People typically define DE genes using cutoffs like pval_adj < 0.05, log2fc>1, pct>0.1.

Edit: Just noticed similar comments from @ivirshup, sorry about replicating :)

falexwolf:
Thank you all for this! In particular, @Koncopd!

I know that this will cause a little more headache, but could we consider renaming to `rank_genes`?

The function has its name with a `_groups` suffix from very early considerations about distinguishing it from a second function that would rank genes by fitting a model on time or pseudotime.

But I don't see that function ever coming within the current Scanpy main API.

So, it'd be nice to have a simple and short name.

As backward compat will be broken to some degree anyway in 1.5, and a small converter for legacy `h5ad` (in the scanpy read function - renaming `.uns['rank_genes_groups']` to `.uns['rank_genes']` and making entries a dataframe instead of structured arrays). It'd be a good point in time, now.

Please slack me if you really want my feedback. Sorry that I'm so absent. 😔

falexwolf:
Instead of having `self.d` as the central attribute, why not having a central property `self.stats` and that's a DataFrame with a multi-index for the columns. The outer index is the group that you're comparing against a reference, the inner index loops over `gene`, `score`, `pval`, `pval_adj`, etc.

Meaning, there is a class for this besides the convenience function:
```
rg = sc.RankGenes()
```

And if you just want a dataframe and not store something in `adata`, you can do
```
rg.compute_stats(...)
```

After this `rg.stats` contains your results.

Within the convenience wrapper `sc.tl.rank_genes`, this `stats` attribute will be written to `adata`.

LuckyMD:
> I know that this will cause a little more headache, but could we consider renaming to `rank_genes`?

Would you not maybe do this in stages with a `DeprecationWarning` first for a couple of releases? This change would break nearly everyone's pipelines and published notebooks. It's not a lot of work to change... but it might warrant a longer warning time.

Koncopd:
@falexwolf 
image
Looks fine with miltindex dataframe. However a bit awkward with this names column when `n_genes` is set. I will set `n_genes` to all by default but do we need this at all? If there is no `n_genes`, then names can be moved to dataframe's index as they are the same.

Koncopd:
`adata.rename_categories` doesn't work with dataframes in uns.

scanpy/tools/_rank_genes_groups.py
Koncopd: This should signify that this class is not intended for use outside of `rank_genes_groups` function.

scanpy/tools/_rank_genes_groups.py
Koncopd: Set is not needed, but your suggestion doesn't seem to work properly.

scanpy/tools/_rank_genes_groups.py
Koncopd: Changed.

scanpy/tools/_rank_genes_groups.py
Koncopd: Changed.

ivirshup:
> adata.rename_categories doesn't work with dataframes in uns.

* What problem is this causing?
* @falexwolf, why did this method get un-deprecated?

> Looks fine with miltindex dataframe. However a bit awkward with this names column when n_genes is set. I will set n_genes to all by default but do we need this at all?

I personally find MultiIndexes a bit hard to work with. Could you show how they would be used here? For example, how would I just get a dataframe for the naive T cells vs. rest?

I'm also not sure I get why we'd order genes by rank, when there are multiple comparisons in the table. What operations does this make easier?

**Most importantly**, I don't think we have support for reading and writing multi indexes in anndata.

An alternative would be to just have an entry in `uns` that looked like:

```python
adata.uns[key_added] = {
    "params": {
        "groupby": "leiden",
        "reference": "rest",
        "test": "wilcoxon",
        "rep": "X"
    },
    "results": {
        "1": ...,  # pd.DataFrame, with index of .var_names
        "2": ...,  #etc
    },
}
```

scanpy/tools/_rank_genes_groups.py
ivirshup: Ah, I must have tested it for the array not the series. Apparently `__contains__` for a series looks in the index. Just removing the `set` would be good.

fidelram:
I second the suggestion by @falexwolf to rename the function to something simpler but also to keep the previous functionality with a Deprecate message as suggested by @LuckyMD.

@Koncopd The changes also requires adapting the corresponding `sc.pl.rank_genes_groups*` functions. I can take over that once the PR is ready.

scanpy/tools/_rank_genes_groups.py
```diff
@@ -18,18 +18,347 @@
...
+
+        df = pd.DataFrame(data=get_chunk(X, left, right))
+        ranks = df.rank()
+        yield ranks, left, right
+
+        left = right
+
+
+class _RankGenes:
+    def __init__(self, adata, groups, groupby, reference, use_raw, layer):
```
ivirshup: Could these values get documented? At the very least, what their types should be?

ivirshup:
@Koncopd @falexwolf, I have a proposal for this PR.

This PR makes a huge improvement to the internals of the differential expression code. Could we aim to limit the scope here to just improving the internals (i.e., no API introductions or changes)? Then we can introduce and change APIs in later PRs?

My interest here is getting the improved internals onto master ASAP, so that other people can hack on it. This would allow correctness improvements to happen in parallel with API design.

Koncopd:
Sorry for the delay, i'll return to this soon.

@ivirshup 
I have nothing against this suggestion, actually this was my original plan.

scanpy/tools/_rank_genes_groups.py
Koncopd: These are the same parameters as in the main function, so the docs are there.

scanpy/tools/_rank_genes_groups.py
Koncopd: Removed yield.

Koncopd:
I've restored the original output format. 
However, there are still a few changes to the api. 
if `pts=True` (default `False`) the function also computes the fraction of cells expressing the genes for each group, and if `reference='rest'` then `pts_rest` is added with fraction of cells expressing the genes for the complement of each group.
`n_genes` is set to all genes by default.

falexwolf (approved):
Thank you. This is great!

Koncopd:
Docs are built fine locally, i don't know why readthedocs fails.

PR #1160, opened 2020-04-12
bugfix: dense zero centering

simonwm:
zero_center argument was ignored for the dense expression matrix path, leading to zero centering even if zero_center=False was explicitly given as argument


ivirshup:
Thanks for catching this! Could you add a test so this doesn't happen again in the future?

Also, I believe the tests that were failing were due to a umap release, not anything you changed.

simonwm:
Sounds reasonable, I'll look into it.

simonwm:
Just added a test and changed the behaviour of scale a little more.
The case of zero variance was until now replace with an arbitrary tiny variance, which arbitrarily blew up the scaled value and made it completely meaningless `scale[scale == 0] = 1e-12`.
Now I put instead `scale[scale == 0] = 1`. This yields the same result for `zero_center == True`: all values set to `0`, anyway (but with less arbitrary magic numbers and maybe less rounding errors). But if `zero_zenter == False`, unscalable values are untouched. This only affected the dense codepath where zero-centering was done afterwards anyway due to the original bug. Therefore this is no code breaking change.
But I also moved this statement before the sparse check to have consistent handling of sparse and dense data. Before that the sparse path wrote infs in the values (unchecked divison by zero) - this is a potentially code breaking change, but it only leads to the behaviour already stated in the documentation. I personally think that code relying on this undocumented behaviour should be rewritten, anyway...
In the new test I explicitly check for this behaviour to make it well defined.
Similar for integer datatypes (resulted in an error), they are now converted to floating point for scaling and return a copy.

BTW: In order to make the tests run in my conda environment, I had to remove every reference to compare_images from matplotlib.testing.compare. There seems to be a version conflict in the version checking... It always gave errors like the following:
`________________ ERROR collecting scanpy/tests/test_plotting.py ________________
scanpy/tests/test_plotting.py:16: in <module>
    from matplotlib.testing.compare import compare_images
~/.conda/envs/custom/lib/python3.8/site-packages/matplotlib/testing/compare.py:240: in <module>
    _update_converter()
~/.conda/envs/custom/lib/python3.8/site-packages/matplotlib/testing/compare.py:222: in _update_converter
    mpl._get_executable_info("gs")
~/.conda/envs/custom/lib/python3.8/site-packages/matplotlib/__init__.py:364: in _get_executable_info
    return impl([e, "--version"], "(.*)", "9")
~/.conda/envs/custom/lib/python3.8/site-packages/matplotlib/__init__.py:346: in impl
    if min_ver is not None and version < min_ver:
~/.conda/envs/custom/lib/python3.8/distutils/version.py:52: in __lt__
    c = self._cmp(other)
~/.conda/envs/custom/lib/python3.8/distutils/version.py:337: in _cmp
    if self.version < other.version:
E   TypeError: '<' not supported between instances of 'str' and 'int''`
I have the current matplotlib version 3.2.1.

simonwm:
@ivirshup Do you have any more comments or request regarding this fix?

ivirshup:
Sorry for the late response!

In general, it's good to keep PR's focussed on the initial topic, and make any other changes in new PRs. This will make the review process go a bit quicker.

I've merged some code (already in progress) for simplifying the control flow of `scale`, and am going to ask that you rebase on that. Please let me know if you run into any trouble there.

For the change in substituted values (from `1e-12` to `1`), could you provide an example I could run of `inf` values being generated? I just get zeros.

simonwm:
I'll keep the focused PRs in mind.

The new control flow seems reasonbale. I was unfamiliar with the `@singledispatch` decorator, so it was not directly self explanatory for me. I expected infinity loops and unreachable code, but it turned out to be correct (:

`inf` is there for the sparse case `zero_center=False` and a gene with zero variance but finite mean. Here is the example (slightly modified from the new `tests/test_scaling.py`), with the four cases for genes `(mean==0,mean!=0) x (var==0,var!=0)`:
```
X = csr_matrix([[-1,2,0,0],[1,2,4,0],[0,2,2,0]])
X = sc.pp.scale(Xtest, copy=True, zero_center=False)
X
```
If `std[std == 0] = eps` (`eps!=0`) is only in the dense path, I get: `array([[-1., inf,  0.,  0.],  [ 1., inf,  2.,  0.],  [ 0., inf,  1.,  0.]])`
if `std[std == 0] = 1` is before the sparse/dense split, I get: `array([[-1.,  2.,  0.,  0.],  [ 1.,  2.,  2.,  0.],  [ 0.,  2.,  1.,  0.]])`
if `std[std == 0] = 1e-12` is before the sparse/dense split, I get:  `array([[-1.,  2.e+12,  0.,  0.],  [ 1.,  2.e+12,  2.,  0.],  [ 0.,  2.e+12,  1.,  0.]])`

This suggests, that `0/0` in a sparse setting remains `0` (I guess thats what you see); it makes sense for an efficient sparse matrix implementation, as the `0` is not even represented in the sparse data, so scaling with anything is optimized away. If it were not, it should probably yield `nan` and not `0`.
But if you have something finite with zero variance, you get an explicit `<finite>/0=inf`. [This IS an edge case, and probably never the case in real expression data, but still the behaviour should be consistent and well defined.]
Now, if you have the statement `std[std == 0] = eps` before the sparse/dense split, the `inf` is caught in both cases.

The change from `eps=1e-12` to `eps=1` only makes the values keep their original values without zero centering, instead of having these values multiplied by the arbitrary `1e12`. I read the intent for this behaviour into the Note in the docs "Variables (genes) that do not display any variation (are constant across all observations) are retained". Setting them to zero makes no sense to me without zero centering. With zero centering setting the values to zero is automatic - if they are not infinity before. An argument can be made for setting these values to `inf` without zero centering and to `0` or `nan` with zero centering. That is something for you to decide (I guess).

simonwm:
@ivirshup PS: Just added a different PR #1194 for another fix, which might be focused enough (:

scanpy/preprocessing/_simple.py
```diff
@@ -772,15 +772,26 @@ def scale_array(
...
 
+    if np.issubdtype(X.dtype, np.integer):
+        logg.info(
+            '... scale_data: as scaling leads to float results, integer '
+            'input is cast to float, returning copy.'
+        )
+        if issparse(X):
+            X = X.__class__(X, dtype=np.float) # keep the identical sparse matrix type but with float data
+        else:
+            X = np.array(X, dtype=np.float) # keep dense array but with float data
```
ivirshup: This could be a little more simple:

```suggestion
    if np.issubdtype(X.dtype, np.integer):
        logg.info(
            '... as scaling leads to float results, integer '
            'input is cast to float, returning copy.'
        )
        X = X.astype(float)
```

ivirshup (commented):
One last little change and this all looks good! Thanks for pointing out those other cases, I hadn't tested those 😊. Looking back at the history, I think the choice of 1e-12 was just to avoid `nan` https://github.com/theislab/scanpy/commit/dce2be194a6ff865ecaeb939f3c990f6c3b0e244.

In terms of what values should be returned we are mainly aiming for returning the same values that `sklearn.preprocessing.StandardScalar` would.

simonwm:
Cool! I knew the .astype method only from pandas...

ivirshup (approved):

PR #1173, opened 2020-04-21
Getters and setters for X

### LINKED ISSUE #1089: Subset and Rescale?
We often identify a subset of cells as irrelevant noise cells and hope to discard them during the analysis. I am a bit confused about how to perform such operations in Scanpy.

It's a common practice in other analysis tool like Seurat to do ScaleData across cells so that the relative expression level is adjusted without uninteresting cells' influences. 

Such operation is supported by Seurat by providing multiple "Assay", such as `counts`, `data`, and `scale.data`, which stores the raw UMI counts, column normalized data (across genes, log1p), and row normalized data (across samples, zscores). 

I noticed the scaled data are stored in `adata.X` in scanpy. With this design, how can we access the raw UMI count and do re-scaling if I hope to subset the data?

Looking forward your reply. Thanks.

ivirshup:
This PR adds some utilities for choosing which values to use for `X` inside scanpy functions. These utilities include a getter and setter for representations of the observations and some checks for testing these functions.

This approach has been applied to `sc.pp.scale` and `sc.pp.log1p` as demonstrations, and because that code was in need of a refactor. I think it results in cleaner code with more standardized behaviors.

The implementation of `sc.pp.scale` is just a hammered out version of what I commented here: https://github.com/theislab/scanpy/pull/1135#issuecomment-608200735

Still todo:

- [x] Figure out how documentation should work. In particular automatic documentation with type variables
- [x] Figure out tab completion for arguments not in the default/ fallback definition

This PR would:

* Close #1089
* Supercede #1135

### DIFF
--- a/docs/extensions/param_police.py
+++ b/docs/extensions/param_police.py
@@ -36,8 +36,9 @@ def show_param_warnings(app, exception):
             file_name,
             line,
         )
-    if param_warnings:
-        raise RuntimeError('Encountered text parameter type. Use annotations.')
+    # TODO: Figure out with @flying-sheep what to do about these
+    # if param_warnings:
+    #     raise RuntimeError('Encountered text parameter type. Use annotations.')
 
 
 def setup(app: Sphinx):
--- a/docs/extensions/param_police.py
+++ b/docs/extensions/param_police.py
@@ -36,9 +36,8 @@ def show_param_warnings(app, exception):
             file_name,
             line,
         )
-    # TODO: Figure out with @flying-sheep what to do about these
-    # if param_warnings:
-    #     raise RuntimeError('Encountered text parameter type. Use annotations.')
+    if param_warnings:
+        raise RuntimeError('Encountered text parameter type. Use annotations.')
 
 
 def setup(app: Sphinx):
--- a/scanpy/_utils.py
+++ b/scanpy/_utils.py
@@ -169,6 +169,19 @@ def _doc_params(**kwds):
     return dec
 
 
+def _check_array_function_arguments(**kwargs):
+    """Checks for invalid arguments when an array is passed.
+
+    Helper for functions that work on either AnnData objects or array-likes.
+    """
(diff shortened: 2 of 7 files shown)


docs/extensions/param_police.py
```diff
@@ -36,8 +36,9 @@ def show_param_warnings(app, exception):
             file_name,
             line,
         )
-    if param_warnings:
-        raise RuntimeError('Encountered text parameter type. Use annotations.')
+    # TODO: Figure out with @flying-sheep what to do about these
+    # if param_warnings:
+    #     raise RuntimeError('Encountered text parameter type. Use annotations.')
```
flying-sheep: Hi, I’m currently on holidays, but I’m checking out mentions, thanks for notifying me here.

Every deactivation of an automation is a step backward in maintainability. 
If one needs to, of course we can add a case-by-case deactivation with some mechanism, but never flat out like this.

I don’t know how documenting singledispatch stuff works, and why one needs to copy the parameters that aren’t used for some variants. Can you link me to the docs of some singledispatch function to show me?

flying-sheep (changes requested):

docs/extensions/param_police.py
ivirshup: [Here's what the docs for the function look like in this PR](https://icb-scanpy--1173.com.readthedocs.build/en/1173/api/scanpy.pp.log1p.html#scanpy.pp.log1p)  (readthedocs PR builds are already useful!)

The issue is that the most generic method (i.e. the fallback) for a function is the one that get's the doc-string. There can be other methods which take more arguments that we would want to document. The question is how do we document the arguments, and how do we get tab completion?

ivirshup: I've just added the arguments to the function declaration and throw an error if they are passed. It would be nice to set default values and tab completion with a bit more granularity, but this PR does not need to wait on that.

PR #1179, opened 2020-04-21
Add neighbors_key to paga

Koncopd:
And changes to neighbors return section of the docstring.

### DIFF
--- a/scanpy/neighbors/__init__.py
+++ b/scanpy/neighbors/__init__.py
@@ -101,10 +101,12 @@ def neighbors(
     -------
     Depending on `copy`, updates or returns `adata` with the following:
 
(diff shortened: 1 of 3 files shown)


Koncopd:
@ivirshup @flying-sheep 
Hi, do you know how to find out why `docs/readthedocs.com` fails?

ivirshup:
Not sure. Do the docs build fine for you locally?

Koncopd:
@ivirshup yes, they build fine but there are a lot of warnings of the form `"WARNING: py:class reference target not found: scanpy._compat.Literal_"`.

Koncopd:
Ok, seems like the problem was on the side of readthedocs.com.

PR #1182, opened 2020-04-22
Seurat v3 HVG method

### LINKED ISSUE #993: Seurat v3 VST highly variable gene method
I find this method to be the most conceptually straightforward and it gives great results in my tests.

I have a rough implementation in python. I see that making a PR would be more involved as the code relies on log-transformed data, while the Seurat method should be on the raw counts.

I also understand that adding `rpy2` to scanpy could be a bit challenging so I have a close approximation with the stats models library.

```python
import statsmodels.api as sm
def seurat_v3_highly_variable_genes(adata, n_top_genes=4000, use_lowess=False):
    norm_gene_vars = []
    del_batch = False
    if "batch" not in adata.obs_keys():
        del_batch = True
        adata.obs["batch"] = np.zeros((adata.X.shape[0]))
    for b in np.unique(adata.obs["batch"]):
        var = adata[adata.obs["batch"] == b].X.var(0)
        print(var.shape)
        mean = adata[adata.obs["batch"] == b].X.mean(0)
        estimat_var = np.zeros((adata.X.shape[1]))

        y = np.log10(var)
        x = np.log10(mean)
        if use_lowess is True:
            lowess = sm.nonparametric.lowess
            # output is sorted by x
            v = lowess(y, x, frac=0.15)
            estimat_var[np.argsort(x)] = v[:, 1]
        else:
            estimat_var = loess(y, x)

        norm_values = (adata[adata.obs["batch"] == b].X - mean) / np.sqrt(10 ** estimat_var)
        # as in seurat paper, clip max values
        norm_values = np.clip(
            norm_values, None, np.sqrt(np.sum(adata.obs["batch"] == b))
        )
        norm_gene_var = norm_values.var(0)
        norm_gene_vars.append(norm_gene_var.reshape(1, -1))

    norm_gene_vars = np.concatenate(norm_gene_vars, axis=0)
    ranked_norm_gene_vars = np.argsort(np.argsort(norm_gene_vars, axis=1), axis=1)
    median_norm_gene_vars = np.median(norm_gene_vars, axis=0)
    median_ranked = np.median(ranked_norm_gene_vars, axis=0)

    num_batches_high_var = np.sum(
        ranked_norm_gene_vars >= (adata.X.shape[1] - n_top_genes), axis=0
    )
    df = pd.DataFrame(index=np.array(adata.var_names))
    df["highly_variable_n_batches"] = num_batches_high_var
    df["highly_variable_median_rank"] = median_ranked

    df["highly_variable_median_variance"] = median_norm_gene_vars
    df.sort_values(
        ["highly_variable_n_batches", "highly_variable_median_rank"],
        ascending=False,
        na_position="last",
        inplace=True,
    )
    df["highly_variable"] = False
    df.loc[:n_top_genes, "highly_variable"] = True
    df = df.loc[adata.var_names]

    if del_batch is True:
        del adata.obs["batch"]

    adata.var["highly_variable"] = df["highly_variable"].values
    adata.var["highly_variable_n_batches"] = df["highly_variable_n_batches"].values
    adata.var["highly_variable_median_variance"] = df[
        "highly_variable_median_variance"
    ].values


def loess(y, x, span=0.3):
    from rpy2.robjects import r
    import rpy2.robjects as robjects

    a, b = robjects.FloatVector(x), robjects.FloatVector(y)
    df = robjects.DataFrame({"a": a, "b": b})
    loess_fit = r.loess("b ~ a", data=df, span=span)

    return np.array(loess_fit[loess_fit.names.index("fitted")])
```

adamgayoso:
Fixes #993.

This is an approximate implementation of the Seurat v3 hvg method. The only difference should be the use of lowess instead of loess (which is not available in python as far as I know).

This method takes the UMI counts as input. The way HVGs from batches are merged is also different from the other flavors. As such, I didn't see a straightforward way to integrate this in the existing HVG code.


adamgayoso:
Please let me know if you have any recommendations @ivirshup !

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -338,12 +369,123 @@ def highly_variable_genes(
...
+        lightweight batch correction method.
+    lowess_frac
+        The fraction of the data (cells) used when estimating the variance in the lowess model fit.
+    """
+    import statsmodels
+
+    lowess = statsmodels.nonparametric.lowess
+
+    if batch_key is None:
+        batch_info = pd.Categorical(np.zeros((adata.X.shape[0])).astype(int))
```
```suggestion
        batch_info = pd.Categorical(np.zeros(adata.shape[0], dtype=int))
```

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -338,12 +369,123 @@ def highly_variable_genes(
...
+    else:
+        batch_info = adata.obs[batch_key]
+
+    norm_gene_vars = []
+    for b in np.unique(batch_info):
+
+        mean, var = materialize_as_ndarray(_get_mean_var(adata[batch_info == b].X))
```
ivirshup: What does the `materialize_as_ndarray` do here?

```suggestion
        mean, var = _get_mean_var(adata[batch_info == b].X)
```

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -338,12 +369,123 @@ def highly_variable_genes(
...
+        not_const = var > 0
+        estimat_var = np.zeros((adata.X.shape[1]))
```
```suggestion
        estimat_var = np.zeros(adata.shape[1])
```

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -338,12 +369,123 @@ def highly_variable_genes(
...
+        # get normalized variance
+        reg_std = np.sqrt(10 ** estimat_var)
+        batch_counts = adata[batch_info == b].X.copy()
+        # clip large values as in Seurat
+        N = np.sum(batch_info == b)
+        vmax = np.sqrt(N)
+        clip_val = reg_std * vmax + mean
+        # could be something faster here
+        for g in range(batch_counts.shape[1]):
+            batch_counts[:, g][batch_counts[:, g] > vmax] = clip_val[g]
```
ivirshup: Does this work for sparse arrays?  I think you'd be setting values on a copy here

```python
from scipy import sparse
import numpy as np

s = sparse.csr_matrix(np.zeros((10, 10)))
s[:, 0][1:5] = 1
assert s.nnz == 0
```

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -338,12 +369,123 @@ def highly_variable_genes(
...
+    batch_key
+        If specified, highly-variable genes are selected within each batch separately and merged.
+        This simple process avoids the selection of batch-specific genes and acts as a
+        lightweight batch correction method.
+    lowess_frac
+        The fraction of the data (cells) used when estimating the variance in the lowess model fit.
+    """
+    import statsmodels
+
+    lowess = statsmodels.nonparametric.lowess
```
ivirshup: I don't think this import will work. [`statsmodels` import api is a bit strange](https://www.statsmodels.org/stable/api-structure.html#import-paths-and-structure), but I think this should work:

```suggestion
    from statsmodels.nonparametric.smoothers_lowess import lowess
```

ivirshup (changes requested):
Thanks for the PR!

I've made some comments on the code about correctness. I'm still working on making sure I completely understand the code. I'd like for @gokceneraslan and @LuckyMD to take a look at this too, since they're more familiar with this.

Thanks for including that link to the overleaf page, that will be useful in understanding this. Do you think some of that could just go directly in the doc-string?

Could you add:

* Tests
* An example for us, showing that it works/ is useful
* An short example of usage in the doc-string

scanpy/preprocessing/_highly_variable_genes.py
```suggestion
        if sp_sparse.issparse(batch_counts):
            batch_counts = sp_sparse.csr_matrix(batch_counts)
            mask = batch_counts.data > vmax
            batch_counts.data[mask] = clip_val[batch_counts.indices[mask]]
        else:
            for g in range(batch_counts.shape[1]):
                batch_counts[:, g][batch_counts[:, g] > vmax] = clip_val[g]
```

I believe this should work. Could probably be optimized further by using a `csc` matrix, and not iterating over all the genes. This version takes makes the whole call take about half a second for 2700 cells x 16634 genes in csr format, and 1.15 seconds for a dense array.

Edit: Seems like optimizing further might not be worth it. With this implementation, 90% of the compute time is in statsmodels for the sparse array example.

Edit 2: Figured out a possible way to vectorize the dense case. You'll also want to change the definition of the `batch_counts` to make sure it's of the same dtype as `clip_val`.

```python
        batch_counts = adata[batch_info == b].X.astype(np.float64).copy()
        ...
        if sp_sparse.issparse(batch_counts):
            batch_counts = sp_sparse.csr_matrix(batch_counts)
            mask = batch_counts.data > vmax
            batch_counts.data[mask] = clip_val[batch_counts.indices[mask]]
        else:
            np.putmask(
                batch_counts,
                batch_counts > vmax,
                np.broadcast_to(clip_val, batch_counts.shape)
            )
```

This cuts the dense array case to ~0.8 seconds from 1.15.

scanpy/preprocessing/_highly_variable_genes.py
adamgayoso: This I'm not sure about, I see it in the current hvg function, as well as a call to `sanitize_adata`. Is it necessary?

scanpy/preprocessing/_highly_variable_genes.py
adamgayoso: good catch, this looks like a nice solution

scanpy/preprocessing/_highly_variable_genes.py
ivirshup: `sanitize_anndata` should change views of anndata objects into actual ones, and will clean up categorical values.

`materialize_as_ndarray` is mostly used for forcing distributed or lazy arrays into memory. It's only necessary if this function would work for dask arrays.

adamgayoso:
Thanks for the review. I will add the other requested changes/additions soon.

So also the way I interpreted the way they clip values based on the Stuart '19 publication is different than what it appears they do in their code, which I linked in the overleaf doc in the docstring. My original interpretation is the first code in the linked issue.

adamgayoso:
@ivirshup I added a custom implementation of loess, which uses a degree 2 polynomial (closer to seurat implementation). It's a bit slow, I'm sure it can be sped up with numba, but it's having a problem with this line:

https://github.com/adamgayoso/scanpy/blob/b48d5729a075bc6f7fa0bbbf38a856d4c2e720ba/scanpy/preprocessing/_highly_variable_genes.py#L607

Any idea why?

Edit: I got it to work by not using clip, but it's still a bit slow. Please let me know if you have any ideas!

Edit 2: I found a stable python implementation of degree 2 loess

adamgayoso:
I'm also not sure why the test is failing -- it works interactively locally for me.

adamgayoso:
@ivirshup is it possible that Travis has cached pbmc3k and that's what's causing the error? I really don't have it running pytest locally either. 

Also as far as the code review -- I understand code is duplicated, but this code does not really fit in the existing implementation because it works a bit differently and requires raw data. Let me know how you'd like to address this.

Thanks!

adamgayoso:
This got a bit messy -- travis is defeating me because the tests work locally. I'm going to close this and figure it out on my fork and make a new PR with a clean history.

ivirshup:
Sorry for the lack of feedback! I got caught up in my own work.

* Have you made any progress on the travis issue?
* Is it an issue of getting different results from computing variance?

adamgayoso:
No worries - I really can't tell what the issue is since it works perfectly fine locally. My sense now is that
```
mean, var = _get_mean_var(adata[batch_info == b].X)
not_const = var > 0
```
is not working properly on travis because based on this method, if the regularized std is 1, which occurs when it's constant, the new variance would just be the same as in the variances column of the df, which is what's happening on travis.

gokceneraslan:
Hey @adamgayoso ,

I really love this HVG method, but sometimes I get the following error from the loess fit:

```pytb
<ipython-input-244-764977f87ce6> in <module>
----> 1 sc.pp.highly_variable_genes(ad_sub, n_top_genes=1500, flavor='seurat_v3', layer='counts')
      2 sc.pp.pca(ad_sub)
      3 sc.pp.neighbors(ad_sub)
      4 sc.tl.umap(ad_sub)
      5 sc.tl.leiden(ad_sub, resolution=2.0)

~/.miniconda3/lib/python3.8/site-packages/scanpy/preprocessing/_highly_variable_genes.py in highly_variable_genes(adata, layer, n_top_genes, min_disp, max_disp, min_mean, max_mean, span, n_bins, flavor, subset, inplace, batch_key)
    413 
    414     if flavor == 'seurat_v3':
--> 415         return _highly_variable_genes_seurat_v3(
    416             adata,
    417             layer=layer,

~/.miniconda3/lib/python3.8/site-packages/scanpy/preprocessing/_highly_variable_genes.py in _highly_variable_genes_seurat_v3(adata, layer, n_top_genes, batch_key, span, subset, inplace)
     82         x = np.log10(mean[not_const])
     83         model = loess(x, y, span=span, degree=2)
---> 84         model.fit()
     85         estimat_var[not_const] = model.outputs.fitted_values
     86         reg_std = np.sqrt(10 ** estimat_var)

_loess.pyx in _loess.loess.fit()

ValueError: b'reciprocal condition number  7.4971e-16\n'
```

This is due to the very low but non-zero variance genes, I think. It goes away when I run `sc.pp.filter_genes(ad_sub, min_cells=5)` but not when I run only `sc.pp.filter_genes(ad_sub, min_cells=1)`.

Maybe we can make the variance check more stringent, or we can print a better error message for the users?

PR #1188, opened 2020-04-29
Make normalize_total actualize views

### LINKED ISSUE #1183: normalize_total throws ValueError when passed a large view
When I run sc.pp.normalize_total(adata, target_sum=1e4),I got the error:ValueError: could not convert integer scalar,and How can I fixed this issue?

ivirshup:
Fixes #1183

Only thing I'm not sure of here is if this breaks any out-of-core workflows.


ivirshup:
Thought about it a bit more. I'm pretty sure we've dropped the "modify a subset of a backed dataset" use-case since most other functions won't respect that.

PR #1196, opened 2020-05-03
improved memory efficiency in _score_genes

redst4r:
In short: Avoids creating dense matrices for score calculation.

Currently, `_score_genes()` unnecessarily creates dense matrices to use the `np.nanmean` function (which doesn't work on sparse matrices out of the box).
This causes memory problems for larger datasets (anything 50k cells got me into trouble) and can be completely avoided with a `nanmean()` implementation of sparse matrices.

Not sure why the build is failing now in completely unrelated functions though!


ivirshup (commented):
Thanks for the PR!

A couple questions,

* Does score genes even work if the are `nan` values in the array?
* Could any tests please be added to the test suite?

redst4r:
hi,

- yeah, it should work if the original array contain `nan` (i.e. it would just ignore these entries)
- ah, just saw that there's a `tests/test_score_genes.py` file, I'll add a few tests the next days!

redst4r:
hi,

finally managed to add some tests. Had to refactor the original `test_score_genes.py` a little, I hope that's ok: The one test that was already there still exists, I just pulled out the creation of the adata into a separate function.

ivirshup:
LGTM, thanks!

PR #1200, opened 2020-05-06
Bugfix scale centering

ivirshup:


ivirshup:
@simonwm I've merged #1160 here since github and travis were being difficult about merging

PR #1204, opened 2020-05-06
Seurat v3 hvg implementation

### LINKED ISSUE #993: Seurat v3 VST highly variable gene method

adamgayoso:
Fixes #993.


adamgayoso:
@ivirshup -- I still can't tell why Travis is failing. For some reason on Travis, loess is outputting a zero for the gene mentioned in the error message, but this doesn't happen locally for me.

adamgayoso:
I think the inconsistency I'm seeing is that the loess function gives different results on my Mac than it does on linux.

https://github.com/has2k1/scikit-misc/issues/8

adamgayoso:
@ivirshup I'm not sure why this change makes it work, but looks like it works. Anyway, let me know how you would like to handle the repeated code for storing the outputs! Thanks.

requirements.txt
```diff
@@ -20,3 +20,4 @@ umap-learn>=0.3.10
 legacy-api-wrap
 setuptools_scm
 packaging
+scikit-misc
```
ivirshup: I'm not sure I want this to be a hard dependency. It doesn't look like it's been a particularly active repo.

I think I'm okay with it being an optional dependency since (looking through it's dependents) I see a fair amount of packages in our ecosystem.

ivirshup (commented):
Just a quick review for the moment:

Could the repeated code just get put into a separate function?

I'm not sure I love the name. Could it be more descriptive of what it does?

I might leave a more in depth review to people more familiar with this code. @LuckyMD or @gokceneraslan do you think you could take a look at this?

LuckyMD:
> I might leave a more in depth review to people more familiar with this code. @LuckyMD or @gokceneraslan do you think you could take a look at this?

Sorry, I don't think I will have time for this in the next 2 weeks. I'm finishing up a 2 larger projects atm.

requirements.txt
has2k1: @ivirshup, `scikit-misc` is still actively maintained but it is mostly used through [plotnine](https://github.com/has2k1/plotnine), where it is optional dependency.

adamgayoso:
@ivirshup this should be good to go now. I don't thin Codacy failing has anything to do with my code. @gokceneraslan would you be able to review?

.travis.yml
```diff
@@ -24,7 +24,7 @@ cache:
   - directories: data
 install:
   - pip install docutils sphinx
-  - pip install -e .[test,louvain,leiden,magic,scvi]
+  - pip install -e .[test,louvain,leiden,magic,scvi,skmisc]
```
has2k1: The package on pypi is called `scikit-misc`.

adamgayoso: yes this is already addressed in setup.py, this is just a shortname

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -323,11 +344,184 @@ def highly_variable_genes(
...
+    )
+    df["highly_variable"] = False
+    df.loc[: int(n_top_genes), "highly_variable"] = True
+    df = df.loc[adata.var_names]
+
+    if inplace or subset:
+        logg.hint(
+            'added\n'
+            '    \'highly_variable\', boolean vector (adata.var)\n'
+            '    \'highly_variable_rank\', int vector (adata.var)\n'
```
gokceneraslan: This looks float when I run the function, can you double check?

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -323,11 +344,184 @@ def highly_variable_genes(
...
+    """\
+    Annotate highly variable genes (HVGs) as in Seurat v3 [Stuart19]_.
+
+    Expects raw count data.
+
+    Computes for each gene a normalized variance. First, the data is standardized
+    (i.e., z-score normalization per feature) with a regularized standard deviation.
+    Next, the normalized variance is computed as the variance of each gene after the transformation.
+    Genes are ranked by the normalized variance. Optionally, HVGs are merged from different batches
+    as a form of lightweight batch correction.
```
gokceneraslan: Maybe it's also good to shortly explain how batch-specific HVGs are merged?

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -323,11 +344,184 @@ def highly_variable_genes(
...
+            adata.var['highly_variable_intersection'] = df[
+                'highly_variable_intersection'
+            ].values
+        if subset:
+            adata._inplace_subset_var(df['highly_variable'].values)
+    else:
+        return df
+
+
+def highly_variable_genes_norm_variance(
```
gokceneraslan: One concern is that it's implemented as a separate function. On the one hand I like that the interface is cleaner and specific to the method, on the other hand we already have a "flavor" option in the existing HVG function. So I was wondering if it'd make sense to add this as a "seuratv3" flavor. 

Implementing it as a separate function might also complicate things like `sc.pl.highly_variable_genes()`. If we unify two HVG functions and save the flavor under e.g. `adata.uns['hvg']`, `sc.pl.highly_variable_genes` might look there and plot it accordingly.

More like a minor thing but a new function would also be harder to "discover" for the new users. Flavor option of the existing HVG function has a better visibility.

scanpy/preprocessing/_highly_variable_genes.py
gokceneraslan: Do you see any improvement when you used the batch_key option? In my experience, it was worse than the existing batch_key + n_top_genes=2000 combination. It's not a big deal though. I wonder if it's due to the different merging strategy.

scanpy/tests/test_highly_variable_genes.py
```diff
@@ -125,45 +159,35 @@ def test_filter_genes_dispersion_compare_to_seurat():
 
 def test_highly_variable_genes_batches():
```
gokceneraslan, line 182: Can you add similar batch HVG tests? I might have missed because of the "black" modifications.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -323,11 +344,184 @@ def highly_variable_genes(
...
+    df = pd.DataFrame(index=np.array(adata.var_names))
+    df["highly_variable_nbatches"] = num_batches_high_var
+    df["highly_variable_rank"] = median_ranked
+    df["variances_norm"] = np.mean(norm_gene_vars, axis=0)
+    df["means"] = mean
+    df["variances"] = var
+
+    df.sort_values(
+        ["highly_variable_nbatches", "highly_variable_rank"],
+        ascending=False,
```
gokceneraslan: Should this be `[False, True]`? Not clear if bigger highly_variable_rank value means less or more variable.

gokceneraslan (commented):
It's an amazing contribution, thanks so much @adamgayoso. I added some comments.

scanpy/preprocessing/_highly_variable_genes.py
adamgayoso: Yeah, I'll double check

scanpy/preprocessing/_highly_variable_genes.py
adamgayoso: I can put explanation, honestly I'm not sure which is better (batch_key or not), though the implementation is the same as seurat v3.

scanpy/preprocessing/_highly_variable_genes.py
adamgayoso: I wanted to implement it in the same function, but wasn't sure how to:

1. deal with the fact that this requires count input and the others want log normalized (is there anything to check this?)
2. the main function is really hardcoded with the "dispersion" type keys and it seemed like adding this function would make the docstring really complicated.

any thoughts on this @gokceneraslan ?

scanpy/tests/test_highly_variable_genes.py
adamgayoso: sure!

scanpy/preprocessing/_highly_variable_genes.py
adamgayoso: higher rank means higher normalized variance (i.e., more variable)

scanpy/preprocessing/_highly_variable_genes.py
gokceneraslan: > I wanted to implement it in the same function, but wasn't sure how to:
> 
> 1. deal with the fact that this requires count input and the others want log normalized (is there anything to check this?)

We store log base under `.uns`, see: 

https://github.com/theislab/scanpy/blob/master/scanpy/preprocessing/_simple.py#L353

but it's not super reliable because there might be older anndata files that are generated before we introduced that feature.

Actually, we already exponentiate the data in the seurat flavor of the HVG function, so it also expects data in the non-logarithmized scale, see:

https://github.com/theislab/scanpy/blob/master/scanpy/preprocessing/_highly_variable_genes.py#L45

I think the reason we require log-transformed data is that cellranger flavor uses `mean(log(X))` but the seurat flavor uses `log(mean(X))`, so if flavor is seurat we technically do `log(mean(exp(log(X))))` 😄 

> 2. the main function is really hardcoded with the "dispersion" type keys and it seemed like adding this function would make the docstring really complicated.

I see, yes it'd be a bit ugly. But I think if we write that min/max_dispersion and min/max_mean parameters are ignored if flavor is seurat_v3, it'd be ok. We already have n_top_genes parameter that the new flavor can require.

This overall requires big changes though, @ivirshup wdyt?

adamgayoso: > We store log base under .uns, see:

> 
> https://github.com/theislab/scanpy/blob/master/scanpy/preprocessing/_simple.py#L353
> 
> but it's not super reliable because there might be older anndata files that are generated before we introduced that feature.

But for absolute consistency with seurat v3, we would have to check that user has inputted non-negative integers (i.e., no library size normalization). Is there a method for this? It seems like it could be a useful util down the road as already some external methods require this as well.

gokceneraslan: Hmm I see. There is no nice way to do this I think, but we can check if values are whole numbers via 

```python
np.all(np.mod(adata.X.data[:100], 1) == 0)
```

or we can also check if data is TPM normalized:

```python
sums = adata.X.sum(1)
np.all(sums == sums[0])
```

Also we can make `sc.pp.normalize_total` function add something to adata.uns like 

```python
adata.uns['normalize_total'] = {}
adata.uns['normalize_total']['totals'] = adata.X.sum(1)
adata.uns['normalize_total']['target_sum'] = 10000
```

This way we can also add an `sc.pp.unnormalize` function. It would be nice to have sc.pp.expm1() and sc.pp.unnormalize() functions in general :)

Would these help?

gokceneraslan: (There is also https://github.com/theislab/scanpy/blob/master/scanpy/external/pp/_scvi.py#L118)

adamgayoso: I can do this, just want to confirm we have consensus on these changes? This will indeed be a big change.

gokceneraslan: I think it'd be useful to have an internal "_is_raw()" function in utils.py or so to check if adata.X has counts. This can then be utilized in many other functions like existing HVG function (to check if it returns false), my sctransform PR, or DCA or scVI etc.

adamgayoso:
@gokceneraslan why are the defaults for the main function all None (e.g., dispersion cutoffs)? It seems like if scanpydoc is picking up the defaults then we can make them not None and reduce some code?

scanpy/preprocessing/_highly_variable_genes.py
adamgayoso: Indeed it's a float

scanpy/preprocessing/_highly_variable_genes.py
adamgayoso: I added an explanation here.

scanpy/preprocessing/_highly_variable_genes.py
adamgayoso: I added `check_nonnegative_integers()` util and decided for now to add a `layer` option, where a user can select a layer to use for computing HVGs. As computing HVGs with `seurat_v3` flavor would be one of the first steps after basic gene and cell filtering, it seems that this is a suitable compromise for now. An issue with what you propose for storing the size factors is that `normalize_total` currently has an option to store the size factor for you and it's off by default. So I can still do this and add `sc.pp.unnormalize` but it seems like it would involve changing the default of that function. Thoughts?

scanpy/tests/test_highly_variable_genes.py
adamgayoso: I added the test in the seurat_v3 test function.

scanpy/preprocessing/_highly_variable_genes.py
gokceneraslan: `check_nonnegative_integers()` looks great, can you add a test too if it's not too much to ask? Layer option is also great I think, because many people store counts in `.layer['counts']`.

I didn't know about the key_added option of normalize_total, interesting. Then sc.pp.unnormalize can require a mandatory "key" option which is an obs key where size factors are stored. So it can only be used if the user used the key_added option of normalize_total. 

Btw, I think sc.pp.unnormalize idea is beyond the scope of this PR so it's completely up to you. I was just thinking out loud about all options regarding raw counts :) But if you feel like implementing it then a separate PR would be a better place.

gokceneraslan:
> @gokceneraslan why are the defaults for the main function all None (e.g., dispersion cutoffs)? It seems like if scanpydoc is picking up the defaults then we can make them not None and reduce some code?

Honestly, I don't know. But @ivirshup also brought it up here https://github.com/theislab/scanpy/pull/1180#discussion_r412871280 and here https://github.com/theislab/scanpy/pull/1180#discussion_r413445806, I just didn't have time to address it. I am fine with making them not None.

gokceneraslan:
But I think scanpydoc is very confused now for some reason. Documentation build is broken, it's visible in ~all~ some recent PRs too and there is not much we can do without the help of @falexwolf or @flying-sheep or @ivirshup, because we cannot even see the error message. My local builds are just fine 🤷

adamgayoso:
There's a strange issue I'm having where the gene sets when combining from different batches are not equal between this code and Seurat v3. The problem appears to be due to the fact that many genes have the same normalized variance in a given batch and the merging method uses ranks. So I believe the difference is due to genes being sorted differently with the same normalized variance. Perhaps this merging scheme is not ideal.

Otherwise, this PR is good to go.

ivirshup:
The docs should now be fixed on master. Updating the branch should fix that.

I've added a commit to fix that formatting test.

@adamgayoso could you compress the data files you're using for testing, and remove to uncompressed files from the git history? This is to keep the repo size as small as possible.

Otherwise, @gokceneraslan were you happy with the state this PR is in?

adamgayoso:
I think I messed something up when trying to rewrite the history @ivirshup :(

adamgayoso:
There's one more little issue with nan warnings that I need to handle.

gokceneraslan:
Any ideas why doc build still fails?

ivirshup:
I believe the latest plotting PR broke it (https://github.com/theislab/scanpy/pull/1210#issuecomment-651510328)

fidelram:
@ivirshup @gokceneraslan Do you have access to the error details for readthedocs? I get  'page does not exists' error

adamgayoso:
@gokceneraslan @ivirshup just checking in on this. Will this be part of the next release? Do you need anything else from me? It might be nice to note somewhere that when `batch_key` is not None, results aren't absolutely consistent with Seurat.

> The problem appears to be due to the fact that many genes have the same normalized variance in a given batch and the merging method uses ranks. So I believe the difference is due to genes being sorted differently with the same normalized variance. Perhaps this merging scheme is not ideal.

gokceneraslan:
Looks good to me. I wish the batch_key option also worked well but we can improve it later on, I guess. Sorry for the delay.

scanpy/_utils.py
```diff
@@ -463,6 +464,21 @@ def update_params(
...
+
+    data = X if isinstance(X, np.ndarray) else X.data
+    # Check no negatives
+    if np.any(data < 0):
+        return False
+    # Check all are integers
+    elif np.any(~np.equal(np.mod(data, 1), 0)):
+        return False
+    else:
+        return True
```
ivirshup:
```suggestion
def check_nonnegative_integers(X: Union[np.ndarray, sparse.spmatrix]):
    """Checks values of X to ensure it is count data
    """
    from numbers import Integral

    data = X if isinstance(X, np.ndarray) else X.data
    # Check no negatives
    if np.signbit(data).any():
        return False
    # Check all are integers
    elif issubclass(data.dtype.type, Integral):
        return True
    elif np.any(~np.equal(np.mod(data, 1), 0)):
        return False
    else:
        return True
```

Minor update, just checking if the type is already an integer.

ivirshup:
Awesome, thanks for this! I've added something to the release notes, let me know if you'd like to say more.

PR #1216, opened 2020-05-14
Update install instructions to use leiden

ivirshup:


YubinXie:
leiden install via conda code is wrong in the current page. it should be:

```
conda install -c conda-forge leidenalg 
```

PR #1217, opened 2020-05-14
fix inverted coords and test

### LINKED ISSUE #1059: Flipped y-axis in sc.pl.spatial without img_key

### LINKED ISSUE #1148: sc.pl.spatial plots vertically flipped spots when  `img_key=None`

giovp:
This PR supersedes theislab/scanpy#1149 and closes theislab/scanpy#1148 and theislab/scanpy#1059

I have changed couple of if statements from previous version, now it plots circles and not scatterplot.


scanpy/plotting/_tools/scatterplots.py
```diff
@@ -206,7 +206,7 @@ def embedding(
             and len(size) == adata.shape[0]
         ):
             size = np.array(size, dtype=float)
-    elif img_key is not None:
+    elif basis is "spatial":
```
ivirshup: Could this behavior be determined by whether `sc.pl.spatial` was called, and not the key in `obsm`?

That is, if someone were to call `sc.pl.embedding(adata, basis="spatial")` I think that should use a scatter plot, but if someone called `sc.pl.spatial` that should plot the circles.

ivirshup (changes requested):
Thanks for the quick turn around!

My main issue here is the use of the key as determining how the plot is made. Could you instead make this predicated on `sc.pl.spatial` being called?

giovp:
Yes, I understand the point and agree. Now coordinates are inverted whenever `basis='spatial'` but circled are plotted only when `library_id is not None`, which is passed by `sc.pl.spatial` to embedding. Otherwise, scatterplot is used

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -369,7 +370,7 @@ def embedding(
                     in_groups_size = not_in_groups_size = size
 
                 # only show grey points if no image is below
-                if img_key is None:
+                if library_id is None:
```
ivirshup, line 373: What does this branch do? It seems like the next statement does the same thing?

giovp: when groups are defined, then only cells in that group are colored, the others are light gray. When `sc.pl.spatial` is plotted, we only want spots in the specific groups, and the others should simply not be plotted. 
You can have a look at cell 12 in the [tutorial](https://scanpy-tutorials.readthedocs.io/en/latest/spatial/basic-analysis.html). If this if statement is not passed, gray spots would be plotted instead of leaving it blank.

ivirshup: Ah, I see. I was missing the `~`

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -893,7 +896,7 @@ def _get_data_points(
                 tuple(number - 1 for number in comp) for comp in components_list
             ]
     else:
-        data_points = [adata.obsm[basis_key][:, offset : offset + n_dims]]
+        data_points = [np.array(adata.obsm[basis_key])[:, offset : offset + n_dims]]
```
giovp, line 902: Just wanted to point out that this is needed otherwise the flipping would change at every call of the function (meaning it would be flipped forward and reverse at every call)

giovp:
Can I merge this ?

ivirshup:
@giovp, I'll merge this. I'm merging a couple other things first though.

I'm not super happy with the logic flow here at the moment. Could we aim for separating out the code for scatter plots, and overlaying grids on-top of images in the next release cycle?

giovp:
Yes for sure

PR #1218, opened 2020-05-14
Update release notes

ivirshup:
Added a bunch of stuff to the release notes.

PR #1219, opened 2020-05-14
update with spatial functions

giovp:


ivirshup:
Thanks!

So that I could edit this all together, I've added this commit to #1218.

PR #1224, opened 2020-05-16
adds totalVI to external for CITE-seq analysis

adamgayoso:
This mostly uses the same API as the scvi external function. I'm not quite sure how you all will handle CITE-seq data within anndata, so I imagine we might have to change the data input a bit. 

I also added a lightly filtered CITE-seq dataset that can be used in your tutorial. This is basically doublet removal (using DoubletDetection) followed by the same filters used in the basic tutorial with an additional protein library size filter. Though I understand also if you'd like the scanpy dataset to be the unfiltered one.


adamgayoso:
@falexwolf @ivirshup please let me know what your plans are for the "scanpy" way of handling CITE-seq data are. Also, if somehow you could compare to totalVI in your tutorial that would be great! It takes about 20-25 minutes to run on CPU for the pbmc5k and HVGs.

ivirshup:
Thanks for this PR, this looks interesting! Sorry for taking a while to get back to you, we've been quite busy getting a release out. We'll try and get back to you with more in the next couple weeks.

PR #1226, opened 2020-05-17
add log1p param to calculate_qc_metrics

falexwolf:


scanpy/tests/test_qc_metrics.py
```diff
@@ -106,6 +106,14 @@ def test_qc_metrics():
...
     for col in adata.var:
         assert np.allclose(adata.var[col], old_var[col])
+    # with log1p=False
+    adata = AnnData(X=sparse.csr_matrix(np.random.binomial(100, 0.005, (1000, 1000))))
+    adata.var["mito"] = np.concatenate(
+        (np.ones(100, dtype=bool), np.zeros(900, dtype=bool))
+    )
+    adata.var["negative"] = False
+    sc.pp.calculate_qc_metrics(adata, qc_vars=["mito", "negative"], log1p=False, inplace=True)
+    assert "log1p_total_counts" not in adata.obs.keys()
```
ivirshup:
```suggestion
    assert not np.any(adata.obs.columns.str.startswith("log1p_"))
    assert not np.any(adata.var.columns.str.startswith("log1p_"))
```

How about this?

gokceneraslan:
Do we need to consider the `base` parameter of log1p here?

falexwolf:
In my opinion, we'll likely move away from logging everything. Isaac built this in so that one can conveniently visualize things in seaborn; I added the switch to turn it off so that the basic tutorial of v1.5 doesn't lead to a completely cluttered AnnData object.

But, I guess, we all agree that this here isn't the final solution. 😄

PR #1236, opened 2020-05-20
Fix docstring for pca, fixes #1234

### LINKED ISSUE #1234: misleading docs: docs & docstring for sc.pp.pca incorrectly states pca_sparse as a keyword arg
Both the docs here: https://scanpy.readthedocs.io/en/stable/api/scanpy.pp.pca.html and the docstring on the function scanpy.pp.pca incorrectly states pca_sparse as a keyword arg.  It looks like maybe at one point in the PR that produced this that was the case (been tracking the sparse PCA PRs), but it now looks like it is just the default now.  

Should be removed from the docstring.

ivirshup:


ivirshup:
Superseded by #1240

PR #1240, opened 2020-05-21
Fix pca on sparse data reproducibility

ivirshup:
Bugfix for the sparse pca.

It looks like we forgot to pass a random seed when this is used... But we also never had a test that checks if you run the function twice with the same random seed it returns the same result.

This PR fixes both these issues. The new tests are a bit slow, but are definitely needed.

I've also added a fixture for returning a copy of the pbmc3k dataset which has been normalized and had `highly_variable_genes` run on it. Preparation of the object should only happen once per run of the suite, but a new copy will be provided for each test that uses it. This was done to speed up the new tests.


ivirshup:
To limit the amount of time we have to wait for rebuilding, I've added the change from #1236 here

PR #1241, opened 2020-05-21
Ignore RandomState references so rtd will build docs

ivirshup:
Per title


ivirshup:
Ping @flying-sheep

flying-sheep:
can you link a doc build failure, please? Then I can check out why it fails.

I assume they changed the documented location of the class. We have it in `qualname_overrides` and should probably just update the location there (or remove it that line if the new documented location now matches the qualified name)

https://github.com/theislab/scanpy/blob/e5d246aacc71fb9ed71d49e2e7d5e26743fd4acb/docs/conf.py#L136-L140

PR #1245, opened 2020-05-22
Update palantir external

awnimo:
Critical updates to Palantir external tool wrapper. Includes updated DocString.


Koncopd:
@awnimo , everything seems fine for me. However, shouldn't `palantir_results` function also be in the docs [here](https://github.com/awnimo/scanpy/blob/update_palantir_external/scanpy/external/__init__.py)?

PR #1247, opened 2020-05-25
Bump scipy requirement to >1.4

ivirshup:
Fix for https://github.com/theislab/scanpy/issues/1246. It looks like the `rmatmat` argument for `LinearOperator` was only added in scipy 1.4: https://github.com/scipy/scipy/pull/10849

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -11,7 +11,7 @@
 /* Custom classes */
 
 .small { font-size:40% }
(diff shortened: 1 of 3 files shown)


gokceneraslan:
I think it's very likely people will hit this bug, because 1) they typically don't update packages like scipy very often 2) most pipelines use sparse datasets + PCA. I think it deserves a new release.

PR #1270, opened 2020-06-08
Get rid of top and bottom spaces in dot plots

falexwolf:
@fidelram, I'm not sure whether I'm breaking things you intended differently. But for us, this removes the strange scaling of the white space above and below the dots, and makes it behave like the x-axis.

Just a suggestion, no need to follow up.

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -2011,15 +2011,12 @@ def dotplot(
     dot_ax.set_xticklabels([mean_obs.columns[idx] for idx in x_ticks], rotation=90)
     dot_ax.tick_params(axis='both', labelsize='small')
     dot_ax.grid(False)
-    dot_ax.set_xlim(-0.5, len(var_names) + 0.5)
     dot_ax.set_ylabel(groupby)
 
     # to be consistent with the heatmap plot, is better to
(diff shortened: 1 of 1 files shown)


fidelram:
@falexwolf I added a padding argument to dotplot (already merged into master):

```PYTHON
markers = {'T-cell': ['CD3D', 'CD3E', 'IL32'], 'B-cell': ['CD79A', 'CD79B', 'MS4A1'], 'myeloid': ['CST3', 'LYZ']}
fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(12,5), gridspec_kw={'width_ratios': [1.2, 1], 'wspace':0.6}, sharex=False, sharey=False)
ax1 = sc.pl.dotplot(adata, markers, groupby='bulk_labels', ax=ax1, dendrogram=True, show=False)
ax2 = sc.pl.dotplot(adata, markers, groupby='bulk_labels', ax=ax2, dendrogram=True, show=False, return_fig=True).\
    style(x_padding=0.5, y_padding=0.5, cmap='Reds').show()
```
image

The units for the `x_padding` and `y_padding` are in ticks distance.

ivirshup:
I think we're good to close this PR then, right? Let me know if I should reopen it.

PR #1271, opened 2020-06-09
Add sctransform as sce.pp.sctransform and revive external.rtools

gokceneraslan:
[sctransform](https://genomebiology.biomedcentral.com/articles/10.1186/s13059-019-1874-1) uses Pearson residuals from “regularized negative binomial regression,” to correct for the sequencing depth. After regressing out total number of UMIs (and other variables if given) it ranks the genes based on their residual variances and therefore also acts as a HVG selection method.

This function replaces `sc.pp.normalize_total` and `sc.pp.highly_variable_genes` and requires raw counts in ``adata.X``.


LuckyMD:
This is awesome! And much more elaborate than my version of the same:
https://github.com/normjam/benchmark/blob/master/normbench/methods/ad2seurat.py

gokceneraslan:
Do you guys think this PR makes sense or is it too much to add R packages to the travis setup? @ivirshup @flying-sheep 

There are bunch of useful R packages out there that will most likely not be reimplemented in Python (limma-voom pseudobulk DE, Liger, MAST etc.). I think it'd be cool to revive rtools and add access to such packages. I'm not sure if this is the right way but, any guidance is appreciated :)

ivirshup:
I don't think we should have to worry about dealing with CI for R in this project. This only becomes harder if the intent is to have many other dependencies.  I think doing anything like this makes much more sense in seperate dedicated package.

PNorvaisas:
Are there any updates concerning this PR? scTransform functionality in scanpy would be much appreciated :)

PR #1305, opened 2020-07-05
Fix `_check_datafile_present_and_download` FileExistsError

michalk8:
Hi,
currently, our notebooks tests fail (see https://travis-ci.com/github/theislab/cellrank_notebooks/jobs/357647707) because of missing `exists_ok` in `mkdir`.
`_check_datafile_present_and_download` eventually calls `_download` (where `exists_ok=True`), so I've deferred creation + logging of the directory to `_download`.


ivirshup:
I like that CI for notebooks!

Question, what should be happening here? I'm a little confused about how the path isn't a directory, but you're going to write a file inside of it? If this is the problem, I think having `exists_ok=True` will still error.

In general, I believe we'd previously decided to not create parent directories for writing automatically (mentioned in https://github.com/theislab/anndata/pull/364, and I think talked about in a zoom call?). This follows other tools better, which just throw an error if your parent directory doesn't exist. I personally like this approach better because it'll throw an error for typos, instead of doing the wrong thing.

The only exception here is for the `datasets` module, which should create the directory it will store data in. I would recommend you explicitly create the directory in your code if needed.

@falexwolf @flying-sheep, I would like to remove the directory creation code from `_check_datafile_present_and_download`. Do you think this will be a problem? It's not documented, so I'm leaning towards just removal instead of deprecation, but could be convinced to just warn for a cycle.

ivirshup:
I'm pretty sure this has been superseded. Let me know if it it hasn't.

PR #1306, opened 2020-07-07
Add support for Harmony integration

esrice:
I've had better luck integrating data from multiple experiments using [Harmony](https://portals.broadinstitute.org/harmony/) than the current integration methods in scanpy.external.pp. This PR adds a wrapper for Harmony to the external API.


gokceneraslan:
Hi @esrice,

This looks great, actually I was considering adding support for the PyTorch implementation of Harmony https://github.com/lilab-bcb/harmony-pytorch. I only used this version so I don't know if it brings any speedup but results were very nice usually.

esrice:
Thanks @gokceneraslan. This implementation takes very little time (a couple seconds per iteration on ~10k cells), so a speedup might not make much difference, but if you think using the pytorch implementation would be better, I can certainly switch to that one easily enough.

flying-sheep:
Looks good then! Great that you followed https://github.com/theislab/scanpy/pull/503#issuecomment-471331400 and named it `harmony_integrate` instead of generally `harmony`. Not sure if `obsm_{in,out}_field` are good names. Maybe use something more speaking? I think we use `basis` or `rep` for something like `X_pca` elsewhere.

Please also fix doc build errors and the `.travis.yml` conflict. (“Kor**s**unsky19” is a typo I guess, and I think it should be `kwargs` instead of `**kwargs`).

esrice:
Great, thanks for the feedback. Hopefully this merge and commit fix everything. I wasn't able to see what errors were causing the readthedocs build fail as the "Details" link just took me to a page that said "SORRY / This page does not exist yet.", so let me know if there are any other issues.

flying-sheep:
Ah, yeah we need to fix the visibility of the builds.

Hmm, so you can’t see the build [here](https://icb-scanpy--1306.com.readthedocs.build/en/1306/external/scanpy.external.pp.harmony_integrate.html)?

It still has issues:

grafik

you need to add blank lines before the `>>>` i guess.

Also grafik again. You need to either give us acess to WarrenLab/scanpy@harmony or continuously merge master until you pressing the “merge upstream changes” button and us hitting the “squash & merge” button happens fast enough. forcing up-to-date branches this way is a bit annoying, but it’s the only way to be sure conflicts in changes don’t break everything. (well, the only way without switching to [bors](https://bors.tech/))

esrice:
Nope, that link brings me to a login page, and when I log in with my github account it gives me an error. I merged master again and added newlines; hopefully this fixes the issue. I'll give you access to my fork as well.

flying-sheep:
Looks perfect now, merging.

Check it out: https://scanpy.readthedocs.io/en/latest/external/scanpy.external.pp.harmony_integrate.html

PR #1320, opened 2020-07-17
Make sure dependencies are up to date in travis builds

ivirshup:
Travis builds are currently breaking due to incompatible versions of numpy being present, so no compatible version gets installed. This should fix that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,9 @@ matrix:
     - name: "anndata dev"
       python: "3.7"
       install:
-        - pip install -U docutils sphinx pip
-        - pip install -U -e .[test,louvain,leiden,magic,scvi]
-        - pip install -U git+https://github.com/theislab/anndata
+        - pip install docutils sphinx
+        - pip install -e .[test,louvain,leiden,magic,scvi]
+        - pip install git+https://github.com/theislab/anndata
 python:
   - '3.6'
   - '3.7'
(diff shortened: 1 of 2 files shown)


ivirshup:
Not completely sure if this is doing what I intended. I added the `-U` so dependencies would be upgraded, but numpy still isn't being upgraded as shown by these warnings:

```
umap-learn 0.4.6 has requirement numpy>=1.17, but you'll have numpy 1.15.4 which is incompatible.
scvi 0.6.6 has requirement numpy>=1.16.2, but you'll have numpy 1.15.4 which is incompatible.
```

Not sure why this is happening. I'd prefer if we didn't have to manually specify the dependencies of our dependencies. Any ideas @flying-sheep?


------------------

Updating pip doesn't seem to do anything (maybe it has to do with "editable mode"?)

--------------------

An easy fix is just to add a version requirement on `numpy`, but I really feel like dependency resolution should be dealing with that.

ivirshup:
Updating `pip` doesn't seem to do anything

ivirshup:
Okay, this is weird. My current theory is that it's a bug in pip. It looks like if multiple packages depend on a single dependency, pip doesn't necessarily check if all requirements are satisfied. It just checks if one packages version requirements are. It seems non-deterministic which package is used to check.

I was working on an example, but then I just found this: https://github.com/pypa/pip/issues/8218

There's an unstable feature for this (`--unstable-feature=resolver`), which does work, but I think adding a requirement for numpy is a bit more justifiable.

flying-sheep:
I’m really happy that they’re finally adding a resolver.

Without officially having a resolver, it technically wasn’t a bug that incompatible stuff got installed, but instead just a missing (if very vital) feature.

PR #1328, opened 2020-07-22
Separated .show() into .show and .savefig to match matplotlib style

fidelram:


fidelram:
some older commits were included. I will delete this PR and prepare a clean one.

PR #1329, opened 2020-07-22
split BasePlot.show() into show and savefig to match matplotlib style

fidelram:


scanpy/plotting/_baseplot_class.py
```diff
@@ -558,42 +559,28 @@ def _mainplot(self, ax):
 
         return normalize
 
-    def show(
-        self, show: Optional[bool] = None, save: Union[str, bool, None] = None,
-    ):
+    def render(self):
```
ivirshup: I’m not sure `render` is the right name, since rendering doesn’t actually happen until you show or save the figure. Maybe `make_figure`?

ivirshup (commented):
Looks mostly good! Just not sure about the name of the `render` method. Also, are you sure you want that method to be public?

scanpy/plotting/_baseplot_class.py
fidelram: no problem to change this. As for the method being public.. if you pass the `ax` parameter you want to make the image but not call `plt.show`.  Thus the method should be public.

ivirshup: Alright, sounds good

fidelram:
@ivirshup Default for stacked violin and dotplot now looks  like this (changes: colormap=Blues for violins and border for both violins and dotplot, title of each plot shows which average measure is used):

image

Alternatively, the default for stacked_violins could be (not my preference as this takes more effort to interpret but similar to the stable branch output):

image

ivirshup:
That looks great! Thanks!

I also prefer the shaded violin plot, since I don't think the categorical colors are adding a ton of information. If I wanted category colors I think I'd prefer a "barcode" like what `sns.clustermap` does, but that isn't necessary right now.

One very minor quibble. For the dot plot, would it be easy to make the outline of the dots to be the same color/ width in the legend and figure?

PR #1334, opened 2020-07-28
adding User-Agent changer into _download.py

Mirkazemi:
Spatial tutorial gets 'HTTPError: HTTP Error 403: Forbidden' error by running
`adata = sc.datasets.visium_sge(sample_id="V1_Human_Lymph_Node")`. The error raised by calling `urllib.request.urlretrieve` method for downloading data from 10xgenomics. Sergei suggested to change the User-Agent. 

The `_download.py` tries to get the data via `urllib.request.urlretrieve` as before. If it is not possible the User-agent is changed to 'scanpy-user'.

### DIFF
--- a/scanpy/readwrite.py
+++ b/scanpy/readwrite.py
@@ -931,7 +931,16 @@ def _download(url: str, path: Path):
             t.update(b * bsize - t.n)
 
         try:
-            urlretrieve(url, str(path), reporthook=update_to)
+            try:
+                urlretrieve(url, str(path), reporthook=update_to)
+            except:
+                from urllib.request import build_opener, install_opener
+                urlopener = build_opener()
+                _user_agent = 'scanpy-user'
+                urlopener.addheaders = [('User-agent', _user_agent)]
+                install_opener(urlopener)
+                urlretrieve(url, str(path), reporthook=update_to)
+                logg.info(f'Changing user-agent to {_user_agent}')
         except Exception:
             # Make sure file doesn’t exist half-downloaded
             if path.is_file():


Koncopd:
Hi, does it solve the problem?

ivirshup (commented):
@Koncopd, why does adding a user-agent fix this?

@Mirkazemi why try without the user agent instead of just doing that in the first place?

ivirshup:
@Koncopd, this seems to work for me. But also it looks like `install_opener` modifies global state and we shouldn't do that. In theory it shouldn't be too difficult to both start a download with a header, and write the result to a file, but it's not obvious `urllib` has anything to help do that.

Koncopd:
@ivirshup The idea is that if downloading with a browser or wget works but python requests fail, then the problem should be somewhere in the header. Changing user agent (to avoid blacklist \ whitelist on the server) is the most obvious thing to try.

flying-sheep:
As @ivirshup said in #1344, global changes at runtime aren’t a good idea. I think we should proceed there.

linanzhang:
How to avoid the 'HTTP Error 403: Forbidden' exactly? I reinstalled scanpy and still have this issue.

engrodawood:
I installed it now but still have the issue of HTTP Error 403.

ivirshup:
@engrodawood and @linanzhang, are you both sure you're using scanpy `1.6.0`? The requests seem to be working for me. If the problem persists, could you open a new issue?

engrodawood:
Hi @ivirshup ,
It just fixed when I installed the library directly from pip. Since I was following the documentation for library installation, the command mentioned in the documentation is downloading an outdated version. 

Thanks.

PR #1344, opened 2020-07-30
Download header

ivirshup:
As noted in #1334, visium downloads were broken. Setting a header on downloads seems to fix them.

This supersedes  #1334 since that solution modifies global state around `urllib`, which is asking for trouble. This unfortunately means most of the method had to be reimplemented.
The new implementation is based on `urllib.requests.urlretrieve`, but with a modification to let us pass a header.

I also included a couple minor fixes to existing dataset download stuff:

* We don't get a warning from using `@internet` test marker anymore
* One of the downloaded datasets changed, so the test got updated
* `_download` no longer creates all parent directories. That is handled upstream.

@Mirkazemi @Koncopd

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,3 +2,5 @@
 python_files = test_*.py
 testpaths = scanpy/tests/
 xfail_strict = true
+markers =
+    internet: tests which rely on internet resources (enable with `--internet-tests`)
--- a/scanpy/datasets/_datasets.py
+++ b/scanpy/datasets/_datasets.py
@@ -10,6 +10,7 @@ from .. import logging as logg, _utils
 from .._compat import Literal
 from .._settings import settings
 from ..readwrite import read, read_visium
+from ._utils import check_datasetdir_exists
 
 HERE = Path(__file__).parent
 
@@ -52,6 +53,7 @@ def blobs(
     return AnnData(X, obs=dict(blobs=y.astype(str)))
 
 
+@check_datasetdir_exists
 def burczynski06() -> AnnData:
     """\
     Bulk data with conditions ulcerative colitis (UC) and Crohn's disease (CD).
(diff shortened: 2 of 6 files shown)


ivirshup:
Also ping @flying-sheep since I believe you wrote the initial version.

flying-sheep:
Shouldn’t we just depend on `requests` if it’s so complicated and we have to resort to code copying?

Basically every Python user should have it installed anyway.

ivirshup:
I looked into that, but I'm not sure it actually makes this any less complicated. The issue is getting the `tqdm` thing to update, and requests would need all of the same logic to do that as far as I can tell. Plus, at that point it's copying from stack overflow vs. copying from python's stdlib.

You'd think this would be a convenience function somewhere. Or you'd think that `urlretrieve` could take a `Request` object.

flying-sheep:
Requests makes everything rather convenient, that’s its main appeal. Example for our case: https://stackoverflow.com/a/62113293/247482

giovp:
When we first discussed this with @Mirkazemi , we were also thinking about `requests`, but we weren't sure how would you feel guys of having to include this as a requirement. Can't contribute much but would also go on with Isaac version which wouldn't require that.

flying-sheep:
It’s more code we have to maintain. If I had to decide between adding a common dependency, feature regression, or complex code, I’d go with the first one.

ivirshup:
To me, the code looks pretty equivalent. Since I've run into issues with unexpected defaults in some of the requests developers libraries, I'm more comfortable sticking with code that's close to what we've had before. 

I will shorten the code a bit though

flying-sheep:
fair enough! Aw, too bad we can’t use Python 3.8 yet, this would be perfect:

```py
while block := resp.read(blocksize):
    ...
```

PR #1348, opened 2020-07-31
Export concat

ivirshup:
Exports `anndata.concat` so people don't have to add an extra import for a single function.

This follows what we do with several IO functions and the `AnnData` class.

Should `concat` get a page in the scanpy docs, or is it fine that it's in anndata?


flying-sheep:
Exporting the AnnData class was more of a “initially AnnData was in scanpy” thing …

ivirshup:
Alex and I thought that it's pretty unlikely most users import anndata. I prefer how ergonomic it is to have fewer namespaces to deal with. Plus we do re-export almost everything else from anndata here already.

falexwolf:
Yes, we could debate that initial choice, I know. But now that we are here already, we should reexport `ad.concat`. Having just the `sc` namespace when you use scanpy is much more convenient than also having to import anndata on it's own.

Doc only in anndata is fine, I think. We just should clarify the doc in Scanpy around this.

falexwolf:
That is, we should have a section in the API overview that is headed "anndata functionality" or something similar: 
image

falexwolf (approved):

PR #1356, opened 2020-08-04
Make missing values lightgray in scatterplots

### LINKED ISSUE #1355: UMAP plotting can not properly colour np.nan values in categorical obs

Scanpy version: 1.5.1

If  np.nan values are present in obs column used for colouring `sc.pl.umap` they are coloured with the colour of the last non-nan category (see example below). This is very problematic as cells that are not annotated get the color of one of the categories.
This is another reason why https://github.com/theislab/scanpy/issues/1340 is needed - there should be a default for nan (such as gray).

I plotted an UMAP and coloured it by cell type (left). Then I replaced half of alpha cells' cell types (column pre_cell_type) with np.nan and plotted it again (right). Half of the alpha cells (previously coloured blue) are now coloured pink (colour of the last category, here named 'NA'). These cells should not be coloured pink - nan should have its own colour.

slika

ivirshup:
Potentially fixes #1355

* Would still need tests/ further consideration.
* Need to fix missing values not being plotted below present ones.

Using this branch:

```python
import scanpy as sc
import numpy as np
import matplotlib as mpl

pbmc.obs["louvain"].iloc[::2] = np.nan
with mpl.rc_context({"figure.dpi": 150}):
    sc.pl.umap(pbmc, color="louvain")
```

image

```python
with mpl.rc_context({"figure.dpi": 150}):
    sc.pl.umap(pbmc, color="louvain", groups=list(pbmc.obs["louvain"].cat.categories[:3]))
```

image

## Update:

This PR expanded in scope quite a bit, so I'd like to wrap it up. Most things are implemented and seem to work. Regression tests need to be added, for these cases.

- [x] Tests for all fixed cases (probably write these down as well)
- [x] Decide on adding arguments, and default values
- [x] Decide on whether continuous legend update happens in this PR


ivirshup:
I've managed to fix this up a bit. Missing (or masked - for `groups`)  values in categorical arrays are now always plotted on bottom and use a default color. For spatial plots this default color is transparent. This has led to some code simplification. Surprisingly, this didn't break any tests locally, so a bunch of new tests are probably needed.

Continuous values are still a little weird. Right now the points don't show up on embedding plots, and mess up all the colors for spatial plots.

ivirshup:
I think I've got this mostly working. Here are some comparisons between plots on master and this branch. This will show the major cases I've caught and was thinking about. In each of the cases where there has been a significant change, I'd say the previous behavior seemed like a bug.

A few open questions:

* Are these reasonable defaults for null values?
* Are there cases I've missed?
* How should users be able to interact with this/ choose the color for null values? (currently, this is a private parameter which would override the "bad value" for a passed colormap)

@Hrovatin @LuckyMD @fidelram @giovp @flying-sheep 

# Setup

```python
import scanpy as sc, matplotlib.pyplot as plt, numpy as np, pandas as pd
```


```python
# Setup
brain = sc.read("./tmp_brain.h5ad")  # just an example visium dataset with some preprocessing
sc.tl.umap(brain)

# Add some points with misssing values
brain.obs["leiden_missing"] = brain.obs["leiden"].copy()
brain.obs["leiden_missing"].iloc[::2] = np.nan

brain.obs["Bc1_missing"] = brain.obs_vector("Bc1").copy()
brain.obs["Bc1_missing"].iloc[::3] = np.nan
```



## Null values in categoricals



<details>
<summary> Spatial </summary>




```python
with plt.rc_context({"figure.dpi": 150}):
    sc.pl.spatial(brain, color=["leiden", "leiden_missing"])
```

### Current (this branch):

output_5_0


### Master

output_5_0


```python
with plt.rc_context({"figure.dpi": 150}):
    sc.pl.spatial(brain, color=["leiden", "leiden_missing"], groups=["0", "1"])
```

### Current

output_6_0


### Master

output_6_0


</details>

<details>
<summary> Embedding </summary>


```python
with plt.rc_context({"figure.dpi": 150}):
    sc.pl.umap(brain, color=["leiden", "leiden_missing"])
```


### Current

output_8_0

### Master

output_8_0


```python
with plt.rc_context({"figure.dpi": 150}):
    sc.pl.umap(brain, color=["leiden", "leiden_missing"], groups=["0", "1"])
```

### Current

output_9_0

### Master

output_9_0

</details>

## Continuous values


<details>
<summary> Spatial </summary>


```python
with plt.rc_context({"figure.dpi": 150}):
    sc.pl.spatial(brain, color=["Bc1", "Bc1_missing"])
```

### Current

output_12_0

### Master

output_12_0


```python
with plt.rc_context({"figure.dpi": 150}):
    sc.pl.spatial(brain, color=["Bc1", "Bc1_missing"], vmin="p10", vmax="p90")
```

### Current

output_13_0


### Master

output_13_0


</details>

<details>
<summary> Embedding </summary>


```python
with plt.rc_context({"figure.dpi": 150}):
    sc.pl.umap(brain, color=["Bc1", "Bc1_missing"])
```

### Current

output_15_0


### Master

output_15_0

```python
with plt.rc_context({"figure.dpi": 150}):
    sc.pl.umap(brain, color=["Bc1", "Bc1_missing"], vmin="p10", vmax="p90")
```

### Current

output_16_0

### Master

output_16_0


</details>

Hrovatin:
I like the new behaviour. Maybe a parameter like na_colors could be used to specify a different na color if needed. 
Also, should the NaN and its colour be added to the legend (if categorical) or besides it (if continuous)?

flying-sheep:
Looks pretty great! Improvements everywhere! I’m especially surprised that in this pic, the missing values share a color with an existing category! That’s pretty wrong:



At first I was confused because “current” and “master” often mean “before” and “after” in the opposite order than here, and also you put the “after” plots above the “before” ones.

fidelram:
Thanks for the changes and for the clean up of the code. From the output that you show the changes seem to be working well. Maybe the only relevant issue would be to change the default color if this clashes with a color already assigned to a category. 

Now that you are looking into this, any chance that the missing_color can also be used as the default color when the `color` parameter is empty? The current default is light gray .

ivirshup:
> Maybe the only relevant issue would be to change the default color if this clashes with a color already assigned to a category.

I'm not sure how to check if colors are similar. But I think @Hrovatin's suggestion could make this obvious enough to users:

> Also, should the NaN and its colour be added to the legend (if categorical) or besides it (if continuous)?

Any suggestions for how to handle this @fidelram?

> Now that you are looking into this, any chance that the missing_color can also be used as the default color when the color parameter is empty? The current default is light gray .

Yes, but it took a little bit more work than I expected (https://github.com/matplotlib/matplotlib/issues/18294). Basically we can't just pass an array of nulls for the color values here, since matplotlib throws user visible warnings about this at plot time.

ivirshup:
Note: having null values breaks `legend_loc="on data"`, since the median gets calculated wrong.

ivirshup:
@Hrovatin and @fidelram: What should `"on data"` look like if the there is a null category? Do the null points get a label? Is this the same for `groups`?

fidelram:
For me the most informative solution when I do get NaN values (usually after anndata concatenate),  is to visualize them as a new category labelled 'NaN'. (What I do is convert the categorical data to string and then back to categorical (e.g. `adata.obs['label'] = adata.obs['label'].astype(str).astype('category')`). This solution has the advantage of creating a 'nan' category that solves many of the mentioned problems like the 'on data' labels and the potential color clash of nan color with category color. What it does not do and that I like from the new code is to put those NaN values in the back. 

What we can do is to have a parameter like `na_as_category` that will do what I just mentioned. Otherwise, the NaNs are plotted in gray but do not get a label.

giovp:
Hi sorry, for late reply on this, agree with Fidel and this is what I also do usually. Having the `na_as_category` param would definitely be useful.

ivirshup:
When would you want `na_as_category` off? Is it important enough to add a new parameter for?

Also, how should the continuous color bar show that there is a null value? (Suggestions with code very welcome)

Hrovatin:
When data is continuous and legend is a continuous color bar a second legend could be added below it showing the categorical NaN (like the usual categorical legend).

ivirshup:
I guess I'm wondering when I would want results like this with `"on data"`:

<details>
<summary> </summary>

image
image

</details>

And when  I wouldn't want results like this with `"right margin"`:


<details>
<summary> </summary>

image
image

</details>

ivirshup:
@Hrovatin, that sounds good, but I'm not totally sure how I should do that. If you had a prototype of that it would be very useful. Maybe @fidelram has a suggestion based on the DotPlot legends?

fidelram:
For continuous values I don't think we need to add anything to the color bar. If a dot color is not part of the colorbar then is assumed that is a NaN. I searched in matplotlib for a similar case in which a colorbar includes NaN values but could not find any example.

If this feature is wanted, what we can do is to use the option for colorbar extension and use it for NaNs but we need to find a way to set the label for NaN.

```PYTHON
import numpy as np

adata.obs['n_genes'].iloc[::4] = np.nan
cmap = plt.get_cmap('viridis')
cmap.set_under('lightgray')
cmap.set_bad('lightgray')

fig, ax = plt.subplots()
cax = ax.scatter(adata.obsm['X_umap'][:,0], adata.obsm['X_umap'][:,1], 
                 c=adata.obs['n_genes'], s=20, 
                 cmap=cmap, 
                 vmin=1000, 
                 vmax=2000, plotnonfinite=True)
fig.colorbar(cax, extend='min', extendrect=True, extendfrac=0.1)

plt.show()
```
image

ivirshup:
I think this is now ready for review. 

## Legends

I've decided to leave showing the null value in continuous legends for another PR, since I don't have an obvious solution now.

I have added an argument for specifying whether the na value should show up in the legend, `na_in_legend`. It defaults to `True`. Here's an example:

```python
sc.pl.umap(brain, color="leiden", groups=["0", "1"])
sc.pl.umap(brain, color="leiden", groups=["0", "1"], na_in_legend=False)
```

<details>
<summary> Images </summary>

image
image

</details>

The current default of true is a bit weird for "on data":

```python
sc.pl.umap(brain, color="leiden", groups=["0", "1"], legend_loc="on data")
sc.pl.umap(brain, color="leiden", groups=["0", "1"], legend_loc="on data", na_in_legend=False)
```

<details>
<summary> Images </summary>

image
image

</details>

## Missing color

The missing color can now be specified with `na_color`. This defaults to transparent for spatial plots, and light gray for all other embedding based plots.

```python
with plt.rc_context({"figure.dpi": 150}):
    sc.pl.spatial(brain, color=["leiden_missing", "Bc1_missing"])
    sc.pl.spatial(brain, color=["leiden_missing", "Bc1_missing"], na_color=(.8, .8, .8, .2))
```

<details>
<summary> Images </summary>

image
image

</details>

## Tests

I've added a parameterized regression test around a perhaps-too-cute test case.

<details>
<summary> Test case </summary>

```python
sc.pl.spatial(adata, color="label")
```

image

</details>

This test makes a lot of files, so I'll rebase the revisions away before merge.

## Possible problems

* I'm hoping I haven't missed any edge cases, but would appreciate some testing from @giovp and @fidelram.
* What do you think about the interaction between `groups` and `legend_loc="on data"`? I'd like to keep `na_in_legend` as a simple boolean, but this does change the current behavior.

fidelram:
@ivirshup Looks great! I like the new spatial test image ;) well done!

I just give it a try and didn't find any problem.

One little change: can you add to the legend of `na_color`  that this is also the color used when the parameter for `color` is not given. 

I noticed two parameters in the embedding that I think belong only to the spatial.. Those are `bw` and `alpha_img`. In embeddings they do nothing. 

Other issue, that I don't expect to address at the moment, is the increase in parameters because is becoming difficult to go through the list of parameters when browsing through the documentation. To help on this we can start using alphabetical order for all optional parameters. Other suggestion is to add to the documentation in which version a parameter was added. Thus, power users can easily track changes and try new options.

scanpy/plotting/_docs.py
```diff
@@ -89,6 +89,12 @@
     If `None`, `mpl.rcParams["axes.prop_cycle"]` is used unless the categorical
     variable already has colors stored in `adata.uns["{var}_colors"]`.
     If provided, values of `adata.uns["{var}_colors"]` will be set.
+na_color
+    Color to use for null or masked values. Can be anything matplotlib accepts as a
+    color.
```
fidelram: Can you add that this is the default color used when the `color` parameter is not given?

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -67,7 +70,10 @@ def embedding(
     library_id: str = None,
     #
     color_map: Union[Colormap, str, None] = None,
+    cmap: Union[Colormap, str, None] = None,
     palette: Union[str, Sequence[str], Cycler, None] = None,
+    na_color: Any = "lightgray",
```
fidelram: Where you may use `ColorLike` instead of Any: See: https://github.com/theislab/scanpy/blob/090d577982a6afa9a84d143127e7fa53ac282635/scanpy/plotting/_anndata.py#L75

fidelram (approved):

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -741,6 +734,7 @@ def spatial(
     alpha_img: float = 1.0,
     bw: bool = False,
     size: float = None,
+    na_color=(0, 0, 0, 0),
```
ivirshup: @flying-sheep, the signature for `sc.pl.spatial` is being generated incorrectly. The documented default value is the value from `embedding`, not what's defined here.

```python
?sc.pl.spatial
...
    na_color: Any = 'lightgray',
```

It looks like this has to do with how `inspect` defines signatures. By default, if a function is wrapped, it skips the wrapper and just gets the signature of the wrapped object. If you turn that off, it only looks at the signature of the wrapper. What we want is the original signature updated with the wrapping signature.

I think this can be fixed in getdoc pretty easily, but this does not fix the signature you see in ipython. Ideally this could be fixed at the wrapping level, but I haven't figured out how yet. It looks like `inspect` kinda handles this for `partial`s in `_signature_get_partial`, but that can't be used as is here since it relies on the attributes of partial functions.

Any suggestions on how to fix this short of defining a modified `_signature_get_partial` and setting the `__signature__` attribute on these?

scanpy/plotting/_docs.py
ivirshup: Updated, thanks!

scanpy/plotting/_tools/scatterplots.py
ivirshup: Updated, thanks!

ivirshup:
@fidelram Thanks!

> I noticed two parameters in the embedding that I think belong only to the spatial.. Those are bw and alpha_img. In embeddings they do nothing.

Yeah, those probably shouldn't get documented for functions like `sc.pl.umap`.

> Other issue, that I don't expect to address at the moment, is the increase in parameters is becoming difficult to go through

I agree (related: #956). There are so many that I'm not sure alphabetical always makes sense? Perhaps they could be grouped into sections of related parameters? This would require some work on how the docs are generated.

It would definitely be good to note when features were added. Related to this, I want to discuss versioning at the next meeting to figure out when this should go in a release.

scanpy/plotting/_tools/scatterplots.py
giovp: I understand the point but I don't really know how to help sorry.

giovp:
> @fidelram Thanks!
> 
> > I noticed two parameters in the embedding that I think belong only to the spatial.. Those are bw and alpha_img. In embeddings they do nothing.
> 
> Yeah, those probably shouldn't get documented for functions like `sc.pl.umap`.
> 

well, they are passed to functions that are called in `sc.pl.embedding`, for instance [here](https://github.com/theislab/scanpy/blob/c9fbe5c1e00f5a262262cb7dee2c28d9d536ee34/scanpy/plotting/_tools/scatterplots.py#L296) . For sure it doesn't look great though. Honestly when I first approached the spatial plot, I just hhijacked what was there from the scatterplot and added modifications needed. It worked, but indeed it contributed to make the code really long and difficult to read.  

For the rest, sorry for late reply, it looks great for both the umaps and the spatials, especially the fact that they can be left transparent.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -225,28 +244,36 @@ def embedding(
...
-        color_vector, categorical = _get_color_values(
+        color_source_vector = _get_color_source_vector(
             adata,
             value_to_plot,
             layer=layer,
-            groups=groups,
-            palette=palette,
             use_raw=use_raw,
             gene_symbols=gene_symbols,
+            groups=groups,
```
giovp, line 253: the fact that the grouping is handled here it's great!

giovp (approved):

scanpy/plotting/_tools/scatterplots.py
ivirshup: No worries @giovp, this is something @flying-sheep set up so I think he's probably in the best place to understand it.

ivirshup: Made it work by defining a `__signature__` and removing `__wrapped__`. 

`__signature__ ` make `getdoc` work correctly, while removing `__wrapped__` made the sphinx generated documentation have the correct default values.

PR #1359, opened 2020-08-07
Promote ns from machine int to int64.

stuartarchibald:
As title. This should act as a workaround for Numba issue
https://github.com/numba/numba/issues/5955 where the parallel
transform compiler pass cannot handle different sized integers
in the lowering of an array shape.

### DIFF
--- a/scanpy/preprocessing/_qc.py
+++ b/scanpy/preprocessing/_qc.py
@@ -376,7 +376,7 @@ def top_segment_proportions(
         if not isspmatrix_csr(mtx):
             mtx = csr_matrix(mtx)
         return top_segment_proportions_sparse_csr(
-            mtx.data, mtx.indptr, np.array(ns, dtype=np.int)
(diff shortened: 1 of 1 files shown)


stuartarchibald:
CC @flying-sheep this is an untested as I don't have a windows machine handy to trigger the platform-int-size problem. I'm also somewhat guessing at the fix! From looking at the scanpy source, I don't think that changing the `dtype` of `ns` to a platform consistent and wider `int` will do anything catastrophic to performance or alter the logic in the alg in which it's used as it seems to be a simple index. Hope this helps.

flying-sheep:
Since `indptr = indptr.astype(np.int64)` already exists in `top_segment_proportions_sparse_csr`, I assume that inside of that function would also be the best place to do the other promotion.

I guess @ivirshup used `array(ns, int)` because `ns` might be a list or so and not necessarily a numpy array.

stuartarchibald:
@flying-sheep that should also have the same effect.

PR #1362, opened 2020-08-10
Starfish

Mirkazemi:
`scanpy/scanpy/datasets/_datasets.py/starfish_to_anndata()` function coverts starfish expression matrix into AnnData object. Beside storing genes expression in X, var and obs it stores segmentioan/cell spatial data in obsm. 

There are two issues unclear for me:

1. Which module is the best place for such a function? I placed it in `_datasets.py` since it is a tool for converting one data type to another.

2. Travis CI build raises an error: "docstring should start with one-line description" but it has a one-line description.


ivirshup:
I think this would be more appropriate in `anndata`. Since it's not an on disk format, maybe it the function could be called something like `from_starfish`?

* This would also need tests, so some kind of example data.
* Why the differences between this and [starfish's `save_anndata` method](https://spacetx-starfish.readthedocs.io/en/latest/_modules/starfish/core/expression_matrix/expression_matrix.html#ExpressionMatrix.save_anndata)?

giovp:
So I guess basic difference is that this one can convert starfish to anndata "online" without having to read from disk the anndata object. Also, my understanding with `save_anndata` method is that it doesn't export all the features we might want (e.g. area of segmentation masks or others, to be stored in obs). However, this could also be fixed by sending a PR there and modifying what is saved in anndata of the expression matrix.
What do you guys think?

ivirshup:
> So I guess basic difference is that this one can convert starfish to anndata "online" without having to read from disk the anndata object.

Sure. I was also wondering why use literal values here instead of the constant variables that starfish uses (e.g. `"cells"` vs. `Features.CELLS`).

> this could also be fixed by sending a PR there and modifying what is saved in anndata of the expression matrix

I like the idea of having just one version of conversion, instead of them implementing a `to_anndata` method and us implementing an independent `from_starfish` method. Are either of you already in communication with the starfish developers? 

Additionally, it seems like having a `to_anndata` method would be a pretty trivial modification of their code.

giovp:
This can also be closed. What do you think @Mirkazemi @ivirshup 
I'll make a PR in original repo asap

PR #1377, opened 2020-08-18
Add instructions for conda dev install

flying-sheep:
I hope that solves all workflow woes! @ivirshup?

- Users still only need `pip` and can do `pip install scanpy[extras]`
- Installation from source happens via `pip install .[extras]` or `flit install --deps`/`--extras`
- Dev mode install is nonstandard and therefore happens by simply `ln -s scanpy path/to/env/site-packages/` or flit:

grafik

PS: we could also mention `--pth-file` for the 3 windows users who refuse to update to win10 and therefore can’t create symlinks.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -33,11 +33,18 @@ To work with the latest version `on GitHub`_: clone the repository and `cd` into
 its root directory. To install using symbolic links (stay up to date with your
 cloned version after you update with `git pull`) call::
 
-    flit install -s  # from an activated venv
+    flit install -s  # from an activated venv or conda env
     # or
     flit install -s --python path/to/venv/bin/python
 
 .. _on GitHub: https://github.com/theislab/scanpy
+
+If you want to let conda_ handle the installations of dependencies, do::
+
+    pip install beni
+    conda env create -f $(beni pyproject.toml)
+    conda activate scanpy
+    flit install -s
 
 Docker
 ~~~~~~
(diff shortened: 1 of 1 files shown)


Koncopd:
So `python setup.py develop` won't work?

flying-sheep:
“Development installs” aren’t standard. What they do is linking the package into your PYTHONPATH and adding distro metadata (`.egg-info` directories).

You can do that manually using `ln -s package path/to/env/site-packages` or by setting `PYTHONPATH="path/to/package/..:$PYTHONPATH"` (And moving/linking a `.egg-info`/`.dist-info` directory over), or you can use a tool to help you:

- `python3 setup.py delvelop` or `pip install -e .` for a nonstandard setuptools project
- Whatever tool you want to use (in our case `flit install -s` does it)

The reason `pip` doesn’t support it yet for pyproject.toml-based installs is that pip want to support the old nonstandard way and the new standard way – and the new way has no spec for a “dev install” yet, so pip waits until there is one.

Koncopd:
`flit install --deps none -s` breaks `conda list` for me.
using `PYTHONPATH` for scanpy **and** anndata won't allow importing scanpy because `importlib_metadata.PackageNotFoundError: anndata`.

This might be windows specific problems, but `pip setup.py develop` worked perfectly for me.

flying-sheep:
That’s weird, but that might be another issue, please check out #1378.

/edit: seems to be a conda bug that only occurs on windows due to flit ([legally](https://www.python.org/dev/peps/pep-0376/#record)) writing windows newlines into the RECORD file, and conda reading them as two newlines each and then crashing.

---

This PR adds instructions on how to integrate with conda, which I screenshotted. It fails for me with this error:

```
Collecting package metadata (repodata.json): done
Solving environment: failed

ResolvePackageNotFound: 
  - loompy[version='>=3.0.5']
```

But since loompy 3.x isn’t on conda-forge, that’s correct. Seems that resolving anndata’s dependencies on conda is currently not possible and you need to wait until loompy gets upgraded on conda-forge. Or until Quansight-Labs/beni#3 is resolved and you can specify that you don’t want all deps.

PR #1383, opened 2020-08-20
Spatial graph function

ivirshup:
Here's that `spatial_graph` function I had mentioned @giovp and @Mirkazemi.

I figure it should probably be named something different to make sure it's clear it would only work for hex wells.

### DIFF
--- a/scanpy/preprocessing/__init__.py
+++ b/scanpy/preprocessing/__init__.py
@@ -4,7 +4,7 @@ from ._deprecated.highly_variable_genes import filter_genes_dispersion
 from ._highly_variable_genes import highly_variable_genes
 from ._simple import log1p, sqrt, scale, subsample
 from ._simple import normalize_per_cell, regress_out, downsample_counts
-from ._spatial import spatial_connectivity
+from ._spatial import visium_connectivity
 from ._pca import pca
 from ._qc import calculate_qc_metrics
 from ._combat import combat
--- a/scanpy/preprocessing/__init__.py
+++ b/scanpy/preprocessing/__init__.py
@@ -4,6 +4,7 @@ from ._deprecated.highly_variable_genes import filter_genes_dispersion
 from ._highly_variable_genes import highly_variable_genes
 from ._simple import log1p, sqrt, scale, subsample
 from ._simple import normalize_per_cell, regress_out, downsample_counts
+from ._spatial import spatial_connectivity
 from ._pca import pca
 from ._qc import calculate_qc_metrics
 from ._combat import combat
--- a/scanpy/preprocessing/_spatial.py
+++ b/scanpy/preprocessing/_spatial.py
@@ -2,21 +2,37 @@ from ..get import _get_obs_rep
 
 import numpy as np
 from scipy import sparse
+from typing import Optional
 
 
 def spatial_connectivity(
-    adata: "AnnData", obsm: str = "spatial", key_added: str = "spatial_connectivity",
+    adata: "AnnData",
+    obsm: str = "spatial",
+    key_added: str = "spatial_connectivity",
+    n_rings: int = 1,
+    n_neigh: int = 6,
+    radius: Optional[float] = None,
+    coord_type: str = "visium",
 ):
-    """Creates an adjacency matrix of wells for visium data.
+    """
+    Creates graph from spatial coordinates
 
     Params
     ------
     adata
         The AnnData object.
     obsm
-        Key in obsm which stores coordinates of wells.
+        Key to spatial coordinates.
     key_added
-        Key in obsp to add spatial graph as.
+        Key added to connectivity matrix in obsp.
+    n_rings
+        Number of rings of neighbors for Visium data
+    n_neigh
+        Number of neighborhoods to consider for non-Visium data
+    radius
+        Radius of neighbors for non-Visium data
+    coord_type
+        Type of coordinate system (Visium vs. general coordinates)
 
     Usage
     -----
@@ -36,29 +52,49 @@ def spatial_connectivity(
     obsm: 'spatial'
     obsp: 'spatial_connectivity'
     """
-    adata.obsp[key_added] = _spatial_connectivity(_get_obs_rep(adata, obsm=obsm))
+    coords = _get_obs_rep(adata, obsm=obsm)
+    if coord_type == "visium":
+        Adj = _build_connectivity(coords, 6, None, True)
+        if n_rings > 1:
+            # get up to n_rings order connections
+            Adj += Adj ** n_rings
+            Adj.setdiag(0)
+            Adj.eliminate_zeros()
+            Adj.data[:] = 1.0
+        adata.obsp[key_added] = Adj
+    else:
+        adata.obsp[key_added] = _spatial_connectivity(coords, n_neigh, radius, False)
 
 
-def _spatial_connectivity(coords: np.ndarray):
+def _build_connectivity(
+    coords: np.ndarray, n_neigh: int, radius: float, neigh_correct: bool
+):
     """
-    Given the coordinates of hex based cells from a visium experiment, this returns an adjacency matrix for those cells.
-    
-    Usage
-    -----
-    >>> adata.obsp["spatial_connectivity"] = spatial_connectivity(adata.obsm["spatial"])
+    Build connectivity matrix from spatial coordinates
     """
     from sklearn.neighbors import NearestNeighbors
 
     N = coords.shape[0]
-    dists, row_indices = (
-        x.reshape(-1)
-        for x in NearestNeighbors(n_neighbors=6, metric="euclidean")
-        .fit(coords)
-        .kneighbors()
+
+    tree = NearestNeighbors(
+        n_neighbors=n_neigh or 6, radius=radius or 1, metric="euclidean"
     )
-    col_indices = np.repeat(np.arange(N), 6)
-    dist_cutoff = np.median(dists) * 1.3  # There's a small amount of sway
-    mask = dists < dist_cutoff
+    tree.fit(coords)
+
+    if radius is not None:
+        results = tree.radius_neighbors()
+        row_indices = np.concatenate(results[1])
+        lengths = [len(x) for x in results[1]]
+        col_indices = np.repeat(np.arange(N), lengths)
+    else:
+        results = tree.kneighbors()
+        dists, row_indices = (result.reshape(-1) for result in results)
+        col_indices = np.repeat(np.arange(N), n_neigh or 6)
+        if neigh_correct:
+            dist_cutoff = np.median(dists) * 1.3  # There's a small amount of sway
+            mask = dists < dist_cutoff
+            row_indices, col_indices = row_indices[mask], col_indices[mask]
+
     return sparse.csr_matrix(
-        (np.ones(mask.sum()), (row_indices[mask], col_indices[mask])), shape=(N, N)
+        (np.ones(len(row_indices)), (row_indices, col_indices)), shape=(N, N)
     )
--- a/scanpy/preprocessing/_spatial.py
+++ b/scanpy/preprocessing/_spatial.py
@@ -1,0 +1,64 @@
+from ..get import _get_obs_rep
+
+import numpy as np
+from scipy import sparse
+
+
+def spatial_connectivity(
+    adata: "AnnData", obsm: str = "spatial", key_added: str = "spatial_connectivity",
+):
+    """Creates an adjacency matrix of wells for visium data.
+
+    Params
+    ------
+    adata
+        The AnnData object.
+    obsm
+        Key in obsm which stores coordinates of wells.
+    key_added
+        Key in obsp to add spatial graph as.
+
+    Usage
+    -----
+    >>> adata = sc.datasets.visium_sge("V1_Adult_Mouse_Brain")
+    >>> adata
+    AnnData object with n_obs × n_vars = 2698 × 31053
+    obs: 'in_tissue', 'array_row', 'array_col'
+    var: 'gene_ids', 'feature_types', 'genome'
+    uns: 'spatial'
+    obsm: 'spatial'
+    >>> sc.pp.spatial_connectivity(adata)
+    >>> adata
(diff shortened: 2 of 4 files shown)


giovp:
It looks great!
image

One first improvement could be to expose a parameters that explicitly ask for the number of rings in the neighbors (first 6 ring etc.). Btw, how do I push to this specific PR from my local repo? I managed to pull it but can't figure out a way to push here and not on my fork.

ivirshup:
Oh, sorry, I had completely missed your comment here!

> It looks great!

Thanks! Can I ask why you used leiden clustering on this?

> One first improvement could be to expose a parameters that explicitly ask for the number of rings in the neighbors

This should be easy enough. I'm curious as to whether this it's better to leave this up to whatever algorithm is being used however, since the one step graph has some nice properties. It'd probably be important to include distance in the multistep graph.

> Btw, how do I push to this specific PR from my local repo?

This should be fairly straight forward. If you're using the github cli, I think it should just be:

```sh
gh pr checkout 1383
# whatever changes
git push
```

Let me know if that gives you errors, since it might be a repository permissions issue.

Koncopd:
@ivirshup @giovp i have updated the function, now it can process non-visium coordinates and have the number of rings option for visium.
https://github.com/theislab/spatial-tools/blob/graph/notebooks/build_spatial_graph.ipynb

giovp:
That's great, I'll add Isaac to that project so he can see code (repo is private). Let's discuss whether to integrate in scanpy at next meeting! Thank you Sergei !

ivirshup:
Hey! I was just getting back to this, and I'm not sure I agree with all the choices made in the updates. For example, I would probably rather have separate functions for different spatial neighbor strategies. Also this won't work if `coord_type` isn't `"visium"`.

Should I be making changes to this PR, or are you relying on it?

# hexagonal connectivity

I would propose a separate function just for visium data, maybe called `visium_connectivity`. It works with the assumption of a hexagonal grid. This removes the `neigh`, `radius`, and `coord_type` arguments.

I think if we have an argument for `n_rings` there should be some way to weight the connectivity graph by how many steps away each extra point is. Also I'm pretty sure the current implementation of `n_rings` is incorrect when `n_rings>2`.

Without weighting, I think it should be more like this:

```python
def walk_nsteps(adj, n):
    """Expand adjacency matrix adj by walking out n steps from each node."""
    adj = adj.astype(bool)
    cur_step = adj
    result = adj.copy()
    for i in range(n):
        cur_step = adj @ cur_step
        cur_step.setdiag(False)
        result = result + cur_step
    return result
```

<details>
<summary> An example showing this works </summary>

```python
import networkx as nx
from scipy import sparse
from matplotlib import pyplot as plt
import numpy as np

def walk_nsteps(adj, n):
    """Expand adjacency matrix adj by walking out n steps from each node."""
    adj = adj.astype(bool)
    cur_step = adj
    result = adj.copy()
    for i in range(n):
        cur_step = adj @ cur_step
        cur_step.setdiag(False)
        result = result + cur_step
    return result

# Test data (path graph)

G = nx.Graph()
G.add_nodes_from([0,1,2,3])
G.add_edges_from([(0, 1), (1, 2), (2, 3), (3, 4)])
adj = nx.adjacency_matrix(G).astype(bool)

fig, axes = plt.subplots(nrows=3)
# Fixed circle layout
pos = {i: (np.cos(-np.pi + (np.pi * i) / 4), np.sin(-np.pi + (np.pi * i) / 4)) for i in range(5)}

for n, ax in enumerate(axes):
    nx.draw(
        nx.Graph(walk_nsteps(adj, n)),
        pos=pos,
        ax=ax,
    )

plt.show()
```
image
</details>

# Radius/ n-neighbors coordinates

I'm a little less sure about this use case. Since cells come in all shapes and sizes, I'm not sure if representing each cell as a centroid with a fixed radius or number of neighbors is the way to go. That said, maybe it is.

This could be a separate function `spatial_connectivity`? In this case, I think it would also be useful to return a distance matrix.

Koncopd:
About adding all powers of adjacency matrix - i implemented it at first as you did, but then i thought that it was redundant and changed to the present variant. My thought was that the hexagonal connectivity structure would allow to get all paths of less than n_rings with only n_rings power. And this works in practice, but i agree that there can be some edge cases with isolated node blocks.

ivirshup:
@Koncopd here's an example of the current implementation breaking down at `n=3`

```python
def walk_nsteps_current(adj, n=1):
    adj = adj.copy()
    if n > 1:
        # get up to n_rings order connections
        adj += adj ** n
        adj.setdiag(0)
        adj.eliminate_zeros()
        adj.data[:] = 1.0
    return adj

fig, axes = plt.subplots(nrows=3)
# Fixed circle layout
pos = {i: (np.cos(-np.pi + (np.pi * i) / 4), np.sin(-np.pi + (np.pi * i) / 4)) for i in range(5)}

for n, ax in enumerate(axes):
    nx.draw(
        nx.Graph(walk_nsteps_current(adj, n + 1)),  # making sure we start at 1
        pos=pos,
        ax=ax,
    )
    
plt.show()
```

image

For `n=3` you're missing the neighbors at depth=2. Basically, you're just jumping to the 3rd step, instead of accumulating neighbors up to that step.

Koncopd:
Yes, but if we have hexagonal connections, then, for example, n-1, n-2 and so on paths will be counted.
2 examples
image
You could check also the notebook i posted above.

LuckyMD:
Just reading along.... if all you want is to find neighbors within a certain number of hops, then non-zero values of powers of the adjacency matrix is a bit inefficient i think. There should be simple breadth-first-search or depth-first-search algorithms implemented in `networkx` I imagine. And if you're bent on this approach, adding self-loops (diag = 1) will mean you can just do powers.

giovp:
I'm sorry but I'm not sure I'm following you guys in the discussion, although from the figure it does seem that @Koncopd implementation is missing the n=2.

Regarding radius, this is definitely super useful for anything that is not a grid, and also agree that a distance matrix can be useful, maybe then creating a graph by binarizing that distance matrix based on a threshold?

Koncopd:
@LuckyMD thanks for your comment.
So it seems that it is enough to do this
```
adj.setdiag(1) # or 2?
adj += adj ** n
adj.setdiag(0)
adj.eliminate_zeros()
adj.data[:] = 1.0
```
And this should eliminate all edge cases.

Not sure if it is a good idea to use networkx for this.

LuckyMD:
@Koncopd yes, I believe that should cover everything (maybe test to make sure I'm not missing sth here). However, I still think taking adjacency matrix powers will not be as fast as a BFS/DFS.

ivirshup:
@LuckyMD matrix multiplication on sparse matrices is actually a pretty efficient version of breadth first search, as [used by graphBLAS](http://arxiv.org/abs/1606.05790v2). Here's an example for a BFS from a specific point (which you can expand to all points by using the identity matrix):

image

@Koncopd, I think that should work, since you're adding a self edge so you have redundancy at each step. 

A separate point of contention on handling it like this: do you want the first step neighbors to have the same weight as the second step neighbors? My assumption would be no.

ivirshup:
Btw, I've separated the visium and non-visium case (since they don't share much code), added tests for the visium case, and removed the warnings. I've rebased so I could re-use some test data from another PR.

LuckyMD:
> @LuckyMD matrix multiplication on sparse matrices is actually a pretty efficient version of breadth first search, as [used by graphBLAS](http://arxiv.org/abs/1606.05790v2).

Hmm... i had no idea it was more efficient than using queues, but Figure 10 suggests otherwise. Matrix multiplication definitely seems easier.

> do you want the first step neighbors to have the same weight as the second step neighbors?

You could keep these separate by binarizing the adjacency matrix before doing the multiplication. The neighbors that are only reachable via the Nth-hop should always have a 1 in the N-th matrix product that way.

ivirshup:
> You could keep these separate by binarizing the adjacency matrix before doing the multiplication. The neighbors that are only reachable via the Nth-hop should always have a 1 in the N-th matrix product that way.

Thats a problem with this strategy. Since it's an undirected graph, a node is it's own second neighbor. Since it's a hexagonal grid my first neighbors are also my second neighbors as well as my nth neighbors (in most cases, if there are edges or missing cells this might not be the case). We would either have to do our own BFS which precludes back tracking (i.e. for each search from each node remove previously visited edges), or we could take the difference of the edge sets at each update. Taking the difference would probably be easier.

LuckyMD:
With self-loops a node is also its first neighbor. That means if you binarize the adjacency matrix and you have a 1 in the adjacency matrix after taking some power, that means there is only 1 way to get to that neighbor and it must thus only be reachable in the N-th hope (N being the power you just multiplied with). You could therefore also just look for the 1s in the matrix after every multiplication.... that might be a bit easier than taking the difference.

ivirshup:
Just checked this, and some of the members of the nth ring have more than one neighbor in the nth-1, so checking for values of one doesn't quite work. Also I've been [collecting some functions](https://github.com/ivirshup/sparse_wrapper/blob/master/sparse_wrapper/_core/coordinate_ops.py) for set operations on sparse indices so taking the difference was quite easy 😉

Here's the check:

<details>
<summary> Code </summary>

```python
import scanpy as sc
from sparse_wrapper._core.coordinate_ops import difference_indices
import numpy as np
from scipy import sparse

def find_steps_val(adj, n_steps):
    """Finding positions with value of 1."""
    cur = adj.astype(int)
    diffs = [adj.copy()]
    for i in range(n_steps):
        cur.data[:] = 1
        cur = adj @ cur
        diffs.append(cur == 1)
    return diffs


def find_steps(adj, n_steps):
    """Finding differences in sparsity patterns."""
    diffs = [adj.copy()]
    prev = adj + sparse.eye(adj.shape[0])
    for i in range(n_steps):
        cur = prev @ adj
        diffs.append(difference_indices(cur, prev))
        prev = cur
    return diffs

adata = sc.datasets.visium_sge("V1_Adult_Mouse_Brain")
sc.pp.visium_connectivity(adata)
adj = adata.obsp["spatial_connectivity"]

N = 3

diff_res = find_steps(adj, N-1)
val_res = find_steps_val(adj, N-1)

for i in range(N):
    adata.obs[f"val_res_{i}"] = np.ravel(val_res[i][50, :].toarray()).astype(int)
    adata.obs[f"diff_res_{i}"] = np.ravel(diff_res[i][50, :].toarray()).astype(int)

sc.pl.spatial(
    adata,
    color = [f"val_res_{i}" for i in range(N)] + [f"diff_res_{i}" for i in range(N)],
    ncols=N
)
```

</details>

This is a plot of the neighbors of the 50th well at steps 1, 2, and 3 from each method. Top is finding positions that equaled 1, bottom is taking the differences of the sparsity patterns.

image

Either way, these look kinda pretty:

<details>
<summary> </summary>

```python
from operator import add
from functools import reduce

cells = np.random.choice(adata.n_obs, 20, replace=False)
adata.obs["fireworks"] = (
    reduce(add, (diff_res[i][cells] * (3 - i) for i in range(3)))
    .max(axis=0)
    .toarray()
    .ravel()
)
sc.pl.spatial(adata, color="fireworks", cmap="inferno")
```

</details>

image

LuckyMD:
> some of the members of the nth ring have more than one neighbor in the nth-1

Damn... yeah, that won't work then. Back to differences of adjacency matrices then i guess.

It looks as though your functions are replicating what I assume is going on in the backend of `networkx` anyway. Are you trying to avoid the heavy dependency or why replicate the effort? I would assume that once we need things like label propagation (maybe to denoise label assignment after deconvolving spatial spots?) then `networkx` might come in handy, no?

ivirshup:
> It looks as though your functions are replicating what I assume is going on in the backend of networkx anyway.

So that repo is actually me trying to make it so we can have CSR and CSC matrices that act the same as numpy ndarrays. Those functions are just some building blocks for broadcast operations on those arrays, which are otherwise implemented in C++ in scipy.

A lot of this could probably be done with networkx. networkx is just very slow, since it's all pure python.

LuckyMD:
> A lot of this could probably be done with networkx. networkx is just very slow, since it's all pure python.

Ah, I wasn't aware of that. Cool... then just trying to work with arrays I guess. I hope using spatial graphs doesn't require more than some basic operations in that case.

giovp:
quick practical comment on this very interesting discussion.
@ivirshup shall we have this here or moved to the other package under development? We need to take a decision on this because we'll have to see how it works with rest of functions. Pinging @Koncopd as well.
Sorry to put pressure but we are on a tight schedule 😅 

re: networkx discussion, also agree with Isaac on having these operations external to networkx.

ivirshup:
I'd like to add this function plus tests to scanpy. I think I'll leave out `n_rings` argument and the `radius_neighbors` functions until there are clear use-cases. I would recommend just having a copy of the code in spatial-tools, which can be deduplicated later.

giovp:
ok that's great, thank you! The `radius_neighbors` have very clear applications in fish-like data, and we are assmebling a tutorial to show exactly that. The n-rings as well especially in the context of cell-cell communication (although did not check that systematically yet). So shall we add this functionality to spatial-tools ?

ivirshup:
I could probably be convinced to include `radius_neighbors`. I'd mainly need a test case. My initial opposition was that (1) it's a pretty trivial implementation and (2) without a real example I'm not sure if it's missing any obvious edge cases.

For n-rings, I think that walking some steps from each node generalizes beyond graph construction, and might be reasonable to have as a separate method. I also haven't seen any recommendations about how to weigh the edges, which I think is pretty important.

giovp:
Yeah the edge weighing is definitely not clear we would've to test that. ok so only tests are missing ?

giovp:
ok, after talking with @davidsebfischer we definitely want to have both radius as well as n_rings and also have graph transform functions so to weight edges differently (based on different kernels).
I'd suggest to keep the function in spatial-tools and then decide what to port to scanpy.

@ivirshup regarding use cases (for radius, n_rings, and edge weighing) there are multiple ones from graph convolutional networks application, but it could be interesting also just for the other graph analysis that we have implemented in spatial-tools. We'd have to do extensive interpretation of results of course to prove that it really is useful.

giovp:
super interesting discussion btw. 
I think this can be closed

PR #1390, opened 2020-08-25
Implement sc.get.summarized_expression_df

gokceneraslan:
This pr moves the (in)famous `_prepare_dataframe` function from scanpy.plotting._anndata to sc.get as `_indexed_expression_df` (bcs why would it be in plotting anyway) and implements a simple public interface called `sc.get.summarized_expresion_df` which simply provides nonzero mean/var and fraction using `_indexed_expression_df` function. 

As discussed here (https://github.com/theislab/scanpy/pull/1388#issuecomment-678739734) we can use this in rank_genes_groups_df.

### DIFF
--- a/scanpy/get.py
+++ b/scanpy/get.py
@@ -1,12 +1,7 @@
 """This module contains helper functions for accessing data."""
 from typing import Optional, Iterable, Tuple, Mapping, Union, Sequence
 
-try:
-    from typing import Literal
-except ImportError:
-    try:
-        from typing_extensions import Literal
-    except ImportError:
+from ._compat import Literal
 
 import numpy as np
 import pandas as pd
--- a/scanpy/get.py
+++ b/scanpy/get.py
@@ -1,5 +1,12 @@
 """This module contains helper functions for accessing data."""
-from typing import Optional, Iterable, Tuple, Mapping, Union, Sequence, Literal
+from typing import Optional, Iterable, Tuple, Mapping, Union, Sequence
+
+try:
+    from typing import Literal
+except ImportError:
+    try:
+        from typing_extensions import Literal
+    except ImportError:
 
 import numpy as np
 import pandas as pd
--- a/scanpy/get.py
+++ b/scanpy/get.py
@@ -1,11 +1,15 @@
 """This module contains helper functions for accessing data."""
-from typing import Optional, Iterable, Tuple
+from typing import Optional, Iterable, Tuple, Mapping, Union, Sequence
 
 import numpy as np
 import pandas as pd
-from scipy.sparse import spmatrix, issparse
+from pandas.api.types import is_categorical_dtype
+from scipy.sparse import spmatrix
 
 from anndata import AnnData
+from ._utils import sanitize_anndata
+
+_VarNames = Union[str, Sequence[str]]
 
 # --------------------------------------------------------------------------------
 # Plotting data helpers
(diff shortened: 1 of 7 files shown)


scanpy/get.py
```diff
@@ -295,3 +396,150 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+        if layer not in adata.layers.keys():
+            raise KeyError(
+                f'Selected layer: {layer} is not in the layers list. '
+                f'The list of valid layers is: {adata.layers.keys()}'
+            )
+        matrix = adata[:, var_names].layers[layer]
+    elif use_raw:
+        matrix = adata.raw[:, var_names].X
+    else:
+        matrix = adata[:, var_names].X
```
ivirshup, line 552: We've got a function for this:

```suggestion
    matrix = sc.get._get_obs_rep(adata[:, var_names], use_raw=use_raw, layers=layer)
```

scanpy/get.py
```diff
@@ -295,3 +396,150 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+
+    Returns
+    -------
+    Tuple of `pandas.DataFrame` and list of categories.
+    """
+    from scipy.sparse import issparse
+
+    sanitize_anndata(adata)
+    if use_raw is None and adata.raw is not None:
+        use_raw = True
```
ivirshup, line 500: Can this just default to `False`? I think the behavior around acting differently depending on whether `raw` is present is confusing, and makes it hard to reason about what code is going to do.

scanpy/get.py
```diff
@@ -295,3 +396,150 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+    if groupby is not None:
+        if isinstance(groupby, str):
+            # if not a list, turn into a list
+            groupby = [groupby]
+        for group in groupby:
+            if group not in adata.obs_keys():
+                raise ValueError(
+                    'groupby has to be a valid observation. '
+                    f'Given {group}, is not in observations: {adata.obs_keys()}'
+                )
```
ivirshup, line 518: Why is this error better than what pandas would throw?

scanpy/get.py
```diff
@@ -295,3 +396,150 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+            adata_or_raw = adata.raw
+        else:
+            adata_or_raw = adata
+        for symbol in var_names:
+            if symbol not in adata_or_raw.var[gene_symbols].values:
+                logg.error(
+                    f"Gene symbol {symbol!r} not found in given "
+                    f"gene_symbols column: {gene_symbols!r}"
+                )
+                return
```
ivirshup, line 536: Probably just throw a value error?

scanpy/get.py
```diff
@@ -295,3 +396,150 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+    if gene_symbols is not None and gene_symbols in adata.var.columns:
+        # translate gene_symbols to var_names
+        # slow method but gives a meaningful error if no gene symbol is found:
+        translated_var_names = []
+        # if we're using raw to plot, we should also do gene symbol translations
+        # using raw
+        if use_raw:
+            adata_or_raw = adata.raw
+        else:
+            adata_or_raw = adata
```
ivirshup, line 529: There's already some very similar logic to this in `sc.get.obs_df`. Could that be re-used here? If not, could a more similar solution be used?

Also, doesn't this conflict with how `var_names` is being set if no values are passed?

scanpy/get.py
```diff
@@ -295,3 +396,150 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+                    for group in groupby[1:]:
+                        # create new category by merging the given groupby categories
+                        obs_tidy_idx = (
+                            obs_tidy_idx.astype(str) + "_" + adata.obs[group].astype(str)
+                        ).astype('category')
+                obs_tidy_idx.name = "_".join(groupby)
+                idx_categories = obs_tidy_idx.cat.categories
+            else:
+                obs_tidy_idx = [adata.obs[group] for group in groupby] # keep as multiindex
+                idx_categories = [x.cat.categories for x in obs_tidy_idx]
```
ivirshup, line 584: Is this the current behavior?

scanpy/get.py
```diff
@@ -243,6 +249,101 @@ def var_df(
...
+        layer=layer,
+        gene_symbols=gene_symbols,
+        concat_indices=False,
+    )
+
+    if ops is None:
+        ops = ['mean_expressed', 'var_expressed', 'fraction']
+    if isinstance(ops, str):
+        ops = [ops]
+    assert all(np.isin(ops, ['mean_expressed', 'var_expressed', 'fraction'])), 'Undefined op'
```
ivirshup, line 376: Should this also include normal mean and variance?

scanpy/get.py
```diff
@@ -243,6 +249,101 @@ def var_df(
...
+    res = {}
+    # .agg is super slow even for mean and var, so do it separately using .mean and .var
+    if 'mean_expressed' in ops or 'var_expressed' in ops:
+        nonzero_group = df.mask(df<=threshold).groupby(level=df.index.names, observed=True)
+        if 'mean_expressed' in ops:
+            res['mean_expressed'] = nonzero_group.mean()
+        if 'var_expressed' in ops:
+            res['var_expressed'] = nonzero_group.var()
+    if 'fraction' in ops:
+        res['fraction'] = (df>threshold).groupby(level=df.index.names, observed=True).mean()
```
ivirshup, line 388: Calculations should happen on the matrix from anndata, not on a dataframe. This would be much more efficient since it wouldn't require densifying sparse arrays.

ivirshup (changes requested):
Thanks for looking into this!

I like what `summarized_expression_df` does, and think the operations could definitely be expanded. I'm realizing I'm not actually quite sure what I expect this to return, especially when `groupby` is passed. Can you write out some examples of use cases where you show the output?

However, I'm not sure I see the usefulness of `_indexed_expression_df`. I think `_prepare_dataframe` has been a bit overused in the codebase, particularly where it's pretty close to `adata.to_df()`. Instead of moving that function here directly, can we look at what the use cases are for it, and replace it on a case by case basis? We should avoid densify sparse arrays as much as possible, and I don't think it's worth it here.

PR #1402, opened 2020-09-02
Remove list, map, lambda chaining

### LINKED ISSUE #1401: Using list(map(lambda)) is suboptimal
As the title suggests, it can be suboptimal w.r.t computation time to use

```python
list(map(lambda x: ..., SOME_LIST))
```

In `readwrite.py` [here](https://github.com/theislab/scanpy/blob/f704f724529def21769ee6407f9b47b5c161564c/scanpy/readwrite.py#L190) for example, instead of chaining of the three functions, we can simply compare to a `pandas.Series`:

```python
adata = adata[:, adata.var['genome'] == f"{genome}"]
```

This improves readability and is slightly faster.

In a quick check I ran in Jupyter Lab, this approach is three times faster. The following code should run in Jupyter Lab as is:

```python
import random
import string

import numpy as np
import pandas as pd


from anndata import AnnData

letters = string.ascii_lowercase

adata = AnnData(pd.DataFrame(np.random.randn(1000, 30000)))
adata.var["genome"] = np.array([random.choice(letters) for _ in range(30000)]).reshape(-1, 1)

genome = 'g'
%timeit adata[:, list(map(lambda x: x == f"{genome}", adata.var['genome']))]

%timeit adata[:, adata.var['genome'] == f"{genome}"]
```

WeilerP:
## Changes

Instead of chaining `list`, `map` and `lambda`, the same outcome can be achieved by comparing to a `pandas.Series` object. This approach seems to be three times faster.

## Related issue

Closes #1401.


scanpy/readwrite.py
```diff
@@ -179,12 +179,9 @@ def read_10x_h5(
                     f"Could not find data corresponding to genome '{genome}' in '{filename}'. "
                     f'Available genomes are: {list(adata.var["genome"].unique())}.'
                 )
-            adata = adata[:, list(map(lambda x: x == str(genome), adata.var['genome']))]
+            adata = adata[:, adata.var['genome'] == f"{genome}"]
```
ivirshup:
```suggestion
            adata = adata[:, adata.var['genome'] == genome]
```

Why the format string?

WeilerP: Sorry, wanted to ask the same question: The original code transformed it to a string so I kept it.

WeilerP: @ivirshup, do you want me to remove the conversion to string?

WeilerP:
@ivirshup, building the docs failed even though I did not change any documentation. Do you have an idea what the problem may be?

Koncopd:
@WeilerP yes, test fails seem unrelated to your pr, some changes in the plotting packages cause it.

fidelram:
@ivirshup Can we merge this change?

ivirshup:
LGTM

PR #1413, opened 2020-09-12
Use umap's pairwise_special_metric as a fallback for small AnnDatas with metrics unsupported by sklearn

### LINKED ISSUE #1011: New UMAP distances like ll_dirichlet and hellinger throw exception with small AnnDatas
If adata.n_obs < 4096 and umap version >= 0.4 and if `metric` is a distance that is not supported by scikit-learn (like ll_dirichlet or hellinger), we get a ValueError:

Code for reproducing with UMAP >= 0.4:

```python
import scanpy as sc

adata = sc.datasets.paul15()
sc.pp.neighbors(adata, metric='hellinger')
```

```pytb
<ipython-input-5-e2c66b650fd3> in <module>
      2 
      3 adata = sc.datasets.paul15()
----> 4 sc.pp.neighbors(adata, metric='hellinger')

~/.anaconda3/lib/python3.7/site-packages/scanpy/neighbors/__init__.py in neighbors(adata, n_neighbors, n_pcs, use_rep, knn, random_state, method, metric, metric_kwds, copy)
    108         n_neighbors=n_neighbors, knn=knn, n_pcs=n_pcs, use_rep=use_rep,
    109         method=method, metric=metric, metric_kwds=metric_kwds,
--> 110         random_state=random_state,
    111     )
    112     adata.uns['neighbors'] = {}

~/.anaconda3/lib/python3.7/site-packages/scanpy/neighbors/__init__.py in compute_neighbors(self, n_neighbors, knn, n_pcs, use_rep, method, random_state, write_knn_indices, metric, metric_kwds)
    686             # non-euclidean case and approx nearest neighbors
    687             if X.shape[0] < 4096:
--> 688                 X = pairwise_distances(X, metric=metric, **metric_kwds)
    689                 metric = 'precomputed'
    690             knn_indices, knn_distances, forest = compute_neighbors_umap(

~/.anaconda3/lib/python3.7/site-packages/sklearn/metrics/pairwise.py in pairwise_distances(X, Y, metric, n_jobs, **kwds)
   1550         raise ValueError("Unknown metric %s. "
   1551                          "Valid metrics are %s, or 'precomputed', or a "
-> 1552                          "callable" % (metric, _VALID_METRICS))
   1553 
   1554     if metric == "precomputed":

ValueError: Unknown metric hellinger. Valid metrics are ['euclidean', 'l2', 'l1', 'manhattan', 'cityblock', 'braycurtis', 'canberra', 'chebyshev', 'correlation', 'cosine', 'dice', 'hamming', 'jaccard', 'kulsinski', 'mahalanobis', 'matching', 'minkowski', 'rogerstanimoto', 'russellrao', 'seuclidean', 'sokalmichener', 'sokalsneath', 'sqeuclidean', 'yule', 'wminkowski', 'haversine'], or 'precomputed', or a callable
```

Similar to this code in UMAP (https://github.com/lmcinnes/umap/pull/259/files), we should check if scikit's `pairwise_distances` throws a ValueError and fallback to UMAP's own pairwise `pairwise_special_metric` function.

dwnorton:
This is an attempt to fix #1011

I have taken my cue from https://github.com/lmcinnes/umap/pull/259/files as suggested in the issue's thread. I have also added some tests (which focus on ensuring that results are successfully created rather than ensuring correctness).

I hope this helps.


dwnorton:
The CI tests failed due to the image matching problems (in test_violin and test_pbmc3k) fixed by PR #1422 which is now merged into master. Should I merge master into my fix branch?

fidelram:
yes, please merge with the master branch

gokceneraslan:
Thanks so much for your contribution @dwnorton ! Looks good. Two questions: 1) is there any way to do without densifying the matrix? it's not a big deal though since the matrix is small. 2) Is the UMAP API for the `umap.distances.pairwise_special_metric` function version-dependent somehow in other words do we need to constrain our umap dependency to a specific version, or is the API stable across all commonly used UMAP versions?

dwnorton:
Thanks for the feedback. I have now merged master into the fix branch and the CI tests are happy. To address the shrewd questions asked by @gokceneraslan:
1) I pondered over this myself. As I mentioned (above), I based my approach on a pull request in umap's own codebase which also resorts to densification of the matrix. As far as I can see, `pairwise_special_metric` doesn't directly support a sparse matrix and I don't see an alternative approach that wouldn't involve duplicating some of its implementation and (probably) sacrificing some of the benefits (i.e. parallel processing). A deeper analysis by another brain might draw different conclusions.
2) Another good question. Based on my "git blame" detective work, `pairwise_special_metric` was introduced in [a change on 20 Nov 2018](https://github.com/lmcinnes/umap/commit/edade6841bd9b3c80454bf7f4386177c9aa35ab5) which should have seen it incorporated into version [0.3.7](https://github.com/lmcinnes/umap/tree/0.3.7). Since then its signature has remained compatible (with the only change being the addition of the `kwds` argument). scanpy's `requirements.txt` already has `umap-learn` set to a minimum version of 0.3.10 so I believe we're good on that front.

gokceneraslan:
Hi @dwnorton, 

Sorry for the delay. I can see now that there are a few changes about this on the [UMAP side](https://github.com/lmcinnes/umap/blob/master/umap/spectral.py#L98):

image

Few questions:

-  Can we improve the sparse array handling (e.g. mimicking UMAP e.g. using `SPARSE_SPECIAL_METRICS`) 
-  Can we also use the `SKLEARN_PAIRWISE_VALID_METRICS` instead of catching the ValueError? Would that be safer?
- If we start using these keywords, do we need to change any of the requirements of scanpy?

ivirshup:
Would it be easier to just use NNDescent's handling of distances here?

[Also, this was brought up on the discourse recently](https://scanpy.discourse.group/t/using-a-custom-metric-for-scanpy-pp-neighbors/477/3).

Koncopd:
@dwnorton seems fine, but could you please commit anything here to trigger ci?

@ivirshup do we want to merge this or replace the current neighbors computation method with nndescent?

dwnorton:
Hello. Sorry I have been silent but I haven't used scanpy for a few months and thought there was nothing more for me to do with this pull request.

@Koncopd: the commit I made in September last year (e4483e9) triggered the CI and passed the tests. Is there something further you need me to do?

Thanks.

ivirshup:
Long term, I'd prefer to just use pynndescent since it would be a more simple implementation. That could change some results. What do you think?

Koncopd:
@ivirshup i would also prefer pynndescent. I can write it, but i need to look and check that there are no potential problems with this replacement.

ivirshup:
@Koncopd when we talked about this last there were concerns about backwards reproducibility. I'm wondering if this logic would fix that:

* If the dataset is "small" and the metric is defined by scikit-learn, compute complete distances
* For all other cases use pynndescent

Would this be sufficient to keep results the same, or do we compute dense distances for the more esoteric metrics now?

Koncopd:
Yes, we can do like this.
 "do we compute dense distances for the more esoteric metrics now?" - actually i am not sure, it needs to be checked.

flying-sheep:
This is all possible with `sc.pp.neighbors(..., transformer="pynndescent")`, except for metrics that are only implemented in umap, like `ll_dirichlet`.

If we want to support these, we should find a way to get them work nicely with the new API, so I’m closing this for now.

PR #1421, opened 2020-09-17
update ecosystem docs to contain scvi-tools

adamgayoso:
We recently transitioned from scvi to scvi-tools. Here I update the ecosystem docs to reflect this change.

I decided to make a new section for it as we have a diversity of models in our package that don't fit any of the other categories well.


fidelram:
@adamgayoso A recent update of seaborn caused some trouble with the tests but is now fixed. Can you merge with master to trigger again the tests?

PR #1422, opened 2020-09-17
Fix / Unit tests `test_violin` and `test_pbmc3k`

### LINKED ISSUE #1420: Unit tests fail with latest version of seaborn
## Description

Using the `seaborn==0.11` makes `scanpy/tests/test_plotting.py:test_violin` and `scanpy/tests/notebooks/test_pbmc3k.py` fail. The warning

```bash
UserWarning: Vertical orientation ignored with only `x` specified.
```

is thrown by `seaborn` in the stack trace for the failed tests.

The problem seems to originate from [here](https://github.com/theislab/scanpy/blob/f704f724529def21769ee6407f9b47b5c161564c/scanpy/plotting/_anndata.py#L745) and [here](https://github.com/theislab/scanpy/blob/f704f724529def21769ee6407f9b47b5c161564c/scanpy/plotting/_anndata.py#L747), where `x` and `y` are not both specified.

See #1417 and especially [this](https://github.com/theislab/scanpy/pull/1417#issuecomment-693488428) comment by @fidelram for more details.

WeilerP:
## Description

As briefly outlined in #1420, using `seaborn==0.11` causes the unit tests `scanpy/tests/test_plotting.py:test_violin` and `scanpy/tests/notebooks/test_pbmc3k.py` fail as the violin plots are no longer vertically oriented.

## Changes

* Rely on `seaborn.catplot` and `seaborn.stripplot` instead of `seaborn.FacetGrid` in `scanpy/plotting/_anndata.py`

## Related issues

Closes #1420.


WeilerP:
@fidelram, from `seaborn==0.12` on, the only valid positional argument will be `data` which may cause failures or unexpected behaviour. This is relevant at least [here](https://github.com/theislab/scanpy/blob/f704f724529def21769ee6407f9b47b5c161564c/scanpy/plotting/_anndata.py#L774) and [here](https://github.com/theislab/scanpy/blob/f704f724529def21769ee6407f9b47b5c161564c/scanpy/plotting/_anndata.py#L785). Should we add / specify the keyword argument `x` in this PR or open a separate issue and PR?

fidelram:
@WeilerP Thanks for finding about this. Yes, please add explicitly `x=x` in this PR 👍

fidelram:
Thanks. Any idea why the images are now bigger (which I assume is the reason for the very small fonts)?

WeilerP:
Hm, no sorry, I don't. My best guess would be that `seaborn.catplot` and `seaborn.FacetGrid` use different default figure sizes which in turn makes the font look smaller?

WeilerP:
I noticed, however, that the tests pass even when removing the `if stripplot:` part. Any idea on why this is happening and how to prevent it?

fidelram:
The tests have a tolerance parameter that is set high. The problem is that the stripplot shows different results each time. Also, different versions of matplotlib and seaborn have slight differences.

fidelram:
I wonder why the tests are not working now?

WeilerP:
> I wonder why the tests are not working now?

Sorry, I forgot to update `violin.png` after the latest changes to `_anndata.py`. Let's see if the tests pass now.

WeilerP:
> The tests have a tolerance parameter that is set high. The problem is that the stripplot shows different results each time. Also, different versions of matplotlib and seaborn have slight differences.

Ah yes, I see. The stripplot result could be fixed by setting a seed with `np.random.seed`. I doubt it will fix the difference due to the used version, though.

hhhh1230511:
Hi Developers,

I got the same issue when using seaborn==0.11, and fortunately I got the issue solved by replacing the annotate.py with the modified version. Howerver, I was wondering why I tried to use 'pip install scanpy' to update the scripts, it failed? Is there any other easier method to modify the script, not to locate the file and replace it?

WeilerP:
@hhhh1230511, this PR is not part of any release yet (the latest version `scanpy==1.6` was released August 15, 2020). If you want to have the latest version from GitHub you can follow the instructions for a developer installation [here](https://scanpy.readthedocs.io/en/stable/installation.html) in the documentation, for example. Once a new release is available on `pip`, you can install it via `pip install --upgrade scanpy`
In general, you should avoid modifying the code by e.g. simply copying and pasting. This will either easily cause conflicts when updating the package or cause problems when functions from other files which depend on the content you changed but were not updated accordingly.

Hope this helped and clarified things.

PR #1424, opened 2020-09-22
Replace `np.matrix` by `np.array`

### LINKED ISSUE #1423: Use `np.array` instead of `np.matrix`
## Description

Numpy does not recommend the usage of the matrix subclass. Since the introduction of the matmul operator `@` in python 3.5, the biggest advantage of the matrix subclass does not exist anymore. Since `scanpy` works with python 3.6 and python 3.7 and `numpy>=1.17.0`, there is no apparent reason why to use `np.matrix` instead of `np.array`.

Currently, `np.matrix` is used in `scanpy/plotting/_anndata.py` causing, for example, the `PendingDeprecationWarning`

```pytb
PendingDeprecationWarning: the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
    np.matrix([label2code[lab] for lab in obs_tidy.index])
```

WeilerP:
## Changes

Replaces `np.matrix` by `np.array`. This suppresses the `PendingDeprecationWarning` thrown by `numpy`.

## Related issue

Closes #1423.

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -1242,7 +1242,7 @@ def heatmap(
                 arr += [idx] * (pos[1] + 1 - pos[0])
 
             gene_groups_ax.imshow(
-                np.matrix(arr).T, aspect='auto', cmap=groupby_cmap, norm=norm
+                np.array([arr]).T, aspect='auto', cmap=groupby_cmap, norm=norm
             )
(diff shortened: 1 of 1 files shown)


fidelram:
Thanks!

PR #1430, opened 2020-09-24
[WIP] Remove usage of AnnData constructor dtype argument

ivirshup:
Companion PR to https://github.com/theislab/anndata/pull/434

Basically, I would like to deprecate the `dtype` argument of `AnnData._init_as_actual`, since it mostly just makes unexpected copies of `X`. Since other elements of an AnnData object are passed by reference, it makes sense for this to happen with `X` as well.

Right now, this PR will fail CI. What I've done so far is remove all uses of that argument from the scanpy code base, while keeping the tests passing. I'm trying to figure out how to best preserve compatibility with older versions of `anndata`, without throwing too many warnings.

I think the thing to do will be make code work with both (`AnnData(X.astype(dtype), dtype=dtype))` should only make one copy) and catch warnings. This can be removed once scanpy depends on `anndata 0.8`.

### DIFF
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -493,7 +493,7 @@ def normalize_per_cell(
             adata.obs[key_n_counts] = counts_per_cell
             adata._inplace_subset_obs(cell_subset)
             counts_per_cell=counts_per_cell[cell_subset]
-        normalize_per_cell(adata.X, counts_per_cell_after, counts_per_cell)
+        adata.X = _normalize_per_cell(adata.X, counts_per_cell_after, counts_per_cell)
 
         layers = adata.layers.keys() if layers == 'all' else layers
         if use_rep == 'after':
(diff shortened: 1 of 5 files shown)


scanpy/preprocessing/_simple.py
```diff
@@ -515,8 +515,27 @@ def normalize_per_cell(
...
+        )
+
+
+def _normalize_per_cell(
+    X: Union[np.ndarray, spmatrix],
+    counts_per_cell_after: Optional[float] = None,
+    counts_per_cell: Optional[np.ndarray] = None,
+    copy: bool = False,
+    min_counts: int = 1,
+):
```
flying-sheep, line 534:
```suggestion
) -> Union[np.ndarray, spmatrix]:
```

flying-sheep:
is there a followup of this PR?

flying-sheep:
Since anndata 0.10 is not yet here, let’s close this and revisit once it’s here

Issue #1439, opened 2020-10-03
Scanpy==1.6 cannot be imported for anndata<0.7.4


WeilerP:
## Description

`scanpy==1.6` cannot be used with `anndata<=0.7.3` due to an import error in `scanpy/__init__.py` as `concat` cannot be imported from `anndata`.

### Minimal code sample (that we can copy&paste without having any data)

```python
import scanpy as sc
```

```pytb
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/anaconda3/envs/scanpy_bug/lib/python3.8/site-packages/scanpy/__init__.py", line 41, in <module>
    from anndata import AnnData, concat
ImportError: cannot import name 'concat' from 'anndata' (/opt/anaconda3/envs/scanpy_bug/lib/python3.8/site-packages/anndata/__init__.py)
```

#### Versions

<details>

```
-----
anndata     0.7.3
scanpy      1.6.0
sinfo       0.3.1
-----
anndata             0.7.3
cycler              0.10.0
cython_runtime      NA
dateutil            2.8.1
get_version         2.1
h5py                2.10.0
igraph              0.8.2
joblib              0.15.1
kiwisolver          1.2.0
legacy_api_wrap     1.2
leidenalg           0.8.2
llvmlite            0.33.0
louvain             0.7.0
matplotlib          3.2.1
mpl_toolkits        NA
natsort             7.0.1
numba               0.50.0
numexpr             2.7.1
numpy               1.18.5
packaging           20.4
pandas              1.1.2
pkg_resources       NA
pyparsing           2.4.7
pytz                2020.1
scanpy              1.6.0
scipy               1.4.1
setuptools_scm      NA
sinfo               0.3.1
six                 1.15.0
sklearn             0.23.1
tables              3.6.1
texttable           1.6.2
wcwidth             0.2.4
yaml                5.3.1
-----
Python 3.8.3 (default, May 19 2020, 13:54:14) [Clang 10.0.0 ]
macOS-10.15.7-x86_64-i386-64bit
12 logical CPU cores, i386
-----
Session information updated at 2020-10-03 14:22
```

</details>

jhrdt:
Hej,

I stumbled upon your issue.

Test for my PR #1440:

```
python3 -m venv venv
source venv/bin/activate
pip install -e . 
pip install "anndata<=0.7.3"
python3 -c "import scanpy as sc"
```

WeilerP:
Thank you for your PR!

[...]

PR #1440, opened 2020-10-03
Import concat from absolute path, if import fails

jhrdt:
Anndata's __init__.py is a shortcut, a full path
avoids the issue.

This PR uses the full path as fallback if the
shortcut fails.

Issue: #1439

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-anndata>=0.7
+anndata>=0.7.2
 # numpy needs a version due to #1320
 numpy>=1.17.0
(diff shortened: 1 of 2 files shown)


WeilerP:
Thanks! This will, however, only work for `anndata>=0.7.2a1`. So you'd also have to update `scanpy/requirements.txt`. I left some more comments in #1439.

ivirshup:
Thanks for the contribution! For the reasons outlined in the issue, and because there are a number of important bug fixes related to h5py in newer versions of anndata, I'd like to stick with the higher version restriction as the solution to this. Closing as superseded by #1491

PR #1456, opened 2020-10-15
fixes type bug, now tagging exactly n_top_genes

VolkerBergen:
Due to type conversion, often 1999 genes gets tagged as 'highly_variable' even though n_top_genes is set to 2000. 
Now, this minor change guarantees that exactly (!) n_top_genes get tagged.

### DIFF
--- a/scanpy/preprocessing/_highly_variable_genes.py
+++ b/scanpy/preprocessing/_highly_variable_genes.py
@@ -256,7 +256,7 @@ def _highly_variable_genes_single_batch(
             ) / disp_mad_bin[df['mean_bin'].values].values
     else:
         raise ValueError('`flavor` needs to be "seurat" or "cell_ranger"')
-    dispersion_norm = df['dispersions_norm'].values.astype('float32')
+    dispersion_norm = df['dispersions_norm'].values
     if n_top_genes is not None:
         dispersion_norm = dispersion_norm[~np.isnan(dispersion_norm)]
         dispersion_norm[


LuckyMD:
You think this has to do with precision of the dispersion value? Couldn't it be that you just have exactly the same value as you have the same count distribution for 2 genes?

VolkerBergen:
It's due to precision - and has also been reported multiple times for scvelo.

VolkerBergen:
For instance, scvelo's pancreas example:
```
adata = scv.datasets.pancreas()
scv.pp.filter_and_normalize(adata, min_shared_counts=30)
sc.pp.highly_variable_genes(bdata, n_top_genes=2000)
adata.obs['highly_variable'].value_counts()
```
returns 1999. Having two genes with the very same count distributions exactly at the n_top_genes threshold, is theoretically possible, but practically as likely as lotterie jackpot.

LuckyMD:
I've seen this several times as well, but I'm surprised that float32 isn't already sufficient for keeping dispersion values apart... as you should get at least 5 decimal places of precision. I would have thought that's already sufficient to make this very unlikely.

VolkerBergen:
Simply leaving this out, solves it and guarantees exactness. Imho, there's no reason to do this type conversion. It's, the threshold value, thus just a single value that gets converted to float32, which won't yield any considerable speedup.

VolkerBergen:
any objections? Otherwise, we can move forward.

PR #1465, opened 2020-10-21
`groupby` is converted to a list before adding as dendrogram key

mukundvarma:
Right now, if you specify `groupby` to `sc.tl.dendrogram`, but not `dendrogram_key`, storage (and subsequent retrieval) from `adata.uns` is messed up because `groupby` is converted from `str` --> `list` during computation.

To give a more concrete example, if my `groupby` variable was "cell_subtype", I would expect it to be stored in `adata.uns` as "dendrogram_cell_subtype". However, because of the list conversion it's stored as "dendrogram_['cell_subtype']" (shown below)

image

This PR attempts to address that. I'm not sure if this is the way you want to go about fixing it, but it's one option.

Thanks for the great package!

### DIFF
--- a/scanpy/tools/_dendrogram.py
+++ b/scanpy/tools/_dendrogram.py
@@ -156,10 +156,7 @@ def dendrogram(
 
     if inplace:
         if key_added is None:
-            if len(groupby) == 1:
-                key_added = f'dendrogram_{groupby[0]}'
-            else:
-                key_added = f'dendrogram_{"_".join(groupby)}'
+            key_added = f'dendrogram_{"_".join(groupby)}'
         logg.info(f'Storing dendrogram info using `.uns[{key_added!r}]`')
         adata.uns[key_added] = dat
     else:
--- a/scanpy/tools/_dendrogram.py
+++ b/scanpy/tools/_dendrogram.py
@@ -156,7 +156,10 @@ def dendrogram(
 
     if inplace:
         if key_added is None:
-            key_added = f'dendrogram_{groupby}'
+            if len(groupby) == 1:
+                key_added = f'dendrogram_{groupby[0]}'
+            else:
+                key_added = f'{dendropgram_{"_".join(groupby)}'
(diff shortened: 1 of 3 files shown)


scanpy/tools/_dendrogram.py
```diff
@@ -156,7 +156,10 @@ def dendrogram(
 
     if inplace:
         if key_added is None:
-            key_added = f'dendrogram_{groupby}'
+            if len(groupby) == 1:
+                key_added = f'dendrogram_{groupby[0]}'
+            else:
+                key_added = f'dendropgram_{"_".join(groupby)}'
```
gokceneraslan: dendropgram has a typo.

gokceneraslan: Is groupby with more than one element possible?

mukundvarma: Oops, fixed the typo. That's what I get for editing code in a browser. I'll try to add a test for this.

I think the `sc.tl.dendrogram` function allows you to pass a list of groupby variables. I haven't used it myself, but I imagine there's a use case for it?

mukundvarma: I've added a test for this now, and confirmed locally that the test passes.

scanpy/tools/_dendrogram.py
```diff
@@ -156,7 +156,10 @@ def dendrogram(
 
     if inplace:
         if key_added is None:
-            key_added = f'dendrogram_{groupby}'
+            if len(groupby) == 1:
+                key_added = f'dendrogram_{groupby[0]}'
+            else:
+                key_added = f'dendrogram_{"_".join(groupby)}'
```
gokceneraslan: This would be just fine I think:

```suggestion
            key_added = f'dendrogram_{"_".join(groupby)}'
```

gokceneraslan (changes requested):

gokceneraslan:
Thanks, looks good.

Issue #1445, opened 2020-10-05
Variance explained annotation for scanpy.pp.pca


bfurtwa:
- [x] Additional function parameters / changed functionality / changed defaults?
- [ ] New analysis tool: A simple analysis tool you have been using and are missing in `sc.tools`?
- [ ] New plotting function: A kind of plot you would like to seein `sc.pl`?
- [ ] External tools: Do you know an existing package that should go into `sc.external.*`?
- [ ] Other?

Would it be possible to add a parameter 'annotate_var_explained' = True/False to the function scanpy.pp.pca to annotate the x- and y-axis of the PCA plot with the variance explained by the component? E.g. x-axis label = PC1 (29%) and y-axis label = PC2 (3%)

giovp:
similar to #1211

giovp:
Hi @bfurtwa ,

would something like this be useful for your question? https://scanpy.readthedocs.io/en/stable/api/scanpy.pl.pca_overview.html#scanpy.pl.pca_overview
https://scanpy.readthedocs.io/en/stable/api/scanpy.pl.pca_variance_ratio.html#scanpy.pl.pca_variance_ratio
?

bfurtwa:
Hi @giovp ,

although these plots contain very similar information, I specifically need the variance explained annotation in the scatter plot.

giovp:
Have you thought about just adding it to the axis like this?

```python
plt.xlabel(f"PC var: {adata.uns["pca"]["variance_ratio"][0]}")
```

Any scanpy function can return a matplotlib axis, that could be further manipulated, checkout the `return_axis` argument of e.g. pca plotting function https://scanpy.readthedocs.io/en/stable/api/scanpy.pl.pca.html

giovp:
I don't think we'll prioritize adding such functionality, also because it doesn't really translate to other types of embedding. But if you think it's really necessary feel free to open a PR, I'd be happy to help with that!

PR #1470, opened 2020-10-26
Added annotate_var_explained parameter to pl.pca

bfurtwa:
Passing annotate_var_explained to sc.pl.pca() adds the explained variance per component into the axis labels.
#1445

### DIFF
--- a/scanpy/plotting/_tools/scatterplots.py
+++ b/scanpy/plotting/_tools/scatterplots.py
@@ -90,7 +90,6 @@ def embedding(
     hspace: float = 0.25,
     wspace: Optional[float] = None,
     title: Union[str, Sequence[str], None] = None,
-    annotate_var_explained: bool = False,
     show: Optional[bool] = None,
     save: Union[bool, str, None] = None,
     ax: Optional[Axes] = None,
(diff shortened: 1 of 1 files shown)


giovp:
Thanks a lot for submitting a PR!!!
Will add as reviewer core dev @fidelram to see what he thinks. It's a bit an ad-hoc since it's an additional argument just for this function.

Meanwhile, for failing tests, can you reformat with black? That seems the reason why tests are failing.
```bash
pip install black #if you don't have it installed
black scanpy/plotting/_tools/scatterplots.py
```
or you can do the same from within whatever IDE you work with.

Thank you!

fidelram:
Thanks for the PR. 

I am not a big fan of adding more and more parameters to generic functions like `embedding`. Rather, would be better if this code is handled in the `pca` function itself. 

In other words, could be possible to add the code to annotate the axis labels in the `pca` function and not in the `embedding` function?

bfurtwa:
I could change the labels in the `fig` or `axs` objects in `pca `returned from `embedding`, but I think it would break the `savefig_or_show `feature as this saves the figure inside `embedding`, before returning to `pca`.

fidelram:
Two ideas:  
 1.  pass to `embedding` `show=False, save=False` and then call `save_fig_or_show` within `pca` with the user values for `show` and `save`.
 2.  allow `embedding` to rename any axis using a parameter like `component_labels`. Thus, this feature is not specific to PCA but also available to any other embedding.

PR #1473, opened 2020-10-28
update datasets

### LINKED ISSUE #1475: Update function _download_visium_dataset
I noticed that 10x has released some new spatial gene expression datasets. Could you include them? It should be a simple change in this line [https://github.com/theislab/scanpy/blob/ab9247bdf8b7a3decc34a15b26fec813ea8fba0d/scanpy/datasets/_datasets.py#L323](url).

Also, I've encounter errors when using `scanpy.datasets.visium_sge`. It seems that the url is outdated. The link to the datasets is changed to be
`
https://support.10xgenomics.com/spatial-gene-expression/datasets/{version_id}/{sample_id}
`

Thank you!

giovp:
Update `sc.datasets.visium_sge` to include new datasets from 10x genomics and account for spaceranger version.

### DIFF
--- a/scanpy/datasets/_datasets.py
+++ b/scanpy/datasets/_datasets.py
@@ -306,7 +306,9 @@ def pbmc3k_processed() -> AnnData:
         )
 
 
-def _download_visium_dataset(sample_id: str, base_dir: Optional[Path] = None):
+def _download_visium_dataset(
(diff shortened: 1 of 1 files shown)


giovp:
this is a really minor thing, can I merge @Koncopd ?

PR #1476, opened 2020-10-30
Add Scrublet as an external tool

pinin4fjords:
This is a PR attempting to address https://github.com/theislab/scanpy/issues/173, and building on the work of @swolock in his fork (see https://github.com/theislab/scanpy/compare/master...swolock:master). 

I've taken a different approach to that fork:

- Various component functions wrapped individually (core Scrublet functionality, doublet simulation), to allow for custom workflows with different Scanpy preprocessing where required.
- ... but a scrub_doublets() function that does pre-processing analogously with Scrublet (but using Scanpy calls) is what imagine most people will use. 
- I've moved what was sensible to use Scanpy functions.  
- ... but some parts are left alone. In particular the PCA seems like it's used a bit differently in Scrublet, fitting the model to the observed transcriptomes (.fit()) and then applying dimensionality reductions to both observed and simulated matrices (.transform()). That's complex to implement with Scanpy functions (which just uses fit_transform()), so thought I'd leave that in Scrublet (which calls the same libraries anyway). Similarly the neighbour graph generation done in the classifier is quite wired in, so I haven't tried to pull it into Scanpy.

I hope that this is a low-maintenance solution, allowing use of Scrublet with Scanpy functions, without a large amount of re-implementation which might cause drift wrt Scrublet itself. 

It does seem to work, but happy to make any changes you or @swolock want- would just be really useful for us to have this in Scanpy :-).


scanpy/external/pp/_scrublet.py
```diff
@@ -0,0 +1,460 @@
...
+        Rate for sampling UMIs when creating synthetic doublets. If 1.0, each
+        doublet is created by simply adding the UMI counts from two randomly
+        sampled observed transcriptomes. For values less than 1, the UMI counts
+        are added and then randomly sampled at the specified rate. 
+    knn_dist_metric : str, optional (default: 'euclidean')
+        Distance metric used when finding nearest neighbors. For list of
+        valid values, see the documentation for annoy (if `use_approx_neighbors`
+        is True) or sklearn.neighbors.NearestNeighbors (if `use_approx_neighbors`
+        is False).
+    normalize_variance : bool, optional (default: True)
```
giovp: can you remove the colon and typing everywhere and try again? There should be the need to put them here as they are already present in the arguments maybe?

pinin4fjords: Ahh yes, thank you. I was copying from https://github.com/swolock/scrublet/blob/master/src/scrublet/scrublet.py I think, but will try taking those out.

pinin4fjords: Hmm, no change @giovp

giovp: yes saw it, need to ask seniors then cause I'm also lost

giovp:
@Koncopd @ivirshup can you have a look at this? CI fails for some weird docs error that we can't nail down.

scanpy/external/pp/_scrublet.py
```diff
@@ -0,0 +1,511 @@
...
+    Union[AnnData, Tuple[AnnData, AnnData], Tuple[AnnData, AnnData, AnnData]]
+]:
+    """\
+    Predict doublets using Scrublet [Wolock2019]_.
+
+    Predict cell doublets using a nearest-neighbor classifier of observed
+    transcriptomes and simulated doublets. Works best if the input is a raw
+    (unnormalized) counts matrix from a single sample or a collection of
+    similar samples from the same experiment.
+
```
ivirshup:
```suggestion

```

I think this is what's causing the test failures. I'm not sure that I think the test is looking for the right thing, it's just finding that there's a line that starts with 3 spaces here. There's some other white space in the docs that can get cleaned up too.

scanpy/external/pp/_scrublet.py
```diff
@@ -0,0 +1,511 @@
...
+    adata_sim = AnnData(scrub._E_sim)
+    adata_sim.obs['n_counts'] = scrub._total_counts_sim
+    adata_sim.obsm['doublet_parents'] = scrub.doublet_parents_
+    adata_sim.uns['scrublet'] = {}
+    adata_sim.uns['scrublet']['parameters'] = {}
+    adata_sim.uns['scrublet']['parameters']['sim_doublet_ratio'] = sim_doublet_ratio
+    return adata_sim
+
+
+def plot_histogram(
```
ivirshup: If this is going to be exported from a high level, could it get a name that ties it to scrublet? Maybe `sce.pl.scrublet_score_distribution`?

scanpy/external/pp/_scrublet.py
```diff
@@ -0,0 +1,511 @@
...
+
+    if get_doublet_neighbor_parents:
+        adata_obs.uns['scrublet'][
+            'doublet_neighbor_parents'
+        ] = scrub.doublet_neighbor_parents_
+
+    return adata_obs
+
+
+def simulate_doublets(
```
ivirshup: If this function is going to be exported, could its name tie it to scrublet?

ivirshup (changes requested):
Looks promising!

A couple requests:

### Tests

Could this get some tests? They should cover any logic that's included here, as well as making sure this doesn't throw errors.

### Docs/ linking functions

Since there are multiple `scrublet` based functions, could they link to each other via ["See Also" sections](https://numpydoc.readthedocs.io/en/latest/format.html#sections)?

### Naming

I've addressed this in some comments on the code, but you make sure all the `scrublet` specific functions are easily identifiable as such?

I also don't think it's obvious from the names `scrub_doublets` and `scrublet` what the difference between the functions is.

### Number of functions and use cases

I'm open to disagreement here. I'm just trying to think about if there's a more simple API here.

`scrub_doublets(adata, ...)` is essentially `scrublet(adata, simulate_doublets(adata, ...), ...)` right? Could this be collapsed to one function, where the second argument defaults to `None`?

This may be confusing since these functions would have a large amount of arguments, and you'd want to check and make sure that simulation and normalization arguments weren't passed if the simulated data was provided.

I guess it depends on what you think the workflows are here. I see two primary use cases:

1. I just have counts (simple use case)
  * Doublets should be simulated
  * Identical normalization should be applied
  * Doublets should be called from the original
2. I'm providing my own simulated data (advanced use case)
  * I think we can require the user to have done their own normalization, since they're already going for more customization
  * All that needs to be done is doublet calling

scanpy/external/pp/_scrublet.py
pinin4fjords: Now moved to sce.pl and named as suggested.

scanpy/external/pp/_scrublet.py
pinin4fjords: Now renamed as suggested

pinin4fjords:
Thanks for all the good feedback @ivirshup - I'll work on it and re-request review when I'm done.

pinin4fjords:
Okay @ivirshup , think I've addressed your comments:

- old 'scrublet' function now not exposed, has become an internal _scrublet_call_doublets (I like it still being separate, makes the logic easier to read). New sce.pp.scrublet now the main exposed function, with scrublet_simulate_doublets() function available for advanced users.
- plot function moved to scanpy/external/pl.py as scrublet_score_distribution().
- functions linked via 'See also' sections.
- tests added for 'scrublet()' and scrublet_simlulate_doublets().

pinin4fjords:
I checked the new tests locally, can't work out why they're not running in the CI here- hopefully it's obvious to you.

ivirshup:
> old 'scrublet' function now not exposed, has become an internal _scrublet_call_doublets (I like it still being separate, makes the logic easier to read)

Oh, I think I wasn't clear here. I was thinking that there would be three doublet calling functions:

1. Simulate doublets. Receives count anndata, returns simulated doublet count anndata.
2. Given two anndata objects, one source data, one simulated, call doublets in the source data. It's assumed both objects have already been normalized.
3. The full workflow. Takes an AnnData object with count data, simulates doublets, runs normalization on both, and then calls doublets on the source object. Uses the previous two functions as well as the normalization workflow internally.

The simple use case is just to call function 3. The advanced use case is to use function 2, potentially with data from function 1, or generated some other way. The advanced use case also allows you to use your own normalization. By not giving function 2 the ability to normalize, we cut down on arguments, and have more modular functions.

What do you think of that?

scanpy/external/pp/_scrublet.py
```diff
@@ -0,0 +1,487 @@
...
+        adata_obs.uns['scrublet'][
+            'doublet_neighbor_parents'
+        ] = scrub.doublet_neighbor_parents_
+
+    return adata_obs
+
+
+def scrublet_simulate_doublets(
+    adata: AnnData,
+    raw_layer: str = 'raw',
```
ivirshup:
```suggestion
    layer = None,
```

Then:

```python
X = sc.get._get_obs_rep(adata, layer=layer)
```

This gives you arguments that look like other functions in scanpy, and a default error message.

pinin4fjords:
> What do you think of that?

I actually prefer my current layout- it accomplishes your logic without duplication of arguments etc, and from a user perspective I prefer having one main function I can use in the two ways  (one anndata you get 3., Two anndata you get 2). But I'm also very happy for you to tweak things to match your vision!

scanpy/tests/external/test_scrublet.py
```diff
@@ -0,0 +1,51 @@
...
+    Check that scrublet runs and detects some doublets.
+    """
+    adata = sc.datasets.pbmc3k()
+    sce.pp.scrublet(adata)
+
+    errors = []
+
+    # replace assertions by conditions
+    if not set(['predicted_doublet', 'doublet_score']).issubset(adata.obs.columns):
+        errors.append("doublet columns not set in .obs")
```
ivirshup:
```suggestion
assert "predicted_doublet" in adata.obs.columns
assert "doublet_score" in adata.obs.columns
```

scanpy/tests/external/test_scrublet.py
```diff
@@ -0,0 +1,51 @@
...
+
+    if len(adata.obs_names[adata.obs.predicted_doublet]) == 0:
+        errors.append("no doublets identified")
```
ivirshup:
```suggestion
    assert adata.obs["predicted_doublet"].any(), "Expect some doublets to be identified"
```

This is boolean valued, right?

scanpy/tests/external/test_scrublet.py
```diff
@@ -0,0 +1,51 @@
...
+
+    assert not errors, "errors occured in scrublet testing:\n{}".format(
+        "\n".join(errors)
+    )
```
ivirshup:
```suggestion
```

If the above changes are made, this bit can be removed

scanpy/tests/external/test_scrublet.py
```diff
@@ -0,0 +1,51 @@
...
+    sc.pp.filter_genes(adata_obs, min_cells=3)
+    sc.pp.filter_cells(adata_obs, min_genes=3)
+    adata_obs.layers['raw'] = adata_obs.X
+    sc.pp.normalize_total(adata_obs)
+    logged = sc.pp.log1p(adata_obs, copy=True)
+
+    hvg = sc.pp.highly_variable_genes(logged)
+    adata_obs = adata_obs[:, logged.var['highly_variable']]
+
+    adata_sim = sce.pp.scrublet_simulate_doublets(adata_obs, raw_layer='raw')
```
ivirshup: Didn't this `raw_layer` argument get changed?

How is this not causing test failures? I'm a bit worried these tests aren't being run.

ivirshup (changes requested):
I'll have a think about the API. It might be that my ideal api should just be go into the main library as a more flexible option.

Otherwise, I think there are a few parts of the tests that should change. I'm a little worried they are not being run on CI, since there's one bit that should definitely fail. Could you check that?

scanpy/tests/external/test_scrublet.py
pinin4fjords: yup

scanpy/tests/external/test_scrublet.py
pinin4fjords: Argh- sorry. Would probably go through testing anyway, the rogue argument would be ignored and _get_obs_rep() would pick up .X. 

As mentioned above I couldn't work out why the tests aren't running here, was hoping you'd see it. I'll have a dig, but I'm clearly missing something about how this works so I'm not sure I'll be able to fix it by myself.

scanpy/tests/external/test_scrublet.py
```diff
@@ -0,0 +1,47 @@
...
+    Check that scrublet runs and detects some doublets.
+    """
+    adata = sc.datasets.pbmc3k()
+    sce.pp.scrublet(adata)
+
+    errors = []
+
+    # replace assertions by conditions
+assert "predicted_doublet" in adata.obs.columns
+assert "doublet_score" in adata.obs.columns
```
ivirshup:
```suggestion
    assert "predicted_doublet" in adata.obs.columns
    assert "doublet_score" in adata.obs.columns
```

Oops, my bad!

scanpy/tests/external/test_scrublet.py
ivirshup: Oh, sorry, totally missed that message.

ivirshup:
@pinin4fjords the tests aren't running since scrublet isn't actually being installed (I'm surprised the build still worked, apparently this is just a warning?). From the travis logs:

```
203$ pip install .[dev,test,louvain,leiden,magic,scvi,harmony,skmisc,scrublet]
204Processing /home/travis/build/theislab/scanpy
205  Installing build dependencies ... done
206  Getting requirements to build wheel ... done
207    Preparing wheel metadata ... done
208  WARNING: scanpy 0.1.dev67+g3918588 does not provide the extra 'scrublet'
```

You'll need to add a scrublet entry to `extras_require` here: https://github.com/theislab/scanpy/blob/d56d6beacdd951a010bb6a93078db26e1ac904b0/setup.py#L31-L56

ivirshup:
Also sorry about these horrible CI wait times! Hopefully these will be better sometime next week.

scanpy/tests/external/test_scrublet.py
pinin4fjords: Ahh, I should have checked

pinin4fjords:
> @pinin4fjords the tests aren't running since scrublet isn't actually being installed (I'm surprised the build still worked, apparently this is just a warning?). From the travis logs:
> You'll need to add a scrublet entry to `extras_require` here:

Ahh, I see- thank you!

pinin4fjords:
> Also sorry about these horrible CI wait times! Hopefully these will be better sometime next week.

Maybe you're having issues with the new Travis credits system? We've certainly had issues- we're gradually migrating to GitHub actions now.

ivirshup:
> Maybe you're having issues with the new Travis credits system?

That's my assumption. This project is also somehow still on travis-ci.org. We're working on moving to azure.

pinin4fjords:
Tests passing! With exception of silly Black formatting issue I've just corrected.

scanpy/external/pp/_scrublet.py
```diff
@@ -0,0 +1,486 @@
...
+    adata_obs.uns['scrublet']['doublet_scores_sim'] = scrub.doublet_scores_sim_
+    adata_obs.uns['scrublet']['doublet_parents'] = adata_sim.obsm['doublet_parents']
+    adata_obs.uns['scrublet']['parameters'] = {
+        'expected_doublet_rate': expected_doublet_rate,
+        'sim_doublet_ratio': adata_sim.uns['scrublet']['parameters'][
+            'sim_doublet_ratio'
+        ],
+        'n_neighbors': n_neighbors,
+        'random_state': random_state,
+    }
```
ivirshup:
```suggestion
    adata_obs.uns['scrublet'] = {
        'threshold': scrub.threshold_
        'doublet_scores_sim': scrub.doublet_scores_sim_
        'doublet_parents': adata_sim.obsm['doublet_parents']
        'parameters': {
            'expected_doublet_rate': expected_doublet_rate,
            'sim_doublet_ratio': (
                adata_sim.uns
                .get('scrublet', {})
                .get('parameters', {})
                .get('sim_doublet_ratio', None)
            ),
            'n_neighbors': n_neighbors,
            'random_state': random_state,
        }
    }
```

Simplifies the assignment a bit, and allows `adata_sim` to not have `scrublet` parameters

scanpy/external/pp/_scrublet.py
```diff
@@ -0,0 +1,486 @@
...
+        sim_doublet_ratio=sim_doublet_ratio,
+        synthetic_doublet_umi_subsampling=synthetic_doublet_umi_subsampling,
+    )
+
+    adata_sim = AnnData(scrub._E_sim)
+    adata_sim.obs['n_counts'] = scrub._total_counts_sim
+    adata_sim.obsm['doublet_parents'] = scrub.doublet_parents_
+    adata_sim.uns['scrublet'] = {}
+    adata_sim.uns['scrublet']['parameters'] = {}
+    adata_sim.uns['scrublet']['parameters']['sim_doublet_ratio'] = sim_doublet_ratio
```
ivirshup:
```suggestion
    adata_sim.uns['scrublet'] = {'parameters': {'sim_doublet_ratio': sim_doublet_ratio}}
```

scanpy/external/__init__.py
```diff
@@ -34,6 +34,11 @@
    pp.harmony_integrate
    pp.mnn_correct
 
+Multiplet detection
+~~~~~~~~~~~~~~~~~~~
+
+   pp.scrub_doublet
```
ivirshup:
```suggestion
   pp.scrublet
   pp.scrublet_simulate_doublets
   pl.scrublet_score_distribution
```

Looks like this didn't get updated

ivirshup (commented):
Sorry for the long turn around!

A couple minor changes, and a release note is needed, then should be good to go.

ivirshup:
@pinin4fjords, just a couple minor changes requested. I'd add them myself, but it looks like I don't have permissions to push to this branch on your repo.

pinin4fjords:
Thanks @ivirshup ! I'm on xmas leave, but I've merged your changes without looking too carefully! Have also added you as collaborator on our fork so you can make further changes if required.

ivirshup:
Enjoy your Holidays!

No worries! I'm still having some trouble pushing to this, so I'm just going to merge it through a separate PR that points to this one. Thanks for working on this!

PR #1483, opened 2020-11-03
Addresses hashsolo pre_existing_cluster bugs

njbernstein:
Also added a check to make sure the number of noise barcodes is less than the number of sample barcodes.

Fixed the example in the doc string to include the cell_hashing_columns argument


njbernstein:
https://github.com/theislab/scanpy/issues/1482

adamgayoso:
@njbernstein I had also noticed hashsolo wasn't added to the documentation, which you can do by adding `pp.hashsolo` [here](https://github.com/theislab/scanpy/blob/master/scanpy/external/__init__.py). Though I'm unsure which heading it belongs under.

njbernstein:
@adamgayoso  I don't think it fits under the other preprocessing tool headings of Data integration or Imputation.

Maybe add a new one called Call hashing or Sample demultiplexing. @fidelram thoughts? Not sure who else to tag

adamgayoso:
@ivirshup

ivirshup:
I think a new section would good here, maybe "Doublet detection"?

njbernstein:
@ivirshup I think doublet detection is a little too narrow in that we can also associate cells to the original sample they came from. Although many folks use cell hashing solely for the purpose of doublet detection. I'd prefer "Sample demultiplexing".

ivirshup:
I'd be happy with whatever, but it needs to go somewhere!

How about "Demultiplexing/ Doublet detection", so it can go with scrublet (#1476) in the next release, and potentially get split out later if more demultiplexing methods are added?

njbernstein:
Any advice for getting the docs to build? The link above doesn't give any details.

scanpy/external/__init__.py
```diff
@@ -34,6 +34,13 @@
    pp.harmony_integrate
    pp.mnn_correct
 
+Sample demultiplexing / Doublet detection
+~~~~~~~~~~~~~~~~
+.. autosummary::
```
adamgayoso, line 40: add line between line 38 and 39

scanpy/external/__init__.py
```diff
@@ -34,6 +34,13 @@
    pp.harmony_integrate
    pp.mnn_correct
 
+Sample demultiplexing / Doublet detection
+~~~~~~~~~~~~~~~~
```
adamgayoso: the tildes should extend to the end of doublet detection

adamgayoso (commented):
My comments will hopefully fix the docs

njbernstein:
mm still not going. @adamgayoso sorry to bug ya again. or @ivirshup @fidelram is there a way to see why the docs are failing?

ivirshup:
Could you rebase on master? I can't see any info on the docs building since there are conflicts.

Can you build the docs locally?

njbernstein:
The docs build locally just fine. There are some warnings but it looks like they are from across the repo. Not sure why they are not building up here. The issue before was that I needed to remove typing from my docstrings. @adamgayoso @ivirshup

docs/api/scanpy.plotting.rst
```diff
@@ -13,4 +13,11 @@ scanpy.plotting
...
-   
\ No newline at end of file
+   
+
+   
+   
+   
+
+
+
```
ivirshup, line 23:
```suggestion
```

scanpy/external/pp/_hashsolo.py
```diff
@@ -255,30 +252,30 @@ def hashsolo(
...
         for example leiden or cell types, not batches though
-    number_of_noise_barcodes : int,
+    number_of_noise_barcodes
         Use this if you wish change the number of barcodes used to create the
         noise distribution. The default is number of cell hashes - 2.
-    inplace : bool
+    inplace
         To do operation in place
     Returns
     -------
```
ivirshup:
```suggestion
        To do operation in place

    Returns
    -------
```

scanpy/external/pp/_hashsolo.py
```diff
@@ -255,30 +252,30 @@ def hashsolo(
         More information and bug reports `here <https://github.com/calico/solo>`__.
     Parameters
```
ivirshup:
```suggestion
        More information and bug reports `here <https://github.com/calico/solo>`__.

    Parameters
```

ivirshup (changes requested):
Hey @njbernstein, it looks like headings want an empty line ahead of them

njbernstein:
finally! Thank you @ivirshup

ivirshup (approved):
Great, thanks for fixing that!

PR #1490, opened 2020-11-11
Require a minimum groupsize for marker detection to prevent division by zero errors

pinin4fjords:
Currently, if a set of cell groups has any groups with only one cell, attempting to run rank_genes_groups() gets you an error like:

```
>>> sc.tl.rank_genes_groups(ad, 'louvain_resolution_3.0')
WARNING: Default of the method has been changed to 't-test' from 't-test_overestim_var'
ranking genes
    consider 'louvain_resolution_3.0' groups:
    with sizes: [28 13 13 11 10  9  9  8  8  8  8  7  6  6  6  4  3  3  1  1  1]
  File "/path/to/scanpy/tools/_rank_genes_groups.py", line 584, in rank_genes_groups
    method, corr_method, n_genes_user, rankby_abs, tie_correct, **kwds
  File "/path/to/scanpy/tools/_rank_genes_groups.py", line 365, in compute_statistics
    for group_index, scores, pvals in generate_test_results:
  File "/path/to/scanpy/tools/_rank_genes_groups.py", line 187, in t_test
    self._basic_stats()
  File "/path/to/scanpy/tools/_rank_genes_groups.py", line 172, in _basic_stats
    self.means[imask], self.vars[imask] = _get_mean_var(X_mask)
  File "/path/to/scanpy/preprocessing/_utils.py", line 14, in _get_mean_var
    var *= X.shape[axis] / (X.shape[axis] - 1)
ZeroDivisionError: division by zero
```

The fix I've come up with is to filter groups by size when calling select_groups(), happy to help on alternate approaches if required.

### DIFF
--- a/scanpy/tests/test_rank_genes_groups.py
+++ b/scanpy/tests/test_rank_genes_groups.py
@@ -139,13 +139,13 @@ def test_results_layers():
     for name in true_scores_t_test.dtype.names:
         assert not np.allclose(true_scores_t_test[name][:7], adata.uns['rank_genes_groups']['scores'][name][:7])
 
-@pytest.mark.xfail
 def test_singlets():
     pbmc = pbmc68k_reduced()
     pbmc.obs['louvain'] = pbmc.obs['louvain'].cat.add_categories(['11'])
     pbmc.obs['louvain'][0] = '11' 
     
-    rank_genes_groups(pbmc, groupby = 'louvain')
+    with pytest.raises(ValueError, match=rf"Could not calculate statistics.*{'11'}"):
+        rank_genes_groups(pbmc, groupby = 'louvain')
 
 @pytest.mark.xfail
 def test_emptycat():
--- a/scanpy/tests/test_rank_genes_groups.py
+++ b/scanpy/tests/test_rank_genes_groups.py
@@ -139,6 +139,20 @@ def test_results_layers():
     for name in true_scores_t_test.dtype.names:
         assert not np.allclose(true_scores_t_test[name][:7], adata.uns['rank_genes_groups']['scores'][name][:7])
 
+@pytest.mark.xfail
+def test_singlets():
+    pbmc = pbmc68k_reduced()
+    pbmc.obs['louvain'] = pbmc.obs['louvain'].cat.add_categories(['11'])
+    pbmc.obs['louvain'][0] = '11' 
+    
+    rank_genes_groups(pbmc, groupby = 'louvain')
+
+@pytest.mark.xfail
+def test_emptycat():
+    pbmc = pbmc68k_reduced()
+    pbmc.obs['louvain'] = pbmc.obs['louvain'].cat.add_categories(['11'])
+    
+    rank_genes_groups(pbmc, groupby = 'louvain')
 
 def test_wilcoxon_symmetry():
     pbmc = pbmc68k_reduced()
--- a/scanpy/tests/test_rank_genes_groups.py
+++ b/scanpy/tests/test_rank_genes_groups.py
@@ -147,12 +147,12 @@ def test_singlets():
     with pytest.raises(ValueError, match=rf"Could not calculate statistics.*{'11'}"):
         rank_genes_groups(pbmc, groupby = 'louvain')
 
-@pytest.mark.xfail
 def test_emptycat():
     pbmc = pbmc68k_reduced()
     pbmc.obs['louvain'] = pbmc.obs['louvain'].cat.add_categories(['11'])
-    
-    rank_genes_groups(pbmc, groupby = 'louvain')
+
+    with pytest.raises(ValueError, match=rf"Could not calculate statistics.*{'11'}"):
+        rank_genes_groups(pbmc, groupby = 'louvain')
 
 def test_wilcoxon_symmetry():
     pbmc = pbmc68k_reduced()
--- a/scanpy/_utils.py
+++ b/scanpy/_utils.py
@@ -518,7 +518,6 @@ def select_groups(adata, groups_order_subset='all', key='groups', min_groupsize 
         groups_order_subset = adata.obs[key].cat.categories[groups_ids].values
     else:
         groups_counts = adata.obs[key].value_counts()
-        valid_groups  = groups_counts.index[groups_counts >= min_groupsize]
         valid_groups = [ item in groups_counts.index[groups_counts >= min_groupsize] for item in groups_order ]
         groups_order_subset = groups_order[ valid_groups ]
         groups_masks = groups_masks[ valid_groups ]
--- a/scanpy/_utils.py
+++ b/scanpy/_utils.py
@@ -475,7 +475,7 @@ def check_nonnegative_integers(X: Union[np.ndarray, sparse.spmatrix]):
         return True
 
 
-def select_groups(adata, groups_order_subset='all', key='groups', min_groupsize = 1):
+def select_groups(adata, groups_order_subset='all', key='groups'):
     """Get subset of groups in adata.obs[key].
     """
     groups_order = adata.obs[key].cat.categories
@@ -507,21 +507,18 @@ def select_groups(adata, groups_order_subset='all', key='groups', min_groupsize 
                     np.array(groups_order_subset),
                 )
             )[0]
-        if len(groups_ids) < min_groupsize:
+        if len(groups_ids) == 0:
             logg.debug(
                 f'{np.array(groups_order_subset)} invalid! specify valid '
                 f'groups_order (or indices) from {adata.obs[key].cat.categories}',
             )
             from sys import exit
+
             exit(0)
         groups_masks = groups_masks[groups_ids]
         groups_order_subset = adata.obs[key].cat.categories[groups_ids].values
     else:
-        groups_counts = adata.obs[key].value_counts()
-        valid_groups  = groups_counts.index[groups_counts >= min_groupsize]
-        valid_groups = [ item in groups_counts.index[groups_counts >= min_groupsize] for item in groups_order ]
-        groups_order_subset = groups_order[ valid_groups ]
-        groups_masks = groups_masks[ valid_groups ]
+        groups_order_subset = groups_order.values
     return groups_order_subset, groups_masks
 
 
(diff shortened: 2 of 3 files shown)


ivirshup:
My preference would be to throw a more informative error. Something like: "Could not calculate statistics for group {group_name} since it only contains one sample.". I don't like that groups would be implicitly excluded from the results. In general I would expect each category of `adata.obs["cat"]` to be in the results of `sc.tl.rank_genes_groups(adata, "cat")`. If they can implicitly be excluded, I think that will lead to more confusing downstream behavior.

Does that sound reasonable @pinin4fjords, @Koncopd?

pinin4fjords:
So for our workflows, where we pass the clustering step directly to the marker detection, it would be good to be able to get markers for non-singlet groups when we specify 'all', even where there are singlet groups. We can add a workaround to our CLI layer and in fact I'm doing that anyway since we need a fix quickly https://github.com/ebi-gene-expression-group/scanpy-scripts/pull/88, but there I'm having to transiently edit .obs to remove the singlet clusters which seems messier (and we're not sure if that will have unintended consequences).

ivirshup:
We should probably let users provide a set of clusters for the reference. 

Until then, could you do:

```python
groups_to_test = (
    adata.obs["clusters"]
    .value_counts()
    .loc[lambda x: x > 1]
    .index
)
subset_adata = adata[adata.obs["clusters"].isin(groups_to_test)].copy()
sc.tl.rank_genes_groups(subset_adata, ...)
adata.uns["rank_genes_groups"] = subset_adata.uns["rank_genes_groups"]
```

Or, if you want the singlets in the reference:

```python
sc.tl.rank_genes_groups(adata, groups=list(groups_to_test), ...)
```

BTW, about the h5py bytes thing, things written as strings are read as strings with h5py 3 as of anndata 0.7.5, which just got released.

ivirshup:
Looking at your code a bit more, does the `groups` argument work for your issue?

pinin4fjords:
> Looking at your code a bit more, does the `groups` argument work for your issue?

Yep, should do as well- I'll try

pinin4fjords:
Thanks @ivirshup that worked nicely, I should have thought of it. I'll remove our h5py pin once anndata 0.7.5 is available on Conda (have you seen that it's failing right now? https://github.com/conda-forge/anndata-feedstock/pull/13)

ivirshup:
Great!

I'm not sure what's going on with that conda build, but hopefully one of the maintainers there will know what to do.

On the topic of this PR, I do think this case should give a more helpful error message. Would you like to do that?

pinin4fjords:
> Great!
> 
> I'm not sure what's going on with that conda build, but hopefully one of the maintainers there will know what to do.
> 
> On the topic of this PR, I do think this case should give a more helpful error message. Would you like to do that?

Sure- there you go. I reverted the above and just raised an error.

ivirshup:
Great! Could you also add a test to make sure the error is being thrown? And does this handle cases where there are categories with no entries?

pinin4fjords:
> Great! Could you also add a test to make sure the error is being thrown? And does this handle cases where there are categories with no entries?

Sorry yep tests added. And yep, the value_counts() will also catch empty categories (though added a test for that too).

scanpy/tests/test_rank_genes_groups.py
```diff
@@ -139,6 +139,20 @@ def test_results_layers():
     for name in true_scores_t_test.dtype.names:
         assert not np.allclose(true_scores_t_test[name][:7], adata.uns['rank_genes_groups']['scores'][name][:7])
 
+@pytest.mark.xfail
+def test_singlets():
+    pbmc = pbmc68k_reduced()
+    pbmc.obs['louvain'] = pbmc.obs['louvain'].cat.add_categories(['11'])
+    pbmc.obs['louvain'][0] = '11' 
+    
+    rank_genes_groups(pbmc, groupby = 'louvain')
```
ivirshup:
```suggestion
def test_singlets():
    pbmc = pbmc68k_reduced()
    pbmc.obs['louvain'] = pbmc.obs['louvain'].cat.add_categories(['11'])
    pbmc.obs['louvain'][0] = '11' 
    
    with pytest.raises(ValueError, match=rf"Could not calculate statistics.*{'11'}"):
        rank_genes_groups(pbmc, groupby = 'louvain')
```

ivirshup: If an `xfail` test passes, I don't think it counts as a failure. This is a more specific way to check for the correct error being thrown, and will definitely fail if the error isn't thrown.

scanpy/tests/test_rank_genes_groups.py
```diff
@@ -139,6 +139,20 @@ def test_results_layers():
...
+
+@pytest.mark.xfail
+def test_emptycat():
+    pbmc = pbmc68k_reduced()
+    pbmc.obs['louvain'] = pbmc.obs['louvain'].cat.add_categories(['11'])
+    
+    rank_genes_groups(pbmc, groupby = 'louvain')
```
ivirshup: I think we implicitly remove categories that have no entries. Is this not the case?

ivirshup (changes requested):
Minor changes to tests, looks good otherwise!

scanpy/tests/test_rank_genes_groups.py
pinin4fjords: I don't think so- we managed to break things by having an empty category https://github.com/ebi-gene-expression-group/scanpy-scripts/pull/88#issuecomment-726001228

pinin4fjords: Though to be fair we did it by editing .obs directly, so it may not be something you need to expect.

ivirshup: Ah, I see. This  looks like a bug to me, since we normally remove unused categories in AnnData objects.

@Koncopd, would you agree with this? That this example should not error and the unused category should be removed?

```python
from itertools import cycle, islice
import pandas as pd
import scanpy as sc

pbmc.obs["cat"] = pd.Categorical(islice(cycle("ab"), pbmc.n_obs), categories=list("abc"))

 sc.tl.rank_genes_groups(pbmc, groupby="cat")
```

```pytb
WARNING: Default of the method has been changed to 't-test' from 't-test_overestim_var'
ZeroDivisionError                         Traceback (most recent call last)
<ipython-input-7-d4efcd9bc643> in <module>
----> 1 sc.tl.rank_genes_groups(pbmc, groupby="cat")

~/github/scanpy/scanpy/tools/_rank_genes_groups.py in rank_genes_groups(adata, groupby, use_raw, groups, reference, n_genes, rankby_abs, pts, key_added, copy, method, corr_method, tie_correct, layer, **kwds)
    583     logg.debug(f'with sizes: {np.count_nonzero(test_obj.groups_masks, axis=1)}')
    584 
--> 585     test_obj.compute_statistics(
    586         method, corr_method, n_genes_user, rankby_abs, tie_correct, **kwds
    587     )

~/github/scanpy/scanpy/tools/_rank_genes_groups.py in compute_statistics(self, method, corr_method, n_genes_user, rankby_abs, tie_correct, **kwds)
    364         n_genes = self.X.shape[1]
    365 
--> 366         for group_index, scores, pvals in generate_test_results:
    367             group_name = str(self.groups_order[group_index])
    368 

~/github/scanpy/scanpy/tools/_rank_genes_groups.py in t_test(self, method)
    186         from scipy import stats
    187 
--> 188         self._basic_stats()
    189 
    190         for group_index, mask in enumerate(self.groups_masks):

~/github/scanpy/scanpy/tools/_rank_genes_groups.py in _basic_stats(self)
    171                 continue
    172 
--> 173             self.means[imask], self.vars[imask] = _get_mean_var(X_mask)
    174 
    175             if self.ireference is None:

~/github/scanpy/scanpy/preprocessing/_utils.py in _get_mean_var(X, axis)
      6 def _get_mean_var(X, *, axis=0):
      7     if sparse.issparse(X):
----> 8         mean, var = sparse_mean_variance_axis(X, axis=axis)
      9     else:
     10         mean = np.mean(X, axis=axis, dtype=np.float64)

~/github/scanpy/scanpy/preprocessing/_utils.py in sparse_mean_variance_axis(mtx, axis)
     40         )
     41     else:
---> 42         return sparse_mean_var_minor_axis(mtx.data, mtx.indices, *shape, np.float64)
     43 
     44 

ZeroDivisionError: division by zero
```

Koncopd: Hm, i missed this discussion. This is another problem, isn't it? An empty group vs a group with only one element.
This happens because of [this](https://github.com/theislab/scanpy/blob/d20547460bc32882c707b21cdbe646ca8f8ce65f/scanpy/_utils.py#L488) i think.
However, shouldn't anndata care about this cases and not the functions which use `.obs`?

ivirshup: If we could intercept things like adding categories, then we could do this in `anndata`, but then we'd have to create a wrapper class for dataframes.

I was wondering if we should just put a `sanitize_anndata` call at the top of `rank_genes_groups`.

scanpy/tests/test_rank_genes_groups.py
```diff
@@ -139,6 +139,20 @@ def test_results_layers():
...
+    
+    with pytest.raises(ValueError, match=rf"Could not calculate statistics.*{'11'}"):
+        rank_genes_groups(pbmc, groupby = 'louvain')
+
+@pytest.mark.xfail
+def test_emptycat():
+    pbmc = pbmc68k_reduced()
+    pbmc.obs['louvain'] = pbmc.obs['louvain'].cat.add_categories(['11'])
+    
+    rank_genes_groups(pbmc, groupby = 'louvain')
```
ivirshup:
```suggestion
def test_emptycat():
    pbmc = pbmc68k_reduced()
    pbmc.obs['louvain'] = pbmc.obs['louvain'].cat.add_categories(['11'])

    with pytest.raises(ValueError, match=rf"Could not calculate statistics.*{'11'}"):
        rank_genes_groups(pbmc, groupby = 'louvain')
```

ivirshup: @pinin4fjords, could you add this? I'd do it myself but git is giving me some trouble with pushing to this PR.

PR #1492, opened 2020-11-12
Ignore blank lines in docstring test

### LINKED ISSUE #1484: PR problem: docstring error
Hi!

I'm having trouble getting the CI to pass on this PR: https://github.com/theislab/scanpy/pull/1476

This seems to be the issue:

```
>           raise SyntaxError(msg, (filename, lineno, 2, text))
E             File "/home/travis/build/theislab/scanpy/scanpy/external/pp/_scrublet.py", line 9
E               >   <
E                ^
E           SyntaxError: Header of function `scanpy.external.pp._scrublet.scrub_doublets`’s docstring should start with one-line description:
E           
E           ␣␣␣␣"""\
E           ␣␣␣␣My one-line␣description.
E           
E           ␣␣␣␣…
E           ␣␣␣␣"""
scanpy/tests/test_docs.py:38: SyntaxError
```

... but the function does have that one-line description. 

Would you be able to suggest to me what I'm doing wrong?

flying-sheep:
Fixes #1484

### DIFF
--- a/scanpy/tests/test_docs.py
+++ b/scanpy/tests/test_docs.py
@@ -24,13 +24,16 @@ def test_function_headers(f):
     broken = [i for i, l in enumerate(lines) if l.strip() and not l.startswith("    ")]
     if any(broken):
         msg = f'''\
-Header of function `{name}`’s docstring should start with one-line description:
+Header of function `{name}`’s docstring should start with one-line description
+and be consistently indented like this:
 
 ␣␣␣␣"""\\
 ␣␣␣␣My one-line␣description.
 
 ␣␣␣␣…
 ␣␣␣␣"""
+
+The displayed line is under-indented.
 '''
         filename = inspect.getsourcefile(f)
         _, lineno = inspect.getsourcelines(f)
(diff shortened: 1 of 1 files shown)


ivirshup:
> Also isn’t it cool that it points exactly to the problematic line?

Currently, I think the line number reported is the number of lines past the `:` in the function definition. It'd be really nice if it could tell you which line number in the file it was (which might be difficult for manipulated doc-strings).

Also, from what the error message says, isn't the `any(broken)` check testing the same thing as assert lines[0], `f"{name} needs a single-line summary"`? Isn't the first one sufficient?

flying-sheep:
It’s using `__orig_doc__` so the line should be correct.

`assert lines[0]` just asserts that the first line is non-empty. `any(broken)` checks if there’s under-indented lines.

ivirshup:
> any(broken) checks if there’s under-indented lines.

Sure, but the error it throws is: ```"Header of function `{name}`’s docstring should start with one-line description:"```, which suggests that wasn't the intent of the check. If it's a check for white space, it should throw an error about white space.

flying-sheep:
Same thing: If the line is under-indented, the first line summary can’t be properly extracted. I’ll make the message more clear.

PR #1493, opened 2020-11-12
Update tests now that 3d plots work

### LINKED ISSUE #849: Unpin matplotlib version once possible
As @fidelram said in https://github.com/theislab/scanpy/pull/661#issuecomment-496144015, we have to wait until matplotlib/matplotlib#14298 is fixed.

It seems to be in matplotlib’s 3.1.2 milestone, so maybe we can just set the dependency to “matplotlib == 3.0.0 or matplotlib >= 3.1.2”

This originally came up in #663, and then later in e.g. #787

ivirshup:
Fixes #849

This PR updates our tests so that they expect 3d plots to work if matplotlib 3.3.3 is installed.

The plots also look a bit different than they used to:

Old:

master_3dprojection

New:

master_3dprojection

It would be nice if the legend was even more visible, but I think I'll leave it for now.


flying-sheep:
Oh wow, that’s been a long time coming!

PR #1494, opened 2020-11-13
Bugfix: Failure due to cugraph api change in v0.16...

rilango:
Method cugraph.add_adj_list is replaced in RAPIDS v0.16 with from_cudf_adjlist.
This patch will check for existence of 'add_adj_list' in the object before
calling it.


ivirshup:
Thanks!

PR #1499, opened 2020-11-18
Speedup sc.get.obs_df

fidelram:
By using array slicing, this codes improves ~10 fold the speed of `sc.get.obs_df()`.

```PYTHON
%timeit sc.get.obs_df(adata, list(adata.var_names[:100]) + ['louvain'])
````
before:
`40.6 ms ± 2.38 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`

after:
`4.45 ms ± 228 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`

### DIFF
--- a/scanpy/get.py
+++ b/scanpy/get.py
@@ -176,6 +176,9 @@ def obs_df(
     if len(var_names) > 0:
         X = _get_obs_rep(adata, layer=layer, use_raw=use_raw)
         matrix = X[:, adata.var_names.get_indexer(var_names)]
+        from scipy.sparse import issparse
+        if issparse(matrix):
+            matrix = matrix.toarray()
         df = df.join(pd.DataFrame(matrix, columns=var_symbol, index=adata.obs.index))
 
     # add obs values
@@ -251,6 +254,10 @@ def var_df(
     if len(obs_names) > 0:
         X = _get_obs_rep(adata, layer=layer, use_raw=use_raw)
         matrix = X[adata.obs_names.get_indexer(obs_names), :]
+        from scipy.sparse import issparse
+        if issparse(matrix):
+            matrix = matrix.toarray()
+
         df = df.join(pd.DataFrame(matrix.T, columns=obs_names, index=adata.var.index))
 
     # add obs values
--- a/scanpy/get.py
+++ b/scanpy/get.py
@@ -180,8 +180,10 @@ def obs_df(
 
     # add obs values
     if len(obs_names) > 0:
-        df.join(adata.obs[obs_names])
+        df = df.join(adata.obs[obs_names])
 
+    # reorder columns to given order
+    df = df[keys]
     for k, idx in obsm_keys:
         added_k = f"{k}-{idx}"
         val = adata.obsm[k]
@@ -200,6 +202,7 @@ def var_df(
     varm_keys: Iterable[Tuple[str, int]] = (),
     *,
     layer: str = None,
+    use_raw: bool = False,
 ) -> pd.DataFrame:
     """\
     Return values for observations in adata.
@@ -214,6 +217,8 @@ def var_df(
         Tuple of `(key from varm, column index of varm[key])`.
     layer
         Layer of `adata` to use as expression values.
+    use_raw
+        Whether to get expression values from `adata.raw`.
 
     Returns
     -------
@@ -221,13 +226,16 @@ def var_df(
     and `varm_keys`.
     """
     # Argument handling
-    lookup_keys = []
+    obs_names = []
+    var_names = []
     not_found = []
     for key in keys:
-        if key in adata.var.columns:
-            lookup_keys.append(key)
-        elif key in adata.obs_names:
-            lookup_keys.append(key)
+        if key in adata.obs_names:
+            obs_names.append(key)
+        elif use_raw and key in adata.raw.var.columns:
+            var_names.append(key)
+        elif key in adata.var.columns:
+            var_names.append(key)
         else:
             not_found.append(key)
     if len(not_found) > 0:
@@ -236,10 +244,25 @@ def var_df(
             " in `adata.obs_names`."
         )
 
-    # Make df
-    df = pd.DataFrame(index=adata.var_names)
-    for k, l in zip(keys, lookup_keys):
-        df[k] = adata.var_vector(l, layer=layer)
+    # initialize df
+    df = pd.DataFrame(index=adata.var.index)
+
+    # add obs values
+    if len(obs_names) > 0:
+        X = _get_obs_rep(adata, layer=layer, use_raw=use_raw)
+        matrix = X[adata.obs_names.get_indexer(obs_names), :]
+        df = df.join(pd.DataFrame(matrix.T, columns=obs_names, index=adata.var.index))
+
+    # add obs values
+    if len(var_names) > 0:
+        if use_raw:
+            df = df.join(adata.raw.var[var_names])
+        else:
+            df = df.join(adata.var[var_names])
+
+    # reorder columns to given order
+    df = df[keys]
+
     for k, idx in varm_keys:
         added_k = f"{k}-{idx}"
         val = adata.varm[k]
--- a/scanpy/get.py
+++ b/scanpy/get.py
@@ -170,20 +170,17 @@ def obs_df(
         )
 
     # Make df
-    df_var = df_obs = None
+    df = pd.DataFrame(index=adata.obs.index)
 
-    # prepare var values
+    # add var values
     if len(var_names) > 0:
         X = _get_obs_rep(adata, layer=layer, use_raw=use_raw)
         matrix = X[:, adata.var_names.get_indexer(var_names)]
-        df_var = pd.DataFrame(matrix, columns=var_symbol, index=adata.obs.index)
+        df = df.join(pd.DataFrame(matrix, columns=var_symbol, index=adata.obs.index))
 
-    # prepare obs values
+    # add obs values
     if len(obs_names) > 0:
-        df_obs = adata.obs[obs_names]
-
-    # concatenate and reorder after given `keys`
-    df = pd.concat([df_obs, df_var], axis=1)[keys]
+        df.join(adata.obs[obs_names])
 
     for k, idx in obsm_keys:
         added_k = f"{k}-{idx}"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: scanpy/tests/test_get.py [hunk 1]
```diff
@@ -27,22 +31,66 @@ def test_obs_df(adata):
...
-    ))
-    assert np.all(np.equal(
-        sc.get.obs_df(adata, keys=["gene2", "obs1"], layer="double"),
-        pd.DataFrame({"gene2": [2, 2], "obs1": [0, 1]}, index=adata.obs_names),
-    ))
-    assert np.all(np.equal(
-        sc.get.obs_df(adata, keys=["raw2", "obs1"], gene_symbols="gene_symbols", use_raw=True),
-        pd.DataFrame({"raw2": [0, 0], "obs1": [0, 1]}, index=adata.obs_names),
-    ))
+    assert np.all(
```
ivirshup: Looking at this again, I think `pd.testing.assert_frame_equal` should be used here. I believe it will give more helpful error messages.

scanpy/get.py
```diff
@@ -167,12 +170,36 @@ def obs_df(
...
+                raise KeyError(
+                    f'Selected layer: {layer} is not in the layers list. '
+                    f'The list of valid layers is: {adata.layers.keys()}'
+                )
+            matrix = adata[:, var_names].layers[layer]
+        elif use_raw:
+            matrix = adata.raw[:, var_names].X
         else:
-            df[k] = adata.raw.obs_vector(l)
+            matrix = adata[:, var_names].X
```
ivirshup: How about:

```python
X = sc.get._get_obs_rep(adata, layer=layer, use_raw=use_raw)
matrix = X[:, adata.var_names.get_indexer(var_names)]
```

scanpy/get.py
```diff
@@ -167,12 +170,36 @@ def obs_df(
         )
 
     # Make df
-    df = pd.DataFrame(index=adata.obs_names)
-    for k, l in zip(keys, lookup_keys):
-        if not use_raw or k in adata.obs.columns:
-            df[k] = adata.obs_vector(l, layer=layer)
+    df_var = df_obs = None
```
ivirshup: I feel like this is a bit weird, I was not expecting `pd.concat` to allow some values to be `None`.

What about initializing the dataframe, then adding columns? So something like:

```python
df = pd.DataFrame(index=adata.obs_names)
...
# then adding columns like:
df[var_names] = matrix
...
df[obs_names] = adata.obs[obs_names]
...
# then reordering
df = df[keys]
```

ivirshup (changes requested):
Good idea! Could you show some of the benchmarks you got those time improvements from?

Could you do the same for `var_df`?

Also, I think might need a few additional tests. I initially used the `.{dim}_vector` functions since I know they return numpy vectors. Can this be tested against `AnnData` objects with sparse `X`, as well as backed objects?

scanpy/get.py
fidelram: Thanks for the tip. Already added.

scanpy/get.py
fidelram: I just tried this second method but is significantly slower:

First time: current, second time: above solution. 

image

ivirshup: That seems really strange.

Could you share the code you ran? I'd like to take a look and maybe open an issue in pandas.

fidelram: using `join` works equally fast. I will keep this solution instead. 

```PYTHON
df = pd.DataFrame(index=adata.obs.index)
... 
df = df.join(pd.DataFrame(matrix, columns=var_symbol, index=adata.obs.index))
```

`2.42 ms ± 195 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)`

ivirshup: Never mind, I've got a small example now.

```python
In [22]: %timeit newdf = df.copy(); newdf[df2.columns] = df2
2.37 ms ± 31.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [23]: %timeit pd.concat([df, df2], axis=1)
215 µs ± 8.33 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

scanpy/tests/test_get.py
fidelram: Done. Here I had some trouble with the dtype pandas testing now checks. I now set the type when creating the Anndata object specifically.

scanpy/get.py
```diff
@@ -167,12 +170,20 @@ def obs_df(
...
-        if not use_raw or k in adata.obs.columns:
-            df[k] = adata.obs_vector(l, layer=layer)
-        else:
-            df[k] = adata.raw.obs_vector(l)
+    df = pd.DataFrame(index=adata.obs.index)
+
+    # add var values
+    if len(var_names) > 0:
+        X = _get_obs_rep(adata, layer=layer, use_raw=use_raw)
+        matrix = X[:, adata.var_names.get_indexer(var_names)]
```
ivirshup: Do you need to convert matrix to a numpy array if it's a sparse array here?

fidelram: needs to be converted to dense if sparse. Just added the code for this.

ivirshup:
About `use_raw` with `sc.get.var_df`: I didn't include this because the semantics differ significantly from `sc.get.obs_df`, as `raw` can have a different number of variables. I think it makes more sense for a user to call `sc.get.var_df(adata.raw, ...)`, since it's much more explicit that `adata.raw.var` and `adata.raw.varm` will be used.

scanpy/get.py
ivirshup: https://github.com/pandas-dev/pandas/issues/37954

fidelram:
I like the ideas of using `adata.raw` when raw data wants to be used in general.  This is an elegant solution to a source of endless confusion.

Regarding `var_df` I can remove the `use_raw`. However, I consider that since this option is everywhere it should be here as well. I find odd that is not present.  I don't see a problem that the size of the dataframe will be different that would be expected.

ivirshup:
It's not just that the length is different, is that `sc.get.obs_df(adata, ["col"], use_raw=x)["col"]` is the same regardless of the value of `x`, but it's different for `var_df`. I think it's easier to build code around functions with more orthogonal arguments.

> However, I consider that since this option is everywhere it should be here as well. 

Could we add an example of `sc.get.var_df(adata.raw, ...)`, leave out `use_raw` for now, and see if anyone complains?

I've been trying to leave out `use_raw` on functions where variable length matters anyways. For example: `adata.var_vector`.

fidelram:
Ok, will remove that as soon as I can

scanpy/get.py
```diff
@@ -227,10 +247,26 @@ def var_df(
...
-    df = pd.DataFrame(index=adata.var_names)
-    for k, l in zip(keys, lookup_keys):
-        df[k] = adata.var_vector(l, layer=layer)
+    # initialize df
+    df = pd.DataFrame(index=adata.var.index)
+
+    # add obs values
+    if len(obs_names) > 0:
+        X = _get_obs_rep(adata, layer=layer)
+        matrix = X[adata.obs_names.get_indexer(obs_names), :]
```
ivirshup: Just had a thought, I don't think this will work if X is a backed dense array and `obs_names` isn't sorted. `h5py.Dataset`s require that the indices be in order. This should probably get a test case.

An alternative is `_get_obs_rep(adata[obs_names], ...).copy()`, but this will have performance issues with `raw`.

ivirshup: Internally in `anndata`, I index in-order then reorder the array for `h5py.Dataset` objects.

fidelram: How can we test for this? I added a test to read a backed dataset from the included datasets. Why this works? or by chance are the indices ordered?

ivirshup: The indices look ordered, since you're getting them like this: `list(adata.var_names[:10])`.

I think `list(adata.var_names[10::-1]) would be enough to make it fail. In `AnnData` valid indices in random order are generated with functions in `anndata/tests/helpers.py`.

scanpy/get.py
```diff
@@ -253,7 +253,7 @@ def var_df(
     # add obs values
     if len(obs_names) > 0:
         X = _get_obs_rep(adata, layer=layer)
-        matrix = X[adata.obs_names.get_indexer(obs_names), :]
+        matrix = X[adata.obs.index.isin(obs_names), :]
```
ivirshup: Will these be in the right order?

Won't the value from `X` be in whatever order the appear, while the values in `obs_names` are in the order they were passed?

ivirshup: This has made me realize some stuff needs to get fixed in `anndata`, mostly around `raw`. In future, I think this should look like `adata[obs_names].to_df(layer=layer)`, but we're not quite there yet.

Here's what you can do to make backed mode work for now:

```python
idxs = adata.obs_names.get_indexer(obs_names)
idxs_order = np.argsort(idxs_order)
matrix = X[idxs[idxs_order], :][np.argsort(idxs_order)]
```

In the next major release of anndata I'm thinking we should export some of the utilities that are used for giving all these indexing operations a consistent interface.

fidelram: Yes. That is why this should be safe for backed mode. Later in the code, before the result is returned the columns are reorder to match the `keys` order.

fidelram:
The previous test failed but is not clear to me why, as it passes the local tests (anndata 0.7.5). It seems that on travis server,  backed slicing requires integer indices and will not work with a boolean vector. I changed to sorted integers hoping that this will solve the issue.

scanpy/tests/test_get.py
```diff
@@ -51,18 +107,94 @@ def test_obs_df(adata):
...
+def test_column_content():
+    "uses a larger dataset to test column order and content"
+    adata = sc.datasets.pbmc68k_reduced()
+
+    # test that columns content is correct for obs_df
+    query = ['CST3', 'NKG7', 'GNLY', 'louvain', 'n_counts', 'n_genes']
+    df = sc.get.obs_df(adata, query)
+    for col in query:
+        assert col in df
+        np.testing.assert_array_equal(query, df.columns)
```
ivirshup, line 145: This shouldn't happen once for every column

fidelram: True. will put it outside.

ivirshup:
My thinking on my change is that I would like all the code that handles backed mode to be cleanly separated. I think this should be handled more cleanly on the `anndata` side, and once that's been done it's easier to replace the backed mode specific code if it's all together.

ivirshup (approved):

PR #1501, opened 2020-11-19
Adjust Combat for control cells

mbuttner:
I wrote this adjustment for Combat for a recent publication (Boettcher et al, 2020): https://www.biorxiv.org/content/10.1101/2020.08.12.247585v1
It allows to compute a ComBat for control cells and apply this to a second set of cells.

### DIFF
--- a/scanpy/preprocessing/_combat.py
+++ b/scanpy/preprocessing/_combat.py
@@ -211,11 +211,7 @@ def combat(
             if key2 in covariates:
                 raise ValueError('Batch key and covariates cannot overlap.')
         # only works on dense matrices so far    
-        if issparse(adata2.X):
-            X = adata2.X.A.T
-        else:
-            X2 = adata2.X.T
-        data2 = pd.DataFrame(data=X2, index=adata2.var_names, columns=adata2.obs_names,)
+        data2 = adata2.to_df().T
 
         sc.utils.sanitize_anndata(adata2)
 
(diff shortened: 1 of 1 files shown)


scanpy/preprocessing/_combat.py
```diff
@@ -180,19 +189,35 @@ def combat(
...
+            raise ValueError('Could not find the key {!r} in adata2.obs'.format(key))
+        if covariates is not None:
+            if key2 in covariates:
+                raise ValueError('Batch key and covariates cannot overlap.')
+        # only works on dense matrices so far    
+        if issparse(adata2.X):
+            X = adata2.X.A.T
+        else:
+            X2 = adata2.X.T
+        data2 = pd.DataFrame(data=X2, index=adata2.var_names, columns=adata2.obs_names,)
```
ivirshup:
```suggestion
        data2 = adata2.to_df().T
```

ivirshup:
Should the reference object where you learn the transformation (currently `adata`) always be a subset of the data you're going to apply the transformation to (`adata2`)? If so, instead of passing a separate object, could there be a mask of which samples to train on?

If not, what do you think about making this a separate function? Maybe `combat_by_reference`?

mbuttner:
> Should the reference object where you learn the transformation always be a subset of the data you're going to apply the transformation to? If so, instead of passing a separate object, could there be a mask of which samples to train on?
> 
> If not, what do you think about making this a separate function? Maybe `combat_by_reference`?

Thank you for your great suggestions. I think it's easier to add a mask for train/evaluate instead of splitting into 2 objects. 
I don't think it should be a separate `combat_by_reference` function, though, because the chance in the function is small and I preserved the original functionality.

PR #1506, opened 2020-11-24
Option for downloading tissue image for spatial visium dataset

hspitzer:
Following discussions with @giovp I've extended the `scanpy.datasets.visium_sge` function to optionally return a path to the high-resolution tissue image also available in the visium Spatial Transcriptomics datasets.
This makes it easy to leverage `scanpy.datasets` to fully explore visium datasets.


ivirshup:
Why not store the path to the image inside `uns` with the other image metadata?

scanpy/tests/test_datasets.py
```diff
@@ -98,6 +98,12 @@ def test_visium_datasets(tmp_dataset_dir, tmpdir):
     mbrain_again = sc.datasets.visium_sge("V1_Adult_Mouse_Brain")
     assert_adata_equal(mbrain, mbrain_again)
 
+    # Test that downloading tissue image works
+    mbrain, mbrain_img = sc.datasets.visium_sge(
+        "V1_Adult_Mouse_Brain", return_image_path=True
+    )
+    assert mbrain_img == sc.settings.datasetdir / "V1_Adult_Mouse_Brain" / "image.tif"
```
ivirshup: Could this also make sure the file exists, maybe even is an image?

hspitzer:
> Why not store the path to the image inside `uns` with the other image metadata?

yes, good idea - I'll do that instead of returning the path!

hspitzer:
I've updated the code to store the image path in `adata.uns['spatial'][sample_id]['tif_image_path']`. The test now also checks whether the image file exists. 

I can also test whether the image file is a valid tiff image, but for this we'd need to add a tiff reading library like `pillow` as a test dependency to scanpy. If that is ok, I'll update the PR with an additional test.

giovp:
thanks a lot @hspitzer !
I think checking for image is indeed problematic, maybe we can drop it if it requests additional packages.
re: image path. I would store it in the metadata dict:
```python
adata.uns["spatial"]["V1_Adult_Mouse_Brain"]["metadata"]
>>> {'chemistry_description': "Spatial 3' v1",
>>> 'software_version': 'spaceranger-1.0.0'}
```
with some name like key=image_name and value=path ? or something like that.

hspitzer:
> with some name like key=image_name and value=path ? or something like that.

sure! What do you mean with image_name? As downloaded from the 10x website, the tiff image always has the same filename ("image.tif"). Or do you mean the string "image_name" as key?

giovp:
good point, maybe the img format? like `'tif_image':''[PATH]`
so that if it's jpeg than it's visible?

ivirshup:
Should this argument be added to `sc.read_visium` as well? That is, add a path to the tiff if it would be in a standardized place?

ivirshup:
Also, I don't think I would mind `Pillow` too much as a test dependency (it doesn't have a ton of dependencies, right?)

A more lightweight solution would be to call [`file`](https://en.wikipedia.org/wiki/File_(command)) on the path, which should be able to recognize it as a `tiff`. I think this would only look at the head of the file though, and wouldn't check if it was corrupted/ didn't finish downloading.

scanpy/datasets/_datasets.py
```diff
@@ -386,7 +399,9 @@ def visium_sge(
     ----------
     sample_id
         The ID of the data sample in 10x’s spatial database.
-
+    add_image_path
```
ivirshup: I think this could have a better name. Maybe `download_tiff` or `include_tiff`?

giovp:
> Should this argument be added to `sc.read_visium` as well? That is, add a path to the tiff if it would be in a standardized place?

yes that's a very good point actually. It could even make it easier for us to initialize the image container maybe?

hspitzer:
> > Should this argument be added to `sc.read_visium` as well? That is, add a path to the tiff if it would be in a standardized place?
> 
> yes that's a very good point actually. It could even make it easier for us to initialize the image container maybe?

I thought about that already - standardised places for the tiff would be `sample_id"_image.tif` and `image.tif`. I don't think we need to add an extra argument to `sc.read_visium`, couldn't it automatically add the path if it exists?

hspitzer:
> Also, I don't think I would mind `Pillow` too much as a test dependency (it doesn't have a ton of dependencies, right?)
> 
> A more lightweight solution would be to call [`file`](https://en.wikipedia.org/wiki/File_(command)) on the path, which should be able to recognize it as a `tiff`. I think this would only look at the head of the file though, and wouldn't check if it was corrupted/ didn't finish downloading.

I was not aware of `file`. I think it might be a good solution! `readwrite._download` should make sure that downloads are not incomplete, so just reading the header might be enough

giovp:
> > > Should this argument be added to `sc.read_visium` as well? That is, add a path to the tiff if it would be in a standardized place?
> > 
> > 
> > yes that's a very good point actually. It could even make it easier for us to initialize the image container maybe?
> 
> I thought about that already - standardised places for the tiff would be `sample_id"_image.tif` and `image.tif`. I don't think we need to add an extra argument to `sc.read_visium`, couldn't it automatically add the path if it exists?

good point, agree probably best solution, however one thing it's not given is the format of the file (could be tif or jpg, but potentially some else, depending on the lab). But an automatic search would be best, just not sure how extensive we can make it

hspitzer:
> > > > Should this argument be added to `sc.read_visium` as well? That is, add a path to the tiff if it would be in a standardized place?
> > > 
> > > 
> > > yes that's a very good point actually. It could even make it easier for us to initialize the image container maybe?
> > 
> > 
> > I thought about that already - standardised places for the tiff would be `sample_id"_image.tif` and `image.tif`. I don't think we need to add an extra argument to `sc.read_visium`, couldn't it automatically add the path if it exists?
> 
> good point, agree probably best solution, however one thing it's not given is the format of the file (could be tif or jpg, but potentially some else, depending on the lab). But an automatic search would be best, just not sure how extensive we can make it

do you have an example where the image is a jpg? I can account for both tif and jpg, but I'm wondering if its worth it.

giovp:
> > > > > Should this argument be added to `sc.read_visium` as well? That is, add a path to the tiff if it would be in a standardized place?
> > > > 
> > > > 
> > > > yes that's a very good point actually. It could even make it easier for us to initialize the image container maybe?
> > > 
> > > 
> > > I thought about that already - standardised places for the tiff would be `sample_id"_image.tif` and `image.tif`. I don't think we need to add an extra argument to `sc.read_visium`, couldn't it automatically add the path if it exists?
> > 
> > 
> > good point, agree probably best solution, however one thing it's not given is the format of the file (could be tif or jpg, but potentially some else, depending on the lab). But an automatic search would be best, just not sure how extensive we can make it
> 
> do you have an example where the image is a jpg? I can account for both tif and jpg, but I'm wondering if its worth it.

collaborators that generated visium data sent me jpg

scanpy/datasets/_datasets.py
hspitzer: yes, I'll change it to `download_tif`. Or is there a preference for using "tiff" over "tif"?

scanpy/tests/test_datasets.py
hspitzer: I'm now checking with `file` whether the file is a tiff

hspitzer:
I've update the code to 
- test that the file is actually a tiff image
- automatically add the path to the image to `adata.uns['spatial'][library_id]['metadata']['tissue_image_path']`. It's looking for a tiff or jpeg file with the name `"image"` or `library_id"_image"`. This should cover most cases hopefully?

ivirshup:
I think it would be fine to only cover the case of what `space ranger` actually outputs. I was thinking there could be an argument where the user manually passes an alternate path. This could be useful for cases where they've processed the image themselves some modifications to the image.

scanpy/datasets/_datasets.py
giovp: since image can be non-tif, I would choose a different name

giovp:
> I think it would be fine to only cover the case of what `space ranger` actually outputs. I was thinking there could be an argument where the user manually passes an alternate path. This could be useful for cases where they've processed the image themselves some modifications to the image.

space ranger doesn't output this image, as it's taken as input to assign spots and get scalefactors and metadata. This type of image is in the same folder just for chance in the 10x genomics dataset. 
In the `read_visium` function I would simply add an argument to pass the path of the image, and basically just assign it to the `adata.uns` metadata. Otherwise just assign None. THis way it's consistent for the spatial tool whichlater uses it in the image container.
It's also convenient to add it as argument so that `read_visium` could just be passed in that same way as it is now in `datasets.visium_sge`

scanpy/datasets/_datasets.py
hspitzer: maybe `download_tissue_image`?

hspitzer:
I've made a small update to add a `tissue_image_path` parameter to `read_visium` that you can use to specify the tissue image path to put in `adata.uns`

scanpy/readwrite.py
```diff
@@ -418,6 +422,26 @@ def read_visium(
...
+                path / (library_id + "_image" + ext)
+                for ext in ['.tif', '.tiff', '.jpg', '.jpeg']
+            ]
+            for p in cand_image_paths:
+                if p.exists():
+                    tissue_image_path = p.resolve()
+                    break
+        # put image path in uns
+        else:
+            tissue_image_path = Path(tissue_image_path).resolve()
```
ivirshup: Is it very important that you guess where the image is? I would rather this not happen, since it could easily read the wrong thing, or not find an image a user would want to pick up. I'd rather the user had to be explicit about this.

scanpy/datasets/_datasets.py
```diff
@@ -377,6 +389,7 @@ def visium_sge(
         'Targeted_Visium_Human_ColorectalCancer_GeneSignature',
         'Parent_Visium_Human_ColorectalCancer',
     ] = 'V1_Breast_Cancer_Block_A_Section_1',
+    download_tissue_image: bool = False,
```
ivirshup: Instead of `tissue_image`, could this be something like `tiff`, `hires_tiff`, or `raw_tiff`?

I don't like `tissue_image`. The jpegs we load by default could also be called `tissue_image`s, so this seems a little unclear to me.

ivirshup (changes requested):
Sorry for the late review! A few comments about naming conventions and guessing filenames.

> In the read_visium function I would simply add an argument to pass the path of the image, and basically just assign it to the adata.uns metadata. Otherwise just assign None. THis way it's consistent for the spatial tool whichlater uses it in the image container.

I like this approach.

scanpy/readwrite.py
hspitzer: From a user perspective its more convenient if the path is assumed. Not that this is also the case for the png images (assumed to be in `spatial/tissue_hires_image.png`). This is the case for the standardised datasets you can download from visium but many of course be different for datasets that you create yourself. 
I do understand your point though, guessing is rarely a very good idea.

scanpy/datasets/_datasets.py
hspitzer: sure, I can call this flag `download_hires_tiff`. This is only for downloading images from 10x, and they only provide hires tiff images.
@giovp made the point earlier that in general, the hires_tiff image could also be a jpg, so I would still call the entry in `adata.uns` something like `tissue_image_path` or `hires_image_path`.

giovp: agree, argument of download could have `tiff` but argument in `uns` should not. I would also drop hires cause there is already a hires in `adata.uns["spatial"]`. Maybe `full_tissue_image`  or something like that? don't know sorry

ivirshup: What uses these images in `squidpy`? In particular what uses these images that wouldn't work with the `jpeg`s?

scanpy/readwrite.py
ivirshup: I think there are two issues with this:

* If it's wrong, it's not obvious until later in the analysis. If there are other images in the directory, for example: low-res thumbnails or plots from the user's pipeline could get picked up by accident.
* As it stands right now, if a user ever passed `"include_tiff"` (or whatever we end up calling it), it will always be included on subsequent calls. I think it's bad for there to be state like that in these functions.

scanpy/datasets/_datasets.py
giovp: in `squidpy` this image is used by the image container. Don't understand the `jpeg` question sorry

ivirshup: Ha, for `jpeg` I meant the `tissue_hires_image.png` and `tissue_lowres_image.png`.

ivirshup: I think the implication is that this is the `raw`, `full`. or `source` image that you want to do processing on.

I would agree that `hires` would also be a good descriptor, but that's already used by the `png`.

You would be in a better position to comment, but I like `source_image_path` and `full_image_path`. Both imply that they are the complete data, which is what you'll want to do processing on.

giovp: ah ok, so 
> what uses these images that wouldn't work with the jpegs [pngs]?

basically everything that works on imag in squidpy, in particular segmentation, feature extraction and visualization with napari

> You would be in a better position to comment, but I like source_image_path and full_image_path. Both imply that they are the complete data, which is what you'll want to do processing on.

Also agree, actually `source_image_path` is nice as that image is indeed the source image for spaceranger to do the processing

scanpy/readwrite.py
hspitzer: ok, sure. Lets leave it at having the user passing `source_image_path` to `read_visium` when they want to include this information in the adata.

hspitzer:
Updated the PR:
- renamed argument in `visium_sge` to `include_hires_tiff`
- renamed argument in `read_visium` to `source_image_path` and removed guessing of image file if `source_image_path` is None

scanpy/readwrite.py
```diff
@@ -418,6 +421,14 @@ def read_visium(
             inplace=True,
         )
 
+        # put image path in uns
+        if source_image_path is not None:
+            # get an absolute path
+            source_image_path = str(Path(source_image_path).resolve())
+        adata.uns["spatial"][library_id]["metadata"]["source_image_path"] = str(
+            source_image_path
+        )
```
ivirshup:
```suggestion
            adata.uns["spatial"][library_id]["metadata"]["source_image_path"] = str(
                source_image_path
            )
```

Right?

scanpy/tests/test_datasets.py
```diff
@@ -7,6 +7,8 @@
 import pytest
 from pathlib import Path
 from anndata.tests.helpers import assert_adata_equal
+import os
```
ivirshup:
```suggestion
```

This doesn't get used, right?

scanpy/readwrite.py
```diff
@@ -316,6 +317,8 @@ def read_visium(
         'filtered_feature_bc_matrix.h5' or 'raw_feature_bc_matrix.h5'.
     library_id
         Identifier for the visium library. Can be modified when concatenating multiple adata objects.
+    source_image_path
+        Path to the high-resolution tissue image. Will be saved in `uns`.
```
ivirshup:
```suggestion
    source_image_path
        Path to the high-resolution tissue image. Path will be included in
        `.uns["spatial"][sample_id]["metadata"]["source_image_path"]`.
```

ivirshup (changes requested):
Minor changes, then lgtm.

Please also add a release note!

scanpy/datasets/_datasets.py
```diff
@@ -386,7 +399,9 @@ def visium_sge(
     ----------
     sample_id
         The ID of the data sample in 10x’s spatial database.
-
+    include_hires_tiff
+        Download and include the high-resolution tissue image (tiff) in `adata.uns["spatial"][sample_id]["metadata"]["source_image_path"]`.
+
```
ivirshup:
```suggestion
    include_hires_tiff
        Download and include the high-resolution tissue image (tiff) in `adata.uns["spatial"][sample_id]["metadata"]["source_image_path"]`.

```

scanpy/datasets/_datasets.py
```diff
@@ -377,6 +389,7 @@ def visium_sge(
         'Targeted_Visium_Human_ColorectalCancer_GeneSignature',
         'Parent_Visium_Human_ColorectalCancer',
     ] = 'V1_Breast_Cancer_Block_A_Section_1',
+    include_hires_tiff: bool = False,
```
ivirshup:
```suggestion
    ] = 'V1_Breast_Cancer_Block_A_Section_1',
    *,
    include_hires_tiff: bool = False,
```

I think it makes sense for this argument to be keyword only.

scanpy/readwrite.py
hspitzer: yes, thanks, totally missed that!

scanpy/tests/test_datasets.py
hspitzer: no, not anymore

hspitzer:
> Please also add a release note!

sorry for the naive question, but how do I add a release note? As a comment in the PR? As a commit?

giovp:
just mention this PR with brief description and your name here: https://github.com/theislab/scanpy/blob/master/docs/release-latest.rst

nice! only one PR to go, thank you @hspitzer @ivirshup !

hspitzer:
> just mention this PR with brief description and your name here: https://github.com/theislab/scanpy/blob/master/docs/release-latest.rst

ok, done. I put it in the "on master" section

scanpy/readwrite.py
```diff
@@ -316,6 +317,9 @@ def read_visium(
         'filtered_feature_bc_matrix.h5' or 'raw_feature_bc_matrix.h5'.
     library_id
         Identifier for the visium library. Can be modified when concatenating multiple adata objects.
+    source_image_path
+        Path to the high-resolution tissue image. Path will be included in
+        `.uns["spatial"][sample_id]["metadata"]["source_image_path"]`.
```
ivirshup:
```suggestion
        `.uns["spatial"][library_id]["metadata"]["source_image_path"]`.
```

ivirshup (approved):

ivirshup:
Thanks for going through this with me!

PR #1507, opened 2020-11-24
Fix progress bar, use requests, tqdm version

michalk8:
This fixes the small progress when downloading the data as `sc.datasets.paul_15()`, etc..
Also fixes removing the file when `KeyboardInterrupt` occurs (previously, it was just on exceptions and at least for me I sometimes stop a download by interupting the jupyter kernel - now it also correctly removes the partially downloaded file).

I've also opted for `requests` library, since it's much cleaner than urllib3 and as a dependency.
However, this is completely unrelated to the 2 things I've mentioned above and can be quickly reverted.

Relevant PRs I could find:
- tqdm version (I opted for bumping the version [search through open issues, lowest one had 4.32]): https://github.com/theislab/scanpy/issues/1244

Not relevant:
My [reaction](https://www.youtube.com/watch?v=nixR6wVa4HY&t=72) when using the progress bar (slightly NSFW [foul language]).

### DIFF
--- a/scanpy/readwrite.py
+++ b/scanpy/readwrite.py
@@ -913,36 +913,26 @@ def _get_filename_from_key(key, ext=None) -> Path:
 
 
 def _download(url: str, path: Path):
-    try:
-        import ipywidgets
-        from tqdm.auto import tqdm
-    except ModuleNotFoundError:
-        from tqdm import tqdm
-    from urllib.request import urlopen, Request
-
-    blocksize = 1024 * 8
-    blocknum = 0
+    from tqdm.auto import tqdm
+    import requests
 
     try:
-        # This is a modified urllib.request.urlretrieve, but that won't let us
-        # set headers.
-        # The '\'s are ugly, but parenthesis are not allowed here
-        # fmt: off
-        with \
-            tqdm(unit='B', unit_scale=True, miniters=1, desc=path.name) as t, \
-            path.open("wb") as f, \
-            urlopen(Request(url, headers={"User-agent": "scanpy-user"})) as resp:
-
-            t.total = int(resp.info().get("content-length", 0))
-
-            block = resp.read(blocksize)
-            while block:
-                f.write(block)
-                blocknum += 1
-                t.update(blocknum * blocksize - t.n)
-                block = resp.read(blocksize)
-        # fmt: on
-    except Exception:
+        with requests.get(
+            url, headers={"User-agent": "scanpy-user"}, stream=True
+        ) as resp:
+            resp.raise_for_status()
+            total = resp.headers.get("content-length", None)
+            with tqdm(
+                unit="B",
+                unit_scale=True,
+                miniters=1,
+                unit_divisor=1024,
+                total=total if total is None else int(total),
+            ) as t, path.open("wb") as f:
+                for chunk in resp.iter_content(chunk_size=8 * 1024):
+                    t.update(len(chunk))
+                    f.write(chunk)
+    except (KeyboardInterrupt, Exception):
         # Make sure file doesn’t exist half-downloaded
         if path.is_file():
             path.unlink()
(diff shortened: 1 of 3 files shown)


scanpy/readwrite.py
```diff
@@ -913,36 +913,26 @@ def _get_filename_from_key(key, ext=None) -> Path:
...
-    try:
-        import ipywidgets
-        from tqdm.auto import tqdm
-    except ModuleNotFoundError:
-        from tqdm import tqdm
-    from urllib.request import urlopen, Request
-
-    blocksize = 1024 * 8
-    blocknum = 0
+    from tqdm.auto import tqdm
```
ivirshup: Can you keep the check for `ipywidgets`? Using `tqdm.auto` in a notebook errors without it (seems it's not so `auto`)

Using tqdm 4.51.0:

```python
from tqdm.auto import tqdm
from time import sleep

for i in tqdm(range(100)):
    sleep(0.05)
```

```pytb
ImportError                               Traceback (most recent call last)
<ipython-input-3-542c7d78a2ff> in <module>
----> 1 for i in tqdm(range(100)):
      2     sleep(0.05)

~/miniconda3/envs/tqdm_nowidgets/lib/python3.8/site-packages/tqdm/notebook.py in __init__(self, *args, **kwargs)
    221         unit_scale = 1 if self.unit_scale is True else self.unit_scale or 1
    222         total = self.total * unit_scale if self.total else self.total
--> 223         self.container = self.status_printer(
    224             self.fp, total, self.desc, self.ncols)
    225         self.sp = self.display

~/miniconda3/envs/tqdm_nowidgets/lib/python3.8/site-packages/tqdm/notebook.py in status_printer(_, total, desc, ncols)
     94         # Prepare IPython progress bar
     95         if IProgress is None:  # #187 #451 #558 #872
---> 96             raise ImportError(
     97                 "IProgress not found. Please update jupyter and ipywidgets."
     98                 " See https://ipywidgets.readthedocs.io/en/stable"

ImportError: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html
```

scanpy/readwrite.py
```diff
@@ -913,36 +913,26 @@ def _get_filename_from_key(key, ext=None) -> Path:
...
+    import requests
```
ivirshup: I'd rather keep the previous code that doesn't use requests, just because it's worked for a while, and "if it ain't broke don't fix it".

ivirshup (changes requested):
Could you show an example of how your results look before and after?

Otherwise, just a few minor reverts and I think this is good.

Thanks!

ivirshup:
The main change here is passing `None` instead of `0` to `total`, right?

Also: this makes some errors with files still existing make much more sense. I had no idea `KeyboardInterrupt` doesn't inherit from `Exception`.

scanpy/readwrite.py
michalk8: Didn't know this caused an issue. I've reverted this.

scanpy/readwrite.py
michalk8: Ok, also reverted, just thought that using `requests` is a bit cleaner.

michalk8:
Thanks for a quick response and the comments.

> The main change here is passing None instead of 0 to total, right?

It was actually setting it in the contructor, rather than assigning it to the tqdm object (the latter doesn't work).

Here's before:
old
and after:
new

> Also: this makes some errors with files still existing make much more sense. I had no idea KeyboardInterrupt doesn't inherit from Exception.

I didn't know that either, so I looked it up (it actually inherits from `BaseException` among other things:
https://docs.python.org/3/library/exceptions.html#exception-hierarchy

michalk8:
Just noticed the `?` at the end (both-cases), last commit should fix it (it was the updating), here's the final result:

newnew

ivirshup (approved):
Looks good to me. Could you add a bugfix release note (`docs/release-latest.rst`)?

flying-sheep:
Hey @ivirshup, can you please use the “Squash & merge” button?

It’s not super useful to drill through commits called “Rever tqdm/requests change” when searching a file’s history

PR #1511, opened 2020-11-27
Heatmap colormatching

### LINKED ISSUE #1479: Mismatching cluster color codes in sc.pl.heatmap

---

### Description
The issue is visible/reproducibe from the [plotting tutorial](https://scanpy-tutorials.readthedocs.io/en/latest/plotting/core.html#Heatmaps) in cell 23.
Here, as an example, Cluster 3/NK cells are associated with the marker genes GNLY and NKG7.
However, the cluster color code representing the rows and columns of the heatmap are mismatching. Cluster 3 is summarized using violet on the cell level (color bar below the heatmap) and using green for grouping Cluster 3 associated markers GNLY and NKG7 (color bar on the right). I would expect that these color codes match.

### Minimal code sample
Below are a few lines to reproduce this behaviour:
```python
import scanpy as sc

sc.tl.leiden(pbmc, key_added='clusters', resolution=0.5)

marker_genes_dict = {'NK': ['GNLY', 'NKG7'],
                     'T-cell': ['CD3D'],
                     'B-cell': ['CD79A', 'MS4A1'],
                     'Monocytes': ['FCGR3A'],
                     'Dendritic': ['FCER1A']}


sc.pl.heatmap(pbmc, marker_genes_dict, groupby='clusters', vmin=-2, vmax=2, cmap='RdBu_r', dendrogram=True, swap_axes=True)
```

LisaSikkema:
Corrected colouring of groups in heat map, so that row and column colors match up. Related to issue #1479 

- [x] black
- [x] fix scanpy tutorial https://scanpy-tutorials.readthedocs.io/en/latest/plotting/core.html#Heatmaps
- [x] account for case where var_group_labels are not a subset of groupby groups



giovp:
```python
sc.tl.leiden(pbmc, key_added="clusters", resolution=0.5)

marker_genes_dict = {
    "1": ["GNLY", "NKG7"],
    "0": ["CD3D"],
    "2": ["CD79A", "MS4A1"],
    "4": ["FCGR3A"],
    "3": ["FCER1A"],
}


sc.pl.heatmap(
    pbmc,
    marker_genes_dict,
    groupby="clusters",
    vmin=-2,
    vmax=2,
    cmap="RdBu_r",
    dendrogram=True,
    swap_axes=True,
)
```

before:
image

after:
image

as you can see, heatmap colors for rows (genes) is consistent with columns (clusters) 
@wkopp is this helpful? notebook still to be updated

LisaSikkema:
One thing to note:
if the categories of the groupby variable (clusters in this case) don't match up with the categories in the marker_genes_dict, there will be no matching if colors between rows and columns. This is what also happens in the tutorial, where
marker_genes_dict = {'NK': ['GNLY', 'NKG7'],
                     'T-cell': ['CD3D'],
                     'B-cell': ['CD79A', 'MS4A1'],
                     'Monocytes': ['FCGR3A'],
                     'Dendritic': ['FCER1A']}
while the clusters are numbers ("1", "2" etc.).
We could consider throwing a warning in that case, but I don't think it's necessary.

wkopp:
I'm confused by the figure above.

In Cluster 2 (shown in green below the heatmap) the marker gene that is coming up most prominently is CD3D, which was defined as a marker for Cluster 0, though.
The color code on the right seems correct, but why is CD3D most prominently expressed in Cluster 2 and not in Cluster 0?
The same seems to be the case for the other clusters.
Or were the marker genes perhaps just mislabelled?

LisaSikkema:
Yeah the marker assignments here are random, that might be a bit confusing indeed. The only point is that the colors for the clusters (columns) match with the *names* of the marker gene lists (rows). So if, in an actual dataset, you now name the correct clusters "T cells", "B cells" etc. in you obs.clusters variable, and match those names with your marker gene dict keys, everything will match up. 
Does that make sense?
I just didn't go through annotations of the clusters here, that would be a bit tedious. But try it out on your own (correctly annotated) anndata object if you'd like.

wkopp:
Yes, the colorcodes are matching now and the confusion about the marker genes is just because of with mislabelling marker-cluster associations.
You can find the correct markers-cluster associations from cross-checking with umap figures, e.g.

```{python}
sc.pl.umap(pbmc, color=["GNLY",'FCER1A',"CD3D","CD79A","FCGR3A", "clusters"])
```

Thanks!

giovp:
exactly! also this makes a lot of sense in the context of the reversed heatmap (where keys of the mapping are plotted as column annotation). 
This is also particularly useful if you have 2 annotations `cluster1=['1','2','3']` and `cluster2=['foo','bar']`, and you want to check for
```python
markers={
	"foo":["gene1", "gene2"],
	"bar":["gene3"]
}
```
but with respect to `cluster1`.

I would have just thrown an error but this is a much more elegant solution!

LisaSikkema:
Okay true that's not that hard. So here a better example, with correct cluster + marker gene combinations:
before:
Screenshot 2020-11-27 at 15 48 18
after:
Screenshot 2020-11-27 at 15 47 16

and code to reproduce it:
```python
import scanpy as sc

sc.tl.leiden(pbmc, key_added="clusters", resolution=0.5)
sc.pl.umap(
    pbmc,
    color=[
        "GNLY",
        "FCER1A",
        "CD3D",
        "CD79A",
        "FCGR3A",
        "NKG7",
        "CD79A",
        "MS4A1",
        "clusters",
    ],
)
marker_genes_dict = {
    "3": ["GNLY", "NKG7"],
    "1": ["FCER1A"],
    "2": ["CD3D"],
    "0": ["FCGR3A"],
    "4": ["CD79A", "MS4A1"],
}
sc.pl.heatmap(
    adata=pbmc,
    var_names=marker_genes_dict,
    groupby="clusters",
    vmin=-2,
    vmax=2,
    cmap="RdBu_r",
    dendrogram=True,
    swap_axes=True,
)
```

LisaSikkema:
Hi! not sure why the two tests failed, I don't think it's related to my edits

ivirshup:
@LisaSikkema, no worries there's been some flakiness of that test.

Can this get a test case like ? Maybe even two, one where groups match, one where they don't?

I'm thinking something that calls `save_and_compare_images`, like https://github.com/theislab/scanpy/blob/02fc946a8ce3c2e456dbc6e026ee068734f11e1e/scanpy/tests/test_plotting.py#L272

fidelram:
Thanks for looking at this... it is surprising that this bug was not detected earlier.  

I looked at the code and looks fine but, I would like to add a test. @LisaSikkema can you check this? If this is too much trouble I can do it or I can help you because the plot test are difficult as they require similar setup as in the CI tests.

LisaSikkema:
Yes would be happy to look into it and come up with a test! I have no experience with the testing code yet though, so might take me a while to figure out how to write it up.

fidelram:
first try the following and check that you don't have any errors. 
```
cd scanpy/scanpy/tests
py.test test_plotting.py
```

If that is the case, you can go to this line: https://github.com/theislab/scanpy/blob/02fc946a8ce3c2e456dbc6e026ee068734f11e1e/scanpy/tests/test_plotting.py#L36 and add a new test.

I usually run the tests after adding the new test case which will fail the first time. But, it generates an image to compare. Then I copy the new image from `./figures/` to `./_images/` and add it to git. Off course you want to check that the image is what you expect. If you re-run the tests they should work now.

LisaSikkema:
thanks for making the test, was planning to do it today (Friday = scanpy day)! Read through your code, so now I'll know how to do it next time.

PR #1512, opened 2020-11-27
fix logic for empty library_id and images

### LINKED ISSUE #1534: sc.pl.embedding with basis "spatial" has side-effects and is missing one row of points
- [ ] I have confirmed this bug exists on the latest version of scanpy.

---

sc.pl.embedding with basis "spatial" behaves strangely. See code example below. 
- when "spatial" is an `unit` array, the resulting plot is missing the top row of points (I think there might be an overflow happening)
- when plotting edges and using basis "spatial", `adata.obsm["spatial"] is modified after calling `sc.pl.embedding`
@giovp 

```python
import numpy as np
import anndata as ad
import scanpy as sc
print(sc.__version__)

arr = np.array([[1888,   55],
       [1887,   54],
       [1888,   54],
       [1889,   54],
       [1887,   55],
       [1889,   55],
       [1887,   56],
       [1888,   56],
       [1889,   56]], dtype=np.uint8)
conn = np.array([[0., 1., 1., 1., 1., 1., 1., 1., 1.],
       [1., 0., 1., 0., 1., 0., 0., 0., 0.],
       [1., 1., 0., 1., 1., 1., 0., 0., 0.],
       [1., 0., 1., 0., 0., 1., 0., 0., 0.],
       [1., 1., 1., 0., 0., 0., 1., 1., 0.],
       [1., 0., 1., 1., 0., 0., 0., 1., 1.],
       [1., 0., 0., 0., 1., 0., 0., 1., 0.],
       [1., 0., 0., 0., 1., 1., 1., 0., 1.],
       [1., 0., 0., 0., 0., 1., 0., 1., 0.]])
a = ad.AnnData(arr)
a.uns = {'spatial': {'connectivities_key': 'spatial_connectivities', 'distances_key': 'spatial_distances', 'params': {'n_neighbors': 8, 'coord_type': None, 'radius': 1.5}}}
a.obsp['spatial_connectivities'] = conn
a.obsm['coords'] = arr.copy()
a.obsm['spatial'] = arr.copy()

sc.pl.embedding(a, 'coords',  edges=True, neighbors_key = 'spatial')
print('coords', a.obsm['coords'])
sc.pl.embedding(a, 'spatial')
print('spatial no edges', a.obsm['spatial'])
sc.pl.embedding(a, 'spatial',  edges=True, neighbors_key = 'spatial')
print('spatial edges', a.obsm['spatial'])
```

produces following output:

- expected (when plotting "coords")
image
```
coords [[96 55]
 [95 54]
 [96 54]
 [97 54]
 [95 55]
 [97 55]
 [95 56]
 [96 56]
 [97 56]]
```

- when plotting "spatial" without edges (obsm is not modified)
image
```
spatial no edges [[96 55]
 [95 54]
 [96 54]
 [97 54]
 [95 55]
 [97 55]
 [95 56]
 [96 56]
 [97 56]]
```

- when plotting "spatial" with edges (obsm is modified)

image
```
spatial edges [[ 96 255]
 [ 95 254]
 [ 96 254]
 [ 97 254]
 [ 95 255]
 [ 97 255]
 [ 95   0]
 [ 96   0]
 [ 97   0]]
```

#### Versions

<details>

scanpy==1.6.1.dev70+g7f15d22d anndata==0.7.5 umap==0.4.6 numpy==1.19.4 scipy==1.5.3 pandas==1.1.4 scikit-learn==0.23.2 statsmodels==0.12.1 python-igraph==0.8.3 louvain==0.6.1 leidenalg==0.8.3

</details>

giovp:
fix logic in order to accomodate non-visium plotting (e.g. seqfish).

essentially, for non-visium, scatterplot should be used because we can't find heuristic for spot size. We can't find heuristic for spot size because spatial axis could be any dimension (e.g. I just found z-scored ones). With visium instead, spatial axis will always be pixel dimension, and so ok to use circles and thus set the heuristics (not changed, but just for references it's line 328 of scatterplots).


giovp:
not ready yet, there is a problem with `sc.pl.embedding`

giovp:
the problem this PR try to address is the following.
We construct the spatial graph and put keys of obsp in `adata.uns["spatial"]`. Problem is that there is also the library id in case of Visium. The logic before would work only if there was a single key in `adata.uns["spatial"]`. Now instead, we need to account for `adata.uns["spatial"]` being empty, and also having `adata.uns["spatial"]["connectivities_key"]` or `adata.uns["spatial"]["distances_key"]`.
It doesn't look really good because of the hard coded keys. Any suggestion?

giovp:
This shouldbe ready now @ivirshup , thank you!

ivirshup:
Question for this, what heuristics have you tried? My guess would be that `min(distances_between_points) / 3`  should be fine for an upper bound.

Second, I think this logic is a little convoluted, and I don't know that `library_id` will always be associated with visium only. Would a better check be for `["metadata"]["software_version"]` or something like that?

It might help for me to know what exactly you're planning on putting in the `"spatial"` entry.

giovp:
all very good points, and I don't think I have a clear solution, it's more that we have to decide how to go about this:
> Question for this, what heuristics have you tried? My guess would be that min(distances_between_points) / 3 should be fine for an upper bound.

this indeed could be solved, but we still would have to set this heuristics differently according to the spatial data type in question. E.g. for visium `size=1` is correct, because coordinates are in pixel measure. In the dataset I have now (seqFISH) the coordinates are essentially z-score and so would have to change for instance to the one you proposed. However, why would we want to have `circle` at all in that t case, and not just scatterplot? Since there is no real notion of size, I think a scatterplot is actually more appropriate. We discussed this already but back then we didn't have this example.

> Second, I think this logic is a little convoluted, and I don't know that library_id will always be associated with visium only. Would a better check be for ["metadata"]["software_version"] or something like that?

It definitely is, there might be a better solution but I couldn't come up with it. The problem stems in the fact that we have a `library_id`  key in `adata.uns["spatial"]`. In `spatial` we also put this
```python
{'connectivities_key': 'spatial_connectivities',
 'distances_key': 'spatial_distances',
 'params': {'n_neighbors': 6, 'coord_type': None, 'radius': None}}
``` 
this is needed for plotting. The default in `sc.pl.spatial` now is that if library_id is `_empty`, then it iteratively search for some keys in `adata.uns["spatial"]`.
```python
try:  # check if key is empty
            spatial_data = adata.uns['spatial']
            library_id = next(
                (
                    i
                    for i in spatial_data.keys()
                    if i not in ["connectivities_key", "distances_key"]
                )
            )
except (KeyError, StopIteration) as e:
	library_id = None
```
The point is that it should only assign whatever key it finds that is not `["connectivities_key", "distances_key"]`. It could also be though that it's empty altogether (e.g. `spatial_connecitivty`) has not been called yet. Then, it should still set `library_id=None`

> It might help for me to know what exactly you're planning on putting in the "spatial" entry.

1 milllion dollars question 😅 , for now I would say only `["connectivities_key", "distances_key"]`.

ivirshup:
> the coordinates are essentially z-score

I'm not totally sure what this means. The coordinates have been z-score transform across each axis? How is this useful? In particular, how is it useful to completely replace the original coordinates with this?

-------------------

For spatial neighbour information, I feel like this should go in a separate top level key in `uns`. For example, with `"params"`, it's not obvious to me that the parameters in `"spatial"` should only refer to how neighbours were computed.

Can the spatial neighbours be based off multiple library ids? If so, could you have:

```python
uns = {
    "spatial": {
        "library1": {...},
        "library2": {...},
        ...
    },
    "spatial_neighbors": {
        "library_ids": ["library1", ...],
        "connectivities_key": ...,
        "distances_key": ...,
        "params": {...},
    },
}
```

> for now I would say only ["connectivities_key", "distances_key"].

I think once you start adding stuff, it's easy to add more, and this code will stop working if that happens. For example, the `"params"` key.

------------

For figuring out which case this is:

* If no library key was provided:
    * If there is only one key, use that one
    * If there is more than one key, use none, and warn that the user should pass this.

giovp:
Thanks for bearing with me Isaac 😅 🙏  took some of your suggestions and here's current status:

- reverted back to look for `library_id` in spatial, but still added the exception that `adata.uns["spatial"]` does not exist. This is in order to use `sc.pl.spatial` with non-visium data.
  - if that's the case, then spatial should simply wrap embedding. This also refers to your point.
	> I'm not totally sure what this means. The coordinates have been z-score transform across each axis? How is this useful? In particular, how is it useful to completely replace the original coordinates with this?

	this is very likely to happen for anything that it's not visium. In that case, users will share already processed data that contains coordinates in some type of system, and this is the case for whatever processing they had to undertake (would suggest you to have a look at https://github.com/spacetx/starfish for examples of those processing steps.). Anyway, in short, it's much easier for us to just wrap embedding in that case, and I also think it's more correct cause then is the user to choose whatever heuristics they want for point sizes.

- fixed a problem in #1534 , that is that the coordinate systems in non-visium has bottom left origin (whereas in visium is top-left, which makes sense because it's in image pixel coordiantes). For this reason, I added the y coordinate inversion in `sc.pl.spatial`, and only in the case where visium is selected, but with img_key = None. Note that this happens because if an img is plotted (before the spots with `circle`), then the origin automatically swap. But if `img_key` is None, then it reverts to default (bottom left). This made it easier as I could remove it from `def _get_data_points` and from `utils._get_edges`. Also added couple of tests for this case.

This should be ready for another review, let me know if logic is clearer or I could add more comments in code.

re
> Can the spatial neighbours be based off multiple library ids? If so, could you have:
```python
uns = {
    "spatial": {
        "library1": {...},
        "library2": {...},
        ...
    },
    "spatial_neighbors": {
        "library_ids": ["library1", ...],
        "connectivities_key": ...,
        "distances_key": ...,
        "params": {...},
    },
}
```

yes indeed, I will change all occurrences in squidpy so that `sc.pl.spatial` can simply work as before.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -820,24 +820,37 @@ def spatial(
...
+                    (k for k in ['hires', 'lowres'] if k in spatial_data['images']),
+                    None,
+                )
+            elif img_key is None:  # invert coordinate if img_key is None
+                adata.obsm["spatial"][:, 1] = np.abs(
+                    np.subtract(
+                        adata.obsm["spatial"][:, 1],
+                        np.max(adata.obsm["spatial"][:, 1]),
+                    )
+                )
```
ivirshup: Plotting functions shouldn't mutate data

ivirshup (commented):
> reverted back to look for library_id in spatial

So, my understanding of the `spatial` function is that it's for the case when you have an image you would like to plot on top of. To me it, it doesn't sound like this z-score normalized case falls into that.

>  that is that the coordinate systems in non-visium has bottom left origin (whereas in visium is top-left, which makes sense because it's in image pixel coordiantes). For this reason, I added the y coordinate inversion in sc.pl.spatial

Yeah, that mutation is pretty bad, and definitely should not happen.

But aren't you mutating again here? It looks like you'd end up just flipping it again if you called `spatial` again.

---------------------------------

Do you think you could provide me with some example objects that are giving you trouble?

I'm quite certain there is a better solution here, but I think I need to be able to try out some solutions.

I think it's going to be close to [inverting the y-axis of the plot](https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.invert_yaxis.html#matplotlib.axes.Axes.invert_yaxis) if there's an image.

scanpy/plotting/_tools/scatterplots.py
giovp: yes good point. I should just create a temporary adata object?

ivirshup: I think the plot should just get it's y-axis inverted if you're expecting coordinates for an image.

Alternatively, a copy of just the spatial coordinates could be made.

giovp:
> Do you think you could provide me with some example objects that are giving you trouble?

you can take any `sc.datasets.visium_sge` and play around with inverting/not inverting second axis,and plotting using `sc.pl.spatial` or `sc.pl.embedding` where coordinates are in `adata.obsm["coords"]`

I'll give another summary on current situation and goals:

**Type of spatial data**
1. data with coordinates centered bottom left and no image (non visium)
2. data with coordinates centered top left and no image (visium)
3. data with coordinates centered top left and image (visium)

`sc.pl.spatial` should support all of the above cases. In all cases it wraps embedding but in 1. it uses scatterplot, in 2. and 3. it uses circles (with a specified radius, present only in visium).

The inversion is needed in case 2., because in case 3. this is already handled by the image axis plot.
My solution for this is to pass inverted coordinates for y axis in 2. to `embedding`, so that everything can be handled by the function independently. 
What do you think is a good way to go about this ?

ivirshup:
Thanks for the new explanation, this makes much more sense to me now.

----------------------------

For case 1, when there is no image, what is the advantage to using `sc.pl.spatial` over just `sc.pl.embedding`?

Also what about non-visium data with an image?

-----------------------------

For case 2, I think there are two options. You can plot a blank image of the correct size, or you could invert the y axis manually with something like [`Axes.set_ylim`](https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.set_ylim.html#matplotlib.axes.Axes.set_ylim) or [`Axes.invert_yaxis`](https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.invert_yaxis.html#matplotlib.axes.Axes.invert_yaxis). I believe this is what `Axes.imshow` is doing internally (see `mpl.image.AxesImage.set_extent`).

giovp:
thank you for bearing with me 😅 I understand I should have been clear like that from beginning, sorry.

So now by default everything that calls spatial inverts y axis. I also added two lines in the doc to clarify for user. I think this account for all the cases above, which are also presents in tests.

> For case 1, when there is no image, what is the advantage to using sc.pl.spatial over just sc.pl.embedding?

indeed none, but I still like that user could use `sc.pl.spatial` which in that case is a simple call `sc.pl.embedding(adata, basis="spatial", **kwargs)`

> Also what about non-visium data with an image?

in that case, we essentially don't strictly have a direct mapping to our observation uni (i.e. cell/spot) unless the user also specify a segmentation mask or some other way of annotating molecular probes in the image to observation units (e.g see [this](https://www.biorxiv.org/content/10.1101/800748v2.abstract) and [this](https://www.biorxiv.org/content/10.1101/2020.02.12.945345v1) paper).
It can be also more complicated if the data has subcellular resolution, like [this](https://science.sciencemag.org/content/361/6401/eaar7042)
For all these cases, we'll rely on Napari, I'm in the process of building a class that maps anndata+img container to napari https://github.com/theislab/squidpy/pull/184

ivirshup:
> thank you for bearing with me 😅 I understand I should have been clear like that from beginning, sorry.

No worries!

I think communicating about the ideas we have for these tools can be fraught.

> in that case, we essentially don't strictly have a direct mapping to our observation uni (i.e. cell/spot)

I don't think this is the case. 

First, I believe there are non-visium grid based spatial methods (I remember seeing a product page for one, but can't find it atm)

Second, I think you don't need segmentation info to use this function. You just need coordinates (probably derived from segmentation) and possibly an image. Like this:

> unless the user also specify a segmentation mask or some other way of annotating molecular probes in the image to observation units

But I think a user already having done the segmentation, then coming to scanpy is a reasonable workflow.

giovp:
@ivirshup here's the final version, I think it's the best compromise wrt what we discussed (i'll quote from slack the following points)

> All image handling moved into the spatial function, e.g. `img_key`, `crop_coord`, `alpha_img`, `bw`, and `library_id` don’t get passed to embedding. Instead just an image array gets passed, potentially also a scaled coordinate array as basis.

this is partially addressed, what is now passed to embedding is the image and a bunch of args that are though relevant to the image itself (so nothing to do with spatial). That is:
```python
img: Union[np.ndarray, None, Empty] = None, # the image
cmap_img: Optional[str] = None, # cmap for ax.imshow
alpha_img: Optional[float] = 1.0, # alpha for the image
crop_coord: Tuple[int, int, int, int] = None, # crop coordinate for the image
scale_basis: Optional[float] = None, # scaling of adata.obsm[basis] wrt to image
```
This for example allows to do something like this:
```python
img = adata.uns["spatial"]["V1_Adult_Mouse_Brain"]["images"]["hires"]
scalef = adata.uns["spatial"]["V1_Adult_Mouse_Brain"]["scalefactors"][
    "tissue_hires_scalef"
]
sc.pl.embedding(
    adata,
    color="leiden",
    scale_basis=scalef,
    img=img,
    basis="spatial",
    size=10,
    groups=["0"],
)
```

<details>
<summary>Details</summary>

image

</details>

Note that if I was using source image (veryhigh res) I wouldn't need the scale_basis arg (default to None). Also note that in this case the scatterplot is `scatter` and not `circles`.
The same operation with spatial would look like this (and here it would be `circles`):

```python
sc.pl.spatial(adata, color="leiden", groups=["0"])
```
<details>
<summary>Details</summary>

image

</details>

Finally, all the image processing part is removed from embedding and only present in spatial

--------------------

>  No behaviour changes in embedding if the basis is called "spatial" vs anything else, this should be triggered by calling the spatial function.

this is addressed, embedding changes behaviour only if img is passed, but has nothing to do with spatial, there is a small trick, and has to do with `ax.invert_yaxis()`. See following point

--------------------

> When spatial is called, it’s always shapes being drawn on an image. If there isn’t an image passed, an empty image would be generated. There would be no scatter plot case here.

I played around with this and decided to go against. Here's the following reasons
- if no img is passed, then we should assume that also no `scale_basis` is provided/available. Thus, the empty img to be created has to be of the size of the spatial coordinates system. In the case of visium (but would be even worse for larger field of views) the "blank source image" would be very often a 10k * 10k empry array. This slows down the plotting and create an unneccesary large object
- if no img is passed, there really shouldn't be any need for using `circles` instead of `scatter` , since there is no notion of "spot radius" or "spot size" (this was my first idea since the very beginning, but eventually agreed to still use scale factor. This is also the reason why test is failing with empty visium).

However, if no img is passed, when calling spatial the scatterplot should still have inverted coordinates (because we assume origin to be top left). I ended up simply setting `img = _empty` and adding it in embedding:
```python
if img is _empty:
	ax.invert_yaxis()
```
This is the behviour
```python
sc.pl.embedding(adata, color="leiden", basis="spatial")
```

<details>
<summary>Details</summary>

image


</details>

```python
sc.pl.spatial(adata, color="leiden", img_key=None)
```

<details>
<summary>Details</summary>

image


</details>

----------------

TO summarize, what `sc.pl.spatial` does is:
- if an image is present, process and scale accordingly and use `circles` instead of `scatter`
- if an image is not present, use `scatter` but invert coordinate since expected origin is top left

Furthermore, `sc.pl.embedding` now simply support the possibility to add an image in the background and accepts the relevant arguments needed for the image to be displayed correctly and modified, as well as a scale_basis argument to match the image coordinate if needed.

That's it, looking forward to hear what are your thoughts and if you agree with current behaviour I'll go on and changing the docs as well as writing more tests (especially for first example where `groups` is used, where the background color is different)

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -1142,25 +1138,48 @@ def _basis2name(basis):
...
+    else:
+        raise KeyError(
+            f"Could not find basis {basis} in `adata.obsm`\n"
+            f"while calling plotting._tools._process_image."
+        )
+
+    # get scale factor
     scalef_key = f"tissue_{img_key}_scalef"
+    if scalef_key in spatial_data['scalefactors'].keys():
+        scale_basis = spatial_data['scalefactors'][scalef_key]
```
ivirshup: Why `scale_basis` and not `scale_factor`? I feel like `scale_factor` would be more obvious since it's more directly related to the key used.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -314,23 +314,20 @@ def embedding(
...
-                ax.set_ylim(img_coord[3], img_coord[2])
-            elif img_key is None and library_id is not None:
-                # order of magnitude similar to public visium
-                size_spot = 70 * size
+            # plot img in background
+            if img is not None and img is not _empty:
+                ax.imshow(img, cmap=cmap_img, alpha=alpha_img)
+                if crop_coord is not None:
+                    ax.set_xlim(crop_coord[0], crop_coord[1])
+                    ax.set_ylim(crop_coord[3], crop_coord[2])
```
ivirshup: I think it might actually be okay to allow `crop_coord` with non-image plotting. What do you think?

ivirshup (commented):
Great, thank you for doing this! I'm much happier with how this is looking.

Two main changes I would like you to consider:

* Can `scale_factor` be removed as an argument to `embedding`, and instead have that handling occur inside `spatial`?
* Can an image be passed directly into `spatial`? I'd prefer this as being the "expert users" interface for plotting over an image, and think passing an image to `embedding` could be removed altogether in the future.

------------------------

In this example:

```python
sc.pl.embedding(
    adata,
    color="leiden",
    scale_basis=scalef,
    img=img,
    basis="spatial",
    size=10,
    groups=["0"],
)
```

Can we remove `scale_basis` as an argument to `embedding`? I would prefer that the coordinates could be scaled and passed in transformed in this case. I don't like that the scale factor doesn't get applied to the size here, just the coordinates.

There would just be a some normalization in  `embedding` so:

* If `isinstance(basis, str)` get the coordinates from the `adata`
* Otherwise we assume it's already an array, and make sure it's the right shape

---------------------------

Re: not passing an image: the size is a good point. I was hoping we could pass an implicit array (like `np.broadcast_to(np.uint8(1), (1000, 1000, 4))`), but it looks like Matplotlib make it real.

One possible alternative is to make it so if `spatial` is called and neither an `image` or `scale_factor` can be resolved an error is we throw an error.

> if no img is passed, there really shouldn't be any need for using circles instead of scatter , since there is no notion of "spot radius" or "spot size" (this was my first idea since the very beginning, but eventually agreed to still use scale factor. This is also the reason why test is failing with empty visium).

My thinking here was:

* It would make it easier to get a plot that looks the same, just minus the background
* It would make it easier to allow different shapes that circles in the future, e.g. cell polygons/ hexagons.

I can live with this not changing this version, however.

--------------------------------

> Furthermore, sc.pl.embedding now simply support the possibility to add an image in the background and accepts the relevant arguments needed for the image to be displayed correctly and modified, as well as a scale_basis argument to match the image coordinate if needed.

So, I'm not sure that long-term `embedding` should support passing an image. Right now I think that should be a separate function, that happens to share a lot of the internals. I see `spatial` calling `embedding` as more of an implementation detail that we can change.

Basically, my thought here is that a lot of the internals of `sc.pl.embedding` are more generally useful, and don't need to be tied to a specific plot type (or plotting library). It would be nice if we could use these internals to implement other functions instead of adding more features into one function.

This is what I meant by:

> I think the `spatial` function should be used just for the case when you want to overlay points on an image, while `embedding` would only be used for scatter plots.

scanpy/plotting/_tools/scatterplots.py
giovp: yes, done

scanpy/plotting/_tools/scatterplots.py
giovp: ok changed it

giovp:
> Can scale_factor be removed as an argument to embedding, and instead have that handling occur inside spatial?

no, otherwise I would have to modify the adata or pass a copy, and this would break other functionalities (first that come to mind, categorical colors saved in adata.uns)

> Otherwise we assume it's already an array, and make sure it's the right shape

what do you mean by that>? What you are proposing is to pass the adata.obsm as array in question and not as a string basis right? Is that possible now>? Would be happy to do that but don't want to create and pass an adata copy

> Can an image be passed directly into spatial? I'd prefer this as being the "expert users" interface for plotting over an image, and think passing an image to embedding could be removed altogether in the future.

done, also there is no image in embedding now. Everything is handled by spatial. It is possible to do something like this:

```python
img = adata.uns["spatial"]["V1_Adult_Mouse_Brain"]["images"]["hires"]
scalef = adata.uns["spatial"]["V1_Adult_Mouse_Brain"]["scalefactors"][
    "tissue_hires_scalef"
]
sc.pl.spatial(
    adata,
    color="leiden",
    scale_factor=scalef,
    img=img,
    size=100,
    basis="spatial",
    groups=["0"],
)
```
image

I would still like to have this in 1.7 if possible, I can write docs and additional tests real quick tomorrow early morning

ivirshup:
Oh, wow, sorry! I completely missed your comment here!

> What you are proposing is to pass the adata.obsm as array in question and not as a string basis right? 

What I was thinking: if it's a string get the array from `obsm`, if it's an array, check that it's shape is right, then use the array directly.

>  It is possible to do something like this

That looks great, thanks!

> I would still like to have this in 1.7 if possible, I can write docs and additional tests real quick tomorrow early morning

For sure! I was waiting on this actually, just managed to miss any notifications about it. Sorry again about that!

giovp:
> What I was thinking: if it's a string get the array from obsm, if it's an array, check that it's shape is right, then use the array directly.

I'm really sorry but I still don't get it 😅 . What I unnderstand is to modify `adata.obsm[spatial]` in `pl.spatial` and pass that to emebdding. However, I don't want to modify the adata in place or pass a copy. Maybe I'm missing something fundamental, but I could see doing this only if modifying the adata or passing the spatial coordinates as array (but is it possible in sc.pl.embedding? I couldn't find a way).

For the rest, I've added docs and added a test, should be ready to go (?). Thanks again for bearing with me.

ivirshup:
Are the squidpy notebooks ready? It would be good to make sure any changes I make don't mess with how you're using this function over there.

------------------------------

A few changes I would like to get your opinion on:

## Similar cropping regardless of img (even with img=None)

This code:

```python
with plt.rc_context({"figure.dpi": 120}):
    sc.pl.spatial(glio, color="log1p_total_counts", img_key="hires")
    sc.pl.spatial(glio, color="log1p_total_counts", img_key="lowres")
    sc.pl.spatial(glio, color="log1p_total_counts", img_key=None)
```

<details>
<summary> Currently generates these plots (using this PR) </summary>

image

</details>

Would it be better if the circles were consistently layed out, the plots were the same shape, and all had the same size?

<details>
<summary> Like this </summary>

image

</details>

These changes are:

* Getting the spot size even if the image isn't used
* Using `ax.set_aspect("equal")` when there is no image, so that the aspect ratio is equivalent (coordinates are assumed to be pixel space)
* If `crop_coords` is not passed, use the cropping matplotlib would have used if there was no image. This is done by getting the axis limits before the image is added


## Easier manual plotting

I would like it to be reasonable to get similar images via manually passing the img, and not. So would this be a reasonable API:

```python
adata_no_spatial = adata.copy()
del adata_no_spatial.uns["spatial"]

library_id = "Parent_Visium_Human_Glioblastoma"
img_key = "hires"

spatial_dict = adata.uns["spatial"][library_id]
img = spatial_dict["images"][img_key]
scale_factor = spatial_dict["scalefactors"][f"tissue_{img_key}_scalef"]
size = spatial_dict["scalefactors"]["spot_diameter_fullres"]

# These are equivalent
sc.pl.spatial(adata_no_spatial, color="log1p_total_counts", img=img, scale_factor=scale_factor, size=size)
sc.pl.spatial(adata, color="log1p_total_counts", img_key=img_key)

# As are these
sc.pl.spatial(adata_no_spatial, color="log1p_total_counts", size=size)
sc.pl.spatial(adata, color="log1p_total_counts", img_key=None)
```

This largely follows from the previous, with the exception that if `size` is passed when no image is present, it's the radius.

It might actually just make more sense to have a parameter for `spot_size` when I think about it more. The presence or absence of `spot_size` (and whether it could be found) would control whether `circles` are used.


## Crop coord being in data-space, not pixel space

I feel like it would make sense for these to crop to the same part of the image or embedding:

```python
crop_coord = (...)
sc.pl.spatial(adata, crop_coord=crop_coord, img_key="hires")
sc.pl.spatial(adata, crop_coord=crop_coord, img_key="lowres")
sc.pl.spatial(adata, crop_coord=crop_coord, img_key=None)
```

giovp:
looks really great, I'd say Yes! to pretty much everything, thanks for looking into this.
> Getting the spot size even if the image isn't used

so this is something we are going back and forth a lot, I still think it has pros and cons, and also agree with the point below re having a separate argument `spot_size`. 
If `sc.pl.spatial` *always* plot `circles` and not `scatter`, then the size of the radius needs to be inferred from the data: this makes me a bit worried for the different cases that could arise. I understand that is much nicer that the function returns always the same type of plot (circles), but it might be a bit forcing in this context (given the heterogeneity of the data).

What I would agree instead is to pass the `size_spot` and use circles, otherwise simply use scatter (and then set size). What do you think? 

> Using ax.set_aspect("equal") when there is no image, so that the aspect ratio is equivalent (coordinates are assumed to be pixel space)
this is really nice, I'd say yes in principle, would like to try it out though for couple of plots.

> If crop_coords is not passed, use the cropping matplotlib would have used if there was no image. This is done by getting the axis limits before the image is added

this is also fine and probably cleaner than having an heuristic for the offset.

> I feel like it would make sense for these to crop to the same part of the image or embedding:

missed this, ok yes it makes sense, then metric of `crop_coord` is the same as `adata.obsm["spatial"]`.

Thank you!

ivirshup:
Merging this, docs are failing to build from intersphinx since numpy's docs are down. This commit has built on read the docs before, so I'm happy to assume it still does.

Thanks @giovp!

giovp:
thank you @ivirshup 🙏 🎉

PR #1516, opened 2020-12-01
Set up Azure Pipelines with initial configuration

Mirkazemi:


ivirshup:
This is fantastic, thank you!

A few things I'm unclear on:

* Why is this PR getting a build if there is no [`pr` trigger entry](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#pr-triggers) in the `yaml`?
* Why isn't travis running on this PR? It might be that we've turned off branch CI since it was causing double runs with branches on this repo which were being used in PRs, but I thought it would still trigger once a pr was made.

I think I'm just going to try and merge this, since it seems to be working. We can fine tune it via PRs as we go.

flying-sheep:
> Why is this PR getting a build if there is no `pr` trigger entry in the yaml?

See 3 paragraphs down:

> If no pr triggers appear in your YAML file, pull request validations are automatically enabled for all branches, as if you wrote the following pr trigger. This configuration triggers a build when any pull request is created, and when commits come into the source branch of any active pull request.
> 
> ```
> pr:
>   branches:
>     include:
>     - '*'  # must quote since "*" is a YAML reserved character; we want a string
> ```

PR #1517, opened 2020-12-02
Modifications to azure pipeline

ivirshup:
I'm going to use this PR to test some modifications to the azure pipelines. First off, can we lint in the same job as running tests.


ivirshup:
@Mirkazemi

PR #1520, opened 2020-12-02
New pip resolver

flying-sheep:
@ivirshup why Ubuntu 16.04? Isn’t that super old?

### DIFF
--- a/scanpy/__init__.py
+++ b/scanpy/__init__.py
@@ -22,6 +22,7 @@ set_figure_params = settings.set_figure_params
 
 # has to be done at the end, after everything has been imported
 import sys
+sys.modules.update({f'{__name__}.{m}': globals()[m] for m in ['tl', 'pp', 'pl']})
 from ._utils import annotate_doc_types
 annotate_doc_types(sys.modules[__name__], 'scanpy')
 del sys, annotate_doc_types
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -28,7 +28,6 @@ jobs:
   - displayName: 'Install dependencies'
     script: |
       python -m pip install --upgrade pip
-      pip install docutils sphinx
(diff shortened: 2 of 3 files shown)


scanpy/__init__.py
```diff
@@ -22,6 +22,7 @@
 
 # has to be done at the end, after everything has been imported
 import sys
+sys.modules.update({f'{__name__}.{m}': globals()[m] for m in ['tl', 'pp', 'pl']})
```
ivirshup, line 25: Why was this added?

flying-sheep: This makes it so that `sc.tl` and so on is a real module python knows about, instead of a global variable that happens to point to a module object. Sphinx 3.2 got more strict about where it allows imports from. It won’t use attribute access for something we claim is a module, it’ll import it as a module.

Also people can now do `from scanpy.tl import ...` which is consistent to what we claim in our modules (I.e. that `sc.tl` is a module)

ivirshup: Is this better than having a `tl.py` which has: `from .tools import *`?

ivirshup:
AFAICT `ubuntu 16.04` is what they use in most of their examples. Considering many of our users will be on academic clusters, I think old-ish versions of linux are reasonable to test against. The alternative would probably be `ubuntu-18.04`, which we should switch to when `16` is out of support.

[Here are the options](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software).

scanpy/__init__.py
flying-sheep: I think so. That way the modules are actually identical and not two different modules containing references to the same objects.

flying-sheep:
We should have used the newer LTS from the start, 16.04 is out of support in 5 months.

PR #1524, opened 2020-12-02
Skip louvain-dependent tests

flying-sheep:
… if not installed

### DIFF
--- a/scanpy/tests/notebooks/test_paga_paul15_subsampled.py
+++ b/scanpy/tests/notebooks/test_paga_paul15_subsampled.py
@@ -2,7 +2,6 @@
 # Hematopoiesis: trace myeloid and erythroid differentiation for data of [Paul *et al.* (2015)](http://doi.org/10.1016/j.cell.2015.11.013).
 #
 # This is the subsampled notebook for testing.
-from importlib.util import find_spec
 from pathlib import Path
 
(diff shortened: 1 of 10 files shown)


scanpy/tests/notebooks/test_paga_paul15_subsampled.py
```diff
@@ -17,6 +18,7 @@
 FIGS = HERE / 'figures'
 
 
+@pytest.mark.skipif(not find_spec('louvain'), reason='needs package `louvain`')
 def test_paga_paul15_subsampled(image_comparer, plt):
     save_and_compare_images = image_comparer(ROOT, FIGS, tol=25)
```
ivirshup: Why not `importorskip`?

```suggestion
def test_paga_paul15_subsampled(image_comparer, plt):
    pytest.importorskip("louvain")
    save_and_compare_images = image_comparer(ROOT, FIGS, tol=25)
```

flying-sheep: Good point! Hmm, too bad that one can’t be used as a mark/decorator, would be more consistent.

Zethson:
Louvain is dead and this PR has tons of conflicts. I think it's fine if we close it for now.

PR #1529, opened 2020-12-03
new options to rank_genes_groups plots

### LINKED ISSUE #1796: pl.rank_genes_groups_[heatmap|dotplot|matrixplot|stacked_violin] don't work with gene_symbol in the same way as pl.rank_genes_groups

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

```python
import scanpy as sc
import pandas as pd
import numpy as np
import urllib.request
   
url = "https://humancellatlas.usegalaxy.eu/datasets/11ac94870d0bb33a8d1d2e6bc37a120e/display?to_ext=h5ad"
# 75 MB anndata
urllib.request.urlretrieve(url, '11ac94.h5ad')

adata = sc.read('11ac94.h5ad')

sc.settings.figdir = '.'

sc.pl.rank_genes_groups_dotplot(
    adata,
    save='.png',
    show=False,
    gene_symbols='Symbol',
    n_genes=10,
    log=False,
    use_raw=False,
    )
```

```pytb
  File "/private/var/folders/23/wwh84vd95rncnf5mx753cpgh0000gp/T/tmpenni0vju/job_working_directory/000/11/configs/tmp4tnm4xxb", line 15, in <module>
    sc.pl.rank_genes_groups_dotplot(
  File "/usr/local/lib/python3.8/site-packages/scanpy/plotting/_tools/__init__.py", line 647, in rank_genes_groups_dotplot
    return _rank_genes_groups_plot(
  File "/usr/local/lib/python3.8/site-packages/scanpy/plotting/_tools/__init__.py", line 426, in _rank_genes_groups_plot
    _pl = dotplot(
  File "/usr/local/lib/python3.8/site-packages/scanpy/plotting/_dotplot.py", line 913, in dotplot
    dp = DotPlot(
  File "/usr/local/lib/python3.8/site-packages/scanpy/plotting/_dotplot.py", line 127, in __init__
    BasePlot.__init__(
  File "/usr/local/lib/python3.8/site-packages/scanpy/plotting/_baseplot_class.py", line 105, in __init__
    self.categories, self.obs_tidy = _prepare_dataframe(
  File "/usr/local/lib/python3.8/site-packages/scanpy/plotting/_anndata.py", line 1845, in _prepare_dataframe
    obs_tidy = get.obs_df(
  File "/usr/local/lib/python3.8/site-packages/scanpy/get/get.py", line 272, in obs_df
    obs_cols, var_idx_keys, var_symbols = _check_indices(
  File "/usr/local/lib/python3.8/site-packages/scanpy/get/get.py", line 167, in _check_indices
    raise KeyError(
KeyError: "Could not find keys '['ENSMUSG00000000682', 'ENSMUSG00000000782', 'ENSMUSG00000001403', 'ENSMUSG00000002222', 'ENSMUSG00000003038', 'ENSMUSG00000004207', 'ENSMUSG00000004642', 'ENSMUSG00000005732', 'ENSMUSG00000015176', 'ENSMUSG00000016256', 'ENSMUSG00000020423', 'ENSMUSG00000020641', 'ENSMUSG00000020649', 'ENSMUSG00000020914', 'ENSMUSG00000021213', 'ENSMUSG00000022051', 'ENSMUSG00000022528', 'ENSMUSG00000023087', 'ENSMUSG00000023274', 'ENSMUSG00000023944', 'ENSMUSG00000024397', 'ENSMUSG00000025270', 'ENSMUSG00000025362', 'ENSMUSG00000025647', 'ENSMUSG00000025889', 'ENSMUSG00000025980', 'ENSMUSG00000026234', 'ENSMUSG00000026238', 'ENSMUSG00000026581', 'ENSMUSG00000026605', 'ENSMUSG00000026770', 'ENSMUSG00000027306', 'ENSMUSG00000027447', 'ENSMUSG00000027863', 'ENSMUSG00000028212', 'ENSMUSG00000028639', 'ENSMUSG00000029322', 'ENSMUSG00000029530', 'ENSMUSG00000029580', 'ENSMUSG00000029922', 'ENSMUSG00000031004', 'ENSMUSG00000031239', 'ENSMUSG00000032231', 'ENSMUSG00000034165', 'ENSMUSG00000034634', 'ENSMUSG00000036371', 'ENSMUSG00000038642', 'ENSMUSG00000038871', 'ENSMUSG00000038943', 'ENSMUSG00000039221', 'ENSMUSG00000040274', 'ENSMUSG00000044533', 'ENSMUSG00000045328', 'ENSMUSG00000046364', 'ENSMUSG00000049775', 'ENSMUSG00000052305', 'ENSMUSG00000053044', 'ENSMUSG00000053977', 'ENSMUSG00000054717', 'ENSMUSG00000056290', 'ENSMUSG00000057113', 'ENSMUSG00000057322', 'ENSMUSG00000060802', 'ENSMUSG00000061232', 'ENSMUSG00000061311', 'ENSMUSG00000061787', 'ENSMUSG00000061983', 'ENSMUSG00000062328', 'ENSMUSG00000063316', 'ENSMUSG00000063856', 'ENSMUSG00000064351', 'ENSMUSG00000066551', 'ENSMUSG00000067288', 'ENSMUSG00000069917', 'ENSMUSG00000069919', 'ENSMUSG00000073411', 'ENSMUSG00000073421', 'ENSMUSG00000073940', 'ENSMUSG00000076498', 'ENSMUSG00000076749', 'ENSMUSG00000076752', 'ENSMUSG00000076928', 'ENSMUSG00000078812', 'ENSMUSG00000079435', 'ENSMUSG00000079523', 'ENSMUSG00000094777']' in columns of `adata.obs` or in adata.var['Symbol']."
```

The same happens with the equivalents `pl.rank_genes_groups_*` for heatmap, matrixplot and stacked violin. In contrast, with the same dataset:

```
sc.pl.rank_genes_groups(
    adata,
    save='.png',
    show=False,
    n_genes=10,
    gene_symbols='Symbol',
    fontsize=8,
    ncols=4,
    sharey=True)
``` 

works well. I suspect that this is due to the dataset not having symbols as index:

```
>>> adata.var
                                    ID          Symbol    NA.  ...  dispersions_norm      mean       std
index                                                          ...
ENSMUSG00000103922  ENSMUSG00000103922          Gm6123  False  ...         -2.190900  0.001387  0.043064
ENSMUSG00000033845  ENSMUSG00000033845          Mrpl15  False  ...         -0.469772  0.349803  0.824321
ENSMUSG00000025903  ENSMUSG00000025903          Lypla1  False  ...         -0.396586  0.136585  0.531308
ENSMUSG00000104217  ENSMUSG00000104217         Gm37988  False  ...         -1.059987  0.036154  0.246548
ENSMUSG00000033813  ENSMUSG00000033813           Tcea1  False  ...         -0.342200  0.578174  1.001880
...                                ...             ...    ...  ...               ...       ...       ...
ENSMUSG00000094915  ENSMUSG00000094915      AC168977.2  False  ...         -3.593359  0.000265  0.015330
ENSMUSG00000079808  ENSMUSG00000079808      AC168977.1  False  ...         -0.756427  0.023002  0.199670
ENSMUSG00000095041  ENSMUSG00000095041      AC149090.1  False  ...          0.816554  0.149080  0.578574
ENSMUSG00000063897  ENSMUSG00000063897  CAAA01118383.1  False  ...          0.403835  0.041207  0.304460
ENSMUSG00000095742  ENSMUSG00000095742  CAAA01147332.1  False  ...         -0.307310  0.001870  0.058347
```

So when `pl.rank_genes_groups` is being called, some adequate mapping is being done from the index of `adata.var` to Symbol, whereas when the other methods are called, this mapping is skipped, hence the error. I would expect the different `pl.rank_genes_groups_*` methods to behave as `pl.rank_genes_groups` when it comes to symbols not being used as indexes.

Thanks!

#### Versions

<details>

>>> sc.logging.print_versions()
WARNING: If you miss a compact list, please try `print_header`!
-----
anndata     0.7.5
scanpy      1.7.2
sinfo       0.3.1
-----
PIL                 8.1.2
anndata             0.7.5
cairo               1.20.0
cffi                1.14.5
colorama            0.4.4
cycler              0.10.0
cython_runtime      NA
dateutil            2.8.1
get_version         2.1
h5py                2.10.0
igraph              0.8.3
joblib              1.0.1
kiwisolver          1.3.1
legacy_api_wrap     0.0.0
leidenalg           0.8.3
llvmlite            0.36.0
louvain             0.7.0
matplotlib          3.3.4
mpl_toolkits        NA
natsort             7.1.1
numba               0.53.1
numexpr             2.7.3
numpy               1.19.5
packaging           20.9
pandas              1.1.5
pkg_resources       NA
pyparsing           2.4.7
pytz                2021.1
scanpy              1.7.2
scipy               1.5.3
setuptools_scm      NA
sinfo               0.3.1
six                 1.15.0
sklearn             0.24.1
sphinxcontrib       NA
tables              3.6.1
texttable           1.6.3
typing_extensions   NA
zipp                NA
-----
Python 3.6.13 | packaged by conda-forge | (default, Feb 19 2021, 05:36:01) [GCC 9.3.0]
Linux-4.19.121-linuxkit-x86_64-with-debian-10.8
2 logical CPU cores
-----
Session information updated at 2021-04-12 15:14

</details>

fidelram:
This PR adds the following to `rank_genes_groups_*` plots:
 * Allows `n_genes` to be a negative number to plot the bottom ranked `n_genes`. Useful to check what is not being expressed on a cluster.
 * Added `gene_names` to  `rank_genes_groups_matrixplot` and `rank_genes_groups_dotplot`. This option is for checking a 
 given list of genes instead of the top or bottom ranked genes. This allows to check for example log fold change of p-values for
 the given genes.
 * `gene_symbols` was not working properly. Now it is.


LuckyMD:
Are the bottom ranked really not expressed, or just not differentially expressed? The former could still have significant p-values. I guess I wonder if you rank by logFC or by adjusted p-value.

fidelram:
@LuckyMD  genes at the bottom simply have the lowest rank but they could be expressed. By default the ranking is taking directly from  `sc.get.rank_genes_groups_df` which ranks the genes by log fold change. Bottom genes tend to have significant p-value. 

To make this more transparent we can add a parameter to  select how to rank for example by  p-value or log fold change. 

But, first I need to figure out what is this mess with the new tests....

LuckyMD:
Yeah, I recently found out that `rank_genes_groups` doesn't just filter for +ve logFC, but ranks by it. I used to think that it's a filtering and you needed to do A vs B and B vs A to get all results ;).

fidelram:
@LuckyMD Your impression is right, but after changes to `sc.tl.rank_genes_groups` were introduced, now by default the full list of genes is returned and is not necessary to do A vs. B and then B vs A. In my impression this change opened new opportunities, like looking at specific genes or looking at the bottom ranked.

However, I think it is worth to make the ranking and selection more transparent and I am open here for suggestions. For background the current state is:
 * `sc.tl.filter_rank_genes_groups` can be used to filter the results in different ways like fold change or fraction of cells expressing the gene in a given cluster or outside a given cluster. The goal was to allow identification of markers quite specific to a cluster. Although, I made this function I think we should not use it as it is not up to date and creates confusion because it replaces genes  by NaNs to allow the filtering. This was pre `sc.get.rank_genes_groups_df` and some other changes. Also is complicated to use because is run, a new rank_genes_groups key is created with the filtering and this key has to be added to the plotting functions to see the results. 
 * The `sc.pl.rank_genes_groups_*` plots have the option `min_logfoldchage` for filtering. I find this useful but limited because is not possible to filter by p-value for example.

As a solution, plots could have a filtering option that uses pandas query syntax like: `filtering='logfoldchange>1 & p-value<0.0001'`  and for the sorting something like  `sortby=('logfoldchange', 'ascend')`. What do you think?

LuckyMD:
I like your suggestions. Especially the `filter_rank_genes_groups` use makes a lot of sense to me. The one thing I would suggest to take into account is that some of these filtering steps can be done before significance testing and therefore you would not have to perform multiple testing correction on the filtered out genes. This may be quite useful to some. That precludes filtering on p-value though. It also makes a case for filtering already in `rank_genes_groups` rather than in `sc.get`.

fidelram:
My suggestion is to do filtering on the fly. 

What I am not so sure is how to nicely achieve this without creating too many parameters and/or too much typing that is difficult to remember.

scanpy/plotting/_tools/__init__.py
```diff
@@ -382,26 +388,41 @@ def _rank_genes_groups_plot(
...
-            # select genes with given log_fold change
-            genes_list = df[df.logfoldchanges > min_logfoldchange].names.tolist()[
-                :n_genes
-            ]
-        else:
-            # get all genes that are 'non-nan'
-            genes_list = [
-                gene for gene in adata.uns[key]['names'][group] if not pd.isnull(gene)
-            ][:n_genes]
+    gene_symbols = kwds.get('gene_symbols', None)
```
ivirshup: This should just be an argument if you want it to be an argument.

scanpy/plotting/_tools/__init__.py
```diff
@@ -382,26 +388,41 @@ def _rank_genes_groups_plot(
...
+        var_names = {}
+        gene_names = []
+        for group in group_names:
+            df = rank_genes_groups_df(adata, group, key=key, gene_symbols=gene_symbols)
+            if min_logfoldchange is not None:
+                # select genes with given log_fold change
+                df = df[df.logfoldchanges > min_logfoldchange]
+
+            if gene_symbols is not None:
+                df['names'] = df['symbol']
```
ivirshup: `"symbol"` shouldn't be hardcoded, as it's not hardcoded in the output of `sc.get.rank_genes_groups`

```suggestion
                df['names'] = df[gene_symbols]
```

scanpy/plotting/_tools/__init__.py
```diff
@@ -382,26 +388,41 @@ def _rank_genes_groups_plot(
...
+            gene_names = var_names
+    else:
+        # dict in which each group is the key and the n_genes are the values
+        var_names = {}
+        gene_names = []
+        for group in group_names:
+            df = rank_genes_groups_df(adata, group, key=key, gene_symbols=gene_symbols)
+            if min_logfoldchange is not None:
+                # select genes with given log_fold change
+                df = df[df.logfoldchanges > min_logfoldchange]
```
ivirshup: Why not pass this to `rank_genes_groups_df`?

scanpy/plotting/_tools/__init__.py
```diff
@@ -382,26 +388,41 @@ def _rank_genes_groups_plot(
...
-        gene_names.extend(genes_list)
+    if gene_names is not None:
+        var_names = gene_names
+        if isinstance(var_names, Mapping):
+            # get a single list of all gene names in the dictionary
+            gene_names = sum([list(x) for x in var_names.values()], [])
+        elif isinstance(var_names, str):
+            gene_names = [var_names]
+        else:
+            gene_names = var_names
```
ivirshup: I think it could be confusing how the values of the names are changing here.

Could they either create new values with new names, or can you add a comment saying what the normalized values should be after this code block? Also, is validation/ normalization generalizable, and could it go into a separate function?

scanpy/plotting/_docs.py
```diff
@@ -229,3 +229,39 @@
...
+
+doc_rank_genes_groups_values_to_plot = """\
+values_to_plot
+    The mean gene values are plotted by default. Alternatively, any other
+    values computed by `sc.rank_genes_groups` can be plotted. For example
+    log fold change or p-value.
+gene_names
+    Genes to plot. Sometimes is useful to pass a specific list of genes to check their
+    fold changes or p-values instead of the top/bottom genes. If a list of
+    genes is given, this overwrites the `n_genes` option. The gene_names
```
ivirshup: If the arguments are incompatible, I think it should be an error if they are both passed.

scanpy/plotting/_tools/__init__.py
```diff
@@ -1332,7 +1289,9 @@ def _get_values_to_plot(
 
         df_list = []
         for group in groups:
-            df = rank_genes_groups_df(adata, group, key=key)
+            df = rank_genes_groups_df(adata, group, key=key, gene_symbols=gene_symbols)
+            if gene_symbols is not None:
+                df['names'] = df['symbol']
```
ivirshup: Similar to above

```suggestion
                df['names'] = df[gene_symbols]
```

ivirshup (commented):
Initial review

Could you include some example of usage of the new stuff? There are a few parts I don't quite follow, and figure playing around with it would be the fastest way to check it out.

scanpy/plotting/_tools/__init__.py
fidelram: solved. This required adding the gene_symbols parameter to many functions. Thats probably the change that takes more lines.

scanpy/plotting/_tools/__init__.py
fidelram: Solved

scanpy/plotting/_tools/__init__.py
fidelram: Solved. I took the change to  rename `gene_names` to `var_names` as this is consistent with general usage. However, there is this mix all over between `gene` and `var`.  

After the code modification I hope that is is now clear the goal of that code which is to turn the different data types of var_names into a list.

scanpy/plotting/_tools/__init__.py
```diff
@@ -382,26 +389,41 @@ def _rank_genes_groups_plot(
...
+        # dict in which each group is the key and the n_genes are the values
+        var_names = {}
+        var_names_list = []
+        for group in group_names:
+            df = rank_genes_groups_df(
+                adata,
+                group,
+                key=key,
+                gene_symbols=gene_symbols,
+                log2fc_min=min_logfoldchange,
```
fidelram, line 466: Would be great to homogenise the parameters. Here we have `log2fc_min` and `min_logfoldchange`. Do we have a strategy to make some parameters obsolete?

Also, would be good to allow the same filtering in the plots as `sc.get.rank_genes_groups_df`, namely `log2fc_min`, `log2fc_max` and `pval_cutoff`.

scanpy/plotting/_docs.py
fidelram: the problem with raising and error is that `n_genes` is not optional and defaults to 10. One solution is to make this parameter optional and if both `n_genes` and `var_names` are None, then the default is `n_genes=10`. Thus we don't break any previous functionality.

scanpy/plotting/_tools/__init__.py
fidelram: solved.

scanpy/plotting/_tools/__init__.py
fidelram: done. Would be great to homogenize the parameter names.

scanpy/plotting/_docs.py
fidelram: solved.

scanpy/plotting/_tools/__init__.py
fidelram: I think we can leave this as an open question.

ivirshup: We have the `deprecated_arg_names` decorator for this

ivirshup (commented):
What do you mean by this?

> gene_symbols was not working properly. Now it is.

To me, the new test looks wrong. You are updating `adata.var` with gene symbols, but `adata.raw` is being used for DE testing. Since `adata.raw` can have a completely different `.var` and `.var_names` than `adata`, gene symbols in `adata.var` should not be used when the tests were on `adata.raw`.

------------------------

> The sc.pl.rank_genes_groups_* plots have the option min_logfoldchage for filtering. I find this useful but limited because is not possible to filter by p-value for example.

Could these just be arguments which are equivalent to those from `sc.get.rank_genes_groups_df`?

------------------------

Side note: @LuckyMD, ordering is done by `score`, not `logfc`, right?

ivirshup:
Re: #1649

Does this still need a max fold change argument?

More generally, how complex do we want the filtering available through these functions (and `sc.get.rank_genes_groups_df`) to be? Is it most straight forward to recommend passing the gene names, and recommend users generate these by manipulating the dataframe returned by `rank_genes_groups_df`?

scanpy/plotting/_tools/__init__.py
```diff
@@ -799,25 +832,52 @@ def rank_genes_groups_matrixplot(
...
+
+    A list specific genes can be given to check their log fold change. If a
+    dictionary, the dictionary keys will be added as labels in the plot.
+    >>> var_names = {{"T-cell": ['CD3D', 'CD3E', 'IL32'],
+    ...               'B-cell': ['CD79A', 'CD79B', 'MS4A1'],
+    ...               'myeloid': ['CST3', 'LYZ'] }}
+    >>> sc.pl.rank_genes_groups_matrixplot(adata,
+    ... var_names=var_names, values_to_plot="logfoldchanges", cmap='bwr',
+    ... vmin=-4, vmax=4, min_logfoldchange=3, colorbar_title='log fold change')
+
```
ivirshup: Could these example be formatted a little more nicely?

ivirshup:
Just checking back in on this PR. Did we want to include the new options in this, or are we happy with keeping the scope to just a general clean-up + making the `gene_symbols` argument work?

ivirshup (commented):
### Failing tests:

- [x] master_ranked_genes_dotplot_gene_names
  - looks like order of groups changed
- [x] master_ranked_genes_matrixplot_n_genes_negative
  - Orders of groups on both axis changed
- [x] master_ranked_genes_matrixplot_gene_names_symbol
  - Orders of groups on rows changed

All three of the above were added in this PR

This one was not:

- [x] ranked_genes_matrixplot
  - very different, not sure what's going on
  - I think the existing image was wrong, because it was showing expression from `raw.X` and `use_raw=False`. I assume this is what `use_raw` is meant to do here, but find the presence of the argument a bit confusing. Anyways, fixing by updating the reference image.


<details>
<summary> What was going on </summary>

Expected: 
master_ranked_genes_matrixplot

Actual:
master_ranked_genes_matrixplot

Basically it was plotting the values from `raw` when `use_raw` was `False`. Overall the interaction between `use_raw` and `rank_genes_groups`, especially with the addition of `gene_symbols` is confusing.


As evidence, when use_raw is False most expression is "washed out" by the high expression of a few celltype specific genes:

```python
sns.heatmap(
    (
        sc.get.obs_df(pbmc, ["LYZ", "louvain", "CST3", "CD74", "MZB1"], use_raw=True)
        .groupby("louvain").mean()
    ),
    cmap="viridis"
)
```

image


```python
sns.heatmap(
    (
        sc.get.obs_df(pbmc, ["LYZ", "louvain", "CST3", "CD74", "MZB1"], use_raw=False)
        .groupby("louvain").mean()
    ),
    cmap="viridis"
)
```

image

</details>

### Adding tests

- [x] Test that gene_names is working
- [x] Test that n_genes is working
  - I'm not actually sure how to do this. Ideally I do a test with two groups, then `n_genes` and `-n_genes` show the same set of genes in the plot, but the order is different.
  - Ended up checking this against `var_names`, since they should generate the same plots if you explicitly pass the top or bottom `n_genes`. This also checks that these functions work when `var_names` is passed, which not all of them did.

### Failing doc builds

- [x] I think it was some invalid rst, see if fixing that does the trick.

ivirshup:
@fidelram, I've updated this so the tests pass, and think I've caught a few more bugs. Hopefully I didn't misinterpret your intent here, but I'm merging as we'd like to get a release out. Please let me know if I've messed anything up!

PR #1538, opened 2020-12-10
Simplify embeddings a bit

ivirshup:
Some effort to simplify how the embedding plots are handled.

Right now this consists of simplifying the `components` argument to `dimensions`. Dimensions is a list of collections of ints. Each element in this list has the length of the number of dimensions to be plotted. `components` is no longer used once it can be transformed into dimensions.

What does this do?

* Let's us delete `_get_data_points`, an awful function
* Get rid of `data_points` a list of coordinates that most code assumed would only ever have one element
* Move spatial specific code to spatial specific functions
* Gets rid of edge cases where `components` was either `None` or `[None]` (not sure how)

Side note, also made a modification to a testing fixture that had been making tests fail when run out of order.

### Should `dimensions` be exposed? If so, should `components` be deprecated?

I think it's weird to pass the dimensions as strings `"1,2"` as opposed to dimensions `(0, 1)`.

* Why is it one indexed?
* Isn't it the same amount of keystrokes?
* How useful is `"all"`?

I also think it's weird that the amount of plots generated is the product of `components` and `color`. This differs from every other "vectorized" argument to `embedding`. Changing arguments and deprecating `components` would be an opportunity to change this.

### If dimensions should be exposed, how many places does this need to be implemented?

We use `components` as an argument in a number of places in the codebase. Should we think about doing a large-scale replacement?

### I broke a plotting test. I can't tell the difference

@giovp, does this look fine to you? It's the spatial plots with no image.

<details>
<summary> </summary>

Expected

master_spatial_visium_empty_image

Actual

master_spatial_visium_empty_image

Diff

master_spatial_visium_empty_image-failed-diff

</details>

TODO:

- [ ] Deprecate components repeated behaviour
- [x] Finalize `dimensions` arg


fidelram:
It is great that you are looking into this. Can you check if it is possible to remove the need to type 'color' and simple accept the second parameter as the 'color' parameter (eg instead of `sc.pl.umap(adata, color='clusters')` -> `sc.pl.umap(adata, 'clusters')`

About the changes that you suggest: I have concerns with breaking previous functionality and I wonder what is your position with respect to this. The reason why the dimensions is a string like "1,2", was to avoid breaking previous usage from the very firsts versions of scanpy but maybe you have some good ideas for transitioning this from a string to a tuple. 

The starting number is not 0 because is consistent with usage as in 'principal component 1' or 'UMAP-1'. I don't think this should be changed even though it requires a bit of extra coding. 

For the plots being the product of `color` and `components`: this was to solve the unlikely case in which you want to plot n colors using m dimensions. I don't have an opinion on this as I think is a corner case and have never used this functionality. 

For your question about replacing `components` by `dimensions`. We need to be careful here because in many places the use of components is in the context of PCA as in `sc.pp.neighbors` with the parameter `n_pcs`. I think that  the replacement of `components` by `dimensions` should only be done for the embedding functions.

ivirshup:
> It is great that you are looking into this. Can you check if it is possible to remove the need to type 'color' and simple accept the second parameter as the 'color' parameter (eg instead of sc.pl.umap(adata, color='clusters') -> sc.pl.umap(adata, 'clusters')

👍 

> The reason why the dimensions is a string like "1,2", was to avoid breaking previous usage

The idea is to parse `components`, but allow you to directly pass the indices with dimensions.

> The starting number is not 0 because is consistent with usage as in 'principal component 1' or 'UMAP-1'. I don't think this should be changed even though it requires a bit of extra coding.

To me, I think it makes more sense to be consistent with python. I feel like it's very clear what is happening if the default argument is `sc.pl.pca(adata, dimensions=(0, 1))`. It makes it easier to work with programmatically if the values are equivalent to what you could use to index the array directly. For example, say you find the dimension which is maximally correlated with some gene. You can just pass the result of that into dimensions without having to remember to add 1.

> For the plots being the product of color and components: this was to solve the unlikely case in which you want to plot n colors using m dimensions. I don't have an opinion on this as I think is a corner case and have never used this functionality.

Cool. I feel like this can be useful, but it would be useful if I could choose which arguments it worked with. I think this is a different function call though.

For example, I might was to look at a gene under multiple embeddings, so pairwise combinations of `basis` and `colors`.

fidelram:
Regarding the use of components using zero based index I would like to know what @falexwolf opinion is.

falexwolf:
I'm fine with changing the convention to zero-pre-fix-based.

The original design was also due to diffusion maps having a "zero-component" that is not meant for visualization. But diffusion maps is a highly artificial case given the more modern embedding these days, and so I'd suggest to consistently implement the 0-based pythonic convention.

In diffusion maps, we can default to `(1, 2)` whereas in all other methods we can default to `(0, 1)`.

ivirshup:
> In diffusion maps, we can default to (1, 2) whereas in all other methods we can default to (0, 1).

👍, plus this way handling of the dimensions can be done in `sc.pl.diffmap` function and not `sc.pl.embedding` (e.g. `embedding` can be agnostic to the kind of embedding it's plotting).

ivirshup:
I've put off deprecating `components`, since that's a bunch more work and would need changes throughout the codebase. Will create an issue for it.

PR #1540, opened 2020-12-13
SAM: Updated default parameters to suit large datasets

atarashansky:
The original set of default parameters used by SAM were geared more towards smaller datasets. As scRNAseq throughput is ever-increasing, I tweaked the default parameters to be better suited for large datasets.


ivirshup:
Could this get a notice in the docs/ something in the release notes?

ivirshup:
Thanks for the update! Merged via #1595

Issue #1509, opened 2020-11-27
umap 0.5 compatibility


ivirshup:
It'd be good to make sure everything works with umap 0.5 before it get's released.

ivirshup:
It looks like the new release breaks most of our usage from (at least) a change in arguments to `simplicial_set_embedding` (ping @Koncopd).

<details>
<summary> Example error </summary>


```pytb
            n_epochs = 0 if maxiter is None else maxiter
>           X_umap = simplicial_set_embedding(
                X,
                neighbors['connectivities'].tocoo(),
                n_components,
                alpha,
                a,
                b,
                gamma,
                negative_sample_rate,
                n_epochs,
                init_coords,
                random_state,
                neigh_params.get('metric', 'euclidean'),
                neigh_params.get('metric_kwds', {}),
                verbose=settings.verbosity > 3,
E               TypeError: simplicial_set_embedding() missing 3 required positional arguments: 'densmap', 'densmap_kwds', and 'output_dens'
```

</details>

It looks like there is also a lot of cool stuff in the new release!

Koncopd:
Thanks, i'll check.

gokceneraslan:
This seems related to the new Densmap feature https://umap-learn.readthedocs.io/en/latest/densmap_demo.html (see https://www.biorxiv.org/content/10.1101/2020.05.12.077776v1)

Would be cool to support it in scanpy.

ivirshup:
@gokceneraslan, have you tried `densmap`? I'm curious to hear someone's thoughts on it.

I'm also wondering if adding it as a feature would be more complicated than adding a few pass-thru parameters to `sc.tl.umap`.

ivirshup:
Closed by #1601 and #1589

PR #1543, opened 2020-12-13
Restrict umap to <0.5

ivirshup:
Restricting umap to below `0.5` so we can fix changes before it's released (see #1509).


flying-sheep:
can you please make sure changes like this also end up in the flit branch?

PR #1551, opened 2020-12-18
Asymmetrical diverging colormaps and vcenter

gokceneraslan:
Hi,

I started playing around with `vcenter` and adding support for it in plotting functions. This allows us to do things like:

```python
import scanpy as sc

adata = sc.datasets.paul15()

sc.pp.log1p(adata)
sc.pp.scale(adata)
genes = ['Cst3', 'Car1', 'Cd34', 'Apoe', 'Top2a', 'Ccl5', 'Ctsw']
sc.pl.matrixplot(adata, genes, groupby='paul15_clusters', vmin=-2, vmax=5, vcenter=0, cmap='RdBu_r')
```

image

I haven't gone further before asking you if it makes sense at all, or not. If yes, I can try to write tests too. What do you think?


ivirshup:
I like this idea.

* While we should be conservative about adding new keywords, this fits well with `vmin` and `vmax`
* The docs for this argument should mention that users should pass a diverging palette with it, and probably have an example
* If `norm` is passed along at the same time, an error should be thrown
* It looks like there is a bunch of repeated code handling generating the `norm`, could this get put into a common utility function?

ivirshup:
> If norm is passed along at the same time, an error should be thrown

Following up on this a bit, I realized I didn't actually know what matplotlib would do if you passed `norm` and a bound, so I checked it out. Turns out they currently allow it, but it's deprecated, so throwing an error is the right thing to do.

```python
import vega_datasets
import matplotlib as mpl, matplotlib.pyplot as plt

iris = vega_datasets.data.iris()

norm = mpl.colors.LogNorm()

plt.scatter(
    iris["sepalLength"],
    iris["sepalWidth"],
    c=iris["petalLength"],
    norm=norm,
    vmin=3,
)
plt.colorbar()
```

```
 MatplotlibDeprecationWarning: Passing parameters norm and vmin/vmax 
simultaneously is deprecated since 3.3 and will become an error two minor releases 
later. Please pass vmin/vmax directly to the norm when creating it.
```

gokceneraslan:
> ```
>  MatplotlibDeprecationWarning: Passing parameters norm and vmin/vmax 
> simultaneously is deprecated since 3.3 and will become an error two minor releases 
> later. Please pass vmin/vmax directly to the norm when creating it.
> ```

Yeah, that actually re-ignited my idea of adding support for vcenter after upgrading mpl :)

gokceneraslan:
@fidelram Some help with the plotting tests is very much appreciated :) I don't get why only Python 3.6 test passes and others fail.

For example; why does [this plot](https://github.com/theislab/scanpy/blob/master/scanpy/tests/_images/master_matrixplot_with_totals.png) which is produced by [this code](https://github.com/theislab/scanpy/blob/master/scanpy/tests/test_plotting.py#L357) where the color legend is defined [here](https://github.com/theislab/scanpy/blob/master/scanpy/plotting/_matrixplot.py#L74) as `Mean expression in group` have the old title which is `Expression level in group`? If the "ground truth plots" in the "_images" folder are outdated, aren't the tests supposed to fail since the [this commit](https://github.com/theislab/scanpy/commit/d4d373ea58b9add4451091c5650d4da245d025dc#diff-421b3afbcd51c81c45f23dd7e8483697b68668de9172f124ec6bf2f0523840d9L95) i.e. `Expression level in group` -> `Mean expression in group`, but they don't fail, why?

Another example: I have updated [this file](https://github.com/theislab/scanpy/blob/master/scanpy/tests/_images/master_matrixplot_with_totals.png) but not [that file](https://github.com/theislab/scanpy/blob/master/scanpy/tests/_images/master_matrixplot.png) which seem both outdated. But Python 3.6 test still seems to pass. How come? Are plotting tests disabled in Python3.6?

Plotting tests are extremely tricky and make it pretty difficult to contribute I must say :/ I think twice or three times when I want to change anything about plotting... (not a complaint addressed to @fidelram but something we should consider as a team)

scanpy/plotting/_anndata.py
```diff
@@ -1607,7 +1608,8 @@ def correlation_matrix(
     **kwds
         Only if `show_correlation` is True:
         Are passed to :func:`matplotlib.pyplot.pcolormesh` when plotting the
-        correlation heatmap. Useful values to pas are `vmax`, `vmin` and `cmap`.
+        correlation heatmap. Useful values to pass are `vmax`, `vmin`,
+        `vcenter` and `cmap`.
```
ivirshup: Since we're doing special handling of these values, can they be explicit keyword arguments? I don't really like modifying the `**kwargs` dict. Also, `vcenter` isn't an argument to `pcolormesh`, right?

gokceneraslan: OK, this is done I think, we don't modify kwargs any more. Please have a look @ivirshup

scanpy/plotting/_anndata.py
```diff
@@ -911,6 +920,9 @@ def heatmap(
     show: Optional[bool] = None,
     save: Union[str, bool, None] = None,
     figsize: Optional[Tuple[float, float]] = None,
+    vmin: Optional[float] = None,
+    vmax: Optional[float] = None,
+    vcenter: Optional[float] = None,
```
ivirshup, line 924: Since we handle `norm` directly as well with these, I think it should be an explicit argument, that is documented.

I think it's bad style to interact directly with the `**kwargs` dict, and it can be avoided here.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -294,13 +297,18 @@ def embedding(
...
-                vmin, vmax, count, color_vector
+        if not categorical:
+            vmin_float, vmax_float, vcenter_float = _get_vminmaxcenter(
+                vmin, vmax, vcenter, count, color_vector
             )
+            normalize = _setup_colornorm(
+                vmin_float,
+                vmax_float,
+                vcenter_float,
+                kwargs.get('norm'),
```
ivirshup: Similarly, I think `norm` should be an explicit key word argument.

scanpy/plotting/_utils.py
```diff
@@ -1192,3 +1193,24 @@ def _get_basis(adata: anndata.AnnData, basis: str):
         basis_key = f"X_{basis}"
 
     return basis_key
+
+
+def _setup_colornorm(vmin=None, vmax=None, vcenter=None, norm=None):
```
ivirshup: Could this be called something like `check_colornorm`? I think it fits better with my idea of this function as argument validation/ processing.

scanpy/plotting/_baseplot_class.py
```diff
@@ -127,6 +130,9 @@ def __init__(
         self.groupby = [groupby] if isinstance(groupby, str) else groupby
         self.log = log
         self.kwds = kwds
+        self.vmin = vmin
+        self.vmax = vmax
+        self.vcenter = vcenter
```
ivirshup: Since these values (and `norm`) all get used together, what do you think about putting them all in one parameter? Just thinking about keeping the related values explicit linked by structure, and reducing the number of attributes.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -481,15 +493,16 @@ def _panel_grid(hspace, wspace, ncols, num_panels):
     return fig, gs
 
 
-def _get_vmin_vmax(
-    vmin: Sequence[VMinMax],
-    vmax: Sequence[VMinMax],
+def _get_vminmaxcenter(
+    vmin: Sequence[VMinMaxCenter],
+    vmax: Sequence[VMinMaxCenter],
+    vcenter: Sequence[VMinMaxCenter],
```
ivirshup: Could this be called inside of `_setup_colornorm` (/ `check_colornorm`) so all norm handling occurs in one call?

Whether this is a good idea depends on how common the case where only `float` values are allowed is.

scanpy/tests/test_plotting.py
```diff
@@ -816,19 +876,38 @@ def test_scatter_embedding_add_outline_vmin_vmax(image_comparer):
...
+            'X_umap',
+            color=['percent_mito', 'n_counts'],
+            norm=norm,
+            vmin=0,
+            vmax=1,
+            vcenter=0.5,
+            cmap='RdBu_r',
+        )
+    except ValueError:
+        pass
```
ivirshup: Use `pytest.raises` to check that an error is thrown.

I would like to see tests that passing `norm` directly works. Two ideas for this:

* If you construct the norm manually, it's equivalent to using the `v{min, max, center}` arguments
* Passing different norms generates different figures (e.g. the argument has an effect)

These could both be parameterized over different plotting functions and don't require plots added to the repo.

scanpy/tests/test_embedding_plots.py
```diff
@@ -115,8 +115,8 @@ def groupsfunc(request):
 
 
 @pytest.fixture(
-    params=[(None, None), (0, 5), ("p15", "p90")],
-    ids=["vbounds.default", "vbound.numbers", "vbound.percentile"],
+    params=[(None, None, None), (0, 5, None), ("p15", "p90", None), (0, "p99", 0.1)],
+    ids=["vbounds.default", "vbound.numbers", "vbound.percentile", "vbound.vcenter"],
 )
 def vbounds(request):
     return request.param
```
ivirshup: Could this be a bit more explicit about the values? This is mostly an ease of reading thing. Here's how I'd do it:



```suggestion
    params=[
        pytest.param(
            {"vmin": None, "vmax": None, "vcenter": None},
            id="vbounds.default"
        ),
        ...  # Fill in the rest
)
def vbounds(request):
    return request.param
```

It's more text, but it's very easy to tell what happening. It also allows just splatting the parameter into the call, or `update`-ing the parameter `dict`.

What do you think about also including `norm` into this `fixture`?

scanpy/plotting/_stacked_violin.py
```diff
@@ -317,15 +323,17 @@ def _mainplot(self, ax):
...
-        import matplotlib.colors
 
-        norm = matplotlib.colors.Normalize(
-            vmin=self.kwds.get('vmin'), vmax=self.kwds.get('vmax')
-        )
         cmap = pl.get_cmap(self.kwds.get('cmap', self.cmap))
         if 'cmap' in self.kwds:
             del self.kwds['cmap']
-        colormap_array = cmap(norm(_color_df.values))
+        normalize = _setup_colornorm(
```
ivirshup: Why rename the variable?

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -24,6 +24,8 @@
     circles,
     ColorLike,
     check_projection,
+    _setup_colornorm,
+    VMinMaxCenter,
```
ivirshup: Could the type import go next to the other type import?

scanpy/plotting/_utils.py
```diff
@@ -31,6 +31,7 @@
 _FontSize = Literal[
     'xx-small', 'x-small', 'small', 'medium', 'large', 'x-large', 'xx-large'
 ]
+VMinMaxCenter = Union[str, float, Callable[[Sequence[float]], float]]
```
ivirshup: How about calling this `VBound` instead of `VMinMaxCenter`?

Also a short docstring would be nice.

scanpy/plotting/_anndata.py
ivirshup, line 924: Why are functions/ shorthand not allowed here? It's non-obvious to me, and seems like it could be useful.

scanpy/plotting/_docs.py
```diff
@@ -123,7 +123,20 @@
     as defined by matplotlib `scatter` function. When making multiple plots, vmin can
     be a list of values, one for each plot. For example `vmin=[0.1, 'p1', None, my_vmin]`
 vmax
-    Maximum value to plot. The format is the same as for `vmin`\
+    Maximum value to plot. The format is the same as for `vmin`.
```
ivirshup: I don't think "Maximum value to plot" is an accurate description of what this does. This doesn't filter which points are plotted, it determines which value gets the color-maps highest value. There's probably a snippier way to phrase that.

"The value representing the center of the color scale" is better to me.

ivirshup (changes requested):
Thanks for addressing my initial comments. I would definitely like @fidelram to take a look at it as well.

One addition I would like to see is a documented example with the expected plot and some prose. Maybe one for DE plots, one for embedding plots? This might be at home in the plotting tutorials.

I've requested some more changes in comments on the code

scanpy/plotting/_docs.py
ivirshup: How matplotlib does it:

```
vmin, vmax : float, default: None
    *vmin* and *vmax* are used in conjunction with the default norm to
    map the color array *c* to the colormap *cmap*. If None, the
    respective min and max of the color array is used.
    It is deprecated to use *vmin*/*vmax* when *norm* is given.
```

scanpy/plotting/_anndata.py
fidelram: for shorthand you mean like 'p99' for the scatter plots?

fidelram:
@gokceneraslan Regarding the tests: yes, they are annoying particularly because is not possible to actually check why a test failed on the server while passes locally. I agree that this limits contribution because the mountain of work to get the tests working puts one off.

For the particular question about the title difference: the test may be passing because of the 'threshold' used to call the images as different. Why we use a threshold? This is to avoid tests from failing due to small differences between matplotlib or other graphic libraries versions or fonts installed. However, sometimes the threshold may be masking some small problems, although in general I am quite happy because important differences not missed. 

BTW: The image that you point out is clearly wrong but I updated it recently for other reason (PR #1584).

Regarding the issue about adding  `norm` as explicit parameter. I would suggest to add it if this just mean changing very few lines but I know this is lot of work (do we want tests for this?) for something that is already working. 

Besides the very good review by Isaac I don't have much to add and will be happy to merge once some of the changes are taken care.

scanpy/plotting/_anndata.py
ivirshup: Yes

scanpy/plotting/_tools/scatterplots.py
gokceneraslan: If you mean `from ..._utils import sanitize_anndata, _doc_params, Empty, _empty` line it's ..._utils not .._utils, so I should keep it I guess.

Edit: Oh I see what you mean now.

scanpy/plotting/_stacked_violin.py
gokceneraslan: To make it more consistent with the variable in _dotplot and _matrixplot files.

scanpy/tests/test_embedding_plots.py
gokceneraslan: Done. Added norm too.

scanpy/plotting/_anndata.py
gokceneraslan: Done.

scanpy/plotting/_baseplot_class.py
gokceneraslan: I did it with `namedtuple`, I hope that's what you meant.

scanpy/plotting/_tools/scatterplots.py
gokceneraslan: This sounds good but requires more modifications e.g. _get_vminmaxcenter has some indexing function which is  necessary only for embedding plots. Also this requires access to the color vector itself to calculate percentiles. I am leaving this to another PR, this one already grew more than I expected :)

scanpy/plotting/_anndata.py
gokceneraslan: It is doable but percentile stuff is integrated to an embedding plot-specific function, namely _get_vminmaxcenter (which is renamed to _get_vboundnorm). It must be moved to _utils.py too. So, good idea but leaving it to another PR.

PR #1553, opened 2020-12-21
Allow `groupby` adata.obs.index in dotplot, matrixplot, heatmap and stacked_violin plots

fidelram:
This PRadds the option to use `adata.obs.index` as groupby.

To allow this change some changes were required to `sc.get.obs_df` to handle non unique adata.obs.index and duplicated obs column names and var_names.


fidelram:
something is wrong after the merge. I will make a fresh PR.

PR #1554, opened 2020-12-22
add deprecation warning for scvi and pin version

### LINKED ISSUE #1443: deprecate scvi in external
I was wondering if we could deprecate the scvi external wrapper as we now have `scvi-tools`. I could also update the wrapper to have minimal functionality, but I think it would be better for people to use our API now that it's tightly integrated with scanpy anyway.

galenxing:
fixes #1443

### DIFF
--- a/scanpy/external/pp/_scvi.py
+++ b/scanpy/external/pp/_scvi.py
@@ -1,3 +1,4 @@
+import warnings
 import numpy as np
 import pandas as pd
 import scipy as sp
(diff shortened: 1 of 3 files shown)


scanpy/external/pp/_scvi.py
```diff
@@ -1,11 +1,20 @@
...
-MIN_VERSION = "0.6.5"
+MIN_VERSION = "0.6.7"
+
+warnings.filterwarnings("default", category=DeprecationWarning, module=__name__)
+
+warnings.warn(
+    "scvi via scanpy external API is no longer supported. "
+    + "Please use the new scvi-tools package from `scvi-tools.org`",
+    DeprecationWarning,
+)
```
ivirshup: * Could this go inside the `scvi` function (ahead of the imports) instead?
* A [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning) can be used to make deprecation warnings that are user visible
* Please try to avoid side effects like modifying the global warnings filter.

ivirshup (changes requested):
Thanks for the fix! Minor change, then should be good to go.

galenxing:
Thanks for reviewing! Just pushed requested changes.

ivirshup (approved):

PR #1555, opened 2020-12-22
Add scrublet to external

ivirshup:
Having some trouble pushing to the branch from the original PR (#1476) so I'm modifying and merging through this PR.


pinin4fjords:
Thanks for finishing this off @ivirshup

PR #1558, opened 2020-12-24
Do not use the cached rp_forest

Koncopd:
As @falexwolf noticed, caching `rp_forest` from previous neighbors calculations can cause problems.


gokceneraslan:
Do we need to backport?

Koncopd:
I think yes.

PR #1564, opened 2021-01-04
Better CI test results

ivirshup:
Testing some azure stuff for better CI test results. Mainly:

* I don't like what `pytest-azurepipelines` does with warnings
* I'd like to have test coverage
  * Azure's is a bit meh (no diffs), maybe we should use codecovs ([example usage](https://github.com/codecov/example-python/blob/74883884e480b523e0db9e92e97264908ecb9b8f/azure-pipelines.yml#L32-L34))
* I like having the test results be easy to read


ivirshup:
I think I'll leave codecov to a separate PR, since this immediately improves test reporting.

PR #1580, opened 2021-01-12
Fix empty img isaac

ivirshup:
Some changes to @giovp #1512, just pushing here so they are visible. Still needs going through the tests to update offsets, and some doc tweaks (behaviour of `na_color`, what `spot_size` is, move `scale_factor` to be `spatial` only).


scanpy/plotting/_tools/scatterplots.py
```diff
@@ -1181,6 +1166,32 @@ def _check_scale_factor(
...
 
+def _check_spatial_data(
+    uns: Mapping, library_id: Union[Empty, None, str]
+) -> Tuple[Optional[str], Optional[Mapping]]:
+    """
+    Given a mapping, try and extract a library id/ mapping with spatial data.
+
+    Assumes this is `.uns` from how we parse visium data.
+    """
+    spatial_mapping = uns.get("spatial", {})
```
giovp, line 1177: I like this, however, in spatial data without image we don't even have `adata.uns[["spatial"]` so this wil break

giovp, line 1204: I like this! Much cleaner than `try: except:`

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -1181,6 +1166,32 @@ def _check_scale_factor(
...
+    if library_id is _empty:
+        if len(spatial_mapping) > 1:
```
giovp, line 1206: also this is really nice check

giovp:
looks really great Isaac, I think it adresses all the issues from before!

PR #1583, opened 2021-01-13
Allow plots to use adata.obs index as groupby

fidelram:
I simplified the `_prepare_dataframe code` by using `sc.get.df`. However, this change uncovered two issues with `sc.get.obs_df` that I have now addressed in this PR. 

The most relevant is the case when the call to  `sc.get.obs_df` contains keys with duplicates (e.g. `keys=['gene1', 'gene1']`). This case is not rare as for example in `sc.pl.dotplot` the same gene can be visualized several times, which requires calling `sc.get.obs_df` with keys that contain duplications.  An example is when `sc.pl.rank_genes_groups_dotplot` is called and, frequently, the same gene appears as top up-regulated for more than one category. To address this, `sc.get.obs_df` removes all duplicates (which correspond to DataFrame columns) and after the DataFrame is complete, the duplicates are added back.

A second problem was for non-unique adata.obs indices which should be a rare situation. However,  it turns out that one of the test adata object used in `test_plotting` have this issue. Also, for the goal of this PR (allow adata.obs.index as groupby option) it could be expected that the index may not be unique.  

In general, non unique obs indices are ok as long as `.obs` DataFrame is not joined or merge based on index. However, because internally in `sc.get.obs_df` the DataFrames are merged using `adata.obs.index` this non-unique indices caused an increase in rows due to multiple matching. To fix this, the code now checks for unique index, and if it is not unique then a temporary index is added to allow proper join operations and then the non-unique index is put back.


scanpy/plotting/_anndata.py
```diff
@@ -1790,64 +1792,39 @@ def _prepare_dataframe(
...
     if groupby is not None:
         if isinstance(groupby, str):
             # if not a list, turn into a list
             groupby = [groupby]
         for group in groupby:
+            if group == adata.obs.index.name:
+                # if grouping is using the index, reset the index
+                # to have the group as a column.
+                adata.obs.reset_index(inplace=True)
+                orig_index_col = group
```
ivirshup: I think we should avoid modify the anndata object when plotting. This would create a copy if it were a view, and may have other unintended side effects. For example, if this errors, now the anndata object is in a bad state.

Instead, could this happen when you're getting the groupby values below, and look more like:

```python
if group == df.index.name:
    categorical = df.index
```

scanpy/get.py
```diff
@@ -194,6 +204,23 @@ def obs_df(
...
+        # lines which will erroneously expand the length of the obs adata frame.
+        # To solve the problem, a unique index is temporarily created which is
+        # afterwards replaced by the original index
+        logg.warning(
+            'Warning: the adata.obs index is not unique. This can be problematic for '
+            'operations with adata.obs '
+        )
+
+        orig_index = adata.obs.index
+        adata.obs_names_make_unique()
```
ivirshup: This shouldn't modify the anndata object.

scanpy/get.py
```diff
@@ -194,6 +204,23 @@ def obs_df(
             f" {gene_error}."
         )
 
+    # check that adata.obs.index is unique
+    orig_index_map = None
+    if not adata.obs.index.is_unique:
+        # non unique obs index is problematic in the `join` operation used in the next
+        # lines which will erroneously expand the length of the obs adata frame.
+        # To solve the problem, a unique index is temporarily created which is
+        # afterwards replaced by the original index
```
ivirshup: `pd.concat` should work here (instead of `join`), since it doesn't take the product of the repeated indices.

scanpy/plotting/_anndata.py
```diff
@@ -1790,64 +1792,39 @@ def _prepare_dataframe(
...
     if isinstance(var_names, str):
         var_names = [var_names]
 
+    orig_index_col = None
     if groupby is not None:
         if isinstance(groupby, str):
             # if not a list, turn into a list
             groupby = [groupby]
         for group in groupby:
+            if group == adata.obs.index.name:
```
ivirshup: This should error if `group` is both the index name and a column name, like pandas does:

```python
pd.DataFrame({"a": [1,2,3]}, index=pd.Index(list("abc"), name="a")).groupby("a")
```

```pytb
ValueError: 'a' is both an index level and a column label, which is ambiguous.
```

scanpy/plotting/_baseplot_class.py
```diff
@@ -109,6 +111,11 @@ def __init__(
             layer=layer,
             gene_symbols=gene_symbols,
         )
+        if len(self.categories) > self.MAX_NUM_CATEGORIES:
+            raise ValueError(
+                f"Over {self.MAX_NUM_CATEGORIES} categories found. "
+                f"Plot would be too large."
+            )
```
ivirshup: I'm not sure we should restrict this. If someone wants a huge plot, why should we stop them?

scanpy/get.py
```diff
@@ -194,6 +204,23 @@ def obs_df(
...
+        logg.warning(
+            'Warning: the adata.obs index is not unique. This can be problematic for '
+            'operations with adata.obs '
+        )
```
ivirshup: Use actual `warnings`, they are easier for users to deal with.

scanpy/get.py
```diff
@@ -166,9 +168,17 @@ def obs_df(
...
+    # use only unique keys, otherwise duplicated keys will
+    # further duplicate when reordering the keys later in the function
+    for key in np.unique(keys):
         if key in adata.obs.columns:
             obs_names.append(key)
+            if key in gene_names.index:
+                logg.warning(
+                    f'The key `{key}` is found in both adata.obs and adata.var_names.'
+                    'Only the adata.obs key will be used.'
+                )
```
ivirshup: Use actual warnings, they are easier for users to deal with.

scanpy/tests/test_plotting.py
```diff
@@ -90,6 +90,8 @@ def test_heatmap(image_comparer):
     )
     save_and_compare_images('master_heatmap_std_scale_var')
 
+    # test standard_scale_obs and groupby = index
+    adata.obs.set_index('cell_type', inplace=True)
```
ivirshup: This case looks really weird to me, making the index very non-unique is a bit of an anti-pattern to me.

Could this be tested more directly with something like:

```
subset = adata[:10].copy()
subset.obs_names.name = "index"
sc.pl.heatmap(
    subset
    ...
    groupby="index",
    ....,
)
```

I also don't love that if a case gets added after this, the `adata` would be in a weird state. Maybe this test should get divided up into separate cases?

If you want to check that non-unique index values get grouped together, could that test look more like:

```python
adata = AnnData(
    np.arange(16).reshape(4, 4),
    obs=pd.DataFrame(
        {"cat": list("aabb")},
         index=pd.Index([f"cell-{i}" for i in range(4)], name="cell_id")
    )
)
adata_nonuniq_index = adata.copy()
adata_nonuniq_index.obs.set_index("cat", inplace=True)
```

And then check that plots using either are equivalent? This wouldn't need to add any more reference images. If you like #1587, that can get merged pretty quick and you can use the `check_same_image` fixture for this.

ivirshup (changes requested):
Big point: don't modify the AnnData object when you're getting values out of it.

Otherwise, mostly good.

Needs docs and examples though.

scanpy/get.py
fidelram: This solution is much simpler.

scanpy/get.py
fidelram: done

scanpy/plotting/_baseplot_class.py
fidelram: I changed this to a warning in case someone ends up making a plot with lot of categories by accident.

scanpy/plotting/_anndata.py
fidelram: done

scanpy/get.py
fidelram: done

scanpy/plotting/_anndata.py
fidelram: done

scanpy/tests/test_plotting.py
fidelram: I added a new test but I didn't use yet the new `check_same_image` as I noticed after the fact that this was already merged into master.

scanpy/plotting/_anndata.py
```diff
@@ -1790,90 +1792,72 @@ def _prepare_dataframe(
...
             # if not a list, turn into a list
             groupby = [groupby]
         for group in groupby:
-            if group not in adata.obs_keys():
+            if group not in list(adata.obs_keys()) + [adata.obs.index.name]:
                 raise ValueError(
                     'groupby has to be a valid observation. '
-                    f'Given {group}, is not in observations: {adata.obs_keys()}'
+                    f'Given {group}, is not in observations: {adata.obs_keys()} or '
+                    f'index name "{adata.obs.index.name}"'
```
ivirshup: The index name will often be `None`. Should the error message look different in this case?

scanpy/plotting/_anndata.py
```diff
@@ -1790,90 +1792,72 @@ def _prepare_dataframe(
...
+    if groupby_index is not None:
+        # reset index to treat all columns the same way.
+        obs_tidy.reset_index(inplace=True)
+        groupby.append(groupby_index)
 
-    obs_tidy = pd.DataFrame(matrix, columns=var_names)
     if groupby is None:
         groupby = ''
         categorical = pd.Series(np.repeat('', len(obs_tidy))).astype('category')
     else:
```
ivirshup, line 1850: `elif`?

scanpy/plotting/_anndata.py
```diff
@@ -1790,90 +1792,72 @@ def _prepare_dataframe(
...
-    elif use_raw:
-        matrix = adata.raw[:, var_names].X
-    else:
-        matrix = adata[:, var_names].X
+            if group == adata.obs.index.name:
+                groupby_index = group
+    if groupby_index is not None:
+        # obs_tidy contains adata.obs.index
+        # and does not need to be given
+        groupby.remove(groupby_index)
```
ivirshup, line 1829: `groupby` should be a value directly passed by a user, right? You should make a copy if you're going to modify it (so the user doesn't unexpectedly lose an entry in their list)

scanpy/plotting/_anndata.py
```diff
@@ -1790,90 +1792,72 @@ def _prepare_dataframe(
...
-            categorical = adata.obs[groupby[0]]
+            categorical = obs_tidy[groupby[0]].astype('category')
             if len(groupby) > 1:
                 for group in groupby[1:]:
                     # create new category by merging the given groupby categories
                     categorical = (
-                        categorical.astype(str) + "_" + adata.obs[group].astype(str)
+                        categorical.astype(str) + "_" + obs_tidy[group].astype(str)
                     ).astype('category')
             categorical.name = "_".join(groupby)
```
ivirshup: Instead, you could probably do something like this to get them all at once:

```python
groupby[0].str.cat(groupby[1:], sep="_").astype("category")
```

Might need a small helper function for making sure a series is either strings or a categorical of strings.

scanpy/plotting/_anndata.py
```diff
@@ -1790,90 +1792,72 @@ def _prepare_dataframe(
...
-    from scipy.sparse import issparse
 
     sanitize_anndata(adata)
     use_raw = _check_use_raw(adata, use_raw)
+    if layer is not None:
+        use_raw = False
     if isinstance(var_names, str):
         var_names = [var_names]
 
+    orig_index_col = None
```
ivirshup: I don't think this variable is used.

ivirshup (changes requested):
Looking much cleaner! I'm really happy with the simplification of `_prepare_dataframe`.

I've realized I missed a few things with `obsdf` though:

* Shouldn't `var_df` should get similar updates to `obs_df`?
* Could we get tests for `get.obs_df`/ `get.var_df` for the issues you addressed here (repeated indices)?
* I've realized this will need a backport, but that can be a separate pr to the 1.7.x branch where you just copy and paste the new functions

scanpy/plotting/_anndata.py
fidelram: done

scanpy/plotting/_anndata.py
fidelram: indeed..

scanpy/plotting/_anndata.py
fidelram: Sounds like a good solution but at the time being I don't have the time to play with this.

scanpy/plotting/_anndata.py
fidelram: removed

fidelram:
> * Shouldn't `var_df` should get similar updates to `obs_df`?

I would suggest a different PR to address this. 

> * Could we get tests for `get.obs_df`/ `get.var_df` for the issues you addressed here (repeated indices)?

Sure, I added new tests to `get.obs_df` to check duplicated keys.

scanpy/plotting/_anndata.py
```diff
@@ -1836,22 +1839,20 @@ def _prepare_dataframe(
...
-                        categorical.astype(str) + "_" + obs_tidy[group].astype(str)
-                    ).astype('category')
-            categorical.name = "_".join(groupby)
+        categorical = obs_tidy[groupby[0]].astype('category')
+        if len(groupby) > 1:
+            for group in groupby[1:]:
+                # create new category by merging the given groupby categories
+                categorical = (
+                    categorical.astype(str) + "_" + obs_tidy[group].astype(str)
+                ).astype('category')
```
ivirshup: What is the branch for? Doesn't the length have to be greater than 1 to reach this code?

scanpy/plotting/_anndata.py
```diff
@@ -1823,6 +1825,7 @@ def _prepare_dataframe(
     if groupby_index is not None:
         # obs_tidy contains adata.obs.index
         # and does not need to be given
+        groupby = groupby[:]  # copy to not modify user passed parameter
```
ivirshup: `groupby.copy()` or `list(groupby)` if there's a chance it might not be a list (e.g. a tuple or array)?

ivirshup (changes requested):
> Sure, I added new tests to get.obs_df to check duplicated keys.

Oh, I meant the repeated index names that were giving you problems with `DataFrame.join`. This issue you mentioned before:

> However, because internally in sc.get.obs_df the DataFrames are merged using adata.obs.index this non-unique indices caused an increase in rows due to multiple matching. 

Here is a test you could use:

```python
adata = sc.AnnData(
    np.arange(16).reshape(4, 4),
    obs=pd.DataFrame(index=["a", "a", "b", "c"]),
    var=pd.DataFrame(index=[f"gene{i}" for i in range(4)]),
)
df = sc.get.obs_df(adata, ["gene1"])
pd.testing.assert_index_equal(df.index, adata.obs_names)
```

-------------------------------

> > Shouldn't var_df should get similar updates to obs_df?
>
> I would suggest a different PR to address this.

Just because the code changes should be largely equivalent (as should the tests) I think it will be easier to review if these changes are together in one PR.

scanpy/plotting/_anndata.py
fidelram: Fixed and added a better way to join the categorical indices

scanpy/plotting/_anndata.py
fidelram: solved

fidelram:
@ivirshup I think this PR is ready to merge. readthedocs is failing because of a missing link from scvelo unrelated to the PR

ivirshup:
Fidel, sorry to say, but I've run into some issues. Most of these actually didn't have to do with this PR, but were additional things that broke from #1499. I'll give you a few examples of what I've found, mostly by contrast with the current behaviour of 1.6.1.

```python
import scanpy as sc, pandas as pd, numpy as np

M, N = (5, 3)
adata = sc.AnnData(
    X=np.zeros((M, N)),
    obs=pd.DataFrame(
        np.arange(M * 3).reshape((M, 3)),
        columns=["repeated_col", "repeated_col", "var_id"],
        index=pd.Index([f"cell_{i}" for i in range(M)], name="obs_index"),
    ),
    var=pd.DataFrame(
        index=pd.Index(["var_id"] + [f"gene_{i}" for i in range(N-1)], name="var_index"),
    ),
)
```

## Repeated column in `adata.obs`

I think this should be an error. This is because downstream functions (like plotting) currently assume that for each key input here, there will be one output column. Turns out this isn't exactly pandas behaviour with repeated column values, but I do think it's reasonable.

```python
M, N = 5, 3
adata = sc.AnnData(
    X=np.zeros((M, N)),
    obs=pd.DataFrame(
        np.arange(M * 2).reshape((M, 2)),
        columns=["repeated_col", "repeated_col"],
        index=[f"cell_{i}" for i in range(M)],
    ),
    var=pd.DataFrame(
        index=[f"gene_{i}" for i in range(N)],
    ),   
)
sc.get.obs_df(adata, ["repeated_col"])
```

### This pr (gets both columns)

```
           repeated_col  repeated_col
obs_index                            
cell_0                0             1
cell_1                3             4
cell_2                6             7
cell_3                9            10
cell_4               12            13
```

### 1.6 (errors)

```pytb
~/miniconda3/envs/scanpy-1.6/lib/python3.8/site-packages/pandas/core/internals/blocks.py in __init__(self, values, placement, ndim)
    140 
    141         if self._validate_ndim and self.ndim and len(self.mgr_locs) != len(self.values):
--> 142             raise ValueError(
    143                 f"Wrong number of items passed {len(self.values)}, "
    144                 f"placement implies {len(self.mgr_locs)}"

ValueError: Wrong number of items passed 2, placement implies 1
```

Not a great error, could definitley be improved.

## Key in adata.obs.columns and adata.var_names

In this case, the key is ambiguous (should it get the gene values or the column from obs?). I think this means it should error. I feel like this point has been discussed a number of times, but doesn't seem to have been discussed when this behaviour was changed.

```python
M, N = 5, 3
adata = sc.AnnData(
    X=np.zeros((M, N)),
    obs=pd.DataFrame(
        np.arange(M),
        columns=["var_id"],
        index=[f"cell_{i}" for i in range(M)],
    ),
    var=pd.DataFrame(
        index=["var_id"] + [f"gene_{i}" for i in range(N-1)],
    ),   
)
sc.get.obs_df(adata, ["var_id"])
```

### This pr (warns)

```
/Users/isaac/github/scanpy/scanpy/get.py:177: UserWarning: The key `var_id` is found in both adata.obs and adata.var_names.Only the adata.obs key will be used.
  warnings.warn(
Out[58]: 
           var_id
obs_index        
cell_0          2
cell_1          5
cell_2          8
cell_3         11
cell_4         14
```

### 1.6 (errors)

```pytb
<ipython-input-16-69be169f6a4f> in <module>
----> 1 sc.get.obs_df(adata, ["var_id"])

~/miniconda3/envs/scanpy-1.6/lib/python3.8/site-packages/scanpy/get.py in obs_df(adata, keys, obsm_keys, layer, gene_symbols, use_raw)
    171     for k, l in zip(keys, lookup_keys):
    172         if not use_raw or k in adata.obs.columns:
--> 173             df[k] = adata.obs_vector(l, layer=layer)
    174         else:
    175             df[k] = adata.raw.obs_vector(l)

~/miniconda3/envs/scanpy-1.6/lib/python3.8/site-packages/anndata/_core/anndata.py in obs_vector(self, k, layer)
   1362                 )
   1363                 layer = None
-> 1364         return get_vector(self, k, "obs", "var", layer=layer)
   1365 
   1366     def var_vector(self, k, *, layer: Optional[str] = None) -> np.ndarray:

~/miniconda3/envs/scanpy-1.6/lib/python3.8/site-packages/anndata/_core/index.py in get_vector(adata, k, coldim, idxdim, layer)
    156 
    157     if (in_col + in_idx) == 2:
--> 158         raise ValueError(
    159             f"Key {k} could be found in both .{idxdim}_names and .{coldim}.columns"
    160         )

ValueError: Key var_id could be found in both .var_names and .obs.columns
```

## Repeats in var_names

When there are repeats in `var_names` (pretty frequent occurence), getting a dataframe with keys that aren't repeated. I think it's fine for this to work. I do think it should error if the key is one values that is duplicated in the index.

```python
adata = sc.AnnData(
    X=np.ones((2, 3)),
    obs=pd.DataFrame(index=["cell-0", "cell-1"]),
    var=pd.DataFrame(index=["gene-0", "gene-0", "gene-1"]),
)
sc.get.obs_df(adata, ["gene-1"])
``````

### This PR (errors)

```pytb
InvalidIndexError                         Traceback (most recent call last)
<ipython-input-62-405d671e2970> in <module>
----> 1 sc.get.obs_df(adata, ["a", "gene-1"])

~/github/scanpy/scanpy/get.py in obs_df(adata, keys, obsm_keys, layer, gene_symbols, use_raw)
    213             var_idx = adata.raw.var_names.get_indexer(var_names)
    214         else:
--> 215             var_idx = adata.var_names.get_indexer(var_names)
    216 
    217         # for backed AnnData is important that the indices are ordered

/usr/local/lib/python3.8/site-packages/pandas/core/indexes/base.py in get_indexer(self, target, method, limit, tolerance)
   3169 
   3170         if not self.is_unique:
-> 3171             raise InvalidIndexError(
   3172                 "Reindexing only valid with uniquely valued Index objects"
   3173             )

InvalidIndexError: Reindexing only valid with uniquely valued Index objects
```


### 1.6 (suceeds)

```python
        gene-1
cell-0     1.0
cell-1     1.0
```

1.6 does error if I use `"gene-0"` as a key, but the error message could definitley be better.

## What should we do about this?

My current inclination is to revert most changes to `obs_df` and `var_df` from this PR and #1499. This should leave the use of indices as groupby untouched. Also, the loss of perfomance from reverting #1499 should be partially mitigated by improvements in pandas (see https://github.com/pandas-dev/pandas/issues/37954). We would keep all the user facing changes, and all the tests from both PRs. We can then make a release now, and can patch in performance boosts during the release cycle.

Do you agree with this assessment? If not, could you propose an alternative?

fidelram:
@ivirshup regarding the three comments:
 
 * `adata.obs.columns` non_unique: I also think this should be an error as this will cause trouble at some point or another.  I added a check for this which will raise a ValueError if duplicates are found
 * `adata.var.index` Here I wanted to raise an error only if a `key` matched a duplicated `var_name` as you suggested. However, the selection by index, instead of `var_name`, from the matrix causes an error whenever the `var_names` contains duplicates. Actually, doing an AnnData selection when the `var_names` are not unique also raises an error.  Thus, I added a ValueError  if the var_names are not unique
```PYTHON
adata = sc.AnnData(
    X=np.ones((2, 3)),
    obs=pd.DataFrame(index=["cell-0", "cell-1"]),
    var=pd.DataFrame(index=["gene-0", "gene-0", "gene-1"]),
)
adata[:, ['gene-1']]
```

```
--------------------------------------------------------------------------
InvalidIndexError                         Traceback (most recent call last)
<ipython-input-58-8d1a96772653> in <module>
----> 1 adata[:, ['gene-0']]

site-packages/anndata/_core/anndata.py in __getitem__(self, index)
   1085     def __getitem__(self, index: Index) -> "AnnData":
   1086         """Returns a sliced view of the object."""
-> 1087         oidx, vidx = self._normalize_indices(index)
   1088         return AnnData(self, oidx=oidx, vidx=vidx, asview=True)
   1089 

site-packages/anndata/_core/anndata.py in _normalize_indices(self, index)
   1066 
   1067     def _normalize_indices(self, index: Optional[Index]) -> Tuple[slice, slice]:
-> 1068         return _normalize_indices(index, self.obs_names, self.var_names)
   1069 
   1070     # TODO: this is not quite complete...

site-packages/anndata/_core/index.py in _normalize_indices(index, names0, names1)
     33     ax0, ax1 = unpack_index(index)
     34     ax0 = _normalize_index(ax0, names0)
---> 35     ax1 = _normalize_index(ax1, names1)
     36     return ax0, ax1
     37 

site-packages/anndata/_core/index.py in _normalize_index(indexer, index)
     95             return positions  # np.ndarray[int]
     96         else:  # indexer should be string array
---> 97             positions = index.get_indexer(indexer)
     98             if np.any(positions < 0):
     99                 not_found = indexer[positions < 0]

site-packages/pandas/core/indexes/base.py in get_indexer(self, target, method, limit, tolerance)
   3169 
   3170         if not self.is_unique:
-> 3171             raise InvalidIndexError(
   3172                 "Reindexing only valid with uniquely valued Index objects"
   3173             )

InvalidIndexError: Reindexing only valid with uniquely valued Index objects
```
 * `key` in both `adata.obs.columns` and `adata.var.index`: I changed this to raise a ValueError.

I though about reverting back to the original implementation as you suggest but this will not work with `_anndata._prepare_dataframe` as I introduced some changes to work with this function. 

The just  added changes should mimic the response from 1.6 except for duplicate names in var_names which I think should respond similarly like when doing a slicing  on the `AnnData` object.

I added new tests based on your examples.

I added checks to test for unique adata.obs.columns

ivirshup:
@fidelram, what are the changes which are incompatible with `_anndata._prepare_dataframe`? I've tried reverting dealing with `X` and those seemed to work fine for me.

ivirshup:
> The just added changes should mimic the response from 1.6 except for duplicate names in var_names which I think should respond similarly like when doing a slicing on the AnnData object.

Thinking about this more. Considering that no one has complained about this so far. I think I'm actually fine with this being an error.  If there are complaints, I think we should change it back.

I do think it's important that `gene_symbols` can have duplicates as long as those values aren't being accessed (as non-unique identifiers is the whole point of `gene_symbols`).

ivirshup:

Issue #1562, opened 2021-01-04
matrixplot group order changes with pandas 1.2.0


ivirshup:
@fidelram

The test `test_matrixplot_obj` fails if `pandas>1.2.0` as the groups order changes. These groups are meant to be sorted with `plot.add_totals(sort='descending')`. All groups have the same value here, so my assumption is this is fine. I'm going to change the test for now, but it'd be good to hear back from you on whether this is a bug or not.

It would probably be good if we could ensure a stable sort was used so this won't change in future.

fidelram:
If I understand correctly, pandas > 1.2.0 `plot.add_totals(sort='descending')` returns a different sorting compared to previous version. However, most of the categories have the same number of cells and thus a different sorting is equally valid.

![image](https://user-images.githubusercontent.com/4964309/104478840-0a9d5980-55c3-11eb-94b0-b0e2092c55c3.png)

My suggestion is to use other dataset in which the number of cells per category is different and update the image.

PR #1584, opened 2021-01-13
updated test that works with pandas 1.2 #1562

fidelram:
This PR addresses #1562
A test requiring sorting was updated such that it works the same with pandas<1.2.0 or pandas>=1.2.0.

The issue was caused by the test having identical values to sort.


ivirshup:
I think that having the case with identical values is actually useful, since otherwise we're not checking this edge case. Ideally, I think we should be using a stable sort so results will be consistent in the case of similar values.

ivirshup:
Huh, it looks like pandas does not have the option for doing a stable sort here, so maybe we should change the test.

fidelram:
Yeah, I check to see if there was some option to get the same sorting regardless of the pandas version, but could not find anything. I will merge the change.

PR #1586, opened 2021-01-14
Prep release again

ivirshup:
Ran into issues making release (PyPi didn't like the upload), trying again

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,11 +4,11 @@ build-backend = 'setuptools.build_meta'
 
 # uses the format of tool.flit.metadata because we’ll move to it anyway
 [tool.scanpy]
-author = '''
-Alex Wolf, Philipp Angerer, Fidel Ramirez, Isaac Virshup,
-Sergei Rybakov, Gokcen Eraslan, Tom White, Malte Luecken,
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -4,11 +4,11 @@ build-backend = 'setuptools.build_meta'
...
-author = '''
-Alex Wolf, Philipp Angerer, Fidel Ramirez, Isaac Virshup,
-Sergei Rybakov, Gokcen Eraslan, Tom White, Malte Luecken,
-Davide Cittaro, Tobias Callies, Marius Lange, Andrés R. Muñoz-Rojas
-'''
+author = """
+Alex Wolf, Philipp Angerer, Fidel Ramirez, Isaac Virshup, \
+Sergei Rybakov, Gokcen Eraslan, Tom White, Malte Luecken, \
+Davide Cittaro, Tobias Callies, Marius Lange, Andrés R. Muñoz-Rojas\
+"""
```
flying-sheep, line 11: Why change the quotes?

ivirshup: pypi doesn't like it when authors (or anything before long_description) has new lines.

toml has different rules for single and double tick strings. Can't escape in single tick strings.

PR #1587, opened 2021-01-14
Attach failing plots to CI results

ivirshup:
If plots fail on CI, this should let us see the expected, actual, and diff through azure. ~~Hopefully~~ It works!

I think this could make debugging plotting issues much easier. Here's an example of what the results look like:

image

Current issue, if tests are run in parallel, this does not work (https://github.com/pytest-dev/pytest-nunit/issues/40), which is not an immediate problem for CI, but limits applicability for local usage. I believe this is an issue with this particular pytest plugin, not necessarily this strategy.


ivirshup:
> yes, they are annoying particularly because is not possible to actually check why a test failed on the server while passes locally. 

@fidelram, from your comment (https://github.com/theislab/scanpy/pull/1551#issuecomment-761117523), makes me think you'd like to enable this? If you okay this, all this needs to be ready to merge is: 

- [x] Figure out where result xml should live
- [x] `.gitignore` update
- [x] Remove failing test (just there as an example)
- [x] Document where to find this stuff

fidelram:
This is great Isaac! Very welcome feature. Took me a while to find the 'attachments' in the azure webpage.

Regarding the location or results.xml, I honestly don't have an opinion were to locate it.

ivirshup:
Great!

> Took me a while to find the 'attachments' in the azure webpage

I'll document this!

ivirshup:

gokceneraslan:
@ivirshup do you know why are there sometimes three:

image


and sometimes two plots:

image

PR #1601, opened 2021-01-19
fix ingest for umap 0.5

Koncopd:
subj
https://github.com/theislab/scanpy/issues/1509
`rp_forest` storage is still broken though.

### DIFF
--- a/docs/release-latest.rst
+++ b/docs/release-latest.rst
@@ -66,4 +66,5 @@ Version 1.7
 - Fixed bug where calling `set_figure_params` could block if IPython was installed, but not used. :pr:`1547` :smaller:`I Virshup`
 - :func:`~scanpy.pl.violin` no longer fails if `.raw` not present :pr:`1548` :smaller:`I Virshup`
 - :func:`~scanpy.pl.spatial` refactoring and better handling of spatial data :pr:`1512` :smaller:`G Palla`
-- :func:`scanpy.pp.pca` works with `chunked=True` again :pr:`1592` :smaller:`I Virshup`
+- :func:`~scanpy.pp.pca` works with `chunked=True` again :pr:`1592` :smaller:`I Virshup`
+- :func:`~scanpy.tl.ingest` now works with umap-learn 0.5.0 :pr:`1601` :smaller:`S Rybakov`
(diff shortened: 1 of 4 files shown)


ivirshup:
Would making pynndescent a requirement make the code here a bit more simple? I'm wondering if all the branches around umap versions and whether pynndescent is present could be removed.

Koncopd:
For umap 0.5 pynndescent is in requirements, so it is used here also. For backwards compatibility i left the code for the old versions. I think i will remove the old code sometime in the future and just use pynndescent and umap >= 0.5.

ivirshup:
> rp_forest storage is still broken though.

Is this something you are planning on fixing in this pr?

Also, what's broken about it? I would have thought we'd have a failing test.

Koncopd:
I am checking, it seems that the format for `rp_forest `changed significantly.

No, failing forest conversion won't cause any problems because it was deliberately coded to avoid problems. The code is in neighbors, and neighbors is critical functuionality and can work without `rp_forest`.

ivirshup:
> I am checking, it seems that the format for rp_forest changed significantly

Can we validate the format? If so, can we just recalculate the forest if the format isn't what we expect?

Also, can you point me to the code where this is handled in `ingest`? I feel like something user visible should happen if there's malformed data in the object.

Koncopd:
https://github.com/theislab/scanpy/blob/5fc12f4a918e21f0c57937b787d52040db046f01/scanpy/neighbors/__init__.py#L284
https://github.com/theislab/scanpy/blob/5fc12f4a918e21f0c57937b787d52040db046f01/scanpy/neighbors/__init__.py#L436
https://github.com/theislab/scanpy/blob/5fc12f4a918e21f0c57937b787d52040db046f01/scanpy/neighbors/__init__.py#L177
https://github.com/theislab/scanpy/blob/5fc12f4a918e21f0c57937b787d52040db046f01/scanpy/neighbors/__init__.py#L736

`rp_forest` is handled in neighbors.

ivirshup:
Alright, I was more wondering how you were handling the forest right now, but I see that you're just calculating it. I suspect there will be a minimum version bound needed for the `init_graph` parameter.

Let me know when you think this is ready to merge.

Koncopd:
@ivirshup i am merging this and will work on `rp_forest` later.

ivirshup:

PR #1602, opened 2021-01-20
add twine check to azure CI + 3.8 matrix

### LINKED ISSUE #1585: Add build and `twine check` CI task
`twine check` is not great at telling you why it's failing. It would be easier to figure out what caused the break if we were continuously checking for this.

Inspired by finding out that `authors` can't have new lines, via an error that says `long_description` can't have section headings (which definitely isn't true).

Zethson:
Dear everyone,

This PR adds
1. Python 3.8 to the build matrix. We were currently only testing vs 3.6 and 3.7. 3.10 is already coming up in April, so it is in my opinion time to use the more latest Python versions. If you feel like this adds too much to the Azure/CI bill, then I would suggest to remove 3.6. 
2. Solves #1585 

Signed-off-by: Zethson <lukas.heumos@posteo.net>

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -79,7 +79,7 @@ jobs:
       rst2html.py --halt=2 README.rst >/dev/null
     displayName: 'rst2html'
 
-- job: Twine Check
+- job: Twine
   pool:
     vmImage: 'ubuntu-18.04'
   steps:
@@ -88,7 +88,7 @@ jobs:
         versionSpec: '3.8'
 
     - script: |
-      pip install setuptools twine wheel
-      python setup.py sdist bdist_wheel
-      twine check dist/*
-    displayName: 'Twine check'
+        pip install setuptools twine wheel
+        python setup.py sdist bdist_wheel
+        twine check dist/*
+      displayName: 'Twine check'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -87,7 +87,7 @@ jobs:
 
   - script: |
       python -m pip install --upgrade pip
-      pip install setuptools twine
+      pip install setuptools twine pytoml
     displayName: 'Install dependencies'
 
   - script: pip list
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -80,8 +80,7 @@ jobs:
     displayName: 'rst2html'
 
   - script: |
-      python -m pip install --upgrade pip
-      pip install setuptools twine wheel
+      pip install setuptools twine
       python setup.py sdist bdist_wheel
       twine check dist/*
     condition: in(variables['python.version'], '3.8')
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -14,9 +14,9 @@ jobs:
       Python38:
         python.version: '3.8'
       Python37:
-        python.version: '3.7'
+        python.version: '3.6'
       anndata_dev:
-        python.version: '3.7'
+        python.version: '3.8'
         ANNDATA_DEV: yes
         RUN_COVERAGE: yes
   steps:
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -11,8 +11,6 @@ jobs:
     vmImage: 'ubuntu-18.04'
   strategy:
     matrix:
-      Python39:
-        python.version: '3.9'
       Python38:
         python.version: '3.8'
       Python37:
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -15,8 +15,6 @@ jobs:
         python.version: '3.8'
       Python37:
         python.version: '3.7'
-      Python36:
-        python.version: '3.6'
       anndata_dev:
         python.version: '3.7'
         ANNDATA_DEV: yes
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -11,6 +11,10 @@ jobs:
     vmImage: 'ubuntu-18.04'
   strategy:
     matrix:
+      Python39:
+        python.version: '3.9'
+      Python38:
+        python.version: '3.8'
       Python37:
         python.version: '3.7'
       Python36:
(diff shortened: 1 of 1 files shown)


ivirshup:
:+1: to twine check :-1: to the python versions

Ultimately, we do have a limited amount of CI, so I think it's important to be a bit cautious adding many jobs. Of the dependencies I'm worried about being an issue: generally not newer python versions. Minimum python versions are important for catching us using newer features. 

I am up for swapping python 3.7 with 3.8. I don't think 3.9 is going to work for now. Last time I tried to use 3.9 (a month ago) numpy builds weren't working. I believe numba currently isn't working: https://github.com/numba/numba/issues/6345

Higher priorities to me (roughly in order):

* Test on windows
* Test against lower bounds of our requirements
* Test on Mac
* Dev builds

Zethson:
@ivirshup perfectly fine with me.
Removed 3.6 

So we are now testing against 3.7 and 3.8.

Zethson:
@ivirshup is there any reason why we are currently not additionally using Github Actions?

ivirshup:
> Removed 3.6

We should keep 3.6 as long as we support it. It's easy to accidentally add features which only work with 3.7+ otherwise. I'd be happy to drop 3.6 once numpy does (and in general roughly follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) as soon as the ecosystem does).

>  is there any reason why we are currently not additionally using Github Actions?

Depends on the task. Also depends on the definition of github actions I think? We aren't using any of their runners for testing because we'd like the ability to integrate with hosted resources on azure. Also, azure seemed like much more of a standard for numeric python packages at the time we chose it.

I'd be happy to have github actions for other things, like `precommit`. `twine check` could be another one, but I haven't looked in to how "artifact" type things are handled with github actions to know if we'd be able to recover the built objects.

We'd talked about using codecov too, which I'd like to add a check for. I'm not totally clear on the distinction between checks and actions yet.

ivirshup:
btw, it'd be great if this got a sibling PR in anndata

Zethson:
> We should keep 3.6 as long as we support it. It's easy to accidentally add features which only work with 3.7+ otherwise. I'd be happy to drop 3.6 once numpy does (and in general roughly follow NEP 29 as soon as the ecosystem does).

All right.
So do you want
3.6 + 3.7 + 3.8
Or
3.6 + 3.7
?

Zethson:
Regarding Github Actions. It is in my opinion the fastest option out there and jobs spin up really fast.

We could certainly move some of the fast and easy checks to Github Actions and leave the more heavy jobs for Azure.
But let's keep this for a discussion in another issue or Slack.

I'll add a sibling PR to Anndata as soon as we merged this one.

ivirshup:
I think 3.6 + 3.8 would be reasonable, covering the upper and lower bounds.

Zethson:
> I think 3.6 + 3.8 would be reasonable, covering the upper and lower bounds.

Feels weird to me honestly, but I see what you're trying to do.
Is there no possibility that our Azure credits cover 3.6, 3.7 and 3.8?

If not, then I would be fine with 3.6 and 3.8

ivirshup:
10 simultaneous jobs is the supposed limit (though I'm not sure how much this is enforced). I think upper and lower bounds are generally a good way to go:

* lower bounds make sure we don't add incompatible features
* upper bounds give us deprecation warnings

.azure-pipelines.yml
```diff
@@ -79,3 +79,9 @@ jobs:
       python setup.py check --restructuredtext --strict
       rst2html.py --halt=2 README.rst >/dev/null
     displayName: 'rst2html'
+
+  - script: |
+      pip install setuptools twine
+      python setup.py sdist bdist_wheel
+      twine check dist/*
+    displayName: 'Twine check'
```
ivirshup: Could this be a separate job? It would only need to happen for one python version, probably 3.8.

Also can the `python setup.py check --restructuredtext --strict` script task be removed from the pytest job, since this covers that?

ivirshup (changes requested):
Just a minor change, otherwise all good.

.azure-pipelines.yml
```diff
@@ -76,6 +76,12 @@ jobs:
       testRunTitle: 'Publish test results for Python $(python.version)'
 
   - script: |
-      python setup.py check --restructuredtext --strict
       rst2html.py --halt=2 README.rst >/dev/null
     displayName: 'rst2html'
+
+  - script: |
+      pip install setuptools twine
+      python setup.py sdist bdist_wheel
```
flying-sheep: Once this is merged, can you please add a PR that contains this change and targets the `flit` branch? (#1527)

```suggestion
      pip install twine
      flit build
```

Zethson: Guess I'll have to understand your choice of flit over alternatives first. But if the team decides that flit is the way to move forward instead of poetry, then I will be happy to submit the PR.

.azure-pipelines.yml
Zethson: Separate job is bad, since we would need to setup Python again, install the dependencies etc.

I added a condition for this task to only be run if Python==3.8

ivirshup: Dependencies aren't required to build the package or run twine though?

If you mean the dependencies for the twine check, I'm not worried about these. Looking at the CI, this whole process took like 30 seconds (including the failing to build). Setting up python was 2 seconds.

Zethson: Yeah, I don't know what I was thinking when I wrote this :)

I still think that the current solution of just running it for the latest Python version in the same job is 110% fine and the best solution. The condition is simple and readable.
1. No need to setup Python nor install wheels again.
2. It's weird when the twine check is the only check that has its own job.

ivirshup: The black check should have it's own job too, but that will happen separately. It's already been done this way for `anndata`.

Zethson: Yeah, we can extract all simple static analysis tasks into their own job.
However, this should be done in another PR. Are you fine with this one merged?

ivirshup:

PR #1608, opened 2021-01-27
Remove dependency on scvelo for doc builds

ivirshup:
scvelo docs have been changed so the url for the sphinx inventory is different. We also probably don't want to depend on scvelo's documentation for our doc builds, especially since it's pre 1.0, and we weren't really doing much with it. Should fix current doc build problems.


ivirshup:

Issue #1288, opened 2020-06-23
Probably wrong distance matrix used in sc.tl.dendrogram


jingxinfu:
In the `sc.tl.dendrogram` module, I noticed that the correlation matrix was directly inputted into the `scipy.cluster.hierarchy.linkage`. That results in calculating the distance between samples by the Euclidean(X1_cor_with_others, X2_cor_with_others). Shouldn't the distance be the pure correlation value here? Please correct me if I didn't understand it correctly.

The one I would perfer
```python
df = 1-adata.uns['dendrogram_sample']['correlation_matrix']
data_linkage = hierarchy.linkage(ssd.squareform(
        df)
...
```
The one currently in sc.tl.dendrogram
```python
data_linkage = hierarchy.linkage(adata.uns['dendrogram_sample']['correlation_matrix'])
...
```

```pytb
...
```

#### Versions:

> ...

giovp:
mmh true that should probably be 1-corr_matrix. I'll ping @flying-sheep he might have a better answer

PR #1614, opened 2021-01-30
fixes dendrogram issue  #1228

fidelram:
The PR uses `1-correlation` as distance matrix to compute the dendrogram as suggested in #1288 and mentioned in https://github.com/theislab/squidpy/pull/236

I opted for the minimal changes to the code. Other  solution would be to use `scipy.spatial.distance.pdist` that allows a larger number of distance metrics. I am open for a discussion on this topic (ping @michalk8)


giovp:
Hi @fidelram ! 
looks good yes, with `distance.squareform` you are essentially transforming the corr in a condensed distance matrix right? and that's what `pdist` also passes. So I think it looks good now!
Another option would have been to pass `mean_df.T` directly to linkage, but I understand wanting to keep correlation as distances.

giovp (approved):

fidelram:
Hi @giovp, basically I mimic the what the linkage code would do internally when the first argument is not a distance. Because the correlation matrix is used for `sc.tl.correlation_matrix` it is kept.

giovp:
Makes sense! Thanks it's clear now why the corr matrix

ivirshup (changes requested):
Could this get:

- [x] Bugfix release note
- [x] Links to issues that should be closed by this

fidelram:
Issues that should be closed:
 * https://github.com/theislab/scanpy/issues/1288

ivirshup (approved):

ivirshup:

ivirshup:
@fidelram, in future, if you add a line like: 

```
Fixes #x, #y
```

To the initial comment in a PR, those issues will automatically be closed when the PR is merged. I think you can also link PRs to issues with the "Linked issues" element in the righthand sidebar. Using these makes it easier to tell which issues have been addressed already.

Issue #1604, opened 2021-01-22
Show example plots as part of documentation


fidelram:
In seaborn, the docstring example plots are shown as part of the documentation. Would be great to have this as part of scanpy as this will motivate the creation of more examples as part of the docstring while keeping them up-to-date. 

See for example: https://seaborn.pydata.org/generated/seaborn.catplot.html
![image](https://user-images.githubusercontent.com/4964309/105457501-3a192980-5c87-11eb-8a17-050b0dc5b67b.png)

Does anyone has experience on how to achieve this?

giovp:
@fidelram that's a really great point and something I'd like to discuss at next meeting (already put it in the agenda).

Another great example of such examples 😅 is the way @michalk8 set it up for [cellrank](https://cellrank.readthedocs.io/en/latest/auto_examples/index.html) and squidpy [not yet public].
The even nicer thing is that @michalk8 implemented a CI pipeline for the tutorials/examples part of the repo so that every time there is a change in master of the original repo, the examples are refreshed in the notebooks repo, so to have them always up to date.

Would be really cool to concentrate efforts and try to get this logic also in scanpy (makes it both very user friendly and robust from a maintainer perspective)

ivirshup:
I would love to have this! I've been talking about this with @falexwolf and a bit with @flying-sheep.

Seaborn uses a sphinx extension from `matplotlib` to generate the plots ([plot directive](https://matplotlib.org/3.1.3/devel/plot_directive.html)).

I would definitely like to get tutorials running on CI as well, but think this is a little separate (some previous discussion: #302).

1. Tutorials can use a lot more computational resources than plotting.
2. These examples are written in the docstrings, while tutorials are currently in notebooks. It'll probably be a different set of tools to render these.

fidelram:
I tried the sphinx extension but could not get any result. Can somebody take over this?

ivirshup:
Closed, as the initial issue is done, and being tracked elsewhere.

PR #1632, opened 2021-02-09
Plotting examples

ivirshup:
Re #1604, @fidelram, I've got the sphinx extension working for examples (at least locally) here. You can check out a simple example for `sc.pl.embedding`.

Does this solve the problem you were running into?

<details>
<summary> Functions which need examples </summary>

Functions in `sc.pl`

- [ ] Embedding plots
    - [ ] `sc.pl.embedding`
    - [ ] `sc.pl.draw_graph`
    - [ ] `sc.pl.diffmap`
    - [ ] `sc.pl.pca`
    - [ ] `sc.pl.tsne`
    - [ ] `sc.pl.umap`
    - [ ] `sc.pl.spatial`
    - [x] `sc.pl.embedding_density`
- [ ] PCA specific
    - [ ] `sc.pl.pca_loadings`
    - [ ] `sc.pl.pca_overview`
    - [ ] `sc.pl.pca_scatter`
    - [ ] `sc.pl.pca_variance_ratio`
- [ ] PAGA
    - [ ] `sc.pl.paga`
    - [ ] `sc.pl.paga_adjacency`
    - [ ] `sc.pl.paga_compare`
    - [ ] `sc.pl.paga_path`
- [ ] DPT pseudotime
    - [ ] `sc.pl.dpt_groups_pseudotime`
    - [ ] `sc.pl.dpt_timeseries`
- [ ] Groupby
    - [x] `sc.pl.dotplot`
    - [ ] `sc.pl.matrixplot`
    - [ ] `sc.pl.clustermap`
    - [ ] `sc.pl.heatmap`
    - [ ] `sc.pl.dendrogram`
    - [ ] `sc.pl.stacked_violin`
    - [ ] `sc.pl.tracksplot`
    - [ ] `sc.pl.violin`
- [ ] Preprocessing
    - [ ] `sc.pl.filter_genes_dispersion`
    - [ ] `sc.pl.highest_expr_genes`
    - [ ] `sc.pl.highly_variable_genes`
- [ ] DE
    - [ ] `sc.pl.rank_genes_groups`
    - [ ] `sc.pl.rank_genes_groups_dotplot`
    - [ ] `sc.pl.rank_genes_groups_heatmap`
    - [ ] `sc.pl.rank_genes_groups_matrixplot`
    - [ ] `sc.pl.rank_genes_groups_stacked_violin`
    - [ ] `sc.pl.rank_genes_groups_tracksplot`
    - [ ] `sc.pl.rank_genes_groups_violin`
- [ ] Misc/ to be classified
    - [ ] `sc.pl.ranking`
    - [ ] `sc.pl.scatter`
    - [ ] `sc.pl.sim`
    - [ ] `sc.pl.correlation_matrix`
    - [ ] `sc.pl.matrix`
- [ ] Time series (???)
    - [ ] `sc.pl.timeseries`
    - [ ] `sc.pl.timeseries_as_heatmap`
    - [ ] `sc.pl.timeseries_subplot`

Other functions

- [x] `sc.pp.calculate_qc_metrics`
- [x] `sc.tl.embedding_density`
- [ ]  `sc.get.obs_df`
- [ ] `sc.get.rank_genes_groups_df`

</details>

### DIFF
--- a/docs/dev/documentation.rst
+++ b/docs/dev/documentation.rst
@@ -42,7 +42,21 @@ Some key points:
 * When docs exist in the same file as code, line length restrictions still apply. In files which are just docs, go with a sentence per line (for easier `git diff`\ s).
 * Check that the docs look like what you expect them too! It's easy to forget to add a reference to function, be sure it got added and looks right.
 
-Look at `sc.tl.louvain <https://github.com/theislab/scanpy/blob/a811fee0ef44fcaecbde0cad6336336bce649484/scanpy/tools/_louvain.py#L22-L90>`__ as an example for everything mentioned here:
+Look at `sc.tl.louvain <https://github.com/theislab/scanpy/blob/a811fee0ef44fcaecbde0cad6336336bce649484/scanpy/tools/_louvain.py#L22-L90>`__ as an example for everything mentioned here.
+
+Plots in docstrings
+~~~~~~~~~~~~~~~~~~~
+
+One of the most useful things you can include in a docstring is examples of how the function should be used.
+In the case of plotting functions, it's also possible to include show the resulting plots in the rendered docs.
+This is accomplished with `matplotlib's plot directive <https://matplotlib.org/devel/plot_directive.html>`__.
+
+For examples of what this looks like, see the examples sections of :func:`~scanpy.pl.dotplot` or :func:`~scanpy.pp.calculate_qc_metrics`.
+
+Note that anything in these sections will need to be run when the docs are built, so it's important to keep them light.
+
+* If you need computed features (e.g. an embedding, differential expression results) load data that has this precomputed.
+* Try to re-use datasets, this reduces the amount of data that needs to be downloaded to the CI server.
 
 `Params` section
 ~~~~~~~~~~~~~~~~
--- a/docs/dev/documentation.rst
+++ b/docs/dev/documentation.rst
@@ -48,12 +48,11 @@ Plots in docstrings
 ~~~~~~~~~~~~~~~~~~~
 
 One of the most useful things you can include in a docstring is examples of how the function should be used.
-In the case of plotting functions, it's also possible to include show the resulting plots in the rendered docs.
-This is accomplished with `matplotlib's plot directive <https://matplotlib.org/devel/plot_directive.html>`__.
+These are a great way to demonstrate intended usage and give users a template they can copy and modify.
+We're able to include the plots produced by these snippets in the rendered docs using `matplotlib's plot directive <https://matplotlib.org/devel/plot_directive.html>`__. 
+For examples of this, see the `Examples` sections of :func:`~scanpy.pl.dotplot` or :func:`~scanpy.pp.calculate_qc_metrics`.
 
-For examples of what this looks like, see the examples sections of :func:`~scanpy.pl.dotplot` or :func:`~scanpy.pp.calculate_qc_metrics`.
-
-Note that anything in these sections will need to be run when the docs are built, so it's important to keep them light.
+Note that anything in these sections will need to be run when the docs are built, so please keep them computationally light.
 
 * If you need computed features (e.g. an embedding, differential expression results) load data that has this precomputed.
 * Try to re-use datasets, this reduces the amount of data that needs to be downloaded to the CI server.
--- a/scanpy/plotting/_tools/scatterplots.py
+++ b/scanpy/plotting/_tools/scatterplots.py
@@ -106,17 +106,6 @@ def embedding(
     Returns
     -------
     If `show==False` a :class:`~matplotlib.axes.Axes` or a list of it.
-
-    Examples
-    --------
-
-    .. plot::
-        :context: close-figs
-
-        >>> import scanpy as sc
-        >>> pbmc = sc.datasets.pbmc3k_processed()
-        >>> sc.pl.embedding(pbmc, basis="X_umap")
-
     """
     check_projection(projection)
     sanitize_anndata(adata)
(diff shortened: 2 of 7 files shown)


ivirshup:
Example of what the `dotplot` docs look like (locally) once this is working:

image

Does make that `bbox_inches="tight"` option seem important.

fidelram:
@ivirshup  This looks great! Thanks.  The issue with the cropping of the labels is quite annoying and indeed `bbox_inches="tight"` should help. However, I don't think is nice to add that line for each example, but, on the other hand, if we add this to the scanpy code, many figures will be affected and would need to be updated.

The same issue also affects the test images which most are cropped.

ivirshup:
Update: Nvm, I figured this out. Tt had to do with `qualname_overrides`, which I've updated.

<details>
<summary> Old problem </summary>

@flying-sheep, weird sphinx bug I'm running into:

* The readthedocs builds are failing after commit fc83ec3
* The error is:

```pytb
scanpy/scanpy/external/pp/_bbknn.py:docstring of scanpy.external.pp.bbknn:24: WARNING: py:class reference target not found: sklearn.neighbors._dist_metrics.DistanceMetric
```

* The error will still occur as long as this has been added:

```rst
    .. plot::
        :context: close-figs

        >>> import scanpy as sc
        >>> adata = sc.datasets.pbmc68k_reduced()
        >>> sc.tl.umap(adata)
```

* If I remove the `sc.tl.umap` line, the builds work fine, as `sklearn.neighbors.DistanceMetric` resolves and no warning is thrown.

For now, I'm going to remove the type annotation from `sc.external.pp.bbknn`, since it's causing the error. Any ideas why calling `sc.tl.umap` means `sklearn.neighbors._dist_metrics.DistanceMetric` can no longer resolve? I assume it has something to do with packages being imported in an unexpected order, but also this is real weird.

</details>

ivirshup:
@fidelram 

I don't think making a broader change would be so bad right now, especially if it deals with this issue. However, if there was a way to just fix the issue here specifically that would also be good.

Separate issue these examples raise:

It looks like the colormap is changing at some point when it shouldn't:

image

Any idea what's up with this?

ivirshup:
For this PR, I'm thinking I'm going to wait on a response from @flying-sheep to figure out what to do about the sklearn intersphinx problem, then clean it up for a merge. A new issue will be opened up with all the functions that need examples, and those can be added through separate PRs as the release progresses.

fidelram:
@ivirshup the color change is caused because of the way the defaults are used. This is an issue that has to do with the visualisation of defaults in the documentation. 

When you add `.style(....)` as in the second case you report, the colormap changes to `winter` although it should stay as `Reds`. This happens because the method default value of `cmap` is `winter`.  Ideally, it should be `None` in which case it can be decided if the colormap was already set or if a default is needed.

However, setting the default of the colormap to `None` removes the default information from the sphinx documentation.

We had some discussions with @flying-sheep on how to address this that are still pending but ideas on how to address this are welcome.

ivirshup:
Is there an open issue for this colormap problem? If so, could you link it? I'd like to track it for at least 1.8.

docs/dev/documentation.rst
```diff
@@ -42,7 +42,21 @@ Some key points:
...
 * Check that the docs look like what you expect them too! It's easy to forget to add a reference to function, be sure it got added and looks right.
 
-Look at `sc.tl.louvain <https://github.com/theislab/scanpy/blob/a811fee0ef44fcaecbde0cad6336336bce649484/scanpy/tools/_louvain.py#L22-L90>`__ as an example for everything mentioned here:
+Look at `sc.tl.louvain <https://github.com/theislab/scanpy/blob/a811fee0ef44fcaecbde0cad6336336bce649484/scanpy/tools/_louvain.py#L22-L90>`__ as an example for everything mentioned here.
+
+Plots in docstrings
+~~~~~~~~~~~~~~~~~~~
+
+One of the most useful things you can include in a docstring is examples of how the function should be used.
+In the case of plotting functions, it's also possible to include show the resulting plots in the rendered docs.
```
fidelram: Typo: 
".. also possible to include *and*? show the resulting plots in the rendered docs"

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -106,6 +106,17 @@ def embedding(
...
     -------
     If `show==False` a :class:`~matplotlib.axes.Axes` or a list of it.
+
+    Examples
+    --------
+
+    .. plot::
+        :context: close-figs
+
+        >>> import scanpy as sc
```
fidelram: Do you want to also remove the ">>>" prefix here as in other examples?

fidelram:
So far looking good. Maybe we can add to the TO-DO list, to fix issue with the 'cropped' labels.

scanpy/plotting/_tools/scatterplots.py
ivirshup: I was actually planning on removing this one, just because I think it's not a very useful example. I have some idea's about how we should handle example for the embedding plots (since they share so much, and are fairly complex) but I'll tag you in another issue about that.

docs/dev/documentation.rst
ivirshup: I think I meant 

> In the case of plotting functions, it's also possible to ~include~ show the resulting plots in the rendered docs.

But I've reworded the whole thing. It could probably use a rework, but the main point it to link to examples, and make notes about computation.

ivirshup:
I think this is ready to merge. After that, I'll open the meta-issue about adding examples to all the listed functions.

PR #1637, opened 2021-02-10
Fix sc.get.obs_df when keys not passed

### LINKED ISSUE #1634: sc.get.obs_df() gives an error when `obsm_keys` is given and keys are not given
Calling `sc.get.obs_df()` without the `keys` parameter causes an error:

```python
sc.get.obs_df(adata, obsm_keys=[('X_umap', 0,)])
```

```pytb
~/scanpy/scanpy/get/get.py in obs_df(adata, keys, obsm_keys, layer, gene_symbols, use_raw)
    301 
    302     # reorder columns to given order (including duplicates keys if present)
--> 303     df = df[keys]
    304     for k, idx in obsm_keys:
    305         added_k = f"{k}-{idx}"

KeyError: ()
```

Also, if `keys` is not a list, the object returned is not a pandas dataframe but a Series object in which the last row is the obsm values. In other words, instead of adding a column to a dataframe with the obsm values, a row is added to a pandas Series. 

```python
sc.get.obs_df(adata, obsm_keys=[('X_umap', 0,)], keys='CST3')
```
```
index
AAAGCCTGGCTAAC-1                                                0.281
AAATTCGATGCACA-1                                               -0.176
AACACGTGGTCTTT-1                                               -0.818
AAGTGCACGTGCTA-1                                               -0.818
ACACGAACGGAGTG-1                                                0.854
                                          ...                        
TGTGAGTGCTTTAC-8                                               -0.069
TGTTACTGGCGATT-8                                               -0.818
TTCAGTACCGGGAA-8                                               -0.818
TTGAGGTGGAGAGC-8                                                0.428
X_umap-0            [-1.9918625454649166, -3.2486919412134108, -3....
Name: CST3, Length: 701, dtype: object
```

ivirshup:
Fixes #1634

### DIFF
--- a/docs/release-latest.rst
+++ b/docs/release-latest.rst
@@ -9,7 +9,8 @@ Version 1.7
 
 .. rubric:: Bugfixes
 
-- :func:`~scanpy.tl.dendrogram` use `1 - correlation` as distance matrix to compute the dendrogram :pr`1614` :smaller:`F Ramirez`
+- :func:`~scanpy.tl.dendrogram` use `1 - correlation` as distance matrix to compute the dendrogram :pr:`1614` :smaller:`F Ramirez`
+- Fixed :func:`~scanpy.get.obs_df`/ :func:`~scanpy.get.var_df` erroring when `keys` not passed :pr:`1637` :smaller:`I Virshup`
 
 1.7.0 :small:`2021-02-03`
 ~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


ivirshup:

fidelram:
@ivirshup Does this PR also solve the issue when keys is a str? I thought somewhere in the code we needed to add something like

```python
if isinstance(keys, str):
 keys = [keys]
````

ivirshup:
I think we should also do that, but I don't think that's connected to this bug.

fidelram:
It is a different issue but related because if `keys` is an string and `obsm` is also given, the returned DataFrame is wrong because it only has one column and all the `obsm` data is added in one field.

PR #1640, opened 2021-02-10
Remove scanpy.api

### LINKED ISSUE #1627: Remove `scanpy.api`
Finally get around to deleting `scanpy.api`. It's occasionally causing issues, makes docs more complicated and has been deprecated for a while.

ivirshup:
Fixes #1627


flying-sheep:
Cool! Will take a closer look tomorrow morning.

.azure-pipelines.yml
```diff
@@ -41,8 +41,8 @@ jobs:
     displayName: 'Set env'
 
   - script: |
-      python -m pip install --upgrade pip
-      pip install pytest-cov wheel
+      python -m pip install --upgrade pip || exit 1
+      pip install pytest-cov wheel || exit 1
```
flying-sheep: IDK if there’s a more elegant way to do this but without these changes, the pipelines don’t make too much sense.

E.g. you didn’t blacken things and the pipeline setup didn’t catch it. It’s going to be easy to forget to add the `|| exit 1`. Wish there was a global option.

See microsoft/azure-pipelines-tasks#10125

flying-sheep (approved):

.azure-pipelines.yml
ivirshup: Can we keep this these changes to a separate PR?

Anyways, black should be under it's own check, probably through pre-commit.

ivirshup: Also, most of these tasks should fail if the earlier steps didn't complete right? For instance, if we fail to build wheels, check should fail.

flying-sheep: Steps yes, but individual steps will succeed if the last *command* happens to succeed.

Earlier commands will be ignored in determining the success of steps.

ivirshup: Yeah, but that's just bash being bash, right?

It's kinda crazy they don't just have a "checked-script" task. However, I don't think adding `|| exit 1` after every command is better than adding `set -e` to the top of some commands – or splitting the command that's doing two checks into two tasks.

PR #1642, opened 2021-02-10
enable highly_variable_genes_seurat_v3 to work with pseudocounts

giovp:
Hi,

I'm working with pseudocounts data (kallisto/alevin/salomon output). I think they are called "pseudocount": if a read is assigned to two regions (genes) , a probability is assigned (e.g. gene1=0.2, gene2=0.8). Nevertheless, they can still considered counts and so it would be cool to use the `highly_variable_genes flavour=seuratv3` . 
I added an additional argument in case users would like to enforce this, as it was similarly done/discussed in  https://github.com/theislab/scvelo/issues/190

Would like to hear what you guys think, pinging @adamgayoso (thanks for the great overleaf doc! )

### DIFF
--- a/scanpy/preprocessing/_highly_variable_genes.py
+++ b/scanpy/preprocessing/_highly_variable_genes.py
@@ -21,6 +21,7 @@ def _highly_variable_genes_seurat_v3(
     layer: Optional[str] = None,
     n_top_genes: int = 2000,
     batch_key: Optional[str] = None,
+    pseudocount: Optional[bool] = False,
     span: Optional[float] = 0.3,
     subset: bool = False,
     inplace: bool = True,
@@ -57,10 +58,10 @@ def _highly_variable_genes_seurat_v3(
         )
 
     X = adata.layers[layer] if layer is not None else adata.X
-    if check_nonnegative_integers(X) is False:
+    if check_nonnegative_integers(X) is False and pseudocount is False:
         raise ValueError(
             "`pp.highly_variable_genes` with `flavor='seurat_v3'` expects "
-            "raw count data."
+            "raw count data. Consider setting `pseudocount=True` for enforcing it."
         )
 
     if batch_key is None:
@@ -97,7 +98,9 @@ def _highly_variable_genes_seurat_v3(
         else:
             clip_val_broad = np.broadcast_to(clip_val, batch_counts.shape)
             np.putmask(
-                batch_counts, batch_counts > clip_val_broad, clip_val_broad,
+                batch_counts,
+                batch_counts > clip_val_broad,
+                clip_val_broad,
             )
 
         if sp_sparse.issparse(batch_counts):
@@ -299,6 +302,7 @@ def highly_variable_genes(
     flavor: Literal['seurat', 'cell_ranger', 'seurat_v3'] = 'seurat',
     subset: bool = False,
     inplace: bool = True,
+    pseudocount: Optional[bool] = False,
     batch_key: Optional[str] = None,
 ) -> Optional[pd.DataFrame]:
     """\
@@ -417,6 +421,7 @@ def highly_variable_genes(
             layer=layer,
             n_top_genes=n_top_genes,
             batch_key=batch_key,
+            pseudocount=pseudocount,
             span=span,
             subset=subset,
             inplace=inplace,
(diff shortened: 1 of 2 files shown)


adamgayoso:
So, without any evidence, I think it should be fine. The reason I had put an error in the first place is that the typical behavior is to pass log normalized data to this HVG function, and I didn't want people to run this incorrectly.

I think another solution would be to just throw a UserWarning, though in a way I like the idea of having an argument that disables the `check_nonnegative_integer()`. I think I would call it `enforce_counts_seurat_v3` though.

You might also consider bypassing the check if the flag is set, because it can be slowish for large datasets.

ivirshup:
How about a `check_values` argument which defaults to `True`, and replacing the error with a warning like:

"Your data doesn't appear to be have integer values, but this method expects raw count data. Proceed with caution."

`check_values` is nice and generic, and could be used in other functions.

adamgayoso:
> check_values is nice and generic, and could be used in other functions.

Makes sense to me. Though if `check_values` is False, it shouldn't even run the check, correct?

So the warning would be more like,

"'seurat_v3' flavor expects raw count data. Proceed with caution."

ivirshup:
> Though if check_values is False, it shouldn't even run the check, correct?

Yes.

> So the warning would be more like,
>
> "'seurat_v3' flavor expects raw count data. Proceed with caution."

I was thinking the warning would only occur if `check_values=True` and non integer values were found. "partial counts" are common enough I'd agree it's not reasonable for this to error if the results are still meaningful.

giovp:
thanks guys, added the `check_values` argument and triggred the warning.

I tried to add a test for the warning following: https://docs.pytest.org/en/stable/warnings.html
but didn't manage (couldn't find a way to test for `logging` ). not sure if useful anyway.

Thank you!

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -57,11 +58,11 @@ def _highly_variable_genes_seurat_v3(
...
-    if check_nonnegative_integers(X) is False:
-        raise ValueError(
-            "`pp.highly_variable_genes` with `flavor='seurat_v3'` expects "
-            "raw count data."
-        )
+    if check_values:
+        if check_nonnegative_integers(X) is False:
+            logg.warning(
+                "`flavor='seurat_v3'` expects raw count data, but non-integers were found."
+            )
```
ivirshup: Can this be an actual [`UserWarning`](https://docs.python.org/3/library/exceptions.html#UserWarning)?

ivirshup: Logging frameworks should have a way of recording warnings on their own. In general, we should not be using `logg.warning`

ivirshup:
```suggestion
           warn(
                "`flavor='seurat_v3'` expects raw count data, but non-integers were found.",
                UserWarning
            )
```

giovp: done, and added test

giovp:
I will merge this!

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -57,10 +57,10 @@ def _highly_variable_genes_seurat_v3(
...
 
     X = adata.layers[layer] if layer is not None else adata.X
-    if check_nonnegative_integers(X) is False:
-        raise ValueError(
-            "`pp.highly_variable_genes` with `flavor='seurat_v3'` expects "
-            "raw count data."
+    if check_values:
+        warnings.warn(
+            "`flavor='seurat_v3'` expects raw count data, but non-integers were found.",
+            UserWarning,
```
adamgayoso, line 63: I think a bug has been introduced @ivirshup @giovp. Namely, the call to `check_nonnegative_integers(X)` was removed (should be here on line 61) if returns False, raise the warning

adamgayoso: On another note, in this current state `check_nonnegative_integers` is an unused import. Do you guys check for this with flake8?

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -21,6 +20,7 @@ def _highly_variable_genes_seurat_v3(
     layer: Optional[str] = None,
     n_top_genes: int = 2000,
     batch_key: Optional[str] = None,
+    check_values: Optional[bool] = True,
     span: Optional[float] = 0.3,
```
adamgayoso, line 24: shouldn't both **not** be `Optional`?

giovp:
really bad job from my side, thanks @adamgayoso for checking it🙏  reverted PR on master
will fix it shortly

PR #1649, opened 2021-02-12
compare_abs argument in filter_rank_genes_groups

Koncopd:
Argument to compare absolute values of log fold change with `min_fold_change`.
https://github.com/theislab/scanpy/issues/1325
I think `compare_abs` is a better name than `rankby_abs` for this.

### DIFF
--- a/docs/release-notes/1.9.0.rst
+++ b/docs/release-notes/1.9.0.rst
@@ -1,2 +1,6 @@
 1.9.0 :small:`the future`
 ~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. rubric:: Features
+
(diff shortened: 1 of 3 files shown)


fidelram:
I take note that we should allow all this parameters to be passed from the graphical functions to `sc.get.rank_genes_groups_df` in #1529

ivirshup:
@Koncopd @fidelram I've added a discussion of this function to the agenda for the next call. I think it would be good to figure out our plan before making changes.

Koncopd:
@ivirshup @fidelram So what do we do with this?

ivirshup:
I believe @fidelram isn't super available for scanpy stuff until next month. I figured we could talk about this then.

PR #1655, opened 2021-02-16
Drop passing of **kwds

mvdbeek:
This matches the else branch and fixes problem when `figsize` or `color` for instance
are passed in.


ivirshup:
Thanks for the contribution @mvdbeek!

This makes sense to me, passing the same kwargs to violin and stripplot seems like it would cause issues. Any thoughts @fidelram?

fidelram:
Looks good to me as well. I will merge.

ivirshup:

PR #1656, opened 2021-02-16
Fix groupby logic if adata.uns[key]['params']['groupby'] is array

mvdbeek:
Otherwise this fails with:
```
~/miniconda3/envs/anndata/lib/python3.8/site-packages/scanpy/plotting/_anndata.py in _prepare_dataframe(adata, var_names, groupby, use_raw, log, num_categories, layer, gene_symbols)
   1828                 import IPython
   1829                 IPython.embed()
-> 1830                 raise ValueError(
   1831                     'groupby has to be a valid observation. '
   1832                     f'Given {group}, is not in observations: {adata.obs_keys()}' + msg

ValueError: groupby has to be a valid observation. Given ['cell_type'], is not in observations: ['cell_type'] or index name "index"
```

### DIFF
--- a/scanpy/plotting/_tools/__init__.py
+++ b/scanpy/plotting/_tools/__init__.py
@@ -379,7 +379,9 @@ def _rank_genes_groups_plot(
         key = 'rank_genes_groups'
 
     if groupby is None:
-        groupby = str(adata.uns[key]['params']['groupby'])
+        groupby = adata.uns[key]['params']['groupby']
+        if not isinstance(groupby, (str, np.ndarray)):
+            groupby = str(groupby)
     group_names = adata.uns[key]['names'].dtype.names if groups is None else groups
 
     var_names = {}  # dict in which each group is the key and the n_genes are the values


ivirshup:
Thanks for the PR!

How did you end up with an array for that value?

ivirshup (commented):
Thanks for the PR!

How did you end up with an array for that value?

mvdbeek:
That is a good question, I'm just porting our test suite that we used in 1.4.4.post1, so I assume this is an old default ?

ivirshup:
I remember there being a bug at some point with anndata where files written by one version would have scalar values read into an array in a new version, but I'd thought that was fixed. Do you know how this anndata object was generated/ when?

mvdbeek:
Unfortunately I just know that the files were added in https://github.com/galaxyproject/tools-iuc/commit/8dc1fcd2a14f1548249fbc6570bcf399de83703d. If this shouldn't be happening anymore I suppose I could just change `adata.uns[key]['params']['groupby']` to a string in our test data.

mvdbeek:
Yeah, updating our test data solved this, I don't think this is necessary.

PR #1659, opened 2021-02-18
Fix passing of arguments between scrublet routines

### LINKED ISSUE #1644: scrublet ignores threshold
- [] (optional) I have confirmed this bug exists on the master branch of scanpy.

---

```python
import scanpy as sc
adata = sc.datasets.pbmc3k()
sc.external.pp.scrublet(adata, threshold='I am ignored')
```

```pytb
/opt/conda/lib/python3.7/site-packages/scanpy/preprocessing/_normalization.py:138: UserWarning: Revieved a view of an AnnData. Making a copy.
  view_to_actual(adata)

Automatically set threshold at doublet score = 0.27
Detected doublet rate = 1.5%
Estimated detectable doublet fraction = 44.3%
Overall doublet rate:
	Expected   = 5.0%
	Estimated  = 3.4%

```

#### Versions

<details>

scanpy==1.7.0 anndata==0.7.5 umap==0.5.1 numpy==1.20.0 scipy==1.6.0 pandas==1.2.1 scikit-learn==0.24.1 statsmodels==0.12.2 python-igraph==0.8.3 louvain==0.7.0 leidenalg==0.8.3

</details>

pinin4fjords:
Just a small PR to reinstate correct passing of parameters to Scrublet. Addresses https://github.com/theislab/scanpy/issues/1644.

### DIFF
--- a/scanpy/tests/external/test_scrublet.py
+++ b/scanpy/tests/external/test_scrublet.py
@@ -4,6 +4,7 @@ import scanpy as sc
 import scanpy.external as sce
 from anndata.tests.helpers import assert_equal
 
+
 def test_scrublet():
     """
     Test that Scrublet run works.
--- a/scanpy/tests/external/test_scrublet.py
+++ b/scanpy/tests/external/test_scrublet.py
@@ -13,7 +13,7 @@ def test_scrublet():
     pytest.importorskip("scrublet")
 
     adata = sc.datasets.pbmc3k()
-    sce.pp.scrublet(adata, use_approx_neighbors = False)
+    sce.pp.scrublet(adata, use_approx_neighbors=False)
 
     errors = []
 
--- a/scanpy/tests/external/test_scrublet.py
+++ b/scanpy/tests/external/test_scrublet.py
@@ -2,9 +2,7 @@ import pytest
 
 import scanpy as sc
 import scanpy.external as sce
-import io
-from contextlib import redirect_stdout
-
+from anndata.tests.helpers import assert_equal
 
 def test_scrublet():
     """
--- a/scanpy/tests/external/test_scrublet.py
+++ b/scanpy/tests/external/test_scrublet.py
@@ -40,7 +40,7 @@ def test_scrublet_params():
 
     f = io.StringIO()
     with redirect_stdout(f):
-        sce.pp.scrublet(adata)
+        sce.pp.scrublet(adata, use_approx_neighbors=False)
 
     default = f.getvalue()
 
@@ -59,7 +59,7 @@ def test_scrublet_params():
     # Test each parameter and make sure something changes
 
     for param in test_params.keys():
-        test_args = {'adata': adata, param: test_params[param]}
+        test_args = {'adata': adata, use_approx_neighbors=False, param: test_params[param]}
         f = io.StringIO()
         with redirect_stdout(f):
             sc.external.pp.scrublet(**test_args)
--- a/scanpy/tests/external/test_scrublet.py
+++ b/scanpy/tests/external/test_scrublet.py
@@ -34,7 +34,9 @@ def test_scrublet_params():
     """
     pytest.importorskip("scrublet")
 
-    adata = sc.datasets.pbmc3k()
+    # Reduce size of input for faster test
+    adata = sc.datasets.pbmc3k()[:500].copy()
+    sc.pp.filter_genes(adata, min_counts=100)
 
     # Get the default output
 
--- a/scanpy/tests/external/test_scrublet.py
+++ b/scanpy/tests/external/test_scrublet.py
@@ -13,7 +13,7 @@ def test_scrublet():
     pytest.importorskip("scrublet")
 
     adata = sc.datasets.pbmc3k()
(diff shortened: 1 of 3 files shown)


pinin4fjords:
@ivirshup - looks like something cryptic going on here with the Scrublet code, it's also affecting https://github.com/theislab/scanpy/pull/1657 so it's nothing to do with the changes in this PR. Seems to be something new, to do with annoy and its [behaviour on certain processors](https://www.bountysource.com/issues/91392758-initializing-an-annoyindex-crashes-on-amd-processors) in the CI. 

I'm looking into it, but any bright ideas?

pinin4fjords:
I've come across this in [other places](https://github.com/conda-forge/hnswlib-feedstock/pull/11#issuecomment-779420118) recently, it's something to do with 'AVX instruction sets' in CI setups, but I'm a bit out of my depth there.

pinin4fjords:
Never mind, just turned off the problem option for the purpose of the tests :-)

ivirshup (approved):
Looks good to me, thanks for the quick fix!

One possible addition, up to you. Would you like to expand the tests so argument handling is more thoroughly covered? I'm not going to require this as I would for something in the main API, and I would ask that you take pains to make sure these tests don't take to long to run. It would be nice to know have assurances that arguments are being passed around correctly. 

The easiest way to do this would probably be "the results change when the arguments change".

pinin4fjords:
Okay, test added!

Couldn't test for use_approx_neighbors since we know from the above that one version of that breaks the CI. Also, 'stdev_doublet_rate' rate seems to have no impact, but I'm fairly sure it's passed correctly, so I'm going to blame the Scrublet code itself.

scanpy/tests/external/test_scrublet.py
```diff
@@ -24,6 +26,52 @@ def test_scrublet():
...
 
+def test_scrublet_params():
+    """
+    Test that Scrublet args are passed.
+
+    Check that changes to parameters change scrublet results.
+    """
+    pytest.importorskip("scrublet")
+
+    adata = sc.datasets.pbmc3k()
```
ivirshup: As written, this is the longest running test on our CI at 35 seconds.

Reducing the size of the data should make this go much faster.

```suggestion
    # Reduce size of input for faster test
    adata = sc.datasets.pbmc3k()[:500].copy()
    sc.pp.filter_genes(adata, min_counts=100)
```

scanpy/tests/external/test_scrublet.py
```diff
@@ -24,6 +26,52 @@ def test_scrublet():
...
+            'adata': adata,
+            'use_approx_neighbors': False,
+            param: test_params[param],
+        }
+        f = io.StringIO()
+        with redirect_stdout(f):
+            sc.external.pp.scrublet(**test_args)
+        diff_result = f.getvalue()
+
+        assert diff_result != default
```
ivirshup: Instead of testing the difference in what's printed, I'd like to test that the values changed.

```python
# For my suggestion, this will need to be added to imports
from anndata.tests.helpers import assert_equal
```

```suggestion
    default = sce.pp.scrublet(adata, use_approx_neighbors=False, copy=True)

    test_params = {
        'expected_doublet_rate': 0.1,
        'synthetic_doublet_umi_subsampling': 0.8,
        'knn_dist_metric': 'manhattan',
        'normalize_variance': False,
        'log_transform': True,
        'mean_center': False,
        'n_prin_comps': 10,
        'n_neighbors': 2,
        'threshold': 0.1,
    }

    # Test each parameter and make sure something changes

    for param in test_params.keys():
        test_args = {
            'adata': adata,
            'use_approx_neighbors': False,
            'copy': True,
            param: test_params[param],
        }
        curr = sc.external.pp.scrublet(**test_args)
        with pytest.raises(AssertionError):
            assert_equal(default, curr)
```

ivirshup (changes requested):
Thanks for adding the test! Two things:

--------------

It's a bit slow:

image
 
(from the `tests` tab of the azure pipeline). I've added a suggestion on how to cut this down. Basically: subset the anndata object first, gives about a 10X speed up, which is more inline with how long I'd be willing to give this.

--------------------------

I think it'd be better to test that the values changed than what printed changed. I've added a suggestion for testing that the values in the anndata objects are different after these runs.

scanpy/tests/external/test_scrublet.py
pinin4fjords: Got it - thanks.

scanpy/tests/external/test_scrublet.py
pinin4fjords: I was mostly interested in the least expensive/ fiddly way of checking the parameters were having an impact, and the STDOUT seemed to fit the bill (I did check that it remained identical with unchanged parameters). But obviously have no issues with this approach.

ivirshup (approved):
Thanks for the updates @pinin4fjords! LGTM

pinin4fjords:
> Thanks for the updates @pinin4fjords! LGTM

Thanks for the education / help. And for Scanpy of course :-)

ivirshup:

PR #1667, opened 2021-02-19
Cleanup normalize_total

ivirshup:
I was looking over normalize_total and saw some strange behaviour. Since it's such a common function, I think it's important that it has standard scanpy behaviour. To this end, this PR looks at cleanup up it's code.

### Addition

`layer` argument. A specific layer can now be normalized by itself.

### Deprecations

I've deprecated the `layers` and `layer_norm` argument. Normalizing multiple layers at once seems less useful than normalizing a specific layer. These seem like very specific use cases that are easy for user's to implement themselves, and are not common patterns in scanpy functions.

### TODO:

- [x] Tests for deprecations 
- [x] Scheduling of deprecations (deprecate in 1.8, remove in 1.9)

### DIFF
--- a/scanpy/preprocessing/_normalization.py
+++ b/scanpy/preprocessing/_normalization.py
@@ -1,4 +1,5 @@
 from typing import Optional, Union, Iterable, Dict
+from warnings import warn
 
 import numpy as np
 from anndata import AnnData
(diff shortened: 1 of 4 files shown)


scanpy/preprocessing/_normalization.py
```diff
@@ -127,9 +121,30 @@ def normalize_total(
            [ 0.5,  0.5,  0.5,  1. ,  1. ],
            [ 0.5, 11. ,  0.5,  1. ,  1. ]], dtype=float32)
     """
+    if copy:
+        if not inplace:
+            raise ValueError()
```
ivirshup:
```suggestion
            raise ValueError("`copy=True` cannot be used with `inplace=False`.")
```

PR #1672, opened 2021-02-21
Attempt server-side search

ivirshup:
Inspired by @gokceneraslan pointing to [Hail.is]()'s cool doc search I did a little research. It looks like `readthedocs` might offer something similar: https://docs.readthedocs.io/en/stable/server-side-search.html

While those docs make it seem like it should be on by default, that doesn't seem to be the case. This is an attempt to figure out how to get nice search like that with read the docs. An alternative is to use [DocSearch](https://docsearch.algolia.com).

First attempt: add [readthedocs-sphinx-search](https://readthedocs-sphinx-search.readthedocs.io/en/latest/index.html).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -51,6 +51,7 @@ extensions = [
     # 'ipython_directive',
     # 'ipython_console_highlighting',
     'scanpydoc',
+    "sphinx_search.extension",
     *[p.stem for p in (HERE / 'extensions').glob('*.py')],
 ]
 
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setup(
         doc=[
             'sphinx>=3.2',
             'sphinx_rtd_theme>=0.3.1',
+            'readthedocs-sphinx-search',
             'sphinx_autodoc_typehints',
             'scanpydoc>=0.5',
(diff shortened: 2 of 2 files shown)


ivirshup:
Sorta!

image

Not sure if it's not working because something is wrong with the configuration, because it doesn't work with PRs, or that it takes a bit for search results to be available.

One downside of using this over algolia's search is that we get search analytics through algolia, while we'd have to upgrade our readthedocs subscription to have access to that.

ivirshup:
I think it's not showing results because it's a PR build.

If I make the same request to `latest`
```python
In [1]: import requests
   ...: URL = 'https://scanpy.readthedocs.io/_/api/v2/search/'
   ...: params = {
   ...:    'q': 'plotting',
   ...:    'project': 'icb-scanpy',
   ...:    'version': 'latest',
   ...: }
   ...: response = requests.get(URL, params=params)
   ...: print(response.json())
```

There are results. Using the PR build:

image

Not sure how to confirm this besides merging and seeing how it goes. Would be good to get input before that.

ivirshup:
I'm going to merge this to see if it works, and so we can try it out. If it doesn't work, we can revert.

falexwolf:
Awesome! 🎉

Issue #1563, opened 2021-01-04
Pre-commit (dev workflow)


ivirshup:
I'd like to start using [pre-commit](https://pre-commit.com) with scanpy and anndata. Pre-commit is essentially a tool that manages scripts we'd like to run before each commit, e.g. linting and formatting, so it becomes essentially impossible to forget these.

I think this can allow PRs to progress faster since it gives us a way to codify formatting requirements – so we don't have to remember them – and have these checks happen locally – so we don't have to wait on CI. Of course, having these checks run depends on developers installing pre-commit, so we can also run these checks on CI ([example ci script](https://github.com/pandas-dev/pandas/blob/master/.github/workflows/pre-commit.yml), [example run](https://github.com/pandas-dev/pandas/pull/38745/checks?check_run_id=1624558250)).

There is a question of what things we'd like to add here. For sure: `black`. I think import checks (e.g. no unused imports) and `flake8` would be good too. We can also add custom checks for things like slow imports.

I think this would be a good time to run `black` over the whole codebase so we don't have exempted files any more.

My questions for the dev team:

* Does this sound good?
* Do you have more ideas for checks/ tools?

@michalk8, I saw you added this to [`squidpy`](https://github.com/theislab/squidpy/pull/203). How's the experience been there – that is, any major foot guns we should look out for? Also, are there any tools you're using (beyond the basic `black`, `isort`, `flake8`) you'd especially recommend?

LuckyMD:
I like this idea a lot. I recently learned about automated linting and blacking of code per commit and have started using it for the single cell open problems project upon the suggestion of @scottgigante.

michalk8:
@ivirshup 

I'd recommend these:
- `mypy` for type checking (I've caught a few bugs using it)
- `autoflake` to remove unused variables/imports (by default removes only imports from Python's standard library)
- `yesqa` to automatically remove unused `# noqa` for flake8
- `check-yaml` and `pretty-format-yaml` I also found useful
- `requirements-txt-fixer` to sort the reqs.
- `check-added-large-files` have prevented my from commiting some `.h5ad` files that shouldn't be there
- `rstcheck` to check the syntax of .rst files
- some flake8: `pytest-style` (enforces some pytest conventions, like non-returning fixtures beginning with `_`), `blind-except` (no blanket exceptions) and `comprehensions` (helps you rewrite comprehensions)

Here's also an exhaustive list from which I picked the ones I use: https://pre-commit.com/hooks.html

[...]

PR #1679, opened 2021-02-23
enables highly_variable_genes_seurat_v3 to accept pseudocounts

giovp:
fixes #1642 .

I'll reply to the comments here @adamgayoso 

> shouldn't both not be Optional? 

I removed optional `type` from both span and `check_values`

> On another note, in this current state check_nonnegative_integers is an unused import. Do you guys check for this with flake8?

we don't have pre-commit in place, we are discussing it here #1563 . I do check flake8 but clearly didn't do it this time.

> I think a bug has been introduced @ivirshup @giovp. Namely, the call to check_nonnegative_integers(X) was removed (should be here on line 61) if returns False, raise the warning

this should be fixed now

Sorry again for very sloppy handling of this, should be ready to review


adamgayoso (approved):
LGTM

adamgayoso:
> we don't have pre-commit in place, we are discussing it here #1563 . I do check flake8 but clearly didn't do it this time.

I thought at one point you guys were checking flake8 with CI.

ivirshup:
> I thought at one point you guys were checking flake8 with CI.

We were, sorta. The CI tool we were using had pretty stochastic reporting (which isn't really what we want in a CI tool). Hopefully it'll be back in a more reliable form soon: https://github.com/theislab/scanpy/issues/1563#issuecomment-784922713

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -57,11 +57,12 @@ def _highly_variable_genes_seurat_v3(
...
-        raise ValueError(
-            "`pp.highly_variable_genes` with `flavor='seurat_v3'` expects "
-            "raw count data."
-        )
+    if check_values:
+        if check_nonnegative_integers(X) is False:
+            warnings.warn(
+                "`flavor='seurat_v3'` expects raw count data, but non-integers were found.",
+                UserWarning,
+            )
```
```suggestion
    if check_values and (check_nonnegative_integers(X) == False):
        warnings.warn(
            "`flavor='seurat_v3'` expects raw count data, but non-integers were found.",
            UserWarning,
        )
```

* `is` to `==` since `is` can be weird (object identity, not value) 
* `and` since it's short-circuiting and reduces total indentation

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -300,6 +303,7 @@ def highly_variable_genes(
     subset: bool = False,
     inplace: bool = True,
     batch_key: Optional[str] = None,
+    check_values: Optional[bool] = True,
```
ivirshup: Should this just be `bool`?

```suggestion
    check_values: bool = True,
```

ivirshup (approved):
Some minor style comments (up to you on making the changes), but looks good overall.

ivirshup:

PR #1684, opened 2021-02-24
Pre-commit

ivirshup:
Initial PR for pre-commit. 

* Only uses black for checks.
* Adds CI task
* Adds dev docs


ivirshup:
turns out you have to add the github action yml in the first commit for it to register.

PR #1689, opened 2021-02-24
add flake8 pre-commit

Zethson:
I intervened a couple of times manually and switched off a couple of checks, but it should be a strong improvement.

Signed-off-by: Zethson <lukas.heumos@posteo.net>


Zethson:
`E   ImportError: cannot import name 'settings' from partially initialized module 'scanpy' (most likely due to a circular import) (/home/vsts/work/1/s/scanpy/__init__.py)
`

Meh, it's hell to track this down now. I assume that autopep8 removed an unused variable.

Zethson:
```
$ python -m scanpy.tests.blackdiff 10

/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/setuptools_scm/git.py:68: UserWarning: "/home/travis/build/theislab/scanpy" is shallow and may cause errors

  warnings.warn('"{}" is shallow and may cause errors'.format(wd.path))

/home/travis/virtualenv/python3.7.1/bin/python: No module named scanpy.tests.blackdiff
```

Not sure why this happened, but we well we're using black with pre-commit now anyways so w/e. Does this need fixing?

.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+exclude = docs, scanpy/tests
+max-line-length = 120
+ignore = F401, W503, E501, E203, E231, W504, E402, E126, E712, E741
```
ivirshup: For checks that are switched off, could there be some documentation about why? If it's taken from a template it'd be fine to just link to where that came from.

pyproject.toml
```diff
@@ -26,8 +26,8 @@ source = ['scanpy']
 omit = ['*/tests/*']
 
 [tool.black]
-line-length = 88
-target-version = ['py36']
+line-length = 120
```
ivirshup: This shouldn't change, it's definitely out of scope for the pr, and I think 120 is too long.

It also makes reviewing the diffs in this PR difficult. Could you change this and I'll come back for a review?

ivirshup (changes requested):
Main thing: Can you get rid of the line length change and re-request review?

-----------------------

> scanpy.tests.blackdiff

This file got deleted, so I'm actually kinda surprised you're only getting a warning. Whatever this check is can be removed.

.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
+exclude = docs, scanpy/tests
```
ivirshup: Why exclude docs and tests?

I expect some pathological things in tests (unused imports for pytest fixtures), can we disable specific checks by path?

.pre-commit-config.yaml
```diff
@@ -3,3 +3,8 @@ repos:
     rev: 20.8b1
     hooks:
     -   id: black
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: 3.8.4
+    hooks:
+    -   id: flake8
+        exclude: scanpy/tests/
```
ivirshup: Can we just have exclude be defined in one place?

.flake8
```diff
@@ -0,0 +1,4 @@
+[flake8]
```
ivirshup, line 2: A comment here about when this can be moved to the `pyproject.toml` would be useful. Could be [a link to this](https://gitlab.com/pycqa/flake8/-/issues/428#note_251982786)

.pre-commit-config.yaml
Zethson: No, unfortunately not. pre-commit ignores excludes in the .flake8 file. 

https://github.com/pre-commit/pre-commit-hooks/issues/112

.flake8
Zethson: Oh, I did this before to get at least some sanity. There was so much stuff to fix that I excluded them and then forgot.

https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-per-file-ignores

pyproject.toml
Zethson: Then I guess this will a follow up discussion :)

88 makes for some really ugly code imo and 120 is not a lot. But will see what everyone says :)

ivirshup: Haha, last time we went through this I was the one pushing hard for increasing it from 80 😆

scanpy/plotting/_utils.py
```diff
@@ -619,7 +619,7 @@ def setup_axes(
     figure_width = width_without_offsets + left_offset + right_offset
     draw_region_width_frac = draw_region_width / figure_width
     left_offset_frac = left_offset / figure_width
-    right_offset_frac = 1 - (len(panels) - 1) * left_offset_frac
+    right_offset_frac = 1 - (len(panels) - 1) * left_offset_frac  # noqa: F841
```
ivirshup: ?

scanpy/plotting/_utils.py
```diff
@@ -939,8 +937,8 @@ def make_pos(pos, node=root, currentLevel=0, parent=None, vert_loc=0):
     if levels is None:
         levels = make_levels({})
     else:
-        levels = {l: {TOTAL: levels[l], CURRENT: 0} for l in levels}
-    vert_gap = height / (max([l for l in levels]) + 1)
+        levels = {level: {TOTAL: levels[level], CURRENT: 0} for level in levels}
+    vert_gap = height / (max([level for level in levels]) + 1)
```
ivirshup: I might want to disable this rule (which one is this?). Single letter variable names in comprehension expressions are fine.

scanpy/preprocessing/_deprecated/highly_variable_genes.py
```diff
@@ -170,8 +170,10 @@ def filter_genes_dispersion(
         disp_mean_bin[one_gene_per_bin] = 0
         # actually do the normalization
         df['dispersion_norm'] = (
-            df['dispersion'].values  # use values here as index differs
-            - disp_mean_bin[df['mean_bin'].values].values
+            df['dispersion'].values
+            - disp_mean_bin[
+                df['mean_bin'].values
+            ].values  # use values here as index differs
```
ivirshup: I think this comment ended up pointing at the wrong place.

scanpy/neighbors/__init__.py
```diff
@@ -126,7 +124,6 @@ def neighbors(
     **distances** : sparse matrix of dtype `float32`.
         Instead of decaying weights, this stores distances for each pair of
         neighbors.
-
```
ivirshup, line 129: I feel like there should be a blank line here. Was this a flake8 error?

scanpy/plotting/_anndata.py
```diff
@@ -1960,7 +1960,7 @@ def _plot_gene_groups_brackets(
                     va='bottom',
                     rotation=rotation,
                 )
-            except:
+            except Exception:
```
ivirshup: Is this better? If anything, I feel like this should get an issue opened. Maybe a new tag for "scary code"?

scanpy/plotting/_baseplot_class.py
```diff
@@ -121,13 +121,13 @@ def __init__(
         if categories_order is not None:
             if set(self.obs_tidy.index.categories) != set(categories_order):
                 logg.error(
-                    "Please check that the categories given by "
+                    "Please check that the categories given by "  # noqa: F821
```
ivirshup: I don't see how this line could give this code (undefined variable?)

scanpy/plotting/_baseplot_class.py
```diff
@@ -121,13 +121,13 @@ def __init__(
...
                     "the `order` parameter match the categories that "
                     "want to be reordered.\n\n"
                     "Mismatch: "
-                    f"{set(obs_tidy.index.categories).difference(categories_order)}\n\n"
+                    f"{set(obs_tidy.index.categories).difference(categories_order)}\n\n"  # noqa: F821
```
ivirshup: Could you also explain these two?

scanpy/plotting/_tools/__init__.py
```diff
@@ -56,7 +56,7 @@ def pca_overview(adata: AnnData, **params):
     show = params['show'] if 'show' in params else None
     if 'show' in params:
         del params['show']
-    scatterplots.pca(adata, **params, show=False)
+    scatterplots.pca(adata, **params, show=False)  # noqa: F821
```
ivirshup: Also, how is this F821?

If it wants `scatterplots.pca(adata, show=False, **params)`, that seems fine.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -252,7 +252,7 @@ def embedding(
             na_color=na_color,
         )
 
-        ### Order points
+        # Order points
```
ivirshup, line 260: If you're getting this, could you also get rid of the `###` on line 223?

But also, what rule is this?

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -977,7 +976,7 @@ def _get_data_points(
             data_points = []
             for comp in components_list:
                 data_points.append(adata.obsm[basis_key][:, comp])
-        except:
+        except Exception:
```
ivirshup: Again, I'm not sure `except Exception` is actually better than a bare `except:`.

scanpy/plotting/_utils.py
ivirshup: @fidelram, should this line be removed?

scanpy/tools/_dpt.py
```diff
@@ -903,9 +899,9 @@ def _detect_branching_single_haghverdi16(self, Dseg, tips):
         # permutations of tip cells
         ps = [
             [0, 1, 2],  # start by computing distances from the first tip
-            [1, 2, 0],  #             -"-                       second tip
```
ivirshup, line 906: I think the formatting was better before. What code is this?

scanpy/tests/test_pca.py
```diff
@@ -40,7 +38,7 @@
 )
 
 
-def test_pca_transform(array_type):
+def test_pca_transform(array_type):  # noqa: F811
```
ivirshup: What's the error? I don't see how this fits: `redefinition of unused name from line N`

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -237,8 +236,10 @@ def _highly_variable_genes_single_batch(
         disp_mean_bin[one_gene_per_bin.values] = 0
         # actually do the normalization
         df['dispersions_norm'] = (
-            df['dispersions'].values  # use values here as index differs
-            - disp_mean_bin[df['mean_bin'].values].values
+            df['dispersions'].values
+            - disp_mean_bin[
+                df['mean_bin'].values
+            ].values  # use values here as index differs
```
ivirshup: Moving comment

scanpy/preprocessing/_simple.py
```diff
@@ -757,7 +757,9 @@ def scale(
     annotated with `'mean'` and `'std'` in `adata.var`.
     """
     _check_array_function_arguments(layer=layer, obsm=obsm)
-    return scale_array(data, zero_center=zero_center, max_value=max_value, copy=copy)
+    return scale_array(
+        data, zero_center=zero_center, max_value=max_value, copy=copy  # noqa: F821
+    )
```
ivirshup: Wait, this can't be right. I guess it's just never being called?

scanpy/tools/_rank_genes_groups.py
```diff
@@ -35,22 +35,32 @@ def _ranks(X, mask=None, mask_rest=None):
...
-        get_chunk = lambda X, left, right: merge(
-            (X[mask, left:right], X[mask_rest, left:right])
-        )
+
+        def get_chunk(X, left, right):
+            return merge((X[mask, left:right], X[mask_rest, left:right]))
+
     else:
         n_cells = X.shape[0]
-        get_chunk = lambda X, left, right: adapt(X[:, left:right])
```
ivirshup: We should turn this one off.

scanpy/plotting/_utils.py
```diff
@@ -742,9 +742,7 @@ def scatter_base(
             rectangle = [left, bottom, width, height]
             fig = pl.gcf()
             ax_cb = fig.add_axes(rectangle)
-            cb = pl.colorbar(
-                sct, format=ticker.FuncFormatter(ticks_formatter), cax=ax_cb
-            )
+            pl.colorbar(sct, format=ticker.FuncFormatter(ticks_formatter), cax=ax_cb)
```
ivirshup: Why did this line change? I assume black would get this.

scanpy/tests/conftest.py
```diff
@@ -10,6 +10,7 @@
 
 import scanpy
 
+
```
ivirshup, line 13: Is this a different rule than black?

.pre-commit-config.yaml
ivirshup: Aren't you including the tests now?

pyproject.toml
```diff
@@ -27,7 +27,7 @@ omit = ['*/tests/*']
 
 [tool.black]
 line-length = 88
-target-version = ['py36']
+target-version = ['py38']
```
ivirshup: This should be the minimum python version we accept, no? Also, still out of scope. This is already touching so many things I'd prefer that there were no extraneous changes for ease of review.

ivirshup (changes requested):
Left a fair amount of comments and definitely didn't get everything.


## Commit process

I think this review process will be easier if you can limit changes from one flake error type to a single commit. 

This will make it easier to:

* Evaluate rules on their own. Sometimes rules aren't totally clear. I'm quite confused by some of the times you've used `# noqa: F821`
* Lets me know what rule the line was violating
* Revert changes if we decide we want to deactivate a particular rule. There are a few instances of this already, and I'd like to minimize the pain of this process.

This may be kinda hard to do now, but does this sound like a reasonable way to go forward?

## Rules I don't like

* I don't like replacing `x == False` with `not x` in all cases. Sometimes a variable could be a container, and an error should be thrown. I think cases have to be evaluated for this.
* Whats with changing from single letter variables inside expressions? Seems fine to me.
* `lambda's also are generally fine.
* Whats up with removing leading `#`s from comments?

## Process for bugs

So, some of the things you've adding a `# noqa` to look like bugs. I think we need to have a plan in place for doing something about these. Do you have any suggestions?

Zethson:
About the commit process: That's far far too much work to do it like you suggested. I don't have the time for this. 

About the rules: 

1. "I don't like replacing `x == False` with `not x` in all cases. Sometimes a variable could be a container, and an error should be thrown. I think cases have to be evaluated for this." 

This should be covered by tests. In any case it is not good style and a violation.

2. "Whats with changing from single letter variables inside expressions? Seems fine to me."

They are redefinitions of earlier variables and trip up flake8. We can call them whatever we want as long it s not `l` again. 

3. "`lambda's also are generally fine."

See comment at the section.

4. "Whats up with removing leading `#`s from comments?" Not my choice either. What we have now is pep8 and flake8 compliant. If you're not happy with this we can ignore the rule.

5. "So, some of the things you've adding a `# noqa` to look like bugs. I think we need to have a plan in place for doing something about these. Do you have any suggestions?"

The noqa ignore a rule for a specific line. I did not want to "fix" these things myself since Python is a dynamic language and you never know what happens :) Ideally we eventually get rid of all noqas, but not in this PR and not by me. I don't know the internals well enough to know whether this could have any side effects.

scanpy/plotting/_utils.py
Zethson: Yeah I could remove it of course, but I wanted this PR to be rather defensive about manual interventions of mine.

scanpy/plotting/_utils.py
Zethson: It's not about the single letter, but the single letter reusage

scanpy/preprocessing/_deprecated/highly_variable_genes.py
Zethson: True -> 2x black with different line numbers does this. black is not perfect.

scanpy/plotting/_anndata.py
Zethson: No it's not better. It doesn't change anything semantically, but at least it makes flake8 shut up. This PR cannot fix all violations. Scanpy had like 200 flake8 violations. 

So +1 for collecting these in a new issue and fixing them over time, but this PR will not do that.

scanpy/tools/_rank_genes_groups.py
Zethson: The lambdas are not the issue. This was likely an unused variable.

scanpy/plotting/_utils.py
Zethson: unused variable

scanpy/tests/conftest.py
Zethson: I used autopep8 to get rid of many violations. Could be that it introduced this line

scanpy/plotting/_baseplot_class.py
Zethson: I don't know either, but this is the result

```
./scanpy/plotting/_baseplot_class.py:124:26: F821 undefined name 'obs_tidy'
./scanpy/plotting/_baseplot_class.py:124:27: F821 undefined name 'obs_tidy'
```

scanpy/plotting/_baseplot_class.py
Zethson: See above.

scanpy/plotting/_tools/__init__.py
Zethson: `./scanpy/plotting/_tools/__init__.py:59:5: F821 undefined name 'scatterplots'`

Removing the `scatterplots` makes it happy. We're importing pca at the top anyways.

scanpy/plotting/_tools/scatterplots.py
Zethson: ./scanpy/plotting/_tools/scatterplots.py:254:9: E266 too many leading '#' for block comment

scanpy/plotting/_tools/scatterplots.py
Zethson: It's not. But it makes flake8 happy. You can collect all of them and someone has to fix them later. I did not want to change the semantics.

scanpy/tools/_dpt.py
Zethson: ./scanpy/tools/_dpt.py:902:25: E262 inline comment should start with '# '
./scanpy/tools/_dpt.py:904:12: E262 inline comment should start with '# '

scanpy/tests/test_pca.py
Zethson: ./scanpy/tests/test_pca.py:41:24: F811 redefinition of unused 'array_type' from line 6

scanpy/preprocessing/_simple.py
Zethson: It doesn't know `data`. Even Pycharm flags this. I don't know how such code even ended up in Scanpy.

Zethson:
@ivirshup tried to address all comments and made changes when necessary.

scanpy/plotting/_utils.py
ivirshup: But variables defined inside generators aren't defined in outer scopes:

```python
In [1]: [[i for i in range(j)] for j in range(5)]
Out[1]: [[], [0], [0, 1], [0, 1, 2], [0, 1, 2, 3]]

In [2]: i
NameError                                 Traceback (most recent call last)
<ipython-input-2-397d543883c5> in <module>
----> 1 i

NameError: name 'i' is not defined

In [3]: j
NameError                                 Traceback (most recent call last)
<ipython-input-3-3eedd8854d1e> in <module>
----> 1 j

NameError: name 'j' is not defined
```

So, what issue is being avoided here?

scanpy/plotting/_utils.py
ivirshup: Ah, didn't see that

Why is this sometimes changed to `_ = statement` in other places? Was this an automatic tool, or manual?

scanpy/plotting/_utils.py
Zethson: E741. Even Pycharm flags it already.

scanpy/plotting/_utils.py
Zethson: Manual.

_ is the consensus for variables that get thrown away.

__ is not the standard -> I had to rename those as well

scanpy/tools/_rank_genes_groups.py
ivirshup: I'm not sure I understand the change then. How does replacing a `lambda` with a `def` solve this?

scanpy/preprocessing/_simple.py
ivirshup: For this review process, if there is code that looks bad I would prefer that you flagged it for removal rather than adding it to an ignore list.

I think this is where a lot of the value of turning `flake8` on will come from.

scanpy/tools/_rank_genes_groups.py
Zethson: Oh what the hell. Now I see it. This must have been autopep8. This makes no sense. I will readd the lambdas

scanpy/preprocessing/_simple.py
Zethson: Sure. But then you will have to fix it **now** or flake8 will keep failing :)

Whatever you prefer.

scanpy/tests/test_pca.py
ivirshup: Can we turn of rules by module? I expect this will come up a lot and I don't want to add this to every test that uses a fixture.

scanpy/tools/_rank_genes_groups.py
Zethson: I was wrong earlier.

```
./scanpy/tools/_rank_genes_groups.py:38:9: E731 do not assign a lambda expression, use a def
./scanpy/tools/_rank_genes_groups.py:39:9: E731 do not assign a lambda expression, use a def
./scanpy/tools/_rank_genes_groups.py:42:9: E731 do not assign a lambda expression, use a def
./scanpy/tools/_rank_genes_groups.py:48:9: E731 do not assign a lambda expression, use a def
./scanpy/tools/_rank_genes_groups.py:53:9: E731 do not assign a lambda expression, use a def
./scanpy/tools/_rank_genes_groups.py:171:13: E731 do not assign a lambda expression, use a def
./scanpy/tools/_rank_genes_groups.py:173:13: E731 do not assign a lambda expression, use a def
./scanpy/tools/_rank_genes_groups.py:743:13: E731 do not assign a lambda expression, use a def
```

scanpy/plotting/_tools/scatterplots.py
ivirshup: I'd be for turning off E266 then. It's a legitimate way to style comments into sections. Even if it wasn't, this check is not catching bugs.

scanpy/plotting/_tools/__init__.py
ivirshup: My guess  is the intent here was to make it absolutely clear this isn't the `sc.tl.pca`. If we don't import `scatterplots.pca` here, does the warning still come up?

Zethson: Yes.

scanpy/tests/test_pca.py
Zethson: per-file-ignores =
    module: rule

Feel free to push to this branch which rules you want to turn of for whatever module.

scanpy/plotting/_utils.py
ivirshup: Ehh. If the issue is letters that look like numbers (a) in what programming font is this the case (b) they should also be different colors. I'd be for turning this rule off.

scanpy/neighbors/__init__.py
ivirshup: Did this get resolved? I'm not seeing an outdated.

scanpy/plotting/_utils.py
Zethson: I would keep it since the IDE  + flake8 flag it.

scanpy/neighbors/__init__.py
Zethson: Yeah, I added the line back

scanpy/plotting/_utils.py
Zethson: It causes 0 harm to adhere to better naming practices.

Zethson:
@ivirshup I would keep the noqas. They are very easily searchable across the whole project and can be fixed later.

To me the PR is fine now. Any final wishes?

ivirshup:
> About the commit process: That's far far too much work to do it like you suggested. I don't have the time for this.

As a general point about this PR: to me, the fair amount of the work of turning on flake8 deciding on the rules. Perhaps we should start with a subset of files then? I realize you did not come to the meeting where we talked about this, so perhaps there is a difference of expectations here, but we agreed to be conservative about the rules we turned on in `pre-commit`.

Going through everything to make sure changes are correctly reverted is also takes a lot of time for me as the reviewer. I'd also like to limit that.

----------------

You said you used some automated tools to get faster compliance. What were these? In general, I would prefer to have a formatter that automatically ran than a tool that told me I formatted something wrong.

-----------------

> `@ivirshup` I would keep the noqas. They are very easily searchable across the whole project and can be fixed later.

I'm pretty strongly against this. `noqa`s just look like the formatter/ linter was wrong, and I'm not accepting that having no plan to address bugs. I think this should be a discussion with a broader set of the team.

> "Whats up with removing leading #s from comments?" Not my choice either. What we have now is pep8 and flake8 compliant. If you're not happy with this we can ignore the rule.

Yes, lets ignore this.

>> "I don't like replacing x == False with not x in all cases. Sometimes a variable could be a container, and an error should be thrown. I think cases have to be evaluated for this."
>
> This should be covered by tests. In any case it is not good style and a violation.

I will try and take a closer look at these changes. I'm particularly concerned that there will be cases where possible values are `None`, `True`, and `False`.

Zethson:
> You said you used some automated tools to get faster compliance. What were these? In general, I would prefer to have a formatter that automatically ran than a tool that told me I formatted something wrong.

https://pypi.org/project/autopep8/

Someone suggested to use it for pre-commit anyways. However, there is no full flake8 autofixer.


> As a general point about this PR: to me, the fair amount of the work of turning on flake8 deciding on the rules. Perhaps we should start with a subset of files then? 

Possible, but this will be an iterative process that will take a long time. If you are willing to do that I can close this PR and enable flake8 for example on the first folder of Scanpy. I cannot however fix all of them manually in like 20 PRs or something.

> I'm pretty strongly against this. noqas just look like the formatter/ linter was wrong, and I'm not accepting that having no plan to address bugs. I think this should be a discussion with a broader set of the team.

Depends on the view that you have. I see them more as TODOs for later since the "bad code" is in master at the moment anyways. But if you want to discuss that - sure.

scanpy/plotting/_anndata.py
ivirshup: Note to self: this conversation is being continued in the main discussion thread.

scanpy/plotting/_baseplot_class.py
ivirshup: I expect these need `self.obs_tidy` since no variable is called `obs_tidy`

scanpy/tools/_dpt.py
ivirshup: Lets turn off E262 and revert this then.

scanpy/tools/_rank_genes_groups.py
ivirshup: I'm for turning off `E731`.

ivirshup (changes requested):

scanpy/tools/_rank_genes_groups.py
Zethson: Context:
```

    The first form means that the name of the resulting function object is specifically 'f' instead of the generic '<lambda>'. This is more useful for tracebacks and string representations in general. The use of the assignment statement eliminates the sole benefit a lambda expression can offer over an explicit def statement (i.e. that it can be embedded inside a larger expression)

Assigning lambdas to names basically just duplicates the functionality of def - and in general, it's best to do something a single way to avoid confusion and increase clarity.

The legitimate use case for lambda is where you want to use a function without assigning it, e.g:
```

I like lambdas as well so you can make the final call.

.flake8
```diff
@@ -0,0 +1,5 @@
+# Can't yet be moved to the pyproject.toml due to https://gitlab.com/pycqa/flake8/-/issues/428#note_251982786
+[flake8]
+max-line-length = 88
```
giovp, line 3: I am also more happy with max line legnth=120. We have that in squidpy as well

scanpy/plotting/_matrixplot.py
```diff
@@ -219,7 +219,7 @@ def _mainplot(self, ax):
             linewidth=self.edge_lw,
             norm=normalize,
         )
-        __ = ax.pcolor(_color_df, **kwds)
```
giovp, line 235: never seen `__` before 😅

scanpy/plotting/_utils.py
```diff
@@ -126,7 +126,7 @@ def timeseries_subplot(
     else:
         levels, _ = np.unique(color, return_inverse=True)
         colors = np.array(palette[: len(levels)].by_key()['color'])
-        subsets = [(x_range[color == l], X[color == l, :]) for l in levels]
+        subsets = [(x_range[color == level], X[color == level, :]) for level in levels]
```
giovp, line 130: much better

scanpy/preprocessing/_simple.py
giovp: but is this the `sc.pp.scale` ?  I wonder if it's tested?

scanpy/tools/_sim.py
```diff
@@ -410,9 +409,8 @@ def __init__(
         # checks
         if initType not in ['branch', 'random']:
             raise RuntimeError('initType must be either: branch, random')
-        read = False
         if model not in self.availModels.keys():
-            message = 'model not among predefined models \n'
+            message = 'model not among predefined models \n'  # noqa: F841
```
giovp: this is weird

giovp (approved):

giovp:
@Zethson thanks 💯 ! looks really good! Tbh I was expecting much worse, the changes with flake8 are pretty conservative imho.

> Depends on the view that you have. I see them more as TODOs for later since the "bad code" is in master at the moment anyways. But if you want to discuss that - sure.

I totally share this view, I think it's fine to use noqa as flag and come back later.

I admit you guys are 1000x more experts in this, but from what I can judge I think it's fine to merge this as first attempt and then in subsequent PRs improve and address flake 8 fails. 

Maybe worth to merge #1527 first? otherwise there will be potential conflicts more for phil.

.flake8
Zethson: I opened https://github.com/theislab/scanpy/issues/1694 to discuss this. Insert Bernie meme

scanpy/tools/_sim.py
Zethson: Yeah, it can absolutely be removed I guess :)
I didn't want to do it since I did not know whether you have some magic going on somewhere that looks for these "message" strings to log them or something.
But I guess we should remove or replace this with a log statement/exception here

flying-sheep:
Yes please! It’s been soooo long. I’ll happily fix the conflicts here as long as I don’t have to change the flit PR again.

Zethson:
@flying-sheep go ahead :)
I will patiently wait until your PR got merged.

ivirshup:
> I see them more as TODOs for later since the "bad code" is in master at the moment anyways. But if you want to discuss that - sure.

I think it's hard to tell the difference between a `noqa` that was added because: "most of the time, this rule is right, this time it is wrong" vs. "added to get rid of warning". Could any `noqa`s added in this PR get something searchable added to them (like `# noqa: {rule} TODO: fix me`) so we know why it was added?

In a future meeting we can discuss with the whole team how we will actually fix these. I think it's a good task for a hackathon/ sprint.

------------

Looking at this again, I think this is pretty close to done. Just a few documentation/ minor rule changes left.

- [x] Document how to turn off these checks in dev docs
- [x] Document the rules that are turned off (similar to [pandas](https://github.com/pandas-dev/pandas/blob/879cd22dd58b0574cdcaa7a26e396d5ec71a615a/setup.cfg#L71-L79))
- [x] Add autopep8 to precommit. If things can be fixed automatically, they should be. autopep8 should be able to get it's rules from  the flake8 config.
- [x] Add annotation to `noqa`s.
- [x] Also add `TODO` annotation to `except Exception`s that have been added.
- [x] Turn off E731
- [x] Turn off the rule that doesn't allow multiple leading `#` in comments
- [x] Turn off F811 for tests (rule violated by using fixture as test argument)
  - Possible solution: add noqa for this to all files under `tests`, separate check that all files under tests have this.
- [x] Ignore `build` `docs/_build` directories

Zethson:
> Could any noqas added in this PR get something searchable added to them (like # noqa: {rule} TODO: fix me) so we know why it was added?

The noqas already state what they are ignoring. I for now would not try to differentiate between noqas that we want to keep and noqas that we want to get rid of. We want to get rid of all of them in the follow up issue and only when examining all of them we will figure out which ones we want to keep.

> Document how to turn off these checks in dev docs

What do you mean? How to ignore a single line? How to fully ignore whole checks? I would always refer to the flake8 documentation, because it will certainly maintained better than the dev documentation.

> Add autopep8 to precommit. If things can be fixed automatically, they should be. autopep8 should be able to get it's rules from the flake8 config.

I wish it were that easy. autopep8 does not take its configuration from the flake8 config file nor can it fix all pep8 violations nor do Black and autopep8 always work nicely together. Black is an **opiniated** formatter. It formats consistently, but not necessarily compatible with other tools. I would not add autopep8, since I do not see any further benefit to the Black & flake8 combination, only more potential for issues and confused developers.

I agree with your other comments and will take care of them as soon as I got your answers :)

ivirshup:
> I for now would not try to differentiate between noqas that we want to keep and noqas that we want to get rid of. We want to get rid of all of them in the follow up issue and only when examining all of them we will figure out which ones we want to keep.

After this merges new ignore messages can be added for reasons like "this rule is generally good, but not in this specific case". Each of these will go through PR review, so will be vetted.

The ones added here largely have not been vetted, and are just being added so we don't get a failure. I would like to be able to distinguish between these cases. Once more `noqa` cases are added, it gets more complicated to find cases that haven't been vetted if they don't have some associated label.

--------------------------

> What do you mean? How to ignore a single line? How to fully ignore whole checks?

How to disable flake8 errors for a line or file.

> I would always refer to the flake8 documentation, because it will certainly maintained better than the dev documentation.

A link to the section of the flake8 docs on this would be great.

-------------------------

> I wish it were that easy. autopep8 does not take its configuration from the flake8 config file 

`autopep8` says it does this: https://github.com/hhatto/autopep8#configuration

> It formats consistently, but not necessarily compatible with other tools

I would like changes that are automatically applicable to be automatically applied. I'm thinking of things like white space in docstrings. Is there another way to automate these you can suggest?

---------

BTW, I've added a few more points to the checklist above. I would recommend trying to build the package and build the docs in the directory you're working in to see what files get generated so they can be added to the `ignore`.

flying-sheep:
I reverted the setup.py change which hopefully means that an alrady-compliant setup.py from the flit PR will just pass [the check](https://github.com/theislab/scanpy/actions/runs/610117826) (once it runs, seems to be queued for a while now)

Zethson:
I'll fix this PR up in ~10 days. Appreciate your comments @ivirshup and think that some of your feedback could have been mitigated by me not going for short cuts :)

Will request (final hopefully) review then.

Zethson:
@ivirshup I hope that I caught all of your comments.
The flake8 configuration is now minimal and pretty much only contains black violations or what you requested. I added and then removed autopep8 again, because it has other opinions on formatting than the opinionated formatter black. Yes, even with the flake8 configuration file. Black formatted the code then autopep8 and this cycle continues forever.
Added TODOs to exceptions and noqas are still easily searchable and mention what rule they ignore anyways.

I want to get this merged asap since the merge conflicts will just pile up.

.flake8
```diff
@@ -0,0 +1,20 @@
...
+         W503,  # line break before a binary operator -> black does not adhere to PEP8
+         W504,  # line break occured after a binary operator -> black does not adhere to PEP8
+         E501,  # line too long -> we accept long comment lines; black gets rid of long code lines
+         E203,  # whitespace before : -> black does not adhere to PEP8
+         E231,  # missing whitespace after ,', ';', or ':' -> black does not adhere to PEP8
+         E402,  # module level import not at top of file -> required to circumvent circular imports for Scanpys API
+         E126,  # continuation line over-indented for hanging indent -> black does not adhere to PEP8
+         E266,  # E266 too many leading '#' for block comment -> Scanpy allows them for comments into sections
+         E731   # Do not assign a lambda expression, use a def -> Scanpy allows lambda expression assignments
+exclude =
```
ivirshup:
```suggestion
         E731,  # Do not assign a lambda expression, use a def -> Scanpy allows lambda expression assignments
         E741,  # allow I, O, l as variable names -> I is the identity matrix, i, j, k, l is reasonable indexing notation
exclude =
```

scanpy/plotting/_tools/__init__.py
```diff
@@ -165,7 +165,6 @@ def dpt_timeseries(
 ):
     """\
     Heatmap of pseudotime series.
-
```
ivirshup: There should probably be a blank line here?

.flake8
```diff
@@ -0,0 +1,20 @@
...
+    .git,
+    __pycache__,
+    build,
+    docs/_build
+    dist
```
ivirshup:
```suggestion
    docs/_build,
    dist,
per-file-ignores = 
    # F811 Redefinition of unused name from line, does not play nice with pytest fixtures
    tests/test*.py: F811,
```

Supposedly, you're not allowed to have comments on the same lines as entries (even though this clearly works above). See: https://flake8.pycqa.org/en/latest/user/configuration.html

scanpy/neighbors/__init__.py
```diff
@@ -864,7 +860,7 @@ def _compute_connectivities_diffmap(self, density_normalize=True):
             # make the weight matrix sparse
             if not self.knn:
                 mask = W > 1e-14
-                W[mask == False] = 0
+                W[mask == False] = 0  # noqa: E712
```
ivirshup:
```suggestion
                W[~mask] = 0
```

scanpy/neighbors/__init__.py
```diff
@@ -876,7 +872,7 @@ def _compute_connectivities_diffmap(self, density_normalize=True):
                             W[j, i] = W[i, j]
                             mask[j, i] = True
                 # set all entries that are not nearest neighbors to zero
-                W[mask == False] = 0
+                W[mask == False] = 0  # noqa: E712
```
ivirshup:
```suggestion
                W[~mask] = 0
```

scanpy/neighbors/__init__.py
```diff
@@ -1027,24 +1023,24 @@ def _get_dpt_row(self, i):
             mask = self._connected_components[1] == label
         row = sum(
             (
-                self.eigen_values[l]
-                / (1 - self.eigen_values[l])
-                * (self.eigen_basis[i, l] - self.eigen_basis[:, l])
+                self.eigen_values[k]
+                / (1 - self.eigen_values[k])
+                * (self.eigen_basis[i, k] - self.eigen_basis[:, k])
```
ivirshup: This variable should probably be `j` if anything

.flake8
```diff
@@ -0,0 +1,20 @@
...
+ignore = F401,  # module imported but unused -> required for Scanpys API
+         W503,  # line break before a binary operator -> black does not adhere to PEP8
+         W504,  # line break occured after a binary operator -> black does not adhere to PEP8
+         E501,  # line too long -> we accept long comment lines; black gets rid of long code lines
+         E203,  # whitespace before : -> black does not adhere to PEP8
+         E231,  # missing whitespace after ,', ';', or ':' -> black does not adhere to PEP8
+         E402,  # module level import not at top of file -> required to circumvent circular imports for Scanpys API
+         E126,  # continuation line over-indented for hanging indent -> black does not adhere to PEP8
+         E266,  # E266 too many leading '#' for block comment -> Scanpy allows them for comments into sections
+         E731   # Do not assign a lambda expression, use a def -> Scanpy allows lambda expression assignments
```
ivirshup: All of the comments need to be on separate lines from the ignore.

See:

* https://flake8.pycqa.org/en/latest/user/configuration.html
* https://docs.python.org/3/library/configparser.html#customizing-parser-behaviour
* https://docs.python.org/3/library/configparser.html#supported-ini-file-structure

It looks like this is what was causing you problems with autopep8, which should be fine to re-enable after this. It was parsing the config in a way that doesn't allow inline comments.

scanpy/get/get.py
```diff
@@ -96,7 +96,7 @@ def rank_genes_groups_df(
 def _check_indices(
     dim_df: pd.DataFrame,
     alt_index: pd.Index,
-    dim: "Literal['obs', 'var']",
+    dim: "Literal['obs', 'var']",  # noqa: F821
```
ivirshup: Could this be replaced with a `from ._compat import Literal` at the top of the file? Ditto for other occurrences.

scanpy/plotting/_tools/__init__.py
```diff
@@ -62,7 +62,7 @@ def pca_overview(adata: AnnData, **params):
     show = params['show'] if 'show' in params else None
     if 'show' in params:
         del params['show']
-    scatterplots.pca(adata, **params, show=False)
+    scatterplots.pca(adata, **params, show=False)  # noqa: F821
```
ivirshup: Would importing `scatterplots` fix this?

scanpy/plotting/_tools/__init__.py
```diff
@@ -967,7 +966,7 @@ def rank_genes_groups_violin(
         )
         savefig_or_show(writekey, show=show, save=save)
         axs.append(_ax)
-    if show == False:
+    if not show:
```
ivirshup: This is a behavior change since `show` defaults to `False`

scanpy/plotting/_tools/paga.py
```diff
@@ -148,7 +148,7 @@ def paga_compare(
     if suptitle is not None:
         pl.suptitle(suptitle)
     _utils.savefig_or_show('paga_compare', show=show, save=save)
-    if show == False:
+    if not show:
```
ivirshup: Similarly, `show` defaults to `None` so this is a behavior change. I'm starting to think we should just ignore this rule.

scanpy/plotting/_utils.py
```diff
@@ -708,9 +708,9 @@ def scatter_base(
     )
     for icolor, color in enumerate(colors):
         ax = axs[icolor]
-        left = panel_pos[2][2 * icolor]
+        left = panel_pos[2][2 * icolor]  # noqa: F841
         bottom = panel_pos[0][0]
-        width = draw_region_width / figure_width
+        width = draw_region_width / figure_width  # noqa: F841
```
ivirshup: Probably just delete definitions of `width` and `left`

scanpy/plotting/_utils.py
```diff
@@ -940,8 +940,8 @@ def make_pos(pos, node=root, currentLevel=0, parent=None, vert_loc=0):
     if levels is None:
         levels = make_levels({})
     else:
-        levels = {l: {TOTAL: levels[l], CURRENT: 0} for l in levels}
-    vert_gap = height / (max([l for l in levels]) + 1)
+        levels = {level: {TOTAL: levels[level], CURRENT: 0} for level in levels}
+    vert_gap = height / (max([level for level in levels]) + 1)
```
ivirshup:
```suggestion
        levels = {k: {TOTAL: v, CURRENT: 0} for k, v in levels.items()}
    vert_gap = height / (max(levels.keys()) + 1)
```

scanpy/preprocessing/_deprecated/highly_variable_genes.py
```diff
@@ -170,8 +170,10 @@ def filter_genes_dispersion(
         disp_mean_bin[one_gene_per_bin] = 0
         # actually do the normalization
         df['dispersion_norm'] = (
-            df['dispersion'].values  # use values here as index differs
-            - disp_mean_bin[df['mean_bin'].values].values
+            df['dispersion'].values
+            - disp_mean_bin[  # use values here as index differs
+                df['mean_bin'].values
+            ].values
```
ivirshup:
```suggestion
            # use values here as index differs
            df['dispersion'].values
            - disp_mean_bin[df['mean_bin'].values].values
```

Though the comment should probably go outside this expression.

I would also expect these are equivalent:

```python
disp_mean_bin[df['mean_bin'].values].values
disp_mean_bin.values[df['mean_bin']]
```

scanpy/tools/_dpt.py
```diff
@@ -342,7 +342,7 @@ def check_adjacency(self):
         for n_edges in range(1, np.max(n_edges_per_seg) + 1):
             for iseg in range(self.segs_adjacency.shape[0]):
                 if n_edges_per_seg[iseg] == n_edges:
-                    neighbor_segs = self.segs_adjacency[iseg].todense().A1
+                    _ = self.segs_adjacency[iseg].todense().A1
```
ivirshup: Could these get `# noqa`s an `TODO`s? I'm assuming these should just be deleted, but am worried they're here for side effects.

scanpy/tools/_dpt.py
```diff
@@ -903,9 +899,9 @@ def _detect_branching_single_haghverdi16(self, Dseg, tips):
         # permutations of tip cells
         ps = [
             [0, 1, 2],  # start by computing distances from the first tip
-            [1, 2, 0],  #             -"-                       second tip
+            [1, 2, 0],  # -"-                       second tip
```
ivirshup:
```suggestion
            [1, 2, 0],  #              -"-              second tip
```

scanpy/tools/_dpt.py
```diff
@@ -903,9 +899,9 @@ def _detect_branching_single_haghverdi16(self, Dseg, tips):
...
             [2, 0, 1],
-        ]  #             -"-                       third tip
+        ]  # -"-                       third tip
```
ivirshup: Similar to above line, but moved up. Or revert these lines, if easier.

scanpy/tools/_rank_genes_groups.py
```diff
@@ -37,6 +37,7 @@ def _ranks(X, mask=None, mask_rest=None):
     if issparse(X):
         merge = lambda tpl: vstack(tpl).toarray()
         adapt = lambda X: X.toarray()
+
```
ivirshup: ??

Probably safe to revert the whole file.

scanpy/tools/_sim.py
```diff
@@ -1077,16 +1074,6 @@ def sim_givenAdj(self, Adj: np.ndarray, model='line'):
...
-        # nice examples
-        examples = [
-            dict(
-                func='sawtooth',
-                gdist='uniform',
-                sigma_glob=1.8,
-                sigma_noise=0.1,
-            )
-        ]
-
```
ivirshup: Ehh, I'd leave it. I'm not sure what happens in this file so I'd rather leave it untouched.

scanpy/tools/_paga.py
```diff
@@ -534,24 +533,24 @@ def paga_compare_paths(
...
+            path_mapped = [asso_groups2[link] for link in path2]
             path_compare = path1
             path_compare_id = 1
             path_compare_orig_names = [
-                [orig_names1[int(s)] for s in l] for l in path_compare
+                [orig_names1[int(s)] for s in link] for link in path_compare
             ]
             path_mapped_orig_names = [
-                [orig_names1[int(s)] for s in l] for l in path_mapped
+                [orig_names1[int(s)] for s in link] for link in path_mapped
```
ivirshup: Could these get reverted? I'm not sure `link` is an appropriate name. I suspect these are `leafs`, but would rather have someone familiar with this code make that call.

scanpy/tools/_dpt.py
ivirshup: Similar for `prev_connecting_points`

scanpy/preprocessing/_simple.py
```diff
@@ -752,7 +752,9 @@ def scale(
...
```
ivirshup:
```suggestion
    if layer is not None:
        raise ValueError(f"`layer` argument inappropriate for value of type {type(X)}")
    if obsm is not None:
        raise ValueError(f"`obsm` argument inappropriate for value of type {type(X)}")
    return scale_array(X, zero_center=zero_center, max_value=max_value, copy=copy)
```

As a fallback, we'll assume`X` is array-like.

scanpy/external/pl.py
```diff
@@ -332,15 +332,14 @@ def scrublet_score_distribution(
     figsize: Optional[Tuple[float, float]] = (8, 3),
 ):
     """\
-    Plot histogram of doublet scores for observed transcriptomes and simulated doublets. 
+    Plot histogram of doublet scores for observed transcriptomes and simulated doublets.
+
+    The histogram for simulated doublets is useful for determining the correct doublet score threshold.
```
ivirshup:
```suggestion
    The histogram for simulated doublets is useful for determining the correct doublet
    score threshold. 
```

Bit long, seemed fine before

docs/dev/code.rst
```diff
@@ -16,7 +16,13 @@ Code style
...
-`Black <https://black.readthedocs.io/en/stable/the_black_code_style.html>`__
-and Scanpy’s
+`Black <https://black.readthedocs.io/en/stable/the_black_code_style.html>`__,
+and
+`flake8 <https://flake8.pycqa.org>`__.
+We ignore a couple of flake8 checks which are documented in the .flake8 file in the root
+of this repository.
+To learn how to ignore checks per line please read
+`flake8 violations <https://flake8.pycqa.org/en/latest/user/violations.html>`__.
+Additionally, we use Scanpy’s
```
ivirshup:
```suggestion
`Black <https://black.readthedocs.io/en/stable/the_black_code_style.html>`__
and
`flake8 <https://flake8.pycqa.org>`__.
We ignore a couple of flake8 checks which are documented in the .flake8 file in the root of this repository.
To learn how to ignore checks per line please read
`flake8 violations <https://flake8.pycqa.org/en/latest/user/violations.html>`__.
Additionally, we use Scanpy’s
```

ivirshup (changes requested):
Looks good! 

I left a lot of comments/ suggestions, but generally fall into a few categories.

* Places that I understand the code, I suggested style fixes (replacing `noqa`)
* Places where it looks like formatting is held over from mixing formatters/ rules we've abandoned. I've suggested reverting here.
* Fixes/ renamed variables in code I'm not familiar with, I've suggested reverting / adding a noqa.
* Renamed variables where I think the new name isn't appropriate. I've suggested reverting or a different name.
* There were a few places where replacing `var == False` with `not var` would be a behavior change since `var` defaults to `None`, `E712` is fine with `if var is False`.

-----------------------

Bigger points:

* I think the `autopep8` issues you were coming across were a config error. I've suggested a fix in the `.flake8` file. You should be able to turn on `autopep8` again after enabling that.
* Reminder on `# noqa`s, could all of the ones you've added get a `TODO` added inline? I want to be able to tell the difference between `noqa`s added now, and those added later with an editor search. Here, they may be hiding bugs. Later ones may be legit exceptions.

scanpy/get/get.py
Zethson: Not working unfortunately.

scanpy/plotting/_tools/paga.py
Zethson: While you are correct that this is a behavior change I am not really sure whether we should really compare to ==. It is considered "bad style". https://stackoverflow.com/questions/37103705/what-is-the-correct-way-to-check-for-false

"Falsy" value should be good enough here so I would just adhere to the standard.

Final choice is yours, but I would use "not".

scanpy/tools/_paga.py
Zethson: They are redefinitions and should not stay as `l`. I'd be fine with either

1. renaming to _
2. renaming to k/j or something
3. Opening a follow up issue for this.

scanpy/plotting/_tools/__init__.py
```diff
@@ -365,7 +365,7 @@ def _fig_show_save_or_axes(plot_obj, return_fig, show, save):
         plot_obj.make_figure()
         savefig_or_show(plot_obj.DEFAULT_SAVE_PREFIX, show=show, save=save)
         show = settings.autoshow if show is None else show
-        if not show:
+        if show == False:  # noqa: E712  # TODO Does this need fixing? -> is False
```
ivirshup: I'd be fine with this:

```suggestion
        if show is False:
```

scanpy/plotting/_tools/__init__.py
```diff
@@ -967,7 +967,7 @@ def rank_genes_groups_violin(
         )
         savefig_or_show(writekey, show=show, save=save)
         axs.append(_ax)
-    if show == False:
+    if show == False:  # noqa: E712  TODO Does this need fixing?
```
ivirshup:
```suggestion
    if show is False:
```

scanpy/plotting/_tools/__init__.py
ivirshup: I think `flake8` allows this, but pep8 doesn't. Not sure if this would create conflicts with `autopep8`.

scanpy/plotting/_tools/paga.py
```diff
@@ -148,7 +148,7 @@ def paga_compare(
     if suptitle is not None:
         pl.suptitle(suptitle)
     _utils.savefig_or_show('paga_compare', show=show, save=save)
-    if not show:
+    if show == False:  # noqa: E712  # TODO Does this need fixing?
```
ivirshup:
```suggestion
    if show is False:
```

scanpy/tools/_dpt.py
```diff
@@ -899,9 +903,9 @@ def _detect_branching_single_haghverdi16(self, Dseg, tips):
         # permutations of tip cells
         ps = [
             [0, 1, 2],  # start by computing distances from the first tip
-            [1, 2, 0],  # -"-                       second tip
+            [1, 2, 0],  #             -"-                       second tip
             [2, 0, 1],
```
ivirshup:
```suggestion
            [2, 0, 1],  #             -"-                       third tip
```

scanpy/tools/_dpt.py
```diff
@@ -899,9 +903,9 @@ def _detect_branching_single_haghverdi16(self, Dseg, tips):
...
-        ]  # -"-                       third tip
+        ]  #             -"-                       third tip
```
ivirshup:
```suggestion
        ]
```

scanpy/get/get.py
```diff
@@ -6,7 +6,6 @@
 from scipy.sparse import spmatrix
 
 from anndata import AnnData
```
ivirshup, line 8:
```suggestion
from anndata import AnnData

from .._compat import Literal
```

scanpy/get/get.py
```diff
@@ -96,7 +95,7 @@ def rank_genes_groups_df(
 def _check_indices(
     dim_df: pd.DataFrame,
     alt_index: pd.Index,
-    dim: "Literal['obs', 'var']",
+    dim: "Literal['obs', 'var']",  # noqa: F821  # TODO Does this need fixing?
```
ivirshup:
```suggestion
    dim: Literal['obs', 'var'],
```

scanpy/get/get.py
```diff
@@ -176,7 +175,7 @@ def _get_array_values(
     X,
     dim_names: pd.Index,
     keys: List[str],
-    axis: "Literal[0, 1]",
+    axis: "Literal[0, 1]",  # noqa: F821  # TODO Does this need fixing?
```
ivirshup:
```suggestion
    axis: Literal[0, 1],
```

scanpy/tools/_paga.py
ivirshup: I'm not seeing how these are redefinitions, as comprehensions get their own scope. Am I missing something? If anything, wouldn't `s` be the redefinition?

> renaming to _

This would be wrong, since the value is used, right?

> renaming to k/j or something

If there really is a problem with `l`, then I'm fairly confident this variable refers to `leaf`s. My preference is still not to touch it.

scanpy/tools/_dpt.py
```diff
@@ -984,7 +984,7 @@ def __detect_branching_haghverdi16(
                 Dseg[tips[0]][idcs] + Dseg[tips[1]][idcs] + Dseg[tips[2]][idcs]
             )
         # init list to store new segments
-        ssegs = []
+        ssegs = []  # noqa: F841
```
ivirshup:
```suggestion
        ssegs = []  # noqa: F841 # TODO: Look into this
```

scanpy/tools/_dpt.py
```diff
@@ -743,11 +747,7 @@ def _detect_branching(
         tips: np.ndarray,
         seg_reference=None,
     ) -> Tuple[
-        List[np.ndarray],
-        List[np.ndarray],
-        List[List[int]],
-        List[List[int]],
-        int,
+        List[np.ndarray], List[np.ndarray], List[List[int]], List[List[int]], int
```
ivirshup:
```suggestion
        List[np.ndarray],
        List[np.ndarray],
        List[List[int]],
        List[List[int]],
        int,
```

I think this is fine. Seems intentional, due to the trailing comma.

ivirshup (changes requested):
Thanks for the changes, just a few minor things I've added suggestions for and the `paga` variables.

scanpy/tools/_paga.py
Zethson: I honestly don't see it either, but Pycharm and flake8 agree that these are redefinitions. Maybe there's a global imported `l` somewhere. I went with leafs now.

ivirshup (approved):
Looks good, thanks for all the work!

We should add a release note for this at some point, I'm just not sure where yet, probably a section for dev practices. Could you suggest a line for that?

I was unsure about the variable naming for PAGA, so I've decided to revert that. I couldn't get flake8 to call it a redefinition.

Zethson:
> Looks good, thanks for all the work!
> 
> We should add a release note for this at some point, I'm just not sure where yet, probably a section for dev practices. Could you suggest a line for that?
> 
> I was unsure about the variable naming for PAGA, so I've decided to revert that. I couldn't get flake8 to call it a redefinition.

:tada: 
Maybe "Enabled flake8 (https://flake8.pycqa.org/en/latest/) pre-commit to run code style checks"?
Everything else might just be details that people will uncover anyways since the workflows might complain :)

PR #1693, opened 2021-02-25
add codecov so we can have a badge to point to

ivirshup:
Adding upload to code cov so we can point to a coverage badge


ivirshup:
I'm pretty sure the bot won't enabled (changed that by force push). I'll disable if it is.

PR #1695, opened 2021-03-01
Use joblib for parallelism in regress_out

### LINKED ISSUE #1396: regress_out n_jobs question
...

Hi,

I notice that even if I set n_jobs=1 in sc.pp.regress_out, all cpus are utilized. This also happens if I set n_jobs to other numbers. Basically there isn't a noticeable difference in cpu usage no matter what number of n_jobs I set. I'm using CentOS6.8 on a machine with 28 physical cores and hyper threading on (appears as 56 cores in the os).

Is this an intended behavior, or just my installation? I understand that some numpy functions are naturally multi-threaded because of the setup of BLAS libraries.

Thanks.

ivirshup:
Fixes  #1396

Branch which I had sitting on my computer, but forgot to open a PR for (doh).

Uses joblib to parallelize `regress_out` instead of `multiprocessing` in order to prevent oversubscription of threads.

Not sure how to test this.

Should I also allow passing more arguments to joblib?

### DIFF
--- a/docs/release-notes/1.7.2.rst
+++ b/docs/release-notes/1.7.2.rst
@@ -6,7 +6,7 @@
 .. rubric:: Bug fixes
 
 - :func:`scanpy.logging.print_versions` now works when `python<3.8` :pr:`1691` :smaller:`I Virshup`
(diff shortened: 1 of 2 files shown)


ivirshup:

PR #1700, opened 2021-03-01
Temporary PR to check if CI still works with flit

flying-sheep:
To be deleted once we know

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -48,7 +48,7 @@ jobs:
     displayName: 'Install dependencies'
 
   - script: |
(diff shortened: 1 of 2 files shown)


flying-sheep:
All checks passed!

ivirshup:
Can we keep this open until the anndata pr has merged? For instance, I'd like to check this all works after merging #1702.

I'm a bit concerned about `scanpy.tests` using stuff from `anndata.tests` while those are being ignored from the `sdist`.

flying-sheep:
Done! For the record: To reopen a PR whose base branch was deleted (here: `flit`), you need to restore the base branch, if only to switch the base and delete it again.

PR #1702, opened 2021-03-01
Go back to regular pip

flying-sheep:
Hahaha and of course after weeks of this bug, everything gets resolved the day I press the merge button. takluyver/flit#395 should fix the issues where

1. `pip install scvelo` downgrades a `flit install -s` installed scanpy (now the `dist-info` dir name contains the correct version)
2. The wheel built by flit now corresponds to the freshly-changed spec’s [mangling rules](https://packaging.python.org/specifications/binary-distribution-format/#escaping-and-unicode)

So we can remove the workarounds. No problem, `git blame` won’t be affected much, this almost exclusively deletes lines.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -95,8 +95,11 @@ jobs:
 
   - script: |
       python -m pip install --upgrade pip
-      pip install flit twine wheel
-    displayName: 'Install build tools'
+      pip install flit twine wheel pytoml
+      python -c \
+        'import pytoml; c=pytoml.load(open("pyproject.toml")); print("\0".join(c["build-system"]["requires"]), end="")' \
+        | xargs -0 pip install -y
+    displayName: 'Install build tools and requirements'
 
   - script: pip list
     displayName: 'Display installed versions'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -42,7 +42,6 @@ jobs:
 
   - script: |
       python -m pip install --upgrade pip
-      pip install -r .pip-2033.txt
       pip install pytest-cov wheel
       pip install .[dev,doc,test,louvain,leiden,magic,scvi,harmony,scrublet,scanorama]
     displayName: 'Install dependencies'
@@ -103,7 +102,6 @@ jobs:
     displayName: 'Display installed versions'
 
   - script: |
-      pip install flit
       flit build
       twine check dist/*
     displayName: 'Build & Twine check'
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -95,16 +95,13 @@ jobs:
 
   - script: |
       python -m pip install --upgrade pip
-      pip install flit twine wheel pytoml
-      python -c \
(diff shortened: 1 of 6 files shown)


flying-sheep:
Just checked using this dockerfile, works flawlessly:

```dockerfile
FROM continuumio/miniconda

RUN conda install python=3.8
RUN pip install flit>=3.1
RUN git clone https://github.com/theislab/scanpy.git
WORKDIR /scanpy
# Go to the mainline-pip branch if it hasn’t been merged into master yet
RUN git checkout mainline-pip || true
RUN FLIT_ROOT_INSTALL=1 flit install -s --dep=develop  # Make development install of scanpy
# Make sure the dist-info folder has a plus in its name
RUN SCANPY_VERSION=$(python -c 'from importlib.metadata import version; print(version("scanpy"))') && \
    echo $SCANPY_VERSION | grep '+' &&
    test -d /opt/conda/lib/python3.8/site-packages/scanpy-$SCANPY_VERSION.dist-info
# Install project that depends on scanpy
RUN pip install scvelo
# Make sure it’s still a dev install
RUN test -L /opt/conda/lib/python3.8/site-packages/scanpy
```

docs/installation.rst
```diff
@@ -41,22 +41,13 @@ cloned version after you update with `git pull`) call::
...
-.. _bug in pip: https://github.com/pypa/pip/issues/9670
-
 If you want to let conda_ handle the installations of dependencies, do::
 
     pip install beni
     beni pyproject.toml > environment.yml
     conda env create -f environment.yml
     conda activate scanpy
-    flit install -s --deps=develop  # or: pip install -e ".[dev,doc,test]"
+    flit install -s --deps=develop
```
ivirshup, line 50: Can we note that `pip install -e` still works, but is deprecated?

ivirshup (commented):
More something that should have been caught in previous review, for the Check Build job:

* Flit should be installed with the rest of the dependencies.
* Can some of those tools be removed? For example: `setuptools`.

.azure-pipelines.yml
```diff
@@ -96,14 +95,16 @@ jobs:
 
   - script: |
       python -m pip install --upgrade pip
-      pip install setuptools setuptools_scm pytoml wheel twine
-    displayName: 'Install build dependencies'
+      pip install flit twine wheel pytoml
+      python -c \
+        'import pytoml; c=pytoml.load(open("pyproject.toml")); print("\0".join(c["build-system"]["requires"]), end="")' \
+        | xargs -0 pip install
```
ivirshup:
```suggestion
      pip install flit twine wheel pytoml setuptools_scm
```

I think this is going to be easier to figure out.

I take it `flit` isn't using the `build-system.requires` information by itself?

flying-sheep: It is for `flit install`. Remember that `flit build` isn’t usually needed, it’s effectively used as a debug tool to inspect packages in a dev environment.

People usually do `flit install -s` first, which install all the stuff.

Thinking of it, using `python -m build` is the better check here, as it uses an isolated build environment and therefore really checks if the build requirements are well-defined.

ivirshup (approved):

PR #1717, opened 2021-03-04
Fix clipped images in docs

### LINKED ISSUE #1705: Dotplot labels can extend off plot
@fidelram 

As noted in #1632, dotplot labels can often extend outside of the plotted area. Whether the full labels can be seen is dependent on how the plots are being output. It would be great if this always worked.

<details>
<summary> Example from the docs: </summary>

</details>

Could possibly be solved by using `matplotlib`'s `constrained_layout` or `tight_layout`. I think these would require modifying how the grid spec and axes are created.

ivirshup:
Fixes #1705

Found this over on `seaborn`: https://github.com/mwaskom/seaborn/blob/master/doc/matplotlibrc

### DIFF
--- a/docs/matplotlibrc
+++ b/docs/matplotlibrc
@@ -1,0 +1,1 @@
+savefig.bbox : tight


ivirshup:
Proof it worked:

image

flying-sheep:
Very nice!

Issue #1725, opened 2021-03-09
sc.pp.highly_variable_genes with flavor='seurat_v3' returns wrong means/variances when batch_key argument is used


jlause:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of scanpy.
- [x] (optional) I have confirmed this bug exists on the master branch of scanpy.

---

When working on PR #1715, I noticed a small bug when `sc.pp.highly_variable()` is run with `flavor='seurat_v3'` and the `batch_key` argument is used on a dataset with multiple batches:

The columns in the returned data frame `means` and `variances` do not give the correct gene means and gene variances across the whole dataset, but instead give the means and variances of the batch with the last index.

I think this because here means and variances are first computed within each batch:
https://github.com/theislab/scanpy/blob/a085333ead6ff8a0f64c25734060ffc048be56c0/scanpy/preprocessing/_highly_variable_genes.py#L72-L77
..and then not re-computed for the whole dataset before inserting them here:
https://github.com/theislab/scanpy/blob/a085333ead6ff8a0f64c25734060ffc048be56c0/scanpy/preprocessing/_highly_variable_genes.py#L136-L137

Should be an easy fix, I can also prepare it if you want :)

### Code to reproduce:

```python
import numpy as np
import scanpy as sc
import anndata 

import sys
sys.path.append("scanpy/preprocessing")
from _utils import _get_mean_var

np.random.seed(42)
adata = anndata.AnnData(np.random.randint(0,5,(100,100)))
adata.obs['batch'] = np.random.randint(0,5,(100))

#show that without batch_key, everything works fine
true_mean,true_var = _get_mean_var(adata.X)
result = sc.pp.highly_variable_genes(adata,n_top_genes=2,flavor='seurat_v3',inplace=False)
print('Results correct without batch_key?')
print(np.all(result['means']==true_mean))
print(np.all(result['variances']==true_var))

#show that it goes wrong with batch_key..
result_with_batchkey = sc.pp.highly_variable_genes(adata,n_top_genes=2,flavor='seurat_v3',inplace=False,batch_key='batch')
print('Results correct with batch_key?')
print(np.all(result_with_batchkey['means']==true_mean))
print(np.all(result_with_batchkey['variances']==true_var))

#..and that the wrong result comes from the last batch
adata_batch4 = adata[adata.obs['batch']==4,:].copy()
means_batch4,vars_batch4 = _get_mean_var(adata_batch4.X)
[...]

PR #1732, opened 2021-03-10
bugfix for #1725

jlause:
Fixed by computing global mean/variance outside the batch for loop :)

Verification code (same as in #1725):

```python
import numpy as np
import scanpy as sc
import anndata 

import sys
sys.path.append("scanpy/preprocessing")
from _utils import _get_mean_var

np.random.seed(42)
adata = anndata.AnnData(np.random.randint(0,5,(100,100)))
adata.obs['batch'] = np.random.randint(0,5,(100))

#show that without batch_key, everything works fine
true_mean,true_var = _get_mean_var(adata.X)
result = sc.pp.highly_variable_genes(adata,n_top_genes=2,flavor='seurat_v3',inplace=False)
print('Results correct without batch_key?')
print(np.all(result['means']==true_mean))
print(np.all(result['variances']==true_var))

#show that it goes wrong with batch_key..
result_with_batchkey = sc.pp.highly_variable_genes(adata,n_top_genes=2,flavor='seurat_v3',inplace=False,batch_key='batch')
print('Results correct with batch_key?')
print(np.all(result_with_batchkey['means']==true_mean))
print(np.all(result_with_batchkey['variances']==true_var))

#..and that the wrong result comes from the last batch
adata_batch4 = adata[adata.obs['batch']==4,:].copy()
means_batch4,vars_batch4 = _get_mean_var(adata_batch4.X)
print('Wrong results come from last batch?')
print(np.all(result_with_batchkey['means']==means_batch4))
print(np.all(result_with_batchkey['variances']==vars_batch4))
```
Verification code output:
```pytb
Results correct without batch_key?
True
True
Results correct with batch_key?
True
True
Wrong results come from last batch?
False
False

/gpfs01/berens/user/jlause/libs/scanpy/scanpy/preprocessing/_highly_variable_genes.py:148: FutureWarning: Slicing a positional slice with .loc is not supported, and will raise TypeError in a future version.  Use .loc with labels or .iloc with positions instead.
  df.loc[: int(n_top_genes), 'highly_variable'] = True
/gpfs01/berens/user/jlause/libs/scanpy/scanpy/preprocessing/_highly_variable_genes.py:148: FutureWarning: Slicing a positional slice with .loc is not supported, and will raise TypeError in a future version.  Use .loc with labels or .iloc with positions instead.
  df.loc[: int(n_top_genes), 'highly_variable'] = True

```


scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -62,6 +62,7 @@ def _highly_variable_genes_seurat_v3(
             "`flavor='seurat_v3'` expects raw count data, but non-integers were found.",
             UserWarning,
         )
+    mean_fulldataset, var_fulldataset = _get_mean_var(X)
```
ivirshup: Could remove the variables if they're not used anywhere else:

```python
df["mean"], df["var"] =  _get_mean_var(X)
```

ivirshup (changes requested):
Thanks for the fix!

Could this get a test, and a note in the 1.7.2 changelog?

scanpy/preprocessing/_highly_variable_genes.py
jlause: good point, thanks!

jlause:
I updated release notes and added a test for this specific case. I did not write many tests before, so I looked at the other tests and tried to stick to what I saw there.

I noted something unexpected when writing the test: When used `np.mean` and `np.var(.., ddof=1)` to compare against the test failed because some of the variances were off. The current version of the test uses `sc.pp._utils._get_mean_var()` (thats what `highly_variable_genes()` uses internally...), and does not fail..

Is it ok to use that instead? Is it expected that numpy and `_get_mean_var()` are slightly different here?

Test code with numpy ground truth:
```
def test_seurat_v3_mean_var_output_with_batchkey_vs_numpy():
    pbmc = sc.datasets.pbmc3k()
    pbmc.var_names_make_unique()
    n_cells = pbmc.shape[0]
    batch = np.zeros((n_cells), dtype=int)
    batch[1500:] = 1
    pbmc.obs["batch"] = batch

    true_mean = np.mean(pbmc.X.toarray(), axis=0)
    true_var = np.var(pbmc.X.toarray(), axis=0, ddof=1)

    result_df = sc.pp.highly_variable_genes(
        pbmc, batch_key='batch', flavor='seurat_v3', n_top_genes=4000, inplace=False
    )
    np.testing.assert_allclose(true_mean, result_df['means'], rtol=2e-05, atol=2e-05)
    np.testing.assert_allclose(true_var, result_df['variances'], rtol=2e-05, atol=2e-05)
```
Test output:
```
E       AssertionError: 
E       Not equal to tolerance rtol=2e-05, atol=2e-05
E       
E       Mismatched elements: 172 / 32738 (0.525%)
E       Max absolute difference: 0.01117667
E       Max relative difference: 0.00013328
E        x: array([0., 0., 0., ..., 0., 0., 0.], dtype=float32)
E        y: array([0., 0., 0., ..., 0., 0., 0.])

tests/test_highly_variable_genes.py:279: AssertionError
```

giovp:
> Is it ok to use that instead? Is it expected that numpy and _get_mean_var() are slightly different here?

very interesting, just checked `sc.pp._utils._get_mean_var()` and noticed that variance is not calcualted with `np.var`. Would that makes sense to change in `_utils` to use `np.var` ?

https://github.com/theislab/scanpy/blob/af8f323ebca87bc98d51e556742acf3c2cdf56e9/scanpy/preprocessing/_utils.py#L6

jlause:
> > Is it ok to use that instead? Is it expected that numpy and _get_mean_var() are slightly different here?
> 
> very interesting, just checked `sc.pp._utils._get_mean_var()` and noticed that variance is not calcualted with `np.var`. Would that makes sense to change in `_utils` to use `np.var` ?
> 
> https://github.com/theislab/scanpy/blob/af8f323ebca87bc98d51e556742acf3c2cdf56e9/scanpy/preprocessing/_utils.py#L6

I'm not experienced enough to say if what currently happens is maybe more efficient than `np.var()`?!

I think the deviating values I reported above come from what happens in the sparse case (`sparse_mean_variance_axis()`).. I can check next week if the same deviations occurs for the nonsparse case.

PS: As I said before, not even sure if these deviations are maybe even expected because of the different way of computing the variance here?!

ivirshup:
Calculating the variance on 32 bit floats, especially in the value ranges you have for counts, will result in a fair bit of inaccuracy. Because of this, we calculate variance with 64 bit values internally. To demonstrate:

```python
import scanpy as sc, numpy as np
from scanpy.pp._utils import _get_mean_var

pbmc = sc.datasets.pbmc3k()

var_np32 = np.var(pbmc.X.toarray(), axis=0, ddof=1)
var_np64 = np.var(pbmc.X.toarray(), axis=0, ddof=1, dtype=np.float64)
_, var_scanpy = _get_mean_var(pbmc.X)

# These are close
np.testing.assert_allclose(var_np64, var_scanpy)

# Same values are different
assert (np.isclose(var_np32, var_np64) == np.isclose(var_np32, var_scanpy)).all()
```

We don't use the numpy function for variance because then we'd need to create a dense intermediate array. We've previously used `sklearn.utils.sparsefuncs.mean_variance_axis` but that didn't let us control `ddof` or collect to 64 bit values from 32 bit input.

jlause:
Thanks for the demo code! now its clear to me.

I adapted the test to use `np.var(..,dtype=np.float64)` as ground truth, making the internal datatype conversion explicit.

Any other requests? I think everything else is ready :)

ivirshup (approved):
Looks good to me, thanks!

I'm just going to make a little clean up of the existing code, but this should be good to go.

PR #1734, opened 2021-03-10
add_dendrogram c/p doc fix

gokceneraslan:
@fidelram does this look ok?


ivirshup:
I think this makes sense

PR #1735, opened 2021-03-10
Preserve category order when groupby is a list

gokceneraslan:
Let's preserve categories if groupby parameter in _prepare_dataframe is a list:

```python
import scanpy as sc
import numpy as np

ad = sc.datasets.paul15()

sc.pp.log1p(ad)
sc.pp.pca(ad)
sc.pp.neighbors(ad)
sc.tl.leiden(ad)

cats = ad.obs.leiden.cat.categories.tolist()
np.random.seed(1)
np.random.shuffle(cats)
ad.obs.leiden.cat.reorder_categories(cats, inplace=True)
ad.obs.leiden.cat.categories
```

Prints:
```Index(['2', '3', '4', '9', '1', '6', '0', '7', '10', '8', '5'], dtype='object')```

```python
sc.pl.dotplot(ad, ['Cst3', 'Malat1'], groupby=['leiden'])
sc.pl.dotplot(ad, ['Cst3', 'Malat1'], groupby=['leiden', 'paul15_clusters'])
```

Outputs:

image

image

which is very annoying because we lost the category order of leiden. After this PR the output is:

image
image


gokceneraslan:
@fidelram Since we are not using tight layout when we save figures for the plotting tests, axis labels are cut off. I enabled it to avoid that with the following modification:

image

Here is the new plot testing this PR without the tight layout:

master_dotplot_groupby_list_catorder

which is pretty bad because what really matters in this plot for this PR is the labels of the x axis. Here is the same plot with the tight layout:

master_dotplot_groupby_list_catorder-tightlayout

However, many plotting tests fail now due to this change :/ Do you mind helping me with the failing tests?

scanpy/plotting/_anndata.py
```diff
@@ -1865,7 +1865,15 @@ def _prepare_dataframe(
...
-
+        # preserve category order
+        categories = (
+            obs_tidy[groupby]
+            .drop_duplicates()
+            .sort_values(groupby)
+            .agg('_'.join, axis=1)
+            .values
+        )
+        categorical.cat.reorder_categories(categories, inplace=True)
```
ivirshup: Might be nice to figure out categories in a way that scales with number of categories, not number of samples. Maybe something like:


```python
from itertools import product

order = {
    "_".join(k): idx for idx, k in 
    enumerate(product(*(obs_tidy[g].cat.categories for g in groupby)))
}
categorical = categorical.cat.reorder_categories(
    sorted(categorical.cat.categories, key=lambda x: order[x])
)
```

I do wish "categorical" weren't so many characters. Makes these lines so long!

ivirshup (commented):
## Question

Does this interact with group colors and dendrograms at all?

## Test change

All of the plots will start failing because this will change the output for every test. I have a few concerns here:

* I'm worried about repo bloat from the plotting tests. Ideally we could just store the reference images outside of git (git lfs maybe?). Updating all the plots with `tight_layout` would increase repo size by 10%
* Is `tight_layout` deterministic (https://github.com/matplotlib/matplotlib/issues/11809#issuecomment-432726600)? Also, is matplotlib trying to replace it with [`constrained_layout`](https://matplotlib.org/stable/tutorials/intermediate/constrainedlayout_guide.html)?
* Does globally adding `tight_layout` add to test times? My impression was that it basically rendered the plot, fixed the borders, then rendered it again.

Proposed solution:

Can we just explicitly extend the borders for this test? At a later point we can move plots to a different storage system, then have much more freedom in making changes to how they render.

gokceneraslan:
> ## Question
> 
> Does this interact with group colors and dendrograms at all?

Dendrogram and colors seem not affected.

> ## Test change
> 
> All of the plots will start failing because this will change the output for every test. I have a few concerns here:
> 
>     * I'm worried about repo bloat from the plotting tests. Ideally we could just store the reference images outside of git (git lfs maybe?). Updating all the plots with `tight_layout` would increase repo size by 10%
> 
>     * Is `tight_layout` deterministic ([matplotlib/matplotlib#11809 (comment)](https://github.com/matplotlib/matplotlib/issues/11809#issuecomment-432726600))? Also, is matplotlib trying to replace it with [`constrained_layout`](https://matplotlib.org/stable/tutorials/intermediate/constrainedlayout_guide.html)?
> 
>     * Does globally adding `tight_layout` add to test times? My impression was that it basically rendered the plot, fixed the borders, then rendered it again.
> 
> 
> Proposed solution:
> 
> Can we just explicitly extend the borders for this test? At a later point we can move plots to a different storage system, then have much more freedom in making changes to how they render.

Sounds great, I  think I managed to do that.

gokceneraslan:
@Meeseeksdev backport 1.7.x

PR #1740, opened 2021-03-11
Add morans I

### LINKED ISSUE #1698: Adding Moran's I calculation to Scanpy
Could you add Moran's I calculation to Scanpy? It could be used in scIB and to also find variable genes across embedding (could be an alternative to SEMITONES that takes a while to be computed).

giovp:
close #1698 

this is the implementation as proposed here: https://github.com/theislab/squidpy/pull/304
Explicit design choices compared to first proposal in squidpy:
- no joblib present, with low n permutation is fine and saw that you don't even do it in gearys C (which btw, makes a lot of sense in this setting, should consider to skip permutation entirely as well )
- only working on genes. technically it could work on continuos covariates as well, should I add that option?
- I think it could be worth it to add a row wise normalization of the weights (standard in pysal)

@ivirshup would be good to have feedback on those points, I will then add tests. Thank you!

### DIFF
--- a/docs/release-notes/1.8.0.rst
+++ b/docs/release-notes/1.8.0.rst
@@ -15,7 +15,7 @@
 
     - Added :func:`scanpy.metrics.confusion_matrix` for comparing labellings :pr:`915` :smaller:`I Virshup`
     - Added :func:`scanpy.metrics.gearys_c` for spatial autocorrelation :pr:`915` :smaller:`I Virshup`
-    - - Added :func:`scanpy.metrics.morans_i` for global spatial autocorrelation :pr:`1740` :smaller:`I Virshup, G Palla`
+    - Added :func:`scanpy.metrics.morans_i` for global spatial autocorrelation :pr:`1740` :smaller:`I Virshup, G Palla`
 
 .. rubric:: External tools
 
--- a/docs/release-notes/1.8.0.rst
+++ b/docs/release-notes/1.8.0.rst
@@ -15,6 +15,7 @@
 
     - Added :func:`scanpy.metrics.confusion_matrix` for comparing labellings :pr:`915` :smaller:`I Virshup`
     - Added :func:`scanpy.metrics.gearys_c` for spatial autocorrelation :pr:`915` :smaller:`I Virshup`
+    - - Added :func:`scanpy.metrics.morans_i` for global spatial autocorrelation :pr:`1740` :smaller:`I Virshup, G Palla`
 
 .. rubric:: External tools
 
@@ -25,3 +26,8 @@
 .. rubric:: Deprecations
 
 - Deprecated `layers` and `layers_norm` kwargs to :func:`~scanpy.pp.normalize_total` :pr:`1667` :smaller:`I Virshup`
+
+.. rubric:: Docs
+
+- Clarified docs issues for :func:`~scanpy.pp.neighbors`, 
+  :func:`~scanpy.tl.diffmap`, :func:`~scanpy.pp.calculate_qc_metrics` :pr:`1680` :smaller:`G Palla`
--- a/scanpy/metrics/_morans_i.py
+++ b/scanpy/metrics/_morans_i.py
@@ -31,27 +31,21 @@ def morans_i(
     use_raw: bool = False,
 ) -> Union[np.ndarray, float]:
     r"""
-    Calculate `Geary's C <https://en.wikipedia.org/wiki/Geary's_C>`_, as used
-    by `VISION <https://doi.org/10.1038/s41467-019-12235-0>`_.
+    Calculate Moran’s I Global Autocorrelation Statistic.
 
-    Geary's C is a measure of autocorrelation for some measure on a graph. This
-    can be to whether measures are correlated between neighboring cells. Lower
-    values indicate greater correlation.
+    Moran’s I is a global autocorrelation statistic for some measure on a graph. It is commonly used in
+    spatial data analysis to assess autocorrelation on a 2D grid. It is closely related to Geary's C,
+    but not identical. More info can be found `here`<https://en.wikipedia.org/wiki/Moran%27s_I> .
 
     .. math::
 
-        C =
-        \frac{
-            (N - 1)\sum_{i,j} w_{i,j} (x_i - x_j)^2
-        }{
-            2W \sum_i (x_i - \bar{x})^2
-        }
+        I=\frac{n}{S_{0}} \frac{\sum_{i=1}^{n} \sum_{j=1}^{n} w_{i, j} z_{i} z_{j}}{\sum_{i=1}^{n} z_{i}^{2}}
 
     Params
     ------
     adata
     vals
-        Values to calculate Geary's C for. If this is two dimensional, should
+        Values to calculate Moran's I for. If this is two dimensional, should
         be of shape `(n_features, n_cells)`. Otherwise should be of shape
         `(n_cells,)`. This matrix can be selected from elements of the anndata
         object by using key word arguments: `layer`, `obsm`, `obsp`, or
@@ -91,21 +85,21 @@ def morans_i(
     Examples
     --------
 
-    Calculate Gearys C for each components of a dimensionality reduction:
+    Calculate Morans I for each components of a dimensionality reduction:
 
     .. code:: python
 
         import scanpy as sc, numpy as np
 
         pbmc = sc.datasets.pbmc68k_processed()
-        pc_c = sc.metrics.gearys_c(pbmc, obsm="X_pca")
+        pc_c = sc.metrics.morans_i(pbmc, obsm="X_pca")
 
 
     It's equivalent to call the function directly on the underlying arrays:
 
     .. code:: python
 
-        alt = sc.metrics.gearys_c(pbmc.obsp["connectivities"], pbmc.obsm["X_pca"].T)
+        alt = sc.metrics.morans_i(pbmc.obsp["connectivities"], pbmc.obsm["X_pca"].T)
         np.testing.assert_array_equal(pc_c, alt)
     """
     if use_graph is None:
--- a/scanpy/metrics/_morans_i.py
+++ b/scanpy/metrics/_morans_i.py
@@ -11,6 +11,7 @@ import numba.types as nt
 from numba import njit, prange
 
 from scanpy.get import _get_obs_rep
+from scanpy.metrics._gearys_c import _resolve_vals
 
 it = nt.int64
 ft = nt.float32
@@ -211,26 +212,6 @@ def _morans_i_mtx_csr(
 ###############################################################################
 # Interface (taken from gearys C)
 ###############################################################################
-@singledispatch
-def _resolve_vals(val):
-    return np.asarray(val)
-
-
-@_resolve_vals.register(np.ndarray)
-@_resolve_vals.register(sparse.csr_matrix)
-def _(val):
-    return val
-
-
-@_resolve_vals.register(sparse.spmatrix)
-def _(val):
-    return sparse.csr_matrix(val)
-
-
-@_resolve_vals.register(pd.DataFrame)
-@_resolve_vals.register(pd.Series)
-def _(val):
-    return val.to_numpy()
 
 
 @morans_i.register(sparse.csr_matrix)
--- a/scanpy/metrics/_morans_i.py
+++ b/scanpy/metrics/_morans_i.py
@@ -15,11 +15,11 @@ ip = np.int64
 fp = np.float64
 
 
-def moran(
+def morans_i(
     adata: AnnData,
-    connectivity_key: str = Key.obsp.spatial_conn(),
+    connectivity_key: str = "connectivities",
     genes: Optional[Union[str, Sequence[str]]] = None,
-    n_perms: int = 100,
+    n_perms: int = 1000,
     corr_method: Optional[str] = "fdr_bh",
     layer: Optional[str] = None,
     copy: bool = False,
@@ -79,7 +79,7 @@ def moran(
         mi = _moran_score_perms(counts, indptr, indices, data, n_perms)
         moran_list.append(mi)
 
-    df = pd.DataFrame(moran_list, columns=["I", "pval_sim", "VI_sim"], index=gen)
+    df = pd.DataFrame(moran_list, columns=["I", "pval_sim", "VI_sim"], index=genes)
 
     if corr_method is not None:
         _, pvals_adj, _, _ = multipletests(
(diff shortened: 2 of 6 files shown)


ivirshup:
> only working on genes. technically it could work on continuos covariates as well, should I add that option?

Sure. I think it would make sense to mimic the API of `gearys_c` as much as possible here.

> I think it could be worth it to add a row wise normalization of the weights (standard in pysal)

What would this entail? I wonder if this is best left up to the user, who can just chose what values to pass in?

> should consider to skip permutation entirely as well

Yeah, I'm not sure if we need this at the moment. I wonder if calculating p-values should even be a separate method? I would like to understand more about how p-values are calculated, and what you're getting out of this. 

For instance, I would assume it's not appropriate to calculate a p-value for gene expression using a nearest neighbor network based on gene expression.

--------------------

Side note, on performance. So right now it looks like computing one permutation is fairly fast (which makes sense). Most of the time comes from the permutation testing. After that, most of the time looks like it's coming from `adata.obs_vector`, which is a bit slow especially if you're getting many genes with it.

giovp:
damn spent way too much time debugging numba, only to realize that set `parallel=True` in the dot product was clashing with `parallel=True` for the collection. Now it's fast, 10s for 18k genes and 2k cells.
I completely copied over the design from gearys c with the `singledispatch` (btw cool usage for handling different value types!) and simplified a bit the numba part.

Tomorrow I'll add tests and benchmark against pysal and then we are ready to go.

giovp:
reproducibility with pysal: ✅ 

```python
import scanpy as sc
import squidpy as sq

adata = sq.datasets.visium_hne_adata()
adj = adata.obsp["connectivities"]
dat = adata.X.T
moran_scanpy = sc.metrics.morans_i(adj, dat)

import libpysal
import esda

adj = adj.tolil()
neighbors = dict(enumerate(adj.rows))
weights = dict(enumerate(adj.data))
W = libpysal.weights.W(neighbors, weights, ids=adata.obs.index.values)

def _compute_moran(y: np.ndarray, w: W, transformation: str, permutations: int):
    mi = esda.moran.Moran(
        y, w, transformation=transformation, permutations=permutations
    )
    return mi.I


moran_list = []
transformation = "O" # original weights
layer = None
permutations = None
for g in adata.var_names.values:
    mi = _compute_moran(
        adata.obs_vector(g, layer=layer), W, transformation, permutations
    )
    moran_list.append(mi)

moran_esda = np.array(moran_list)
from scipy.stats import pearsonr
import seaborn as sns

g = sns.scatterplot(x=moran_scanpy, y=moran_esda)
g.set_xlabel("scanpy")
g.set_ylabel("esda")
p = pearsonr(moran_scanpy, moran_esda)[0]
g.set_title(f"pearson_{round(p, 5)}")
```

image

have a couple of issues with tests but should fix them by today

giovp:
ok, finished also with tests (I took what you had already for gearys C that tested for different types and consistency). Had to change to float32 cause I was having reproducibility errors (possibly due to overflow). ready to review, thank you!
btw I took a fair bit of code from gearysc re design and tests, so if you think should add better acknowledgment or co-author this PR, please go ahead.

scanpy/metrics/_morans_i.py
```diff
@@ -0,0 +1,263 @@
...
+
+@_resolve_vals.register(sparse.spmatrix)
+def _(val):
+    return sparse.csr_matrix(val)
+
+
+@_resolve_vals.register(pd.DataFrame)
+@_resolve_vals.register(pd.Series)
+def _(val):
+    return val.to_numpy()
```
ivirshup: Could this share the function with `_gearys_c`?

scanpy/metrics/_morans_i.py
```diff
@@ -0,0 +1,263 @@
...
+            raise ValueError("Must run neighbors first.")
+    else:
+        raise NotImplementedError()
+    if vals is None:
+        vals = _get_obs_rep(adata, use_raw=use_raw, layer=layer, obsm=obsm, obsp=obsp).T
+    return morans_i(g, vals)
+
+
+@njit(
+    ft(ft[:], it[:], it[:], ft[:], ft[:], ft, ft),
```
ivirshup: Why use static typing here? Why not just let numba generate specializations?

scanpy/metrics/_morans_i.py
```diff
@@ -0,0 +1,263 @@
...
+    cache=True,
+    parallel=False,
+    fastmath=True,
```
ivirshup: Does this change anything?

I'm not sure if it's good to have on here. I'd be more amenable if this were an approximate method (like nndecent), but feel like statistic computation should probably be precise.

ivirshup (commented):
> btw I took a fair bit of code from gearysc re design and tests, so if you think should add better acknowledgment or co-author this PR, please go ahead.

Ah, no worries! If anything this makes the job of reviewing much easier!

It would be nice if you could share more code, but it's fine for now.

--------------

Any thoughts on the p-value computation? It's got me thinking a bit more about it, though it's probably only valid with spatial data.

scanpy/metrics/_morans_i.py
giovp: mmh only preference is for:
- clarity
- make sure there is nothing weird happening down the line, as types are alwasy consistent (and conversion always happes).

If you have strong dislike I can of course remove

scanpy/metrics/_morans_i.py
giovp: actually,
```
# w/ fastmath
10.6 s ± 146 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# w/o fastmath
12 s ± 1.17 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

again, not much, if you want I can remove.

scanpy/metrics/_morans_i.py
giovp: yes, importing it from `_gearys_c`! very bad copying over from my side

scanpy/metrics/_morans_i.py
ivirshup: So, most of the time the indices are going to be 32 bit (scipy tries to use a small width integer for these), and I do think it's important for the results here to be calculated with 64 bit accuracy since there can be a lot of operations per result value. However, some windows users have 32 bit systems, so locking to 64 bit can be problematic (see numba related issues around `calculate_qc_metrics`).

I'd say this should cast floating point data to system width, and be agnostic about the types of indices. That's what I did for `gearys_c`, where it looks like (from the comments) I came across precision problems with 32 bit values.

scanpy/metrics/_morans_i.py
ivirshup: I'm more concerned about if the results changed. But I'm also not super compelled by the benchmark because the difference between the means is ~ the sum of the std deviations.

I don't know a ton about how this can change results, so I'd like to err on the side of caution and remove it.

giovp: done

scanpy/metrics/_morans_i.py
giovp: ok but I need to stick to np.float32 for values, becuase I hit this error every other test run
```
>       np.testing.assert_array_equal(
            sc.metrics.morans_i(pbmc, obsm="X_pca"),
            sc.metrics.morans_i(g, pbmc.obsm["X_pca"].T),
        )
E       AssertionError: 
E       Arrays are not equal
E       
E       Mismatched elements: 13 / 50 (26%)
E       Max absolute difference: 2.68156172e+154
E       Max relative difference: 1.
E        x: array([ 0.000000e+00, -1.363116e+57, -1.363116e+57, -1.363116e+57,
E              -1.363116e+57, -1.363116e+57, -1.363116e+57,  0.000000e+00,
E              -1.363116e+57, -1.363116e+57, -1.363116e+57, -1.363116e+57,...
E        y: array([ 0.000000e+000, -1.363116e+057, -1.363116e+057, -1.363116e+057,
E              -1.363116e+057, -1.363116e+057, -1.363116e+057,  0.000000e+000,
E              -1.363116e+057, -1.363116e+057, -1.363116e+057, -1.363116e+057,...
```

ivirshup: This seems pretty strange, and I think is worth investigating more. Is it literally alternating test runs?

giovp: yes, am testing now with types but I think only solution is to set float 32

giovp: ok, only solution is to use np.float32 for the obsm_pca test. The other don't fail but I think it makes sense to keep it consistent. I am not really sure why that is, but yes it was failing 4/5 times the test run. tbh I see the point for using int64, but do we really need float 64? probably not i'd say no?

ivirshup: So, the results you're showing in the above test are really wrong, so I think there's a larger issue at play here. I'll take a look into this.

But, I would say impression in these sorts of calculations can cause some pretty bad results. For example, once we get over a couple hundred thousand cells, variance calculation on float32 values can start accumulating significant amounts of imprecision, which can lead to very wrong differential expression results.

ivirshup: So the values that you were getting out reminded me a lot of what was happening when I came across some numba parallelization issues: https://github.com/numba/numba/issues/6774#issuecomment-788789663 (the equivalent test was failing for `gearys_c`).

Looking into it, I think the same problem is occurring, it's just that your getting a consistent wrong answer with `float32`, but a inconsistent wrong answer with `float64`.

For example, on commit 33db5ced726996e292e6:

```python
import scanpy as sc
sc.metrics.morans_i(pbmc.obsp["connectivities"], pbmc.obsm["X_pca"].T)
```

```
array([        0., -26711958., -26711958., -26711958.,         0.,
       -26711958., -26711958., -26711958.,         0., -26711958.,
       -26711958.,         0., -26711958., -26711958.,         0.,
       -26711958., -26711958.,         0., -26711958., -26711958.,
               0., -26711958., -26711958.,         0., -26711958.,
       -26711958.,         0., -26711958., -26711958.,         0.,
       -26711958., -26711958.,         0., -26711958., -26711958.,
               0., -26711958., -26711958.,         0., -26711958.,
       -26711958.,         0., -26711958., -26711958.,         0.,
       -26711958., -26711958.,         0., -26711958., -26711958.],
      dtype=float32)
```

This bug seems to be based on having nested parallelism and certain reductions. We can avoid it by just not having nested parallelism, which is what I've done for `gearys_c`.

scanpy/metrics/_morans_i.py
```diff
@@ -0,0 +1,233 @@
...
+            g.indices,
+            g.indptr,
+            g_data,
+            vals.shape,
+        )
+    elif isinstance(vals, np.ndarray) and vals.ndim == 1:
+        assert g.shape[0] == vals.shape[0]
+        return _morans_i_vec(vals, g.indptr, g.indices, g_data)
+    elif isinstance(vals.astype(fp, copy=False), np.ndarray) and vals.ndim == 2:
+        print(f"here{vals.shape}")
```
ivirshup:
```suggestion
```

ivirshup:
I've made a few changes:

## Numba bug

First, the reason that you were getting different issues with floats is that there's a numba  bug where the generated parallelized code is completely wrong. I ran into this with gearys_c too, so I've just done a similar thing. It seems to be triggered by having a `np.sum` in a `prange` loop, plus some minor other things. You can check the linked comment info for more details.

I believe calculation of `z2ss` in the outer loop was triggering this bug, so I've just moved this into the inner function. Since we're not doing iterations anymore this shouldn't be a performance issue.

## Argument order

So, one bigger organizational change I made is to have consistent argument orders for the elements of a sparse matrix. Basically, always use the same order for positional arguments between functions, otherwise it's very easy to introduce bugs.

## Minor things

* I've modified one of the `morans_i` tests to check that if you pass a dense matrix and a sparse matrix of the same data, you should get the same results.
* I've removed the use of an intermediate array in `_morans_i_vec_W`, since you can just accumulated directly to `inum`.
* Fixed up typing, removed unused exports

giovp:
> a np.sum in a prange loop, plus some minor other things. You can check the linked comment info for more details.

🤦  remember having a look but not quite getting it the first time... thank you!

> Argument order

👍 

> Minor things

👍 🙏 

I'll had release note, than we can merge it?

scanpy/metrics/_morans_i.py
```diff
@@ -0,0 +1,247 @@
...
+    r"""
+    Calculate Moran’s I Global Autocorrelation Statistic.
+
+    Moran’s I is a global autocorrelation statistic for some measure on a graph. It is commonly used in
+    spatial data analysis to assess autocorrelation on a 2D grid. It is closely related to Geary's C,
+    but not identical. More info can be found `here`<https://en.wikipedia.org/wiki/Moran%27s_I> .
+
+    .. math::
+
+        I=\frac{n}{S_{0}} \frac{\sum_{i=1}^{n} \sum_{j=1}^{n} w_{i, j} z_{i} z_{j}}{\sum_{i=1}^{n} z_{i}^{2}}
```
ivirshup:
```suggestion
    .. math::

        I =
            \frac{
                N \sum_{i, j} w_{i, j} z_{i} z_{j}
            }{
                S_{0} \sum_{i} z_{i}^{2}
            } 
```

Suggestion for more consistency with gearys c.

Could also be nice to define the terms more clearly, but that would the case for gearys c as well. Not needed, just a thought.

ivirshup (commented):
Almost, I think you've got to include a reference to the function here: https://github.com/theislab/scanpy/blob/master/docs/api/index.rst#metrics

and I'm not sure the link to wikipedia is formatted right.

docs/release-notes/1.8.0.rst
```diff
@@ -15,6 +15,7 @@
 
     - Added :func:`scanpy.metrics.confusion_matrix` for comparing labellings :pr:`915` :smaller:`I Virshup`
     - Added :func:`scanpy.metrics.gearys_c` for spatial autocorrelation :pr:`915` :smaller:`I Virshup`
+    - - Added :func:`scanpy.metrics.morans_i` for global spatial autocorrelation :pr:`1740` :smaller:`I Virshup, G Palla`
```
ivirshup:
```suggestion
    - Added :func:`scanpy.metrics.gearys_c` for spatial autocorrelation :pr:`915` :smaller:`I Virshup`
    - Added :func:`scanpy.metrics.morans_i` for global spatial autocorrelation :pr:`1740` :smaller:`I Virshup, G Palla`
```

ivirshup (approved):
Great, thanks!

PR #1750, opened 2021-03-18
Check and warn if rank_genes_groups gets raw counts

Koncopd:


ivirshup (commented):
Couple comments, also could this get a test?

Issue #1664, opened 2021-02-19
Inline example plots in docs


ivirshup:
Meta-issue for tracking the addition of rendered plots in the examples sections of the API docs (initial functionality implemented in #1632).

We now allow code samples which generate plots in our docs. See the examples sections of [`calculate_qc_metrics`](https://scanpy.readthedocs.io/en/latest/api/scanpy.pp.calculate_qc_metrics.html#scanpy.pp.calculate_qc_metrics) and [`dotplot`](https://scanpy.readthedocs.io/en/latest/api/scanpy.pl.dotplot.html#scanpy.pl.dotplot) for examples of what this looks like. There is a short section in the dev docs with more information on how to make these: [Plots in docstring](https://scanpy.readthedocs.io/en/latest/dev/documentation.html#plots-in-docstrings)

Here is a checklist of the functions which should have rendered examples. This list is not exclusive and can definitely be expanded.

### Functions in `sc.pl`

- [ ] Embedding plots (these may deserve their own prose section)
    - [ ] `sc.pl.embedding`
    - [x] `sc.pl.draw_graph` #1809 @AnnaChristina 
    - [x] `sc.pl.diffmap` #1809 @AnnaChristina
    - [x] `sc.pl.pca` #1813 @lazappi 
    - [x] `sc.pl.tsne` #1809 @AnnaChristina
    - [x] `sc.pl.umap` #1830 @mayarali
    - [x] `sc.pl.embedding_density` @ivirshup 
- [ ] PCA specific
    - [x] `sc.pl.pca_loadings` #1815 @bio-la 
    - [x] `sc.pl.pca_overview` #1812 @MxMstrmn 
    - [ ] `sc.pl.pca_scatter`
    - [ ] `sc.pl.pca_variance_ratio`
- [ ] PAGA
    - [x] `sc.pl.paga` #1811 @le-ander 
    - [ ] `sc.pl.paga_adjacency`
    - [ ] `sc.pl.paga_compare`
    - [ ] `sc.pl.paga_path`
- [ ] DPT pseudotime
    - [ ] `sc.pl.dpt_groups_pseudotime`
    - [ ] `sc.pl.dpt_timeseries`
- [x] Groupby
    - [x] `sc.pl.dotplot` @ivirshup 
    - [x] `sc.pl.matrixplot` #1808 @mbuttner 
    - [x] `sc.pl.clustermap` #2509 @giuliafrrn
[...]

PR #1753, opened 2021-03-21
Reorganize reference docs

### LINKED ISSUE #1682: Plotting functions should show up in table of contents
It would be useful for plotting functions to show up in the table of contents for the API section of the docs, like they do for all other functions

image

image

ivirshup:
This PR contains a reorganization of the API reference docs, mostly to ease modification and generation.

Fixes #1682

## Case sensitivity

First big issue addressed, `scanpy.plotting.dotplot.rst` and `scanpy.plotting.DotPlot.rst` are the same path on a case insensitive file system. This is what most people on macOS will have. If you have a case sensitive file system, important apps like [Adobe's](https://helpx.adobe.com/creative-suite/kb/error-case-sensitive-drives-supported.html) and [steam](https://support.steampowered.com/kb_article.php?ref=8601-RYPX-5789) won't work. So now the plotting classes are put in `generated/classes`.

For me, this means:

* docs now generate without warnings
* DotPlot and MatrixPlot docs are only rebuilt if I update them

## Separate tracked and generated content

I find it difficult to navigate the rst api docs when there is a large amount of auto generated files mixed in with manually curated ones. This becomes more of a problem since the autogenerated files are not removed by `make clean` and will still generate `html` files and generally be a nuisance while they are still around.

Now all autogenerated API files are put in a `generated` directory. This is also what [xarray](https://github.com/pydata/xarray/blob/e6168ed6b771b7377e89acb47cc7e5aa9f21a574/doc/api.rst), [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.utils.extmath.randomized_range_finder.html#sklearn.utils.extmath.randomized_range_finder), and [seaborn](https://github.com/mwaskom/seaborn/blob/ba4bd0fa0a90b2bd00cb62c2b4a5e38013a73ac6/doc/api.rst) do.

Now `make clean` also deletes all auto generated `rst` files. This also dramatically simplifies our `.gitignore`

## Consolidation

I've also consolidated the managed api docs to a smaller number of files. Namely, instead of `api/*.rst` there's now just `api.rst`. Instead of `external/*.rst` there's just `external.rst`.

## Plotting functions API docs

In the current api docs, some plotting functions have a plot displayed inline instead of a one line description. This is nice because it provides a visual reference for what the plotting function does. It's less nice because:

* It takes up a huge amount of space
* It takes up a different amount of space for each function
* These plots are manually generated, and are often quite out of date

<details>
<summary> example </summary>

image

</details>

I think the reference docs should be handled in the same way for each of the modules. Example plots make more sense in the user guide/ tutorials and within each functions documentation (#1664).

I'd like to go further with this by moving the plotting tutorial's content to the user guide, so we can have more in depth linking to different sections. Additional benefits here include plots being generated with each doc build and having links from functions in the guide to their api docs.

## TODO

- [x] [Redirects](https://docs.readthedocs.io/en/stable/user-defined-redirects.html) (redirects need to be added on readthedocs so old links won't fail) **added**
- [x] Decide where module docs live, as doc-string, or as rst. **rst**


ivirshup:
@flying-sheep, any suggestions on how to automatically clear out all the rst files from `api` and `extension`? I suppose we could add a temporary `git clean`, but that seems a bit harsh.

flying-sheep:
`git clean` can be used with a prefix path no? I’d be OK with `git clean ./docs`

This will break all our links right? Anyone who has ever linked any autogenerated doc page will now have a broken link. Not a fan of that part …

.gitignore
```diff
@@ -5,12 +5,7 @@
 
 # Docs
 /docs/_build/
-/docs/api/scanpy.*.rst
-/docs/external/scanpy.*.rst
-!/docs/api/scanpy.api.rst
-!/docs/api/scanpy.api.AnnData.rst
-!/docs/api/scanpy.plotting.rst
-!/docs/api/scanpy.external.rst
+/docs/generated/
```
flying-sheep, line 8: I don’t think an implementation detail should appear in URLs.

docs/conf.py
```diff
@@ -51,7 +51,13 @@
...
     'scanpydoc',
     "sphinx_search.extension",
-    *[p.stem for p in (HERE / 'extensions').glob('*.py')],
+    # Locally defined
+    "debug_docstrings",
+    "github_links",
+    "param_police",
+    "typed_returns",
+    # "function_images",
+    # *[p.stem for p in (HERE / 'extensions').glob('*.py')],
```
flying-sheep: Yes, this should be generated with some plotting gallery plugin instead.

Let’s move this part to the PR that adds this plugin.

flying-sheep (changes requested):
Regarding `generated`, there’s both no need and actual downsides:

1. `git clean` can be used with a prefix path: `git clean -fx -- ./docs`
2. Changing anything wide-reaching this would break many of our incoming links: Anyone who has ever linked any autogenerated doc page would now have a broken link. Not a fan of that part …
3. For case insensitivity let’s use the feature that specifically exists for that problem instead: [`autosummary_filename_map`](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#confval-autosummary_filename_map)

   > A dict mapping object names to filenames. This is necessary to avoid filename conflicts where multiple objects have names that are indistinguishable when case is ignored, on file systems where filenames are case-insensitive.

This way you can still build the docs without problems, with the only downside being that links to two objects break.

ivirshup:
> Regarding generated, there’s both no need

I disagree with this. I think we definitely should not be intermingling source and generated files, especially when it's one source file to many generated. This is not a pleasant way to navigate files:

<details>
<summary> </summary>

image

</details>

Also, looking around at other big projects with sphinx documentation, this is the way they all handled autosummary stub files.

> `git clean -fx -- ./docs`

This is a bit of a nuke. I've added removing the generated docs to `make clean`, but we wouldn't want to run `git clean -fx -- ./docs` for `make clean`.

> 2. Changing anything wide-reaching this would break many of our incoming links

This would be solved with [redirects](https://docs.readthedocs.io/en/stable/user-defined-redirects.html).

I would be happy to have `docs.scanpy.org/en/latest/api/dotplot.html` be the canonical url, but not if it requires mixing generated and source files.

> 3. For case insensitivity let’s use the feature that specifically exists for that problem instead: autosummary_filename_map

Doesn't this only half solve the issue? Since (AFAIK) the generated html page has to have the same name as the rst file, this would break links (as you mentioned above). While we could do a redirect from here, what do we redirect too? Do we have special names for urls for classes which have an associated function? I don't think that would be cleaner than just having all classes be put in a `classes` subdirectory.

flying-sheep:
> This is not a pleasant way to navigate files:

<details>
<summary>You can fix that with the .ignore plugin!</summary>

Right click the Project view and select “Hide Ignored Files”

grafik

</details>

> I would be happy to have docs.scanpy.org/en/latest/api/dotplot.html be the canonical url, but not if it requires mixing generated and source files.

That would be perfect if it’s possible!

> Since (AFAIK) the generated html page has to have the same name as the rst file, this would break links (as you mentioned above). While we could do a redirect from here, what do we redirect too?

We create manual redirects for the 2 APIs where we failed to add an underscore to the function name, and don’t do that again.

The non-domain parts of URLs *are* case sensitive, therefore having a redirect `.../dotplot.html` → `.../dot_plot.html` works perfectly fine in the presence of `.../DotPlot.html`

docs/conf.py
ivirshup: As mentioned in the main PR comment, I don't think there should be example plots in the table of contents for the reference docs. Examples fit in tutorials, the docs for the function, and examples, but not the table of contents.

ivirshup:
I'm going to merge this since it makes my life much easier.

We can always add more redirects if we want to put the reference docs somewhere else in the future.

PR #1754, opened 2021-03-21
DocSearch

flying-sheep:
[rendered](https://scanpy.readthedocs.io/en/docsearch/)

/edit: huh! weird, seems that [readthedocs-sphinx-search](https://readthedocs-sphinx-search.readthedocs.io/en/latest/) pops up instead. why do we try to use docsearch instead of that one?


ivirshup:
I added the readthedocssphinx search when experimenting with which one to use. DocSearch just seemed a lot nicer. It's got fuzzy matching, so using UK vs US spelling doesn't matter, also gives more context about results.

pyproject.toml
```diff
@@ -90,7 +90,7 @@ doc = [
     'sphinx-rtd-theme>=0.3.1',
     'sphinx-autodoc-typehints',
     'readthedocs-sphinx-search',
```
ivirshup:
```suggestion
```

docs/conf.py
```diff
@@ -94,7 +93,12 @@
 
 
 html_theme = 'scanpydoc'
-html_theme_options = dict(navigation_depth=4, logo_only=True)  # Only show the logo
+html_theme_options = dict(
+    navigation_depth=4,
+    logo_only=True,
+    docsearch_index='scanpy',
+    docsearch_key='fa4304eb95d2134997e3729553a674b2',
```
ivirshup, line 100: I think version needs to be passed, but when I try it locally I still seem to get duplicate results.

flying-sheep: It doesn’t, as documented here: https://icb-scanpydoc.readthedocs-hosted.com/en/latest/scanpydoc.theme.html#confval-docsearch_doc_version

flying-sheep: Must be something else. What’s “duplicated”? From different versions?

ivirshup: Yeah, I was getting results from both latest and stable.

ivirshup (approved):
Looks good now, must have been a cache issue. I'm good with merging, unless I'm forgetting something?

Release note would be good.

flying-sheep:
Looking good! But somehow there’s like 3 times the same for each result … I assume two are expected as one is the page title and one the doc item title (e.g. function name), and we have one page per function/class.

But shouldn’t one side of the vertical line be e.g. the first line of the docstring? 

popup

PR #1765, opened 2021-03-26
highly deviant genes implementation

ktpolanski:
An implementation of highly deviant gene identification from the 2019 GLMPCA paper. I'm rather fond of the method, as it's a straightforward statistical measure, and comes with significance testing as a form of data-driven cutoff.

I put it in a new `highly_deviant_genes()` function, as:
- it comes with a number of unique parameters, and there's only so many different algorithms `highly_variable_genes()` can house
- the paper argues that highly deviant is different from highly variable

I acknowledge that there are no tests, I'm hoping to get some assistance with that if possible.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -36,6 +36,7 @@ For visual quality control, see :func:`~scanpy.pl.highest_expr_genes` and
    pp.filter_cells
    pp.filter_genes
    pp.highly_variable_genes
+   pp.highly_deviant_genes
    pp.log1p
    pp.pca
    pp.normalize_total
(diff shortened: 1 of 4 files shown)


ivirshup:
I like that this method is fairly simple, and could have a meaningful cutoff, but I think I'd like more evidence of it's usefulness before thinking about including it.

I have two main points of concern:

* Are there examples of this method being used outside of the glmPCA paper? I would at least like to know that reasonable results can be found downstream of this.
* In the glmPCA paper, the identified genes are highly correlated (~1) with highly expressed genes, and lowly correlated (~.3 with highly variable gene selection. While I'm not sure which highly variable gene method they compared against, should the low correlation with common practice give us pause?

image

@giovp

Issue #1770, opened 2021-03-30
sc.pl.dotplot() does not use `smallest_dot` correctly


sjfleming:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of scanpy.
- [x] (optional) I have confirmed this bug exists on the master branch of scanpy.

---

I recently realized that the `smallest_dot` argument is no longer working correctly in `sc.pl.dotplot()`, and I think I've identified a small typo in the codebase.

https://github.com/theislab/scanpy/blob/3fb1463a64e91a87854d19e06bcf4efedd0fad66/scanpy/plotting/_dotplot.py#L340
along with the `if` statement on the line above it.

I believe these lines should read
```python
        if smallest_dot != self.smallest_dot:
            self.smallest_dot = smallest_dot
```

Using different values for `smallest_dot` produces the same output, and I think this is why.

I will write a quick PR for this.

PR #1771, opened 2021-03-30
Bug fix for `smallest_dot` in sc.pl.dotplot()

sjfleming:
Solves #1770


ivirshup (changes requested):
Thanks for the PR! Yeah, this looks like a typo, and your fix looks good.

Could you update the reference plot for the tests? This will just entail replacing the reference figure with the new figure generated by the failing test? This may need to happen for all the tests which use `smallest_dot`.

sjfleming:
Hi @ivirshup , I replaced the one test image that was causing a failure, as you suggested.  (And I checked to make sure the image makes sense... it does...)  I think this should do it.

ivirshup:
Looks good, thanks!

PR #1772, opened 2021-03-31
Reorganize plotting test outputs

ivirshup:
We frequently use reference images for plotting tests, that is, we test against a saved copy of a plot from an older version to check that the results haven't changed much.

Currently we store the reference image in git inside a directory like `scanpy/tests/_images`. Tests generate a new file with the same name as the reference image inside `scanpy/tests/figures`. These files are compared. If they're not similar enough, a file with a similar name, appended with `failed-diff` is written to the figures folder.

There are a few annoyances with this setup.

### Opening the relevant files to investigate a failed test

Its not hard to open up the generated file, reference, and diff to see what changed, but it's harder than it should be. You'd have to write a function to make this at all easy.

### It's not obvious from the paths which file is the reference and which file was generated

 E.g. given the paths:

```
scanpy/tests/_images/pca.png
scanpy/tests/figures/pca.png
```

which one is the reference?

### Shared file names cause ambiguity

image

~~This is more of a hunch.~~ There should be three images here, but only two show up. I suspect this has to do with the missing image having a duplicated name.

Update: Tested this in #1773, all images show up now (@gokceneraslan):

image

## This PR

This PR addresses the three points from above. Now each reference image gets its own directory. The reference image is named `expected.png`, and is stored in git. When a test runs the generated image is stored as `actual.png`, and compared with `expected.png`. A failing diff is also written to this directory if the comparison fails.

* Opening all images is now as simple as `open scanpy/tests/_images/{fig_name}/*`
* Naming conventions are obvious
* Per test, each file has a unique name

## TODO

- [x] Dev docs
- [ ] Decide on fixture api
- [ ] Decide how generated outputs are handled (`.gitignore`-d, deleted each time that test is run, both?)


flying-sheep:
This is already useful without the final fixture API design, so let’s just merge it.

PR #1775, opened 2021-03-31
Add scNym as an external tool `sce.tl.scnym`

jacobkimmel:
Hi scanpy folks,

This PR adds our cell identity classification model [`scnym`](https://github.com/calico/scnym) as a tool in `sc.external.tl.scnym`.

The `scnym` API was inspired by `scanpy` and intended to be compatible, so the implementation in `external/_scnym.py` is simply a wrapper.
I also added a test in `tests/external/test_scnym.py` that passes.
Everything was linted with `black,flake8,autopep8` through `pre-commit`.

Please let me know if there are any issues or changes you'd like to see.
Thanks for building a great ecosystem!

Best,
Jacob

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -43,7 +43,7 @@ jobs:
   - script: |
       python -m pip install --upgrade pip
       pip install pytest-cov wheel
-      pip install .[dev,doc,test,louvain,leiden,magic,harmony,scrublet,scanorama,skmisc]
+      pip install .[dev,doc,test,louvain,leiden,magic,harmony,scrublet,scanorama,scnym,skmisc]
     displayName: 'Install dependencies'
 
   - script: |
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ cache:
 - directories:
   - data
 install:
-- pip install .[dev,test,louvain,leiden,magic,harmony,skmisc,scrublet,scanorama]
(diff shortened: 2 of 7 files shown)


ivirshup:
@jacobkimmel thanks for the PR and the kind words!

Question: why do you want this to be in `external` and not `ecosystem`? Generally, I think of `external` as a place to put provide a "scanpy like" API, but `scNym` already provides this kind of API.

jacobkimmel:
Hey @ivirshup,

Thanks for the feedback.

> why do you want this to be in external and not ecosystem? Generally, I think of external as a place to put provide a "scanpy like" API, but scNym already provides this kind of API.

Two reasons:

1. This is my first introduction to `ecosystem`, so I hadn't actually considered it.
2. Given that the `scnym` API is a single function, adding an `sce.tl` endpoint seems like a parsimonious way to improve discoverability for users.

Happy to remove this PR and suggest an edit to `ecosystem.rst` instead if that's preferable to the team.

All the best,
Jacob

ivirshup:
So, we discussed this at the last team meeting and are in agreement that this would be a better fit for the Ecosystem page. We're going to try to increase the visibility of this page to make it more discoverable (#1801).

For the entry, I think including a link to a tutorial as well as the main project page under the entry in ecosystem would be helpful for getting users to check it out.

jacobkimmel:
> this would be a better fit for the Ecosystem page

Understood, I'll go ahead and close this PR and open a clean one that modifies `ecosystem.rst` alone.

Thanks for taking the time to find the right place for this!

Best,
jacob

PR #1805, opened 2021-04-27
Allowing resolution parameter to be used with RAPIDS Louvain clustering

avantikalal:
I allowed the 'resolution' parameter to be used in Louvain clustering when flavor='rapids'. The resolution parameter works with cuGraph Louvain clustering: https://docs.rapids.ai/api/cugraph/stable/api.html#module-cugraph.community.louvain

Tested that the function works and docs are built correctly.


Zethson (approved):
Thanks! Travis fails unrelated.

PR #1811, opened 2021-04-27
Inline example paga

le-ander:
adding an inline example plot for `sc.pl.paga()`
See #1664

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -91,6 +91,7 @@ doc = [
     'sphinx-autodoc-typehints',
     'scanpydoc>=0.6',
     'typing_extensions; python_version < "3.8"',  # for `Literal`
+    'python-igraph',
 ]
 test = [
     'pytest>=4.4',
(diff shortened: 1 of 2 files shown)


le-ander:
Hm, I am a bit lost wrt to the checks failing @giovp 
the doc build doesn't show any logs and travis fails because of black complaining about scanpy/tools/_sim.py and some version checking problem. None of this seems related to my PR 🤔

ivirshup:
I don't think `igraph` or `louvain` are actually being installed on readthedocs. 

I think you'll need to modify the `.readthedocs.yml` for this. But not calling `louvain` could also work.

giovp:
looks good, thanks @le-ander

giovp (approved):

giovp:
I've added igraph to #1809 and seems to work, will ad it here as well

le-ander:
Thanks a lot, giovanni!

PR #1815, opened 2021-04-27
added example for PCA loadings plot

bio-la:
ref  #1664 sc.pl.pca_loadings

### DIFF
--- a/scanpy/plotting/_tools/__init__.py
+++ b/scanpy/plotting/_tools/__init__.py
@@ -108,9 +108,11 @@ def pca_loadings(
         adata = sc.datasets.pbmc68k_reduced()
         sc.tl.pca(adata, svd_solver='arpack')
 
+    Show first 3 components loadings
+
     .. plot::
-        Show first 3 components loadings
         :context: close-figs
+
         sc.pl.pca_loadings(adata, components = '1,2,3')
 
 
--- a/scanpy/plotting/_tools/__init__.py
+++ b/scanpy/plotting/_tools/__init__.py
@@ -96,6 +96,24 @@ def pca_loadings(
         If `True` or a `str`, save the figure.
         A string is appended to the default filename.
         Infer the filetype if ending on {`'.pdf'`, `'.png'`, `'.svg'`}.
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        import scanpy as sc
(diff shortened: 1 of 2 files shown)


scanpy/plotting/_tools/__init__.py
```diff
@@ -96,6 +96,26 @@ def pca_loadings(
...
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        import scanpy as sc
+        import numpy as np
+        import pandas as pd
+        adata = sc.datasets.pbmc68k_reduced()
+        sc.tl.pca(adata, svd_solver='arpack')
```
ivirshup: I'm not sure we actually need to show the PCA being computed. Plus it takes up CI time and it'd be nice to keep these minimal.

ivirshup (changes requested):
Thanks for the PR!

Minor comment on skipping some computation for the example. If you think it's important to show the computation, you can include a code block that doesn't get executed the runs the PCA.

Also, it looks like a `h5ad` file got included by accident. Could you remove that?

scanpy/plotting/_tools/__init__.py
bio-la: hey @ivirshup thanks,
the precomputed sc.datasets.pbmc68k_reduced throws an error so the PCA has to be computed for this to work (see #1816 ). i didn't think it was a big issue to recalc, I do get your point on showing the step without calculating if it's not necessary. 
how do you suggest to solve this? what is a better adata to start from, i guess with pbmc3k_processed the processing would be the same?

sorry for the h5ad will remove and update the commit when you let me know!

ivirshup: Yeah, I'd probably just use `pbmc3k_processed` for this. If you want to show the `pca` call, but not have it run, you can do something like:

```rst
.. code:: python

    import scanpy as sc
    
    sc.pp.pca(pbmc)
```

You could probably even hide the `import` and the data loading if you wanted.

> i didn't think it was a big issue to recalc

It's not, especially for PCA. It becomes more of an issue for things like `umap` or `tsne`. I figure it's better to just be consistent about it, so someone could pretty much copy what's done here for those functions.

giovp:
looks good to me

giovp (approved):

PR #1821, opened 2021-04-29
remove colorbar if legend_loc=None for continuous colorbars

### LINKED ISSUE #1502: Better document `legend_loc=None` for embedding plots
- [x] Additional function parameters / changed functionality / changed defaults?
- [ ] New analysis tool: A simple analysis tool you have been using and are missing in `sc.tools`?
- [ ] New plotting function: A kind of plot you would like to seein `sc.pl`?
- [ ] External tools: Do you know an existing package that should go into `sc.external.*`?
- [ ] Other?

It is a lovely tool, but it would be great to be able to not plot figure legends in, e.g. sc.pl.umap, so that all plots (regardless of what annotation is being colored) are exactly the same size. Perhaps there is already a way to do this, but I've yet to find it. 

Thanks!

AnnaChristina:
closes #1502 

ensure that `legend_loc=None` also removes continuous colorbars for scatterplots like umap


ivirshup:
Thanks for the PR!

Does `legend_loc` ever interact with colorbars otherwise? At least for the embedding plots, I think it only refers to categorical legends. Should colorbar positioning be separate from this argument, especially since options like `"on data"` don't really apply?

AnnaChristina:
would also make sense to have this as `colorbar_loc` as this only really applies for continuous coloring, right? I can definetly see the benefit of removing colorbar from figure if this is wanted by the user

ivirshup:
Sorry for the late reply, thought I responded to this!

> would also make sense to have this as colorbar_loc as this only really applies for continuous coloring

I think that would make sense

LidiaML:
Hi!
Could you finally solve the issue of removing colorbar from the figure? I have tried with the legend_loc=None and legend_loc='none' and they don't remove continuous colorbars. I have tried to fix this issue myself but couldn't find the exact place in the code to do that. Any help would be more than welcome!

MMRES-PyBootcamp:
Hi! 
Was this issue solved? I am trying to remove a continuous colorbar from a umap without success...
Thanks,

Lídia

MMRES-PyBootcamp:
Hi!
Could you finally solve this issue? I am trying to remove a continuous colorbar from a umap and I cannot make it.
Thanks,

Lídia

ivirshup:
Just changed the argument a little and added a test. Will be out soon!

Issue #1174, opened 2020-04-21
pl.violin will not take "hue" kwarg for seaborn?


auesro:
Trying to make a violin plot adding the seaborn hue argument will result in ValueError.
In adata, 'timepoint' and 'replicate' are categorical adata.obs containing floats and ints, respectively. 'timepoint' is the age of the embryo from which cells were isolated (9.5, 10.5, etc) and 'replicate' the number of the replica (1, 2, 3).

```
sc.pl.violin(adata, 'n_genes', jitter=0.4, groupby = 'timepoint', stripplot=False, hue='replicate')
```

```
sc.pl.violin(adata, 'n_genes', jitter=0.4, groupby = 'timepoint', stripplot=False, hue='replicate')
Traceback (most recent call last):

  File "<ipython-input-5-756b321177a2>", line 1, in <module>
    sc.pl.violin(adata, 'n_genes', jitter=0.4, groupby = 'timepoint', stripplot=False, hue='replicate')

  File "/home/auesro/anaconda3/envs/Scanpy/lib/python3.7/site-packages/scanpy/plotting/_anndata.py", line 759, in violin
    **kwds,

  File "/home/auesro/anaconda3/envs/Scanpy/lib/python3.7/site-packages/seaborn/categorical.py", line 2393, in violinplot
    color, palette, saturation)

  File "/home/auesro/anaconda3/envs/Scanpy/lib/python3.7/site-packages/seaborn/categorical.py", line 559, in __init__
    self.establish_variables(x, y, hue, data, orient, order, hue_order)

  File "/home/auesro/anaconda3/envs/Scanpy/lib/python3.7/site-packages/seaborn/categorical.py", line 152, in establish_variables
    raise ValueError(err)

ValueError: Could not interpret input 'replicate'
```

#### Versions:

> scanpy==1.4.6 anndata==0.7.1 umap==0.4.1 numpy==1.18.1 scipy==1.4.1 pandas==1.0.3 scikit-learn==0.22.2.post1 statsmodels==0.11.1 python-igraph==0.8.0

kcArtemis:
Hey,can I look into this issue?

PR #1822, opened 2021-04-29
bug fix: hue in sc.pl.violin

AnnaChristina:
closses #1174 

hue key is now added to keys if hue is in kwds. before obs_tidy did not include this keyword, so this resulted in an ValueError

an alternative would be to add hue as a function parameter to `sc.pl.violin`

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -713,7 +713,7 @@ def violin(
 
     # appending by hue key if provided in kwds
     if kwds.get('hue') is not None:
-        keys = keys.append(kwds.get('hue'))
+        keys.append(kwds.get('hue'))
 
     if groupby is not None:
         obs_df = get.obs_df(adata, keys=[groupby] + keys, layer=layer, use_raw=use_raw)
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -711,6 +711,10 @@ def violin(
             f'found `{len(ylabel)}`.'
         )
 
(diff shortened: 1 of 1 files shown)


ivirshup:
Thanks for the PR.

I think if this is added it should be added as a function parameter. It should also get some tests, I'm a bit concerned how just passing `hue` will work with the groupby and palette parameters.

AnnaChristina:
would it then make generally sense to add an optional hue parameter in seaborn based plotting functions? was checking if hue is anywhere used but does not seem to be the case. 

will check behaviour for this wrt groupby and palette.

ivirshup:
> would it then make generally sense to add an optional hue parameter in seaborn based plotting functions? was checking if hue is anywhere used but does not seem to be the case.

I'm not sure. I don't know off the top of my head how well we can make current functions work with `hue`, since seaborn uses it to group observations in many cases.

ivirshup:
Looking at this again, my current reading is that

```python
sc.pl.violin(adata, keys=keys, groupby=groupby, hue=hue)
```

should translate into something like:

```python
df = sc.get.obs_df(adata, [groupby, hue] + keys)
for k in keys:
    sns.violinplot(data=df, x=groupby, y=k, hue=hue)
```

`sns.catplot` could also be a good model, but I don't think it supports the same range of attributes.

While also passing through the right color palette to use for hue.

* We would need to address where the legend goes, since I believe this would currently repeat the same legend for each plot
* `palette` handling, which I believe is currently resulting in a `KeyError`
* Axis labeling, we would probably want it per groupby, and leave `hue` to the legend.

PR #1844, opened 2021-05-17
Fix ipython detection

### LINKED ISSUE #1841: Failing to detect ipython

---

Initially reported by @VolkerBergen in #1477.

From within ipython:

```python
from scanpy._settings import ScanpyConfig
ScanpyConfig._is_run_from_ipython()
# False
getattr(__builtins__, "__IPYTHON__")
# True
```

I'm not sure if/ how this ever worked, [since `__builtins__` is supposed to be a `dict` unless you're in the `__main__` namespace](https://docs.python.org/3/reference/executionmodel.html#builtins-and-restricted-execution). We can probably replace this with:

```python
import builtins
getattr(builtins, "__IPYTHON__", False)
```

ivirshup:
Fixes #1841

It would be nice to test this, but I'm not sure we're set up for that. E.g., we can test that it doesn't work in a python session, but I'm not sure if we can test that it does work in an ipython session.

TODO:

- [x] Release note
- [ ] ~~Test?~~ Not so sure how to do this, and it's useful now.

### DIFF
--- a/docs/release-notes/1.8.0.rst
+++ b/docs/release-notes/1.8.0.rst
@@ -25,6 +25,7 @@
 .. rubric:: Bug fixes
 
 - Fix :func:`scanpy.pl.paga_path` `TypeError` with recent versions of anndata :pr:`1047` :smaller:`P Angerer`
+- Fix detection of whether IPython is running :pr:`1844` :smaller:`I Virshup`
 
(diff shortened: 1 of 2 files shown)


VolkerBergen (approved):
Works perfectly!

PR #1848, opened 2021-05-21
Add more pre-commit checks

Zethson:
Hey,

finally solves #1563 

I added: 

```
    -   id: trailing-whitespace
    -   id: end-of-file-fixer
    -   id: check-added-large-files
    -   id: check-case-conflict
    -   id: check-toml
    -   id: check-yaml
    -   id: check-merge-conflict
    -   id: detect-private-key
```

They are super super fast, so speed is not a concern.

Signed-off-by: zethson <lukas.heumos@posteo.net>


.pre-commit-config.yaml
```diff
@@ -12,3 +12,15 @@ repos:
     hooks:
     -   id: autopep8
         args: ["-i"]
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.0.1
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-added-large-files
+        exclude: docs/data
```
ivirshup: ?

We don't want any large files in there. Hopefully none got through with all the plotting example PRs...

It does look like you might have added one here: `docs/data/pbmc3k_raw.h5ad`

ivirshup (changes requested):
Only thing that needs to change is the added large file. 

Could you try building the docs and checking if anything gets added? The `scanpy.plotting.rst` file has been weird in the past. I think it's at least started being consistent with newer versions of sphinx, but it'd be good to check.

I would like it if the `yaml` and `toml` checks would also do formatting, but up to you on including that.

.pre-commit-config.yaml
Zethson: Yeah, true.
Think this one origins from me building the docs. I actually thought that this was already there, but apparently not. Hence, I excluded this path.

I will remove the file & remove the exclude.

Zethson:
> Could you try building the docs and checking if anything gets added? The `scanpy.plotting.rst` file has been weird in the past. I think it's at least started being consistent with newer versions of sphinx, but it'd be good to check.

The pbmc data in the docs folder does indeed get downloaded when building the docs.
Do you want me to include this path into the `.gitignore`?

ivirshup:
> Do you want me to include this path into the .gitignore?

:+1:

Zethson:
> > Do you want me to include this path into the .gitignore?
> 
> +1

Done.

ivirshup:
One last thing, could you exclude the test data files from any automatic formatting? I guess it's good to know that new lines at the end of files doesn't matter, but I'd prefer to keep those files exactly as they were written by `spaceranger`/ `zarr`.

Zethson:
@ivirshup 

`exclude: scanpy/tests/_data`

Added for trailing-whitespace and end-of-file-fixer.

ivirshup:
Do revert the changes those made as well please!

Zethson:
@ivirshup Done.

ivirshup:
Great, thanks

Issue #1749, opened 2021-03-18
reproducibility of pca with arpack


fbnrst:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of scanpy.
- [ ] (optional) I have confirmed this bug exists on the master branch of scanpy.

---

Running pca with arpack twice like in the example below sometimes does not give identical results. Sometimes, it does give identical results. Results are reproducible when I use the `randomized` svd solver, the issue is specific to `arpack`.

### Minimal code sample (that we can copy&paste without having any data)

```python
%env PYTHONHASHSEED=0
import numpy as np
np.random.seed(42)
import scanpy as sc

adata = sc.datasets.pbmc3k_processed()
adata1 = sc.pp.pca(adata, copy=True, random_state=42, svd_solver='arpack')
adata2 = sc.pp.pca(adata, copy=True, random_state=42, svd_solver='arpack')

print(np.array_equal(adata1.obsm['X_pca'], adata2.obsm['X_pca']))
print(adata1.obsm['X_pca'] - adata2.obsm['X_pca'])
```

```pytb
env: PYTHONHASHSEED=0
False
[[-4.7683716e-07 -3.5762787e-07 -1.1920929e-07 ... -2.7803779e-03
  -5.5277348e-04  8.6665154e-05]
 [ 9.5367432e-07  4.7683716e-07  6.5565109e-07 ...  3.1501055e-03
  -3.6475658e-03 -1.0871887e-04]
 [ 1.6689301e-06  0.0000000e+00 -1.1920929e-07 ...  1.7441511e-03
   6.9665909e-04  4.2915344e-04]
 ...
 [-5.9604645e-07  4.7683716e-07 -3.5762787e-07 ...  1.5980005e-04
  -1.0134950e-03 -1.2260675e-04]
 [-2.0861626e-07  1.4305115e-06 -8.3446503e-07 ... -7.1705580e-03
  -2.4490356e-03  2.4688244e-04]
 [ 5.3644180e-07 -5.9604645e-07 -6.8545341e-07 ... -1.5603602e-03
   7.9727173e-04  8.6218119e-04]]
```

#### Versions

I am running scanpy in this singularity container: https://singularity-hub.org/collections/5095

<details>

```

-----
anndata     0.7.5
scanpy      1.7.1
sinfo       0.3.1
-----
PIL                 8.1.2
anndata             0.7.5
anyio               NA
attr                20.3.0
babel               2.9.0
[...]

Issue #313, opened 2018-10-19
score_genes reproducibility


coh-racng:
In my preprocessing pipeline, I filtered for the most variable genes and also regress out n_counts and cell cycle. However, it seems like every time I restart my Jupyter notebook and rerun the preprocessing pipeline, I get a different neighborhood graph and UMAP, and therefore different clustering. If I use the same preprocessed data, I can reproduce the same PCA (using svd_solver='arpack'), UMAP, and clusters. So I was wondering if there is randomness introduced during filtering process or regress_out methods? What should I do to ensure reproducibility?

coh-racng:
setting random seed before preprocessing did the trick.

falexwolf:
Ah, that should very likely be due to the cell cyle annotation via `score_genes`. Are you using it? There is a sampling procedure involved that might not set a random seed in the function call.

We should fix that.

falexwolf:
Hm, but it looks like https://scanpy.readthedocs.io/en/latest/api/scanpy.api.tl.score_genes.html should be reproducible with the default settings. In what you describe, only pca, louvain and umap have stochastic elements all of which set a default seed in the function call. I have never observed issues with reproducibility there.

Can we narrow it down to score_genes? Or do you think it's somewhere else?

fidelram:
I think this can be closed as @coh-racng found a solution.

falexwolf:
Yes, but I think that each tool should have the possibility of setting a seed without needing to set it externally. So, I'd like to know why @coh-racng had to do this.

coh-racng:
Thanks for the help. I have traced the randomness to sc.tl.score_genes_cell_cycle. Even if I explicitly state the random state to be 0, two repeats of score_genes_cell_cycle from the same data would give different downstream clustering. On the other hand, if I use one score_genes_cell_cycle output and repeat the downstream clustering methods, the clusters are the same. 
I am currently using scanpy==1.3.1

chris-rands:
Was there any progress on this? I also have an issue of non-reproducible UMAPs when re-starting Jupyter notebook. I have not been able to trace the exact cause of the issue, but it appears unrelated to the PCA step (I am using `svd_solver='arpack'`). In my case setting `random.seed` did not fix the issue. However, surprisingly, combining this with disabling hash randomization  by setting `PYTHONHASHSEED` to `0` did generate reproducible results

Iwo-K:
[...]

PR #1858, opened 2021-05-27
Fixed randomness in tl.diffmap - compute_eigen (v0 argument)

Iwo-K:
Currently tl.diffmap generates slightly different results on consequent runs. This is because the v0 argument is not provided in the scipy.sparse.linalg.eigsh call (inside the compute_eigen function). 

To fix this I set the random_state inside the compute_eigen function and generate a random vector to pass into scipy.sparse.linalg.eigsh as v0. 
I tried to follow how random_state is implemented in other scanpy functions, I hope this is consistent.

To allow user control of the random_state I added random_state arguments to:
- tools/_diffmap.py _diffmap function (tl.diffmap function)
- tools/_dpt.py _diffmap function
- neighbors/__init__.py compute_eigen function

### DIFF
--- a/scanpy/neighbors/__init__.py
+++ b/scanpy/neighbors/__init__.py
@@ -937,6 +937,7 @@ class Neighbors:
         n_comps: int = 15,
         sym: Optional[bool] = None,
         sort: Literal['decrease', 'increase'] = 'decrease',
+        random_state: AnyRandom = 0,
     ):
         """\
         Compute eigen decomposition of transition matrix.
@@ -950,6 +951,8 @@ class Neighbors:
             Instead of computing the eigendecomposition of the assymetric
             transition matrix, computed the eigendecomposition of the symmetric
             Ktilde matrix.
+        random_state
+            A numpy random seed
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


ivirshup (changes requested):
Thanks for the PR!

I think the way we're supposed to be using random state internally (due to recent numpy updates) looks more like:

```python
v0 = random_state.standard_normal(matrix.shape[0])
```

However, I recall there being some weirdness about scipy arpack and random state where it was difficult to ever get a truly deterministic result (see #1749, https://github.com/scipy/scipy/issues/10374).

Could add a test to make sure this works? Something checking that case which is currently non-deterministic becomes deterministic after your patch.

Iwo-K:
Sure, will do! Should I combine this PR with the fix mentioned in issue #313 or would you prefer a separate PR?

ivirshup:
Thanks! Separate PR for that, please!

Iwo-K:
Done! I've changed the random vector generation to use the code you suggested and added a test to test_embedding.

scanpy/neighbors/__init__.py
```diff
@@ -978,8 +981,13 @@ def compute_eigen(
             which = 'LM' if sort == 'decrease' else 'SM'
             # it pays off to increase the stability with a bit more precision
             matrix = matrix.astype(np.float64)
+
+            # Setting the random initial vector
+            random_state = check_random_state(random_state)
+            np.random.set_state(random_state.get_state())
```
ivirshup: Is this line necessary for reproducibility? If not, I don't think we have to modify the global state.

scanpy/tools/_diffmap.py
```diff
@@ -39,6 +41,8 @@ def diffmap(
         .obsp[.uns[neighbors_key]['connectivities_key']],
         .obsp[.uns[neighbors_key]['distances_key']] for connectivities and distances
         respectively.
+    random_state
+            A numpy random seed
```
ivirshup:
```suggestion
        A numpy random seed
```

ivirshup (commented):
Added some minor comments (one formatting, possibly one unnecessary line), but looks great overall.

Could you add a bug-fix release note to `docs/release-notes/1.8.0.rst`?

scanpy/neighbors/__init__.py
Iwo-K: I've just checked, specifying the v0 is enough to get the same results every time. No need for setting global state. I will delete the line, thanks!

Iwo-K:
No problem, all done!

ivirshup:
Thanks for the PR!

PR #1868, opened 2021-06-10
update bbknn arguments and docstring

ktpolanski:
Updated `sc.external.pp.bbknn()` to match current arguments and docstring.

### DIFF
--- a/scanpy/external/pp/_bbknn.py
+++ b/scanpy/external/pp/_bbknn.py
@@ -1,4 +1,5 @@
 from typing import Union, Optional
+import types
 
 from anndata import AnnData
 import sklearn
@@ -14,13 +15,20 @@ del lazy_import
 def bbknn(
     adata: AnnData,
     batch_key: str = 'batch',
+    use_rep: str = 'X_pca',
     approx: bool = True,
-    metric: Union[str, 'sklearn.neighbors.DistanceMetric'] = 'angular',
+    use_annoy: bool = True,
+    metric: Union[
+        str, 'types.FunctionType', 'sklearn.neighbors.DistanceMetric'
+    ] = 'euclidean',
     copy: bool = False,
     *,
+    neighbors_within_batch: int = 3,
     n_pcs: int = 50,
     trim: Optional[int] = None,
-    n_trees: int = 10,
+    annoy_n_trees: int = 10,
+    pynndescent_n_neighbors: int = 30,
+    pynndescent_random_state: int = 0,
     use_faiss: bool = True,
     set_op_mix_ratio: float = 1.0,
     local_connectivity: int = 1,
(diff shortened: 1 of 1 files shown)


ktpolanski:
It seems this is of some relevance to users, as shown by it showing up twice independently over the course of the past week.

For some reason, my tweaks have killed off ReadTheDocs, and I can't check why as upon pressing the "details" button I get the 404 equivalent :) I am now importing `types` so I can correctly define `metric` as also including `types.FunctionType`. This  is probably what is causing whatever hiccup is happening.

@giovp , some assistance with getting the ball rolling on this? The changes are a result of me expanding BBKNN with pynndescent on your recommendation.

pinin4fjords:
@ktpolanski following from https://scanpy.readthedocs.io/en/stable/dev/code.html#code-style and building the docs locally gives me (among some errors that were not relevant):

```
...
/path/to/scanpy/scanpy/external/pp/_bbknn.py:docstring of scanpy.external.pp._bbknn.bbknn:28: WARNING: py:class reference target not found: function
/path/to/scanpy/scanpy/external/pp/_bbknn.py:docstring of scanpy.external.pp._bbknn.bbknn:28: WARNING: py:class reference target not found: function
...
```

Maybe that gives you some pointers?

ktpolanski:
It does not, as an analogous docstring built just fine over on the [BBKNN page](https://bbknn.readthedocs.io/en/latest/).

ktpolanski:
I ran `pip3 install scanpy[doc]`, as instructed by that page, and the doc seems to have built fine locally.

image

The only fishy thing is the `function` is not clickable, while the other two are, but it built.

Also, line 28 is `trim: Optional[int] = None,`, which was there previously already. It's the metric change that's causing this - your argument renaming PR did not hiccup.

pinin4fjords:
Hmm, I'm stumped. @ivirshup ?

scanpy/external/pp/_bbknn.py
```diff
@@ -14,13 +15,20 @@
 def bbknn(
     adata: AnnData,
     batch_key: str = 'batch',
+    use_rep: str = 'X_pca',
```
ivirshup, line 17: Could the representation be chosen with keywords like `obsm`, `layers`? You can use the helper functions `sc.get._get_obs_rep` and `sc .get._set_obs_rep`.

ivirshup (commented):
> The only fishy thing is the function is not clickable, while the other two are, but it built

I think this is it. The error I see is:

```
sphinx.errors.SphinxWarning: /home/docs/checkouts/readthedocs.org/user_builds/icb-scanpy/checkouts/1868/scanpy/external/pp/_bbknn.py:docstring of scanpy.external.pp._bbknn.bbknn:28:py:class reference target not found: function

Warning, treated as error:
/home/docs/checkouts/readthedocs.org/user_builds/icb-scanpy/checkouts/1868/scanpy/external/pp/_bbknn.py:docstring of scanpy.external.pp._bbknn.bbknn:28:py:class reference target not found: function
```

I don't think the `'types.FunctionType'` needs to be quoted, but we've used `collections.abc.Callable` in other places without issue.

ktpolanski:
I can't get it to work. I have unquoted `types.FunctionType`, and `function` remains unclickable. I have tried to switch to an `collections.abc.Callable`, and instead of anything helping the entire argument support vanishes:

image

Please help?

ivirshup:
I think that should do it. I believe the `types` module isn't actually meant to be used for type annotations, that's what the `typing` module is for (which is always confusing to me).

> and instead of anything helping the entire argument support vanishes:

I got this behaviour too, but running `make clean` and building from scratch got things working again.  ¯\_(ツ)_/¯

ktpolanski:
Thanks a lot for the help, assuming things are in order then that's it for me.

ivirshup:
Thanks for the update!

PR #1877, opened 2021-06-16
Dict literal typos in _baseplot_class.py

chris-rands:


ivirshup:
Thanks for the fixes!

PR #1890, opened 2021-06-20
Score genes reproducibility fix

### LINKED ISSUE #313: score_genes reproducibility

Iwo-K:
As discussed in issue #313, score_genes function returns different values on various machines. This is due to using float32 dtype in the `np.nanmean` calls.

This PR fixes this behaviour by changing the dtype to float64 in the relevant sections of code ie. functions `gene_score()` and `_sparse_nanmean`. Following the suggestion of @ivirshup the returned value is now also float64.

I also adapted the tests `test_add_score` and `test_npnanmean_vs_sparsemean` to use and expect float64.


ivirshup (commented):
Looks good overall.

I'm trying to think if there is a good test to add here. Might be difficult if the irreproducibility was only across different machines. Could you suggest a test for this?

Other than that, just needs a bugfix release note.

Iwo-K:
Thank you for checking the code.

I have also been thinking about possible test(s), the only thing that comes to my mind is to check the output against a fixed reference result, which I could verify on a few different machines. But I'm not sure whether this wouldn't cause problems in the future. This could be troublesome to maintain as the matrix will need be stored somewhere and may need re-checking if e.g. numpy has some relevant updates.

I am open to suggestions!

ivirshup:
Hmm, I also consistently get a slightly different result for computing the mean between numpy and numba (I think it's one floating point step).

When you were seeing differences, did they show up as different with `np.allclose`? If they did, I think we can go with adding a small reference test that checks with `np.testing.assert_almost_equal`.

Iwo-K:
I haven't used np.allclose for comparison but if I remember correctly (it was a while ago) the differences were around 4th or 5th decimal point.

I will write a test using np.allclose against a reference and run it on different machines to see how it looks. I will post here the results.

Iwo-K:
I've added the test mentioned above and checked on two machines (MacOS and CentOS) with the same results. 
The differences in output are after 5th decimal point, the test:
- passes np.allclose both with and without the fix (master branch vs this PR)
- passes np.array_equal only with the fix (this PR)
Hence, I've kept only the np.array_equal test.

This series of tests uses _create_sparse_nan_matrix, so I had to set the np.random.seed to make the test reproducible and agree with the reference.
The reference is stored in tests/_data/ as a pkl file, as I've noticed that there were some analogous files stored there for other tests.

ivirshup:
I'm not sure I totally trust that generated test data to be reproducible across scipy versions. Could you use one of the small example datasets modified in a deterministic way instead? Something like `sc.datasets.krumsiek11()`?

Iwo-K:
I see your point, no problem. krumsiek11 dataset does not actually cause inconsistencies float32/float64 (without/with the fix). I suspect this is because it has fewer decimal digits and fewer genes so does not accumulate enough imprecision.
Paul15 data behaves as reported before, so I used that one instead. Hope this is fine!

As previously the test passes with the fix but not without it on the two machines I tested.

Iwo-K:
Sorry forgot to remove the previous .pkl reference and made a bit  of a mess. I squashed everything into a single commit, should be fine now.

scanpy/tests/test_score_genes.py
```diff
@@ -43,6 +47,25 @@ def _create_adata(n_obs, n_var, p_zero, p_nan):
...
+    The reference values had been generated using the same code
+    and stored as a pickle object in ./data
+    """
+
+    np.random.seed(123)
+    adata = sc.datasets.paul15()
+    sc.pp.normalize_per_cell(adata, counts_per_cell_after=10000)
+    sc.pp.scale(adata)
+
+    sc.tl.score_genes(adata, gene_list=adata.var_names[:100], score_name='Test')
```
ivirshup: What is the `np.random.seed(123)` call for?

I don't think there should be any non-determinism in the setup here. Also, this shouldn't have an effect on `sc.tl.score_genes` since randomness there should be controlled by the `random_state` parameter which defaults to 0.

As it stands, removing the `np.random.seed(123)` call shoudn't have any effect on this test. Did you want to do something more like this?

```suggestion
    adata = sc.datasets.paul15()
    sc.pp.normalize_per_cell(adata, counts_per_cell_after=10000)
    sc.pp.scale(adata)

    sc.tl.score_genes(adata, gene_list=adata.var_names[:100], score_name='Test', random_state=123)
```

This would change the output though.

ivirshup (commented):
Thanks! One last thing I noticed about the random numbers, then should be good.

I think test failures are just due to some upstream pandas changes. I've fixed those in #1917

> I squashed everything into a single commit, should be fine now.

We've started just squashing all PRs down to one commit for the merge, but it's useful to keep the git history around once a review starts.

scanpy/tests/test_score_genes.py
Iwo-K: It's a leftover line from the previous test which used randomly generated integers, I just forgot to remove it. There is no need for specifying the seed. Test works fine after removing the line.
The reference values were calculated with default sc.tl.score_genes (=0). I've removed the line in the latest commit.

ivirshup (approved):
Looks good. Thanks for the fix!

PR #1900, opened 2021-06-27
Test rtd failures

### LINKED ISSUE #1899: Latest docs not building
`latest` docs haven't been building for a few days.

Not sure what's causing this yet. There seems to be some issue around malformed doc strings being generated due to special casing around readthedocs. It looks like it might be specific to an import to `legacy_api_wrap`.

ivirshup:
Attempts to fix #1899


ivirshup:
Looks like the issue goes deeper, to `get_version` being called on RTD. Seems like some interaction of that package and the environment.

Ping @flying-sheep

PR #1918, opened 2021-07-05
Fixes for pandas 1.3.0

### LINKED ISSUE #1917: Pandas 1.3.0 compatibility
We have some failing tests due to a couple bugs introduced in pandas 1.3.0:

* https://github.com/pandas-dev/pandas/issues/42380

I think this one is small in scope. Has problems when `df.agg` is called, when all columns are categorical and index is non-unique. Definitely a bug in pandas, and I don't think we do this much. Switching to `df.apply` works around the problem.

* https://github.com/pandas-dev/pandas/issues/42376

Assignment of single columns `np.matrix` to dataframe columns no longer works as if the matrix were a 1d array. I think this is a bug since it's an undocumented behaviour change. Fixes are pretty easy, since we can just wrap occurrences with `np.ravel`, however I wouldn't be surprised if there were many places in the codebase that this happened.

I would be good if these didn't happen.

ivirshup:
Fixes #1917 (at least works around it)

### DIFF
--- a/docs/release-notes/1.8.1.rst
+++ b/docs/release-notes/1.8.1.rst
@@ -3,5 +3,6 @@
 
 .. rubric:: Bug fixes
 
+- Workarounds for some changes/ bugs in pandas 1.3 :pr:`1918` :smaller:`I Virshup`
 
 .. rubric:: Performance Enhancements
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -1938,7 +1938,7 @@ def _prepare_dataframe(
         categorical.name = groupby[0]
     else:
         # join the groupby values  using "_" to make a new 'category'
-        categorical = obs_tidy[groupby].agg('_'.join, axis=1).astype('category')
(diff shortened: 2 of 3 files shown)


ACastanza:
Just a heads up, it looks like Pandas 1.3.3 might break things again, was experiencing errors that I was able to resolve by downgrading. I can create a new issue if you'd like.

Error below so you can determine if this is the same issue or not:
```
    adata.obs['log_counts'] = np.log(adata.obs['n_counts'])
  File "/opt/conda/lib/python3.8/site-packages/pandas/core/frame.py", line 3612, in __setitem__
    self._set_item(key, value)
  File "/opt/conda/lib/python3.8/site-packages/pandas/core/frame.py", line 3784, in _set_item
    value = self._sanitize_column(value)
  File "/opt/conda/lib/python3.8/site-packages/pandas/core/frame.py", line 4509, in _sanitize_column
    com.require_length_match(value, self.index)
  File "/opt/conda/lib/python3.8/site-packages/pandas/core/common.py", line 531, in require_length_match
    raise ValueError(
ValueError: Length of values (1) does not match length of index (38978)
```

PR #1933, opened 2021-07-06
Set NUMBA_THREADING_LAYER in tests

### LINKED ISSUE #1931: pynndescent 0.5.3 breaking parallel numba stuff
Numba can’t correctly detect when a threading backend is available. This was fine in the past, since `pynndescent`/ `umap` were forcing a `workqueue` backend which is always available.

`pynndescent` 0.5.3 recently came out, and started favoring the `tbb` backend. This is a problem since numba thinks this backend is available on our CI (and on my HPC), but it’s actually not. I also can’t get tbb to even install in a way numba sees in on these systems. If tbb isn’t actually available, but numba detects it we get errors with horrible tracebacks anytime parallelized numba code is used https://github.com/lmcinnes/pynndescent/issues/129. These tracebacks are so horrible they break our CI further https://github.com/pytest-dev/pytest-nunit/issues/47.

So what do we do?

## Possible solutions:

* Pin pynndescent below 0.5.3.

This makes pynndescent a required dependency. We’ve previously avoided this since it would change results for people using `umap<0.4` (e.g. anyone with `scvelo` installed) who did not explicitly install pynndescent. However, given the lack of complaints around umap results changing as dependencies have increased, this may not be so bad.

It would be great if we could constrain the version without having it be a dependency. This would be similar to what's possible with `pip` and [constraints files](https://pip.pypa.io/en/stable/user_guide/#constraints-files), I don't see how one would be able to specify this for a package. I don't think it's possible, but maybe I'm missing something about the version string syntax.

* Make sure the numba threading layer is `“workqueue”` after pynndescent is imported

This is tricky. pynndescent<0.5.3 takes a long time to import, so we don’t want to do this at the top level. So we would need to add a check after everytiem pynndescent could possibly be imported to check that it didn’t set the threading backend to anything else. My understanding of the numba threading system is that once you’ve called for parallel compilation, you’re locked into the threading backend for that session.

* Make sure the threading layer is workqueue after pynndescent is imported, but make pynndescent import fast

A kinda simpler solution would be to require pynndescent 0.5.3, import it at the top level (boosts total import time by about ~0.2 seconds out of 3.3 total), and set the threading layer right there. This also has the advantage of letting users specify the threading layer after import scanpy and not having us interfere.

This also requires that pynndescent is installed, which hits the reproducibility issues again.

flying-sheep:
Also move the `MPLBACKEND=agg` to variable definitions

Threading layer docs: https://numba.pydata.org/numba-doc/latest/user/threading-layer.html

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -5,6 +5,8 @@ variables:
   PIP_CACHE_DIR: $(Pipeline.Workspace)/.pip
   ANNDATA_DEV: no
   RUN_COVERAGE: no
+  MPLBACKEND: agg  # make matplotlib work without $DISPLAY set
+  NUMBA_THREADING_LAYER: workqueue  # unified numba backend
 
 jobs:
 - job: PyTest
@@ -34,11 +36,6 @@ jobs:
         python
       path: $(PIP_CACHE_DIR)
     displayName: Cache pip packages
-
-  - script: |
-      export MPLBACKEND="agg"
-      echo $MPLBACKEND
-    displayName: 'Set env'
 
   - script: |
       python -m pip install --upgrade pip
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -5,7 +5,6 @@ variables:
   PIP_CACHE_DIR: $(Pipeline.Workspace)/.pip
   ANNDATA_DEV: no
   RUN_COVERAGE: no
-  MPLBACKEND: agg  # make matplotlib work without $DISPLAY set
   NUMBA_THREADING_LAYER: workqueue  # unified numba backend
 
 jobs:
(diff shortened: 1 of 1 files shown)


ivirshup:
I think pynndescent setting the value at import time will take precedence over the environment variable

flying-sheep:
Oh. Well, that’s just not right. They need to undo that.

ivirshup:
I think it's reasonable that the runtime gets the final say. It's pretty standard for precedence to go: `runtime > environment > config file`, right? I don't think it's reasonable for a library to make the decision, as it should be done by the program.

flying-sheep:
Yes, exactly. The final say should be with the user. I think lmcinnes/pynndescent#130 should apply the workaround only when the buggy one is selected and not generally …

ivirshup:
As in my examples on https://github.com/lmcinnes/pynndescent/issues/129, I didn't see any issues with `omp`, just `tbb`. I'm wondering if `omp` should even be considered "the buggy backend".

PR #1936, opened 2021-07-07
embedding_density: remove limitation <10 categories

falexwolf:
https://github.com/theislab/scanpy/issues/719


LuckyMD:
Looks good to me! Docstring is also clearer now. Do you think it's worth adding a warning at a hard minimum threshold of 5-10 cells per category or sth like this? Although it would probably also just look strange to anyone plotting it.

falexwolf:
I thought about the warning, but I think it's more distraction than actually helpful. The warning in the docstring should do the job, I hope!

ivirshup (approved):
LGTM. I also think this is small enough that we could put it in a bugfix release instead of waiting until 1.9. Preferences on this?

-------------------

I think warning about group size would apply to so many parts of the analysis it would almost be strange to only warn about it here.

ivirshup:
Bumped so it's up to date, and added release note

PR #1939, opened 2021-07-07
Split function in sc.get

Koncopd:
This adds `sc.get.split` to split adata based on an obs column. The examples are in the docstring.

### DIFF
--- a/scanpy/get/get.py
+++ b/scanpy/get/get.py
@@ -1,5 +1,5 @@
 """This module contains helper functions for accessing data."""
-from typing import Optional, Iterable, Tuple, Union, List
+from typing import Optional, Iterable, Tuple, Union, List, Dict
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


ivirshup:
I think this is a really good idea. Definitely a good hold over until a real GroupBy object is worked out. I'll review in depth later but have a few thoughts:

* How about `split_by`? Rhymes with `group_by` which this is quite like.
* I would like to be able to do this along the `var` axis
* What do you think about including some of the grouping options from [`_prepare_dataframe`](https://github.com/theislab/scanpy/blob/d072abd05bda07f280ea91f5e7e4a84f9782c118/scanpy/plotting/_anndata.py#L1842)? E.g. grouping by multiple keys, binning by continuous values? It would be nice if this could be used in place of `_prepare_dataframe`.

The features would not have to be implemented in this PR, but it would be good to think about what the API for them would be and making sure it could be added elegantly.

Koncopd:
Renamed to `split_by`, added `axis` argument.

Koncopd:
What should be names of interval categories in a returned `dict` for binning? Intervals converted to strings?

scanpy/get/get.py
```diff
@@ -437,3 +437,141 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+    if axis not in (0, 1):
+        raise ValueError("axis should be 0 or 1 only.")
+
+    attr_by_key = adata.obs[key] if axis == 0 else adata.var[key]
+    select = [slice(None), slice(None)]
+
+    adatas = {}
+
+    if groups is None:
+        groups = np.unique(attr_by_key)
```
ivirshup: This should be faster for categoricals (since pandas knows to just look at the categories)

```suggestion
        groups = attr_by_key.unique()
```

scanpy/get/get.py
```diff
@@ -437,3 +437,141 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+                                 dict(some=['CD14+ Monocyte', 'CD34+']),
+                                 others_key='others')
+    >>> adatas
+    {'some': View of AnnData object with n_obs × n_vars = 142 × 765
+     ...,
+     'others': View of AnnData object with n_obs × n_vars = 558 × 765
+     ...
+    }
+    """
+    if key not in adata.obs:
```
ivirshup: This wouldn't work for the var axis. Maybe you could abstract over the dimension before this check?

scanpy/get/get.py
```diff
@@ -437,3 +437,141 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+    copy: bool = False,
+) -> Dict[str, AnnData]:
+    """\
+    Split adata by obs key.
+
+    Params
+    ------
+    adata
+        AnnData object to split.
+    key
```
ivirshup: How about calling this `by` to mimic `pd.DataFrame.groupby` or `group` for `xarray.groupby`?

scanpy/get/get.py
```diff
@@ -437,3 +437,141 @@ def _set_obs_rep(adata, val, *, use_raw=False, layer=None, obsm=None, obsp=None)
...
+
+GroupsList = List[str]
+GroupsLists = List[GroupsList]
+GroupsDict = Dict[str, GroupsList]
+
+
+def split_by(
+    adata: AnnData,
+    key: str,
+    groups: Optional[Union[GroupsList, GroupsLists, GroupsDict]] = None,
```
ivirshup:
```suggestion
    key: str,
    *.
    groups: Optional[Union[GroupsList, GroupsLists, GroupsDict]] = None,
```

Could we everything past `key` keyword only? This way we don't have to worry about changing orders.

ivirshup (commented):
I've taken a closer look. The argument I have the most issue with is `groups`, which I think could be replicated by allowing passing `Series` through for `key` (like `df.groupby(series)`).

# Specifying groups more granularly

I see the value in `GroupsList` and `GroupsDict`, but less `GroupsLists`. `GroupsLists` is just not providing a name for `GroupsDict`. I think I'd rather the user just had to specify a name for the grouping.

## `groups` argument usage and precedence

Also, can you provide precedence in other APIs for this kind of argument? I'm wondering if we should instead just allow passing a series as `key`, then the user can create whatever kind of flat grouping they want (example below with numeric values).

I'm also wondering if there is a reason other libraries (`xarray`, `pandas`) don't include this, and if defining the group sets/ subsetting the data makes more sense as a separate operation. Being able to regroup by `dict` seems quite elegant. Filtering the groups used seems like it's largely equivalent to subsetting.

## Non-flat labellings

One interesting part about providing sets of groupings is that they are no longer mutually exclusive sets. In scikit-learn this is represented by a 2d boolean array with dimensions `(obs x groups)` ([sklearn docs](https://scikit-learn.org/stable/modules/multiclass.html#id5)). It would be nice to have interchange with that format.

# Numeric values

> What should be names of interval categories in a returned dict for binning? Intervals converted to strings?

I think using the intervals themselves should be fine.

I'm wondering if we should even directly include this, or make this available through:

```python
sc.get.split_by(adata, pd.cut(adata.obs["total_counts"], bins=10))
```

I think this kind of solution also provides a way for customizing group selections (e.g. could replace the need for the `groups` argument).

# Library choice

Should this go in `anndata`?

scanpy/get/get.py
Koncopd: Renamed.

scanpy/get/get.py
Koncopd: Fixed

scanpy/get/get.py
Koncopd: Changed as you suggested.

Koncopd:
Now `key` is renamed to `by`, it can take a string, list or series with group values for each observation. I still prefer to have the `groups` argument instead of making custom grouping series.

PR #1942, opened 2021-07-09
Fix _rank_genes_groups_plot for filtered genes

Koncopd:
https://github.com/theislab/scanpy/issues/1941

### DIFF
--- a/scanpy/tests/test_plotting.py
+++ b/scanpy/tests/test_plotting.py
@@ -1469,19 +1469,20 @@ def test_color_cycler(caplog):
         # TODO: add other rank_genes_groups plots here once they work
     ),
 )
-def test_filter_rank_genes_groups_plots(plot, check_same_image):
+def test_filter_rank_genes_groups_plots(tmpdir, plot, check_same_image):
+    TESTDIR = Path(tmpdir)
+
     adata = sc.datasets.pbmc68k_reduced()
 
     sc.tl.rank_genes_groups(adata, 'bulk_labels', method='wilcoxon', n_genes=5)
 
-    args = {
-        'adata': adata,
-        'key_added': 'rank_genes_groups_filtered',
-        'min_in_group_fraction': 0.25,
-        'min_fold_change': 1,
-        'max_out_group_fraction': 0.5,
-    }
-    sc.tl.filter_rank_genes_groups(**args)
+    sc.tl.filter_rank_genes_groups(
+        adata,
+        key_added='rank_genes_groups_filtered',
+        min_in_group_fraction=0.25,
+        min_fold_change=1,
+        max_out_group_fraction=0.5,
+    )
 
     var_names = {}
     genes_names = adata.uns['rank_genes_groups_filtered']['names']
@@ -1489,8 +1490,8 @@ def test_filter_rank_genes_groups_plots(plot, check_same_image):
         genes = genes_names[group].astype(str).tolist()
         var_names[group] = [g for g in genes if g != 'nan']
 
-    pth_a = FIGS / f"{plot.__name__}_filter_a.png"
-    pth_b = FIGS / f"{plot.__name__}_filter_b.png"
+    pth_a = TESTDIR / f"{plot.__name__}_filter_a.png"
+    pth_b = TESTDIR / f"{plot.__name__}_filter_b.png"
 
     plot(adata, key='rank_genes_groups_filtered')
     plt.savefig(pth_a)
--- a/scanpy/tests/test_plotting.py
+++ b/scanpy/tests/test_plotting.py
@@ -1471,10 +1471,11 @@ def test_color_cycler(caplog):
 )
 def test_filter_rank_genes_groups_plots(tmpdir, plot, check_same_image):
     TESTDIR = Path(tmpdir)
+    N_GENES = 4
 
     adata = sc.datasets.pbmc68k_reduced()
 
-    sc.tl.rank_genes_groups(adata, 'bulk_labels', method='wilcoxon', n_genes=5)
+    sc.tl.rank_genes_groups(adata, 'bulk_labels', method='wilcoxon', pts=True)
 
     sc.tl.filter_rank_genes_groups(
         adata,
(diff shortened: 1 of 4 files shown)


Koncopd:
No idea why the unrelated plot tests fail.

ivirshup (commented):
Could you add tests for this?

> No idea why the unrelated plot tests fail.

Yeah, not sure. Just re-triggering the build seems to have fixed that.

Koncopd:
I am not sure what king of test. I don't want to add another `save_and_compare_images` test because plots seem to depend on the system at least sometimes (i have 3 failing plotting tests locally but they run fine here).

Koncopd:
This is strange, i also tried to run the tests multiple times at the time of committing this and they failed every time.

ivirshup:
> This is strange, i also tried to run the tests multiple times at the time of committing this and they failed every time.

Maybe a dependency had a bugged release at the time?

> I am not sure what king of test. I don't want to add another save_and_compare_images test because plots seem to depend on the system at least sometimes

You could instead use `check_same_image`.  Check that running `filter_rank_genes_group` then plotting is equivalent to manually passing those genes to `sc.pl.rank_genes_groups_*` plot on an object that hasn't had `filter_rank_genes_group` run on it.

You can search the tests for examples of `check_same_image`.

>  (i have 3 failing plotting tests locally but they run fine here).

Could you open an issue for this and note which tests they are? It would be good to make the tests as resilient as possible on other people's systems.

Koncopd:
Thanks, i will try with `check_same_image`.

Yes, i will check again and open the issue.

scanpy/tests/test_plotting.py
```diff
@@ -1456,3 +1456,48 @@ def test_color_cycler(caplog):
...
+    sc.tl.filter_rank_genes_groups(**args)
+
+    var_names = {}
+    genes_names = adata.uns['rank_genes_groups_filtered']['names']
+    for group in genes_names.dtype.names:
+        genes = genes_names[group].astype(str).tolist()
+        var_names[group] = [g for g in genes if g != 'nan']
+
+    pth_a = FIGS / f"{plot.__name__}_filter_a.png"
+    pth_b = FIGS / f"{plot.__name__}_filter_b.png"
```
ivirshup: My preference would be to write these to a fresh temporary directory so there can't be conflicts with other tests. `test_rank_genes_groups_plots_n_genes_vs_var_names` is an example of this.

scanpy/tests/test_plotting.py
```diff
@@ -1456,3 +1456,48 @@ def test_color_cycler(caplog):
...
+    sc.tl.rank_genes_groups(adata, 'bulk_labels', method='wilcoxon', n_genes=5)
+
+    args = {
+        'adata': adata,
+        'key_added': 'rank_genes_groups_filtered',
+        'min_in_group_fraction': 0.25,
+        'min_fold_change': 1,
+        'max_out_group_fraction': 0.5,
+    }
+    sc.tl.filter_rank_genes_groups(**args)
```
ivirshup: I don't think `args` is reused anywhere, so I think it this would be easier to follow without it

```suggestion
     sc.tl.filter_rank_genes_groups(
        adata,
        key_added='rank_genes_groups_filtered',
        min_in_group_fraction=0.25,
        min_fold_change=1,
        max_out_group_fraction=0.5,
     )
```

scanpy/tests/test_plotting.py
```diff
@@ -1456,3 +1456,49 @@ def test_color_cycler(caplog):
...
+        sc.pl.rank_genes_groups_dotplot,
+        sc.pl.rank_genes_groups_heatmap,
+        sc.pl.rank_genes_groups_matrixplot,
+        sc.pl.rank_genes_groups_stacked_violin,
+        sc.pl.rank_genes_groups_tracksplot,
+        # TODO: add other rank_genes_groups plots here once they work
+    ),
+)
+def test_filter_rank_genes_groups_plots(tmpdir, plot, check_same_image):
+    TESTDIR = Path(tmpdir)
```
ivirshup, line 1473:
```suggestion
    TESTDIR = Path(tmpdir)
    N_GENES = 4
```

scanpy/tests/test_plotting.py
```diff
@@ -1456,3 +1456,49 @@ def test_color_cycler(caplog):
...
+
+    adata = sc.datasets.pbmc68k_reduced()
+
+    sc.tl.rank_genes_groups(adata, 'bulk_labels', method='wilcoxon', n_genes=5)
```
ivirshup:
```suggestion
    sc.tl.rank_genes_groups(adata, 'bulk_labels', method='wilcoxon', pts=True)
```

scanpy/tests/test_plotting.py
```diff
@@ -1456,3 +1456,49 @@ def test_color_cycler(caplog):
...
+        min_in_group_fraction=0.25,
+        min_fold_change=1,
+        max_out_group_fraction=0.5,
+    )
+
+    var_names = {}
+    genes_names = adata.uns['rank_genes_groups_filtered']['names']
+    for group in genes_names.dtype.names:
+        genes = genes_names[group].astype(str).tolist()
+        var_names[group] = [g for g in genes if g != 'nan']
```
ivirshup:
```suggestion
    var_names = {}
    for k, subdf in (
        sc.get.rank_genes_groups_df(adata, group=None, key="rank_genes_groups")
        .query("pct_nz_group >= .25")  # min_in_group_fraction=0.25
        .query("logfoldchanges >= 1")  # min_fold_change=1
        .query("pct_nz_reference < .5")  # max_out_group_fraction=0.5
        .groupby("group")
    ):
        var_names[k] = list(subdf.head(N_GENES)["names"])
```

scanpy/tests/test_plotting.py
```diff
@@ -1456,3 +1456,49 @@ def test_color_cycler(caplog):
...
+
+    pth_a = TESTDIR / f"{plot.__name__}_filter_a.png"
+    pth_b = TESTDIR / f"{plot.__name__}_filter_b.png"
+
+    plot(adata, key='rank_genes_groups_filtered')
```
ivirshup:
```suggestion
    plot(adata, key='rank_genes_groups_filtered', n_genes=N_GENES)
```

ivirshup (commented):
I have one more suggestion about the test, which I was getting confused about.

I wouldn't populate `var_names` with the filtered entries in `uns`, since if there was a bug in how `filter_rank_genes_groups` sets those values this test would still pass. Alternatively, if how we store `filter_rank_genes_groups` changes, this test would fail without any bad behaviour actually occurring. If we instead get the `var_names` separately, we limit how this can fail.

I also set `n_genes` at the plotting call, since we want to know that there's no interference between the filtering done by `filter_rank_genes_groups` and the plotting function itself.

Combined, my suggestions here should implement this change.

Koncopd:
I can add the changes you suggested but

>since if there was a bug in how `filter_rank_genes_groups` sets those values this test would still pass.

this is not a test of `filter_rank_genes_groups`, we have a separate test for this. Do you think we need to double check this?

ivirshup:
> this is not a test of `filter_rank_genes_groups`

To me the point of the test is that `filter_rank_genes_groups` followed by `.pl.rank_genes_groups_*` returns the right image by comparing it to a reference. I mainly prefer that the reference is generated by as orthogonal a method as possible, since then we're closer to testing functionality than implementation.

I also think that this strategy forms a more flexible implementation that can be extended to test other properties. For example, it's easy to modify this to check that the `sc.pl.rank_genes_groups_*(..., min_logfoldchange=)` argument plays well with `filter_rank_genes_groups`. I also like having an example of an equivalent implementation in the test suite.

Koncopd:
Ok, added query.

PR #1985, opened 2021-08-23
Account for case when there are fewer dispersed genes than n_top_genes

pedrofale:
Calling `sc.pp.highly_variable_genes(adata, n_top_genes=n_top_genes)` results in an error if `n_top_genes` is larger than the size of the `dispersion_norm` vector, which is the vector that we want to subset. Before this fix, scanpy just checked if `n_top_genes` was greater than `adata.n_vars`, which is unreliable since `dispersion_norm` can be smaller than that due to the subsetting in line 261: `dispersion_norm = dispersion_norm[~np.isnan(dispersion_norm)]`. This PR fixes this.

All tests in `test_highly_variable_genes.py` pass, but others like `test_plotting.py::test_violin` fail. I'm not sure why -- anyone have an idea?


Zethson:
> All tests in test_highly_variable_genes.py pass, but others like test_plotting.py::test_violin fail. I'm not sure why -- anyone have an idea?

They also fail for me for an unrelated change that does not even change code.
@ivirshup got any idea?

ivirshup (changes requested):
Thanks for the fix!

Could you please add tests for this?

Violin plot failures are unrelated, and should be fixed on master now. I've merged master into this pr to get CI to run.

Would not handling `n_top_genes` being larger than number of non-constant genes also cause issues when accounting for batch? E.g. [here](https://github.com/theislab/scanpy/blob/c488909a54e9ab1462186cca35b537426e4630db/scanpy/preprocessing/_highly_variable_genes.py#L500-L511)?

Zethson:
@pedrofale would you be able to add the suggested test for this, please?
If not, @lazappi  PR will succeed this one and we will close this one.

 Thank you!

Zethson:
Closing in favor of https://github.com/scverse/scanpy/pull/2231

PR #2023, opened 2021-10-28
Bugfix for undefined threshold from Scrublet

pinin4fjords:
This is a bugfix to allow for the fact that Scrublet doesn't always find a threshold and produce predictions by:

   - Detecting when threshold not present, and setting prediction column to False'
   - Not assuming presence of threshold in plot functions.


scanpy/external/pl.py
```diff
@@ -372,7 +372,11 @@ def scrublet_score_distribution(
         simulation separately for advanced usage.
     """
 
-    threshold = adata.uns['scrublet']['threshold']
+    threshold = None
+
+    if 'threshold' in adata.uns['scrublet']:
+        threshold = adata.uns['scrublet']['threshold']
```
ivirshup: A bit shorter:

```suggestion
    threshold = adata.uns["scrublet"].get("threshold", None)
```

ivirshup (commented):
Thanks for the fix. I think these approaches make sense.

Could this get tests? I'm thinking

* Make a case where a threshold can't be found (not sure how this would be done)
  * scrublet warns the user
  * Check plots don't error
  * Check that nothing is predicted as a doublet

pinin4fjords:
> * Make a case where a threshold can't be found (not sure how this would be done)

Obviously I can test the plotting by directly unsetting the relevant things, but I'm not actually sure how to trigger a failure with the test data I'm afraid.

Zethson (approved):
LGTM

PR #2025, opened 2021-10-28
Misc scrublet buxfixes

### LINKED ISSUE #1957: sc.external.pp.scurblet normalization error

---

Hello, according to the documentation, the simulation of doublets with scrublet works best with un-normalized counts. The current wrapper implementation however uses a layer to hold the raw data, which then gets normalized anyway. 

Please have in mind, that this is the first issue I have ever opened. If I made any grievous mistakes please inform me of such. 

---

### Minimal code sample:

```python
import scanpy
adata = scanpy.datasets.pbmc3k()
scanpy.external.pp.scrublet(adata)
```
and around line 177 in scanpy.external.pp.scrublet we need to 'observe' the values:
```python
adata_obs.layers['raw'] = adata_obs.X
print(adata_obs.layers['raw'])
pp.normalize_total(adata_obs)  <--- currently normalizes all layers and X
print(adata_obs.layers['raw'])
```
---
### Impact
The 'raw' layer is later used in line 194 of scanpy.external.pp.scrublet to simulate doublets, however, it doesn't contain raw data anymore.

```python
adata_sim = scrublet_simulate_doublets(
            adata_obs,
            layer='raw',
            sim_doublet_ratio=sim_doublet_ratio,
            synthetic_doublet_umi_subsampling=synthetic_doublet_umi_subsampling,
        )
```

pinin4fjords:
This is a factoring-out of Scrublet bugfixes in https://github.com/theislab/scanpy/pull/1965, namely:

 - A fix for https://github.com/theislab/scanpy/issues/1957, doing a full copy of .X where required
 - Ensure raw counts are used for simulation

### DIFF
--- a/docs/release-notes/1.8.3.rst
+++ b/docs/release-notes/1.8.3.rst
@@ -7,6 +7,8 @@
 
 - Fixed finding variables with ``use_raw=True`` and ``basis=None`` in
   :func:`scanpy.pl.scatter` :pr:`2027` :small:`E Rice`
+- Fixed :func:`scanpy.external.pp.scrublet` to address :issue:`1957`
+  :smaller:`FlMai` and ensure raw counts are used for simulation
 
 .. rubric:: Performance
 
--- a/scanpy/external/pp/_scrublet.py
+++ b/scanpy/external/pp/_scrublet.py
@@ -178,9 +178,9 @@ def scrublet(
 
         # HVG process needs log'd data.
 
-        logged = pp.log1p(ad_obs, copy=True)
+        logged = pp.log1p(adata_obs, copy=True)
         pp.highly_variable_genes(logged)
-        ad_obs = ad_obs[:, logged.var['highly_variable']]
+        adata_obs = adata_obs[:, logged.var['highly_variable']]
 
         # Simulate the doublets based on the raw expressions from the normalised
         # and filtered object.
@@ -193,8 +193,8 @@ def scrublet(
         )
 
         if log_transform:
-            pp.log1p(ad_obs)
-            pp.log1p(ad_sim)
+            pp.log1p(adata_obs)
+            pp.log1p(adata_sim)
 
         # Now normalise simulated and observed in the same way
 
--- a/scanpy/external/pp/_scrublet.py
+++ b/scanpy/external/pp/_scrublet.py
@@ -176,17 +176,11 @@ def scrublet(
         adata_obs.layers['raw'] = adata_obs.X.copy()
         pp.normalize_total(adata_obs)
 
-        # HVG process needs log'd data. If we're not using that downstream, then
-        # copy logged data to new object and subset original object based on the
-        # output.
+        # HVG process needs log'd data.
 
-        if log_transform:
-            pp.log1p(adata_obs)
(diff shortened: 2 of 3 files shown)


ivirshup:
Could you add a test to make sure the correct values are being used?

pinin4fjords:
> Could you add a test to make sure the correct values are being used?

What did you have in mind? A minimum value (e.g. 20?) for the max to check that unlogged values are input?

ivirshup:
Basically just a test to make sure the underlying issue is actually corrected.

In this example:

```python
adata_sim = scrublet_simulate_doublets(
            adata_obs,
            layer='raw',
            sim_doublet_ratio=sim_doublet_ratio,
            synthetic_doublet_umi_subsampling=synthetic_doublet_umi_subsampling,
        )
```

`.layer["raw"]` should contain integer data, which you should be able to reconstruct from the parent cells, right?

pinin4fjords:
Sorry @ivirshup , still not quite getting what you're after. The underlying issues were with a missing .copy() (now added) and with log'd values getting into the simulation process (which is now prevented with a simple code rearrangement). I think those fixes are pretty self-evident.

Non-integer values don't necessarily mean normalised or log'd data, so I can't add a check in the code for that. I could add a check on the values of the simulated doublets, that they are the sum of the parent counts we send to scrublet's simulate function, but that seems outside the scope of these fixes.

ivirshup:
> The underlying issues were with a missing .copy() (now added) and with log'd values getting into the simulation process

So a test could be checking that if you passed correctly simulated data into `sc.external.pp.scrublet` as `adata_sim`, you get the same result as letting the function simulate the data itself. You could recreate the simulation using the `.uns['scrublet']['doublet_parents']` field:

```python
def create_sim_from_parents(adata, parents):
    N_sim = parents.shape[0]
    I = sparse.coo_matrix(
        (
            np.ones(2 * N_sim),
            (np.repeat(np.arange(N_sim), 2), parents.flat),
        ),
        (N_sim, adata.n_obs)
    )
    X = I @ adata.X
    return ad.AnnData(
        X,
        var=pd.DataFrame(index=adata.var_names),
        obs=pd.DataFrame({"total_counts": np.ravel(X.sum(axis=1))}),
        obsm={"doublet_parents": parents.copy()},
    )
```

> (which is now prevented with a simple code rearrangement). I think those fixes are pretty self-evident.


Yeah, I do see from the code what was going wrong. The issue is more that I want a check to be sure it does not go wrong again. These things clearly get through code review, but it's harder for them to get through a test.

pinin4fjords:
@ivirshup okay, I've tried that (thanks for the snazzy function), fingers crossed on tests.

This is somewhat involved. When providing adata_sim I had, by design, made the assumption that the user will have done all their own preprocessing. So to make the comparison you requested (sc.external.pp.scrublet with/without adata_sim) I needed to basically replicate all that preprocessing verbatim in the test.

ivirshup (approved):
Sorry about the late response!

That test looks great, thanks! Could you add a release note? Then this is ready to merge

pinin4fjords:
Hope that's what you mean by a release note @ivirshup , let me know if it needs anything else.

ivirshup:
LGTM, thanks!

PR #2027, opened 2021-10-29
Fix finding variables with sc.pl.scatter(..., use_raw=True)

### LINKED ISSUE #1097: sc.pl.scatter does not find gene names in adata.raw
When I tried to use `sc.pl.scatter` with `use_raw=True` to plot genes whose names are found in adata.raw, but not adata (because they have been filtered out during processing), I got:

`ValueError: 'x', 'y', and potential 'color' inputs must all come from either '.obs' or '.var'`

Plotting genes found in both adata and adata.raw works. Plotting the troublesome genes with `sc.pl.umap` with `use_raw=True` also works. So it seems like a specific issue with the scatter plot function. Am I overlooking anything? Thanks in advance!

esrice:
sc.pl.scatter() is a wrapper for _scatter_obs(). It checks to make sure
the variable names the caller is requesting to plot exist in var and/or
obs, but does not take into account whether it should look in raw based
on the use_raw flag, as _scatter_obs() does. This leads to errors when a
user asks to plot variables that are in the raw but not the filtered
matrix of adata. This commit fixes that bug.


esrice:
It looks like all of the tests that failed have to do with ingest functionality and are thus unrelated to this PR. The same ones failed on the PR after this one as well. Let me know what I can do to fix this.

ivirshup:
Hey @esrice, thanks for the PR! Those tests are failing due to a recent umap release. This should be fixed on master now, so just merge master into this branch and they should be fixed.

esrice:
Thanks! Looks like that fixed the problem.

ivirshup (changes requested):
Could this get tests to make sure it works?

I'm not sure this is going to do what you want for plotting columns from var, since `.raw` isn't maintained through `transpose()` (because `.raw` can have a different `var` axis, but not `obs` axis).

esrice:
Yes, for sure, one of the two commits I just pushed adds a few tests for various cases of `sc.pl.scatter()` with `basis=None`.

You're totally right that plotting the transposed matrix (i.e., cells as points in the scatterplot, instead of genes) does not work. I think this is a separate problem from the one I'm fixing here, in response to the issue #1097. It seems to me there are two problems going on:
1. The wrapper function `sc.pl.scatter()` mistakenly raises a ValueError if x, y, or color are var_names that exist only in raw but not in the base layer, even if `use_raw=True`. The underlying `_scatter_obs()` has no problem dealing with this situation, so to solve this, `sc.pl.scatter()` just needs to call `_scatter_obs()` in this case instead of raising a ValueError. This PR fixes that.
2. When x, y, and/or color are variables found in `obs.index` or `var.keys()`, `sc.pl.scatter()` makes a transposed version of `adata`, but as you said, `adata.raw` does not get transposed. This leads to an `AttributeError` on this line of `_scatter_obs()`:

https://github.com/theislab/scanpy/blob/cab9f781f9fdee2eeebf05a84c2ce5f717afa514/scanpy/plotting/_anndata.py#L250
`AttributeError: 'NoneType' object has no attribute 'obs_vector'`

I'm not sure what the correct way to handle this is, so for now, I'm
* adding a commit so that this PR does not modify anything in the part of `sc.pl.scatter()` dealing with transposition
* adding a parametrized test to `test_plotting.py` that tests all use-cases of `sc.pl.scatter()` where `basis=None`, _except_ for the one where `use_raw=True` and x/y/color are per-obs variables, as that didn't work before this commit or after it. It's still in the parameter list but commented out for now. I'd be happy to help fix it since I've familiarized myself with this code pretty well, but we might need to discuss how this case should be handled first, and my suggestion would be to deal with this in a separate PR/bug fix/commit.

What do you think?

ivirshup:
I'm wondering if `use_raw` should just do something like:

```python
if use_raw:
    adata = adata.raw.to_adata()
```

esrice:
That's a great idea. It might require some reorganization, though, because currently `use_raw` is checked two places: once in `sc.pl.scatter()`, because it needs to know whether to look for variables in raw or not when deciding how to call `_scatter_obs()`, and again in `_scatter_obs()` itself. And it would probably be bad to do `adata = adata.raw.to_adata()` twice?

On another note, some pytests that are in files I did not edit are now failing because they can't find `anndata.tests` to import. I'm not sure if I messed something up by adding tests to `test_plotting.py` or whether this is a different issue.

esrice:
It looks like running `adata.raw.to_adata()` does not preserve the `var` table, so that won't work either:
```
>>> import scanpy as sc
>>> pbmc = sc.datasets.pbmc68k_reduced()
>>> pbmc.var.keys()
Index(['n_counts', 'means', 'dispersions', 'dispersions_norm',
       'highly_variable'],
      dtype='object')
>>> pbmc_raw = pbmc.raw.to_adata()
>>> pbmc_raw.var.keys()
Index([], dtype='object')
```

ivirshup:
> That's a great idea. It might require some reorganization, though, because currently use_raw is checked two places: once in sc.pl.scatter(), because it needs to know whether to look for variables in raw or not when deciding how to call _scatter_obs(), and again in _scatter_obs() itself.

Would it be possible to not call it again in `scatter_obs`? E.g. could `_scatter_obs` not even need to know about the `raw` field?

> On another note, some pytests that are in files I did not edit are now failing because they can't find anndata.tests to import. I'm not sure if I messed something up by adding tests to test_plotting.py or whether this is a different issue.

Aww crap, I think that was me making a new release. On the plus side it means our build system is now working as it's supposed to.

esrice:
> Would it be possible to not call it again in `scatter_obs`? E.g. could `_scatter_obs` not even need to know about the `raw` field?

Sadly not, because `to_adata()` does not preserve `var` columns (see my previous comment) :(

ivirshup:
> It looks like running adata.raw.to_adata() does not preserve the var table, so that won't work either:

That makes sense though right? Since it's a different set of variables. That said, I do suspect there is some weirdness about handling of `use_raw` here that doesn't actually make sense when you think about it, but could break some code to remove.

esrice:
Yes, I agree. My proposal would be to integrate this PR as it currently stands (after the testing system is working), as it fixes #1097 (namely, you can't currently set x, y, or color to something that is in `adata.raw.var.index` but not `adata.var.index`, even if `use_raw=True`) and leaves the logic flow for the transposition case as is. It also adds some test coverage to use cases of `sc.pl.scatter()` that were not covered before. Then, I can start working on coming up with a strategy for solving the second problem (`use_raw=True` with transposition) separately, as it's mostly unrelated, and seems to be a bit more complicated.

ivirshup:
Tests are working now

esrice:
Awesome, thanks. Do I need to do anything still?

scanpy/tests/test_plotting.py
```diff
@@ -1215,6 +1215,57 @@ def test_scatter_specify_layer_and_raw():
...
+    """Test that scatter plots vars in raw if `use_raw in [None, True]`
+
+    If `sc.pl.scatter()` receives labels of vars that are only in raw, and
+    `use_raw in [None, True]`, it should be able to successfully make a
+    scatterplot of them. If `use_raw=False`, it should not be able to make
+    a scatterplot.
+    """
+    save_and_compare_images = image_comparer(ROOT, FIGS, tol=15)
+
+    if value_error:
```
ivirshup: Could cases where this errors be a separate test instead of control flow within the test?

scanpy/tests/test_plotting.py
```diff
@@ -1215,6 +1215,57 @@ def test_scatter_specify_layer_and_raw():
...
+        # test that plotting one variable from var.index vs. another from
+        # obs.keys() works, regardless of use_raw
+        ('HES4', 'percent_mito', None, False, False),
+        ('HES4', 'percent_mito', None, True, False),
+        # test that plotting fails if one axis is a per-var value and the
+        # other is a per-obs value
+        ('HES4', 'n_cells', None, None, True),
+        ('percent_mito', 'AAAGCCTGGCTAAC-1', None, None, True),
+        # test that plotting one variable from obs.index vs. another from
+        # var.keys() works, regardless of use_raw
```
ivirshup: Should this work regardless of `use_raw`?

`pbmc` doesn't have `.var["n_cells"]`, so this should fail with a key / value error for the commented out case, no?

scanpy/tests/test_plotting.py
```diff
@@ -1215,6 +1215,57 @@ def test_scatter_specify_layer_and_raw():
...
+    return pbmc
+
+
+@pytest.mark.parametrize(
+    "x,y,color,use_raw,value_error",
+    [
+        # test that plots work with var_names only found in raw if use_raw
+        # is None or True
+        ('EGFL7', 'F12', 'FAM185A', None, False),
+        ('EGFL7', 'F12', 'FAM185A', True, False),
```
ivirshup: These should be equivalent, right?

Could cases like this instead be tested using `check_same_image` where:

```python
sc.pl.scatter(pbmc, x='EGFL7', y='F12', color='FAM185A', use_raw=None)
# is equivalent to:
sc.pl.scatter(pbmc, x='EGFL7', y='F12', color='FAM185A', use_raw=True)
# and also to:
sc.pl.scatter(pbmc.raw.to_adata() x='EGFL7', y='F12', color='FAM185A')
```

And then another case where you only use values from `obs`?

ivirshup (commented):
First, thanks for adding more tests!

Two questions:

1. Is the file `scanpy/tests/_images/scatter_filtered_genes_raw.png` meant to be here?

2. Could the tests be broken up by what they are asserting? I would prefer to break up what is being tested by test case rather than values of parameters.

 3. Could we cut down on the number of reference images generated since those cause manual maintenance burden on some matplotlib updates. These reference based tests are not great for confirming the correct plot is output, only that their output is consistent across commits.

I think some of these cases could instead be tested with `check_same_image`, e.g. where it doesn't matter whether raw is `True` or `None`. Also testing for checking cases where `use_raw=True` would be equivalent to passing `pbmc.raw.to_adata()`.

----------------

I've added some suggestions for possible changes in the review.

scanpy/tests/test_plotting.py
esrice: It does after the filtering step included in the `pbmc_filtered` fixture.

scanpy/tests/test_plotting.py
esrice: Yes, for sure. See bc949924704ff149e2677c6b68adab6c0eadb353

scanpy/tests/test_plotting.py
esrice: Yes, good idea, see bc949924704ff149e2677c6b68adab6c0eadb353

esrice: However, this does not work for `obs`, before or after this PR. This works as expected:
```python
sc.pl.scatter(pbmc, x="AAAGCCTGGCTAAC-1", y="AAATTCGATGCACA-1", use_raw=False)
```
but this doesn't:
```python
sc.pl.scatter(pbmc, x="AAAGCCTGGCTAAC-1", y="AAATTCGATGCACA-1", use_raw=True)
```
because `sc.pl.scatter()` transposes `pbmc` in both cases, but `raw` doesn't get transposed, resulting in `AttributeError: 'NoneType' object has no attribute 'obs_vector'` because `pbmc.raw` gets wiped out by the transposition.

This is a different bug than the one this PR fixes. I can make a separate issue and work on this separately if you agree that this isn't the correct behavior.

esrice:
> First, thanks for adding more tests!

Sure thing. Thanks for all the great feedback!

> 1. Is the file `scanpy/tests/_images/scatter_filtered_genes_raw.png` meant to be here?

No, thanks for catching that.

> 2. Could the tests be broken up by what they are asserting? I would prefer to break up what is being tested by test case 
rather than values of parameters.

Yes, I've broken both of the tests down into multiple tests.

> 3. Could we cut down on the number of reference images generated since those cause manual maintenance burden on some matplotlib updates. These reference based tests are not great for confirming the correct plot is output, only that their output is consistent across commits.
> I think some of these cases could instead be tested with `check_same_image`, e.g. where it doesn't matter whether raw is `True` or `None`. Also testing for checking cases where `use_raw=True` would be equivalent to passing `pbmc.raw.to_adata()`.

I've cut the number of reference images down to two. I couldn't figure out a clever way to use `check_same_image()` instead of `save_and_compare_images()` for these as you did for the others.

See below for comments about individual suggestions.

scanpy/tests/test_plotting.py
```diff
@@ -1215,6 +1215,76 @@ def test_scatter_specify_layer_and_raw():
...
+    sc.pl.scatter(pbmc_filtered, x='EGFL7', y='F12', color='FAM185A', use_raw=True)
+    plt.savefig(path2)
+    plt.close()
+
+    check_same_image(path1, path2, tol=15)
+
+    # and also to:
+    sc.pl.scatter(pbmc_filtered.raw.to_adata(), x='EGFL7', y='F12', color='FAM185A')
+    plt.savefig(path3)
+    check_same_image(path1, path2, tol=15)
```
ivirshup, line 1260:
```suggestion
    check_same_image(path1, path3, tol=15)
```

?

scanpy/tests/test_plotting.py
```diff
@@ -1215,6 +1215,76 @@ def test_scatter_specify_layer_and_raw():
...
+    # is equivalent to:
+    sc.pl.scatter(pbmc_filtered, x='EGFL7', y='F12', color='FAM185A', use_raw=True)
+    plt.savefig(path2)
+    plt.close()
+
+    check_same_image(path1, path2, tol=15)
+
+    # and also to:
+    sc.pl.scatter(pbmc_filtered.raw.to_adata(), x='EGFL7', y='F12', color='FAM185A')
+    plt.savefig(path3)
```
ivirshup:
```suggestion
    # and also to:
    sc.pl.scatter(pbmc_filtered.raw.to_adata(), x='EGFL7', y='F12', color='FAM185A')
    plt.savefig(path3)
    
    check_same_image(path1, path2, tol=15)
```

This way we'll have up to date output for path3 if the first assertion fails – would be useful info for debugging.

scanpy/tests/test_plotting.py
```diff
@@ -1215,6 +1215,76 @@ def test_scatter_specify_layer_and_raw():
...
+    pbmc = sc.datasets.pbmc68k_reduced()
+    sc.pp.filter_genes(pbmc, min_cells=10)
+    return pbmc
+
+
+def test_scatter_no_basis_raw(check_same_image, pbmc_filtered):
+    """Test scatterplots of raw layer with no basis."""
+    path1 = FIGS / "scatter_EGFL7_F12_FAM185A_rawNone.png"
+    path2 = FIGS / "scatter_EGFL7_F12_FAM185A_rawTrue.png"
+    path3 = FIGS / "scatter_EGFL7_F12_FAM185A_rawToAdata.png"
```
ivirshup:
```suggestion
def test_scatter_no_basis_raw(check_same_image, pbmc_filtered, tmpdir):
    """Test scatterplots of raw layer with no basis."""
    path1 = tmpdir / "scatter_EGFL7_F12_FAM185A_rawNone.png"
    path2 = tmpdir / "scatter_EGFL7_F12_FAM185A_rawTrue.png"
    path3 = tmpdir / "scatter_EGFL7_F12_FAM185A_rawToAdata.png"
```

Could do `tmpdir` instead of `FIGS`? Why `FIGS`?

scanpy/tests/test_plotting.py
ivirshup: Yeah, I think that should work. Would appreciate the PR!

scanpy/tests/test_plotting.py
ivirshup: Ah, gotcha

ivirshup (commented):
Looks good to me! Just a few minor changes.

Could you also add a release note?

docs/release-notes/1.8.3.rst
```diff
@@ -5,6 +5,9 @@
 
 .. rubric:: Bug fixes
 
+- Fixed finding variables with ``use_raw=True`` and ``basis=None`` in
+  :func:`scanpy.pl.scatter` :pr:`2027` :small:`I Virshup`
```
ivirshup:
```suggestion
  :func:`scanpy.pl.scatter` :pr:`2027` :small:`E Rice`
```

ivirshup (approved):

docs/release-notes/1.8.3.rst
ivirshup: @esrice, should probably be your name here!

Btw, I can't seem to commit to this branch, I believe because it's been made on your lab repo and not a personal fork. In future, I can fix small stuff like this if I'm able to commit to the branch.

esrice:
Awesome, just gave you write access to my fork.

PR #2042, opened 2021-11-03
pep-621 pyproject.toml

### LINKED ISSUE #1776: PEP 621/ 631 Metadata
Update `pyproject.toml` to use [PEP-621](https://www.python.org/dev/peps/pep-0621/) and [PEP-631](https://www.python.org/dev/peps/pep-0631/) metadata. Basically, simplify by removing most of the `tool.flit` stuff.

@flying-sheep, would you like to do this/ do you foresee any blockers?

ivirshup:
Testing if I did the pyproject.toml right for pep-621. Still need to decide if it's also worth updating it for the 1.8.x branch, or if this should wait until we're near `1.9`

Fixes #1776 if merged.


ivirshup:
This allows `pip install -e` to do a standardized thing! Unclear how much this helps/ hurts overall though, particularly for asking "what version is this installation". I think we would be able to tell that an install was "editable" through `importlib.metadata.files`.

pyproject.toml
```diff
@@ -1,136 +1,153 @@
...
+    {name = "Tom White"},
+    {name = "Malte Luecken"},
+    {name = "Davide Cittaro"},
+    {name = "Tobias Callies"},
+    {name = "Marius Lange"},
+    {name = "Andrés R. Muñoz-Rojas"},
+]
+maintainers = [
+    {name = "Isaac Virshup", email = "ivirshup@gmail.com"},
+    {name = "Philipp Angerer", email = "philipp.angerer@helmholtz-muenchen.de"},
```
flying-sheep:
```suggestion
    {name = "Philipp Angerer", email = "phil.angerer@gmail.com"},
```

flying-sheep (commented):
This looks wonderful, but I prefer single quotes in TOML, because those allow uninterpreted backslashes.

What’s the rationale behind switching to double quotes?

ivirshup:
> What’s the rationale behind switching to double quotes?

* There is a difference between the two, though I can't think of cases where it would matter off the top of my head.
* The [toml docs](https://toml.io) and flit docs use `"`. I figure they've thought about it more.
* Oh, and you can't use an apostrophe inside a single quote string.

ivirshup:
@flying-sheep, is the new build causing this to fail? I think that's what I'm seeing.

flying-sheep:
Re: quotes: Yes, the difference is that escape sequences work in double quoted strings. So for me a double quoted string in otherwise single quoted TOML means “pay attention, this one has special stuff in it”

Re: Build: The problem is that

1. we’re installing louvain and it
2. [doesn’t have a Python 3.9 wheel](https://pypi.org/project/louvain/#files), which causes us to download the sdist,
3. [Sets `2to3=True` in setup.py](https://github.com/vtraag/louvain-igraph/blob/0.7.0/setup.py#L827-L828), for which [setuptools has removed support](https://setuptools.pypa.io/en/latest/history.html#v58-0-0).

I think the best course of action would be to just port louvain to Python 3 only, and until then make sure our build environment as setuptools 57 installed. See https://github.com/vtraag/louvain-igraph/issues/57

Or we can deactivate louvain tests, skip installing it in the tests, and let people who need it deal with that.

Or we ask @vtraag to upload Python 3.9 and 3.10 wheels, then we kicked the problem back two releases.

flying-sheep (approved):
Looks great! As said, I’d personally do single quotes, but that’s a style choice.

I’m happy with the removal of `__author__` and `__email__`!

Issue #1876, opened 2021-06-15
dotplot with x axis being one variable and y axis being another variable


zhangguy:
- [x] Additional function parameters / changed functionality / changed defaults?
- [ ] New analysis tool: A simple analysis tool you have been using and are missing in `sc.tools`?
- [ ] New plotting function: A kind of plot you would like to seein `sc.pl`?
- [ ] External tools: Do you know an existing package that should go into `sc.external.*`?
- [ ] Other?

Hi,
I'm wondering if it is possible to add a new feature to sc.pl.dotplot if it is not too much of work. Say I'm interested in just one gene, and I want to plot the expression across two conditions. I understand that currently this could be achieved by using groupby = ['var1', 'var2'], but it'll be only one column, and conditions will be coerced into var1_var2. Is it possible to add a feature to the plotting function and change this behavior? I want var1 to be the x axis and var2 to be the y axis.

Thank you very much!

ivirshup:
That's an interesting idea and I see how it would be useful. I don't think it's going to be easy to implement, since I believe our code is heavily based around having groups of observations on one axis, groups of variables on the other. 

Definitely something to keep in mind for a refactor though.

maarten-hifibio:
I've created a hacky workaround by creating new DotPlot class [here](https://gist.github.com/maarten-hifibio/bccbd4436f7bd6b3fc2d218e1d766ea8).
It turned out to be strangely difficult to update the dot colors, so this workaround is essentially a matrix plot with properly sized (but uniformly colored) dots on top. 
It can be used as such:
`from _newdotplot import NewDotPlot`
`NewDotPlot(adata, var_names='CD8A', groupby=['patient','celltype'], dotcolor='powderblue', figsize=(10,10)).style(color_on='square', cmap='Reds').show()`

ivirshup:
@zhangguy, adding on to some thoughts from your PR https://github.com/theislab/scanpy/pull/2055#issuecomment-987012001

From my reading of that PR, you added a boolean argument `groupby_expand` which, when `True`, assumed `group_by` had two values: a grouping variable for the rows of the plot and a grouping variable for the columns of the plot. It also assumed `var_names` was a single variable which would be used to fill cell in the plot. As an example:

```python
pbmc = sc.datasets.pbmc3k_processed().raw.to_adata()
pbmc.obs["sampleid"] = np.repeat(["s1", "s2"], pbmc.n_obs / 2)

[...]

PR #2055, opened 2021-11-20
allowing dotplot to use two variables in groupby as x and y axis

zhangguy:
Hi,

This is for #1876 
I added a new variable `groupby_expand` to dotplot and baseplot to allow the function using two variables from groupby as x and y axis. 
Thanks.


ivirshup:
Thanks for the PR!

While I like the idea of being able to be more flexible around how the dot plots are made, I'm not sure I like the idea of adding more argument specific behavior. That is, I don't like that passing a boolean flag changes the meaning of the values for the `var_names` and `groupby` arguments. E.g. the `group_by` is the selection of `obs` corresponding to the plots rows, while `var_names` was the variable for the columns.

What could maybe be done instead is to have an argument for column grouping. Instead of

```python
sc.pl.dotplot(adata, var_names='C1QA', groupby=['louvain', 'sampleid'], groupby_expand=True)
```

It could be more like

```python
sc.pl.dotplot(adata, var_names='C1QA', group_by='louvain', group_cols='sampleid')
```

I would ideally like a solution here to work for all the other "grouped plots" as well.

I'll write a bit more about this in the parent issue.

PR #2061, opened 2021-11-30
pca_loadings add n_points to account for less than 30 features

Yves33:
When less than 30 features are present in adata.X, pca_loadings will plot some components twice.
The patch solves the problem by adding an n_point parameter (default value 30, same as anndata.ranking())that is then passed to anndata.ranking().

### DIFF
--- a/scanpy/plotting/_tools/__init__.py
+++ b/scanpy/plotting/_tools/__init__.py
@@ -92,6 +92,7 @@ def pca_loadings(
     adata: AnnData,
     components: Union[str, Sequence[int], None] = None,
     include_lowest: bool = True,
+    n_points=30,
     show: Optional[bool] = None,
     save: Union[str, bool, None] = None,
 ):
@@ -143,6 +144,7 @@ def pca_loadings(
         adata,
         'varm',
         'PCs',
+        n_points=n_points,
         indices=components,
         include_lowest=include_lowest,
     )


ivirshup:
Thanks for the PR!

Could you add a test for this?

Also, do you think the behavior should be different for when `n_points > adata.n_vars`? I'm thinking there could be a block like:

```python
if n_points is None and adata.n_vars < 30:
    n_points = adata.n_vars
elif n_points is None:
    n_points = 30
```

Btw, the tests and docs should succeed if you merge changes from master into this branch.

Yves33:
If I catch your point the function prototype would then be

```
def pca_loadings(
    adata: AnnData,
    components: Union[str, Sequence[int], None] = None,
    include_lowest: bool = True,
    n_points: Union[int, None] = None,
    show: Optional[bool] = None,
    save: Union[str, bool, None] = None,
):
```

Then 

```
if n_points is None and adata.n_vars < 30:
    n_points = adata.n_vars
elif n_points is None:
    n_points = 30

## should we also prevent user from plotting more than adata.n_vars?

else:
    n_points = min(n_points, adata.n_vars)
```
    
considering the tests, I don't see what you expect (sorry, I'm quite new to collaborative work on github):
looking at test_rankings in test_plottings.py, the test should look like
sc.pl.pca(pbmc, n_points=10)
save_and_compare_images("pca_loadings_with_10_points")

but what will it compare to?

ivirshup:
We have images which are saved which can be compared against, typically one would add a new image which you have visually confirmed to be correct, so if the behavior changes later we know. There is a little more about this in the [contribution guide](https://scanpy.readthedocs.io/en/stable/dev/testing.html#plotting-tests).

ivirshup:
?

Issue #1468, opened 2020-10-25
scanpy cannot be imported in my windows system


Sophia409:
Hi,
I couldn't import scanpy due to an error: DLL load failed. I have checked pre-existing issues, but all of them seem to be an h5py issue. My error report seems different from them.
```
>>> import scanpy as sc
D:\Anaconda\lib\site-packages\dask\config.py:168: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  data = yaml.load(f.read()) or {}
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\yuhong\AppData\Roaming\Python\Python37\site-packages\scanpy\__init__.py", line 36, in <module>
    from . import tools as tl
  File "C:\Users\yuhong\AppData\Roaming\Python\Python37\site-packages\scanpy\tools\__init__.py", line 17, in <module>
    from ._sim import sim
  File "C:\Users\yuhong\AppData\Roaming\Python\Python37\site-packages\scanpy\tools\_sim.py", line 23, in <module>
    from .. import _utils, readwrite, logging as logg
  File "C:\Users\yuhong\AppData\Roaming\Python\Python37\site-packages\scanpy\readwrite.py", line 10, in <module>
    import tables
  File "C:\Users\yuhong\AppData\Roaming\Python\Python37\site-packages\tables\__init__.py", line 99, in <module>
    from .utilsextension import (
ImportError: DLL load failed: The specified procedure could not be found.

>>> print(sys.version)
3.7.3 (default, Mar 27 2019, 17:13:21) [MSC v.1915 64 bit (AMD64)]
```

Koncopd:
Hi, if you use conda, try `conda install pytables`.
If you don't, try installing from the corresponding wheel here https://www.lfd.uci.edu/~gohlke/pythonlibs/#pytables

Sophia409:
@Koncopd Hi, Thank you for the pointer. It seems to be a problem caused by pytables package. But I still couldn't import tables after installing and uninstalling pytables packages for many times. And I'm in Windows system.
(base) C:\Users\yuhong>python
```
(base) C:\Users\yuhong>conda list | grep pytables
pytables                  3.6.1            py37h14417ae_3    conda-forge  
[...]

Issue #1387, opened 2020-08-21
Spin off read/write into it's own package?


adamgayoso:
As we are refactoring scvi, I'm wondering what the utility would be to spin off the i/o part of scanpy into it's own lightweight package that's more general for reading single cell pipeline outputs into anndata. For example, we'd like to use the scanpy read from 10x/visium functions, but don't necessarily want to have scanpy be a full requirement. It's also a bit confusing why something like `read_umi_tools` is in anndata but not `read_10x_h5`. The same goes for loom to some extent. I could imagine either moving such functionality to anndata or a standalone package that could be expanded to include support for other technologies like scATAC-seq. 

This overall could be a big benefit to methods developers who would like to build on anndata.

ivirshup:
This makes me think of a few things.

### Moving 10x reading functions to `anndata`

Initially the idea was all single cell stuff should go into scanpy. Since we read loom into anndata, I think we'd be okay with putting the 10x readers into anndata, especially if there is broad consensus this would make development of other packages easier. However, they would need to be re-written to use h5py instead of pytables.

### `scanpy` as a requirement

Is the requirement of scanpy so bad? If there are pain points here, should we be trying to make a basic scanpy installation lighter weight?

### Splitting off new modules

Finally, the idea of having IO functions go into their own package. I think this is a much bigger change, and I'd like to see a more fleshed out case for it. This would add a fair bit of complexity to development, so I'd want to be sure it's worth it. Some general questions I have:

* What are the advantages/ disadvantages of having smaller sub-packages?
    * How does this impact users vs. developers?
* Is IO special, or should more parts go into sub-packages?
* What gets re-exported from "main" modules?
* Who manages the sub-packages?

A more specific question: how modular of a component is IO? In some cases, like reading a transcriptomic only datasets, I'd say very. For more recent developments, like visium, I'm not sure this is the case. What we read in, and how we represent it, is very tightly coupled to the methods we have. Until the data's been around for a bit longer, I think it would make sense to keep visium IO in the same package as methods for it.

adamgayoso:
> Moving 10x reading functions to anndata

I haven't worked much with h5py or tables, is it time-consuming to refactor these functions? It seems like moving to anndata is the most straightforward solution at least logically to me.

> scanpy as a requirement

[...]

Issue #1284, opened 2020-06-17
ImportError: libblosc.so.1: cannot open shared object file: No such file or directory


Zifeng-L:
I installed the package by `pip install scanpy`. When I imported it, there was such an error

```python
import scanpy as sc
```

```pytb
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/sibcb1/mxqianlab1/common/anaconda/envs/cnmf_env/lib/python3.6/site-packages/scanpy/__init__.py", line 36, in <module>
    from . import tools as tl
  File "/sibcb1/mxqianlab1/common/anaconda/envs/cnmf_env/lib/python3.6/site-packages/scanpy/tools/__init__.py", line 17, in <module>
    from ._sim import sim
  File "/sibcb1/mxqianlab1/common/anaconda/envs/cnmf_env/lib/python3.6/site-packages/scanpy/tools/_sim.py", line 23, in <module>
    from .. import _utils, readwrite, logging as logg
  File "/sibcb1/mxqianlab1/common/anaconda/envs/cnmf_env/lib/python3.6/site-packages/scanpy/readwrite.py", line 10, in <module>
    import tables
  File "/sibcb1/mxqianlab1/common/anaconda/envs/cnmf_env/lib/python3.6/site-packages/tables/__init__.py", line 93, in <module>
    from .utilsextension import (
ImportError: libblosc.so.1: cannot open shared object file: No such file or directory

```

#### Versions:

> version 1.5.1

ivirshup:
Are you able to run `import tables` in this environment? If not, I think the issue will be with installation of the `pytables` package.

Issue #454, opened 2019-01-30
Windows compatibility issues with dependencies


stefanpeidli:
Hi!
As someone else posted on [stackoverflow](https://stackoverflow.com/questions/54366505/importerror-dll-load-failed-while-file-is-in-working-directory/54441575#54441575), there seem to be problems with the tables dependencies for windows users resulting in the following error when importing scanpy:

```pytb
>>> import scanpy
...
File "C:\Miniconda3\envs\py36\lib\site-packages\scanpy\readwrite.py", line 9, in
import tables
File "C:\Miniconda3\envs\py36\lib\site-packages\tables__init__.py", line 131, in
from .file import File, open_file, copy_file
File "C:\Miniconda3\envs\py36\lib\site-packages\tables\file.py", line 35, in
from . import hdf5extension
ImportError: DLL load failed: The specified procedure could not be found.
```

I've also posted an answer suggestion there. Maybe you could require h5py to have a fixed older version like 2.8 to avoid this problem for other windows users. Downgrading to that version worked for me.

flying-sheep:
hmm, I’m not sure if it’s possible to require versions depending on the platform.

where’s the h5py issue about this?

stefanpeidli:
I think the issue on h5py closest to this might be [this](https://github.com/h5py/h5py/issues/1142). 
I have not opened a seperate issue there yet. Do you think it is caused by scanpy or h5py?

flying-sheep:
definitely in pytables or h5py.

VolkerBergen:
Same issue with OSX python 3.7, solved simply with `conda install pytables`.

flying-sheep:
Pytables is in requirements.txt (the PyPI package is called “tables”), how did y’all get Scanpy installed without all its dependencies?

https://github.com/theislab/scanpy/blob/f252d3a84200cc76060a786ef0589405fc5c9c12/requirements.txt#L7

VolkerBergen:
`conda install` (not `pip`). Perhaps, that is due to pytables' conda dependencies (such as hdf).

VolkerBergen:
Just checked.. same thing applies for windows. It returns an error until you `conda install pytables`.

flying-sheep:
Hm. the conda package doesn’t list availability for windows: https://anaconda.org/bioconda/scanpy, just “conda install linux-64  v1.3.7,  osx-64  v1.3.7”

VolkerBergen:
[...]

PR #2064, opened 2021-12-01
Remove pytables dependency

ivirshup:
Removes the need for a pytables dependency

* pytables has been a source of installation heisenbugs, particularly on windows (#1468, #1284, #454)
* why use two hdf5 libraries
* Makes it easier to move reading 10x files into anndata or elsewhere #1387

I've edited the code as lightly as possible, since these readers were originally contributed by someone at 10X, so I assume they had better knowledge of possible edge cases.

- [ ] ~~Test with h5py 2~~
- [ ] Add release note


ivirshup:
I'm having trouble installing tables into a conda environment. Now may be the time to pull the trigger on this

ivirshup:
I think I would rather just bump the h5py dependency since it's been over a year.

Koncopd:
And what is the problem with 2.10.0 here?

adamgayoso:
@ivirshup any reason now that pytables is gone that we can't move these functions into anndata?

PR #2081, opened 2021-12-13
Add myself to key contributors

Zethson:
Hope you don't mind @ivirshup :)

Was thinking about putting "electronic health records" there instead of e.g. "spatial data" like Giovanni. But wasn't sure.


ivirshup:
> Was thinking about putting "electronic health records" there instead of e.g. "spatial data" 

Good question. I'm happy with this for now.

ivirshup (approved):

Issue #2051, opened 2021-11-17
Show median bars in Violin plots


brianpenghe:
I've been using violinplots to show n_genes. Is there a way to show a horizontal bar of the medians? It would be very useful to have that so that readers can have a better idea of the data quality based on median n_genes.

Thanks!

ivirshup:
Could pass `inner="quartile"`?

brianpenghe:
I tried it but it didn't work?
![image](https://user-images.githubusercontent.com/4110443/145893145-3187a154-8276-49d0-92ed-1f59c9b806a6.png)

ivirshup:
Ah, it works without `multi_panel`. Will have to look into why it doesn't with `multi_panel=True`.

ivirshup:
Looks like this not working with `multi_panel` would be introduced by #1422. At the moment my matplotlib tests are being flaky, so I may have some trouble fixing this.

PR #2083, opened 2021-12-14
Allow users to pass inner to sc.pl.violin when multi_panel=True

ivirshup:
Might fix my proposed solution for #2051

@WeilerP, do you remember why you explicitly set `inner=None` here in #1422?

### DIFF
--- a/scanpy/plotting/_anndata.py
+++ b/scanpy/plotting/_anndata.py
@@ -799,7 +799,6 @@ def violin(
             sharey=False,
             order=keys,
             cut=0,
-            inner=None,
             **kwds,
         )
 


WeilerP:
Hm, no, sorry not entirely sure. I guess it was to mimic the behavior [here](https://github.com/theislab/scanpy/blob/43379e038c5db3f917f45cc69889de0dcb6caa35/scanpy/plotting/_anndata.py#L736-L737). In the original implementation, `kwds` wasn't passed to the stripplot function (see [here](https://github.com/theislab/scanpy/blob/43379e038c5db3f917f45cc69889de0dcb6caa35/scanpy/plotting/_anndata.py#L746-L755)) but it was in #1422. From briefly skimming the docs [here](https://seaborn.pydata.org/generated/seaborn.stripplot.html) I believe `inner` and `cut` aren't valid arguments.

PR #2087, opened 2021-12-18
Add extra descriptions

flying-sheep:
So people looking into this file know what they’re for.

Looking through these, I saw that the `skmisc` isn’t named after the feature it enables. Should we add a more descriptive copy?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,15 +115,15 @@ dev = [
     "black>=20.8b1",
     "docutils",
 ]
-louvain = ["python-igraph", "louvain>=0.6,!=0.6.2"]
-leiden = ["python-igraph", "leidenalg"]
-bbknn = ["bbknn"]
-rapids = ["cudf>=0.9", "cuml>=0.9", "cugraph>=0.9"]
-magic = ["magic-impute>=2.0"]
-skmisc = ["scikit-misc>=0.1.3"]
-harmony = ["harmonypy"]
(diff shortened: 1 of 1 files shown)


ivirshup:
I feel like it's a bit of a weird include anyways.

What would you call it?

------------

Btw, some of the matplotlib tests are being flaky with the size of the plot generated. Not really sure what's up with it.

flying-sheep:
maybe `hvg-seurat`?

> Btw, some of the matplotlib tests are being flaky with the size of the plot generated. Not really sure what's up with it.

It’ so hard to do reproducible plots by pixel. Would be easier if there was a good declarative plot API for python that we used. I don’t know how we can improve this.

ivirshup:
For naming, I'd ask @adamgayoso, since he implemented it.

I'm also fine with leaving it for now and merging this as is (pending tests passing)

> Would be easier if there was a good declarative plot API for python 

Have you seen how `seaborn` does it's tests? Not so many pixel comparison ones.

adamgayoso:
> Looking through these, I saw that the skmisc isn’t named after the feature it enables. Should we add a more descriptive copy?

I don't have strong feelings. I assume most people try to run the method and then it yells at them telling them exactly what to install and how to do it. The comment is sufficient for me

ivirshup:
> The comment is sufficient for me

I'd agree. 

It's good that there's a comment in the `pyproject.toml` saying why it's there now.

PR #2089, opened 2021-12-19
Switch from sinfo to session_info

### LINKED ISSUE #1852: Use session_info instead of sinfo
[`sinfo` has been replaced](https://pypi.org/project/sinfo/) with [`session_info`](https://gitlab.com/joelostblom/session_info), which is definitely a better name. We should switch over to using this.

I think we'll be calling it like: `import session_info; session_info.show(dependencies=True, html=False, **extra_kwargs)`

flying-sheep:
Fixes  #1852. In a notebook, with `rich=None, file=None` or `rich=True`, we display the HTML version.

I also added docs: https://icb-scanpy--2089.com.readthedocs.build/en/2089/generated/scanpy.logging.print_versions.html

It’s just plain `<details>` and `<summary>` tags, no fancy widgets and should therefore survive rendering using nbsphinx or GitHub’s notebook renderer.

Looks like this:

unexpanded

after a click:

expanded once

after another click:

expanded twice


ivirshup:
I'm not sure it should have any sort of html in it, since the main utility for us is to have people copy and paste the results into their bug reports.

Thoughts?

flying-sheep:
There’s two purposes to `print_header` and `print_versions`:

1. Copy into bug reports
2. Have the info in notebook

Ideally, we’d only have one, which has all info output, is unobtrusive, and copyable. I don’t think that’s possible, as collapsed sections don’t get copied.

So maybe we should move the code I added to `print_header`, and leave `print_versions` unchanged?

scanpy/logging.py
```diff
@@ -159,28 +165,56 @@ def print_header(*, file=None):
...
+
+    Parameters
+    ----------
+    file
+        File to write output to. The default `None` means writing to `stdout`
+    rich
+        Rich HTML output. Defaults to `True` only if `file` is `None`
+        and this is called from a jupyter notebook.
+    """
+    in_notebook = 'jupyter_core' in sys.modules
```
ivirshup: Don't we already have a way of doing this? `ScanpyConfig._is_run_from_ipython()` I think?

ivirshup:
> 2. Have the info in notebook

I think I'd be happy to recommend calling `session_info` directly for this.

IIRC, we have these functions at all because a package which did a good job of displaying the imported packages and dependencies didn't really exist. 

> and leave print_versions unchanged?

With the update to use `session_info`?

I'd even be fine to deprecate the `file` argument, since it's not super useful. Plus `session_info` provides this, right?

flying-sheep:
My idea was to have `print_header` output very little, plus an expandable region (as it’s the one called in notebooks), and to revert `print_versions` to just copyable text output.

`session_info` has no `file` argument

ivirshup:
> My idea was to have print_header output very little, plus an expandable region (as it’s the one called in notebooks), and to revert print_versions to just copyable text output.

Could we deprecate `print_header` and instead suggest a way to call `session_info` for the equivalent?

If all we're doing is calling `session_info.show` with a couple default arguments, I'm not sure it's worth keeping here. I'd like users to call it directly because:

* Users get access to all of the session_info options without us having to mediate that
* If something doesn't work, it's not our problem

> session_info has no file argument

It has `write_req_file`, which is the same intent – right?

I assume `file` was there from a time when this wrote something that you could `pip install` from?

PR #2111, opened 2022-01-16
Add azure pipelines plugin

flying-sheep:
This should allow us to use the dedicated test runner UI

See: https://github.com/tonybaloney/pytest-azurepipelines


flying-sheep:
Doesn’t seem to be well maintained, too bad!

PR #2129, opened 2022-02-03
Check dendrogram fix

Koncopd:
The tests and docs now fail due to https://github.com/theislab/scanpy/issues/2125.
For some reason, this https://github.com/theislab/scanpy/blob/b69015e9e7007193c9ac461d5c6fbf845b3d6962/scanpy/tools/_dendrogram.py#L139
has small values instead of 0s on the diagonal and thus fails the check of `is_valid_dm` inside `squareform` if `checks=True`. I don't understand why small values instead of 0s are there. New version of pandas? numpy?
This fix just sets `checks=False`, not sure if it is a good idea actually.

### DIFF
--- a/scanpy/tests/test_plotting.py
+++ b/scanpy/tests/test_plotting.py
@@ -525,11 +525,11 @@ def test_violin_without_raw(tmpdir):
     pbmc = sc.datasets.pbmc68k_reduced()
     pbmc_no_raw = pbmc.raw.to_adata().copy()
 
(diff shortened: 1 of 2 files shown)


jackbrougher:
I'm having the same issue too, but it just started today after having to reinstall python + packages?

ivirshup:
It's possible this was fixed in pandas 1.4.1, released on Friday https://github.com/pandas-dev/pandas/issues/45640

PR #2120, opened 2022-01-24
Fix colorbar mappable error for older matplotlib

Koncopd:
fixes https://github.com/theislab/scanpy/issues/2118

PR #2140, opened 2022-02-15
Colorbar plus

ivirshup:
Based on #2120, but trying to get CI working.


Koncopd:
Hm, looking at the errors, they don't seem to be really related to the changes in this PR, no?

ivirshup:
Don't think so. I think these are closer to errors I was getting locally a few weeks ago, but couldn't get CI to reproduce.

PR #2170, opened 2022-03-07
modify test_read_visium_counts to test read_visium function

dineshpalli:
Here I changed the `test_read_visium_counts()` function to actually test `read_visium()` function instead of `read_10x_h5()`.

### DIFF
--- a/scanpy/tests/test_read_10x.py
+++ b/scanpy/tests/test_read_10x.py
@@ -104,9 +104,9 @@ def test_error_missing_genome():
 
 def test_read_visium_counts():
     # Test that checks the read_visium function
-    h5_pth = ROOT / '../visium_data/1.0.0'
-    spec_genome_v3 = sc.read_visium(h5_pth, genome='GRCh38')
-    nospec_genome_v3 = sc.read_visium(h5_pth)
+    visium_pth = ROOT / '../visium_data/1.0.0'
+    spec_genome_v3 = sc.read_visium(visium_pth, genome='GRCh38')
+    nospec_genome_v3 = sc.read_visium(visium_pth)
     assert_anndata_equal(spec_genome_v3, nospec_genome_v3)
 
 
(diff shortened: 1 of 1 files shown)


dineshpalli:
@giovp @ivirshup I would be glad if you could review and merge, if it passes all the checks.

scanpy/tests/test_read_10x.py
```diff
@@ -103,10 +103,10 @@ def test_error_missing_genome():
...
 
 def test_read_visium_counts():
-    # TODO: What is the purpose of this test?
-    h5_pth = ROOT / '../visium_data/1.0.0/filtered_feature_bc_matrix.h5'
-    spec_genome_v3 = sc.read_10x_h5(h5_pth, genome='GRCh38')
-    nospec_genome_v3 = sc.read_10x_h5(h5_pth)
+    # Test that checks the read_visium function
+    h5_pth = ROOT / '../visium_data/1.0.0'
+    spec_genome_v3 = sc.read_visium(h5_pth, genome='GRCh38')
+    nospec_genome_v3 = sc.read_visium(h5_pth)
```
ivirshup:
```suggestion
    visium_pth = ROOT / '../visium_data/1.0.0'
    spec_genome_v3 = sc.read_visium(visium_pth, genome='GRCh38')
    nospec_genome_v3 = sc.read_visium(visium_pth)
```

ivirshup:
Thanks for the PR! I've just renamed the variable to be a bit more clear.

I do think this test could be a bit better (e.g. check that the structure of the object is correct), but also this is an improvement so LGTM.

giovp:
@ivirshup we still keep this for legacy purposes ? or shall we not since we'll move IO somewhere else ?

ivirshup:
I thought you had mentioned moving this to squidpy? Which is fine to me to do whenever you want. Might be good to do ahead of the next visium releases?

Not sure if another package will happen or a timeline if so.

PR #2182, opened 2022-03-18
Unpin scanpydoc

flying-sheep:
Locally it works, let’s see how it shakes out here.

Fixes https://github.com/theislab/scanpydoc/issues/49

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -4,7 +4,7 @@ build:
 sphinx:
   configuration: docs/conf.py
 python:
-  version: 3.8
+  version: 3.9
   install:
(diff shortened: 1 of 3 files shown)


flying-sheep:
OK, doesn’t work here. Would be great if Sphinx would, like, show the syntax that’s broken.

/edit: found it! https://github.com/theislab/scanpydoc/pull/50

ivirshup:
Ah great! Thank you!

This should get back-ported, yeah?

ivirshup:
Well, maybe not this one. The docs on the backport branch are extra broken.

But the anndata one should get backported

PR #2185, opened 2022-03-18
Rst to myst

ivirshup:
Rst kinda sucks, let's try myst.

🤞

Will probably still need work after merging, but gets most of the docs in a much nicer state.


flying-sheep:
This messed up citations: 
grafik

will myst report broken citations or are we completely on our own in debugging things like this?

PR #2190, opened 2022-03-24
Fix default `number_of_noise_barcodes` for hashsolo

LustigePerson:
It seems like the last PR #1483 1504a36 by @njbernstein broke the way the `number_of_noise_barcodes` worked.

Like in the[ original solo repository ](https://github.com/calico/solo/blob/cd0ec6bd189fafa842af934e3d302715135bc8fc/solo/hashsolo.py#L83-L89) the `number_of_**non**_noise_barcodes` should be `2` as a default OR the `number_of_barcodes` - `number_of_noise_barcodes`. 

However, changing the default `number_of_noise_barcodes` from `None` to `2` and changing:
```python
number_of_non_noise_barcodes = (
        num_of_barcodes - number_of_noise_barcodes
        if number_of_noise_barcodes is not None
        else 2
    )
```

to 
```python
number_of_non_noise_barcodes = num_of_barcodes - number_of_noise_barcodes
```

led to the situation, that the used `number_of_non_noise_barcodes`  was `number_of_barcodes` - `2` as a default instead of `2`.

This leads to problems with demultiplexing, especially with multiple barcodes. Hence, I changed this back.

I also modified the check for a valid input to cope for the `None` default.

``` python
if number_of_noise_barcodes >= len(cell_hashing_columns):
```
changed to
```python
if (number_of_noise_barcodes is not None) and (
        number_of_noise_barcodes >= len(cell_hashing_columns)
    ):
```

The rest of changes was introduced by `black` formatting.


njbernstein:
@LustigePerson yup this was my bad. Thanks for fixing this. much appreciated

Zethson:
Thank you!

Would it be possible to catch and validate this behavior with a test here https://github.com/theislab/scanpy/blob/master/scanpy/tests/external/test_hashsolo.py ?

ivirshup:
@LustigePerson, would you be able to add a quick test here? Then this could get into the next release

LustigePerson:
I just increased the noise in the testdata. With this, the assertion fails with the wrong number of noise barcodes.
The original test data was to "clean".

ivirshup:
LGTM, thanks for the fix!

PR #2192, opened 2022-03-25
slack -> zulip

ivirshup:
Update chat link from slack to zulip


Zethson:
Maybe our README should already be more explicit about bugs/feature requests etc going to the issues and user help to the discourse?

ivirshup:
@Zethson, sounds reasonable, want to PR in something for that?

PR #2204, opened 2022-04-04
Set typehints_defaults and typehints_formatter

flying-sheep:
You already set that, so the config isn’t modified at runtime

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -59,11 +59,7 @@ extensions = [
     'sphinx_autodoc_typehints',  # needs to be after napoleon
     # 'ipython_directive',
     # 'ipython_console_highlighting',
-    # 'scanpydoc',  # scanpydoc.elegant_typehints causes full doc rebuilds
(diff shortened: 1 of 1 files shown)


ivirshup:
Checking locally, this PR causes to docs to be fully rebuilt with no other changes. I think the issue is that `typehints_formatter` is being changed after config init, not `typehints_default`.

```
updating environment: [config changed ('typehints_formatter')] 317 added, 0 changed, 0 removed
```

flying-sheep:
Hm, you’re right! looks like a Sphinx bug, as even setting it manually doesn’t change things.

Maybe pickling the function leads to it having a different object id… 

The only ways to fix this:

1. set the value to a string like `'scanpydoc.elegant_typehints:typehints_formatter'` and implement importing that object in sphinx-autodoc-typehint
2. make it so sphinx compares function-valued settings in a way that doesn’t bust the cache.

PR #2206, opened 2022-04-04
remove pre-commit workflow

flying-sheep:
We’re using pre-commit.ci


adamgayoso:
Why is this necessary? pre-commit ci uses the yaml file, also the file is good to have for local development so CI doesn't have to fix everything

flying-sheep:
Check out the PR. I’m removing the GitHub actions workflow, not pre-commit’s config file. I’m talking about having two checks that do the same thing, so let’s keep the faster one:

image

PR #2235, opened 2022-04-13
Separate test utils from tests

### LINKED ISSUE #2225: Separate testing utils from tests
Pytest supports two test layouts, in-package and outside-of-package. I prefer the outside-of-package mode outlined here: https://docs.pytest.org/en/6.2.x/goodpractices.html

Scanpy currently mixes test utils with tests, but pytest’s test files (`test_*.py` and `conftest.py`) aren’t Python modules one is supposed to import from.

To clean things up, we can refactor scanpy to a in-package structure:

- `pyproject.toml`: add `addopts = ['--import-mode=importlib']` to `[tool.pytest.ini_options]`
- `scanpy/tests/__init__.py` during implementation, make it throw an error on import so we can make sure nobody imports things from there, then delete
- `scanpy/tests/**/__init__.py` delete
- `scanpy/test_utils/` or `scanpy/testing/`
  - `__init__.py`: leave empty for now, later add public, documented test utils
  - `_private.py` add private test utils that can be imported in our tests, such as the `@needs_somepackage` decorators

Later we can decide if we want to keep the in-package layout or switch to the outside-of-package layout

flying-sheep:
This is basically the minimum amount of changes to separate things out and fix some problems with the test setup, plus unification of how we handle optional dependencies in tests.

Fixes #2225


scanpy/testing/_helpers.py
```diff
@@ -99,7 +101,11 @@ def _prepare_pbmc_testdata(sparsity_func, dtype, small=False):
     small
         False (default) returns full data, True returns small subset of the data."""
 
-    adata = pbmc3k().copy()
+    global pbmc3k
+
+    if pbmc3k is None:
+        pbmc3k = sc.datasets.pbmc3k()
+    adata = pbmc3k.copy()
```
ivirshup: I don't understand this change. I also don't think it seems relevant to moving the test utilities.

Could things like this be reverted?

scanpy/testing/_pytest/fixtures/data.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@pytest.fixture
+def krumsiek11(_krumsiek11) -> AnnData:
+    return _krumsiek11.copy()
+
+
+@pytest.fixture
+def paul15(_paul15) -> AnnData:
+    return _paul15.copy()
```
ivirshup: I think this is out of scope and a lot of repeated code.

I would like this to be removed from a pr about moving test utilities.

ivirshup (changes requested):
I've written a bit more about this in the issue thread, but I would be up for a much more narrowly scoped PR. But not something this broad.

Moving utilities out of the test suite, I'm up for. Everything else, not so much.

scanpy/testing/_helpers.py
flying-sheep: That’s the only place where making those “cached dataset” function into fixtures wasn’t beneficial. This helper can be transformed into a fixture in a later PR, it was simply more complex than the straightforward transform of the cached datasets.

scanpy/testing/_pytest/fixtures/data.py
flying-sheep: I created a submodule for marks and one for fixtures. Marks went into one, fixtures into the others. Things that are cached or recomputed across multiple tests are fixtures, so I made the cached datasets pytest fixtures.

I recreated the pattern that I was later pleased to discover was already use in `_pbmc3k_normalized` and `pbmc3k_normalized`. I just replaced the manual cache with Pytest’s session cache. I’m very happy about this change, but if you have any constructive criticism, I’m happy to hear it. (E.g. we could dynamically construct `__all__`)

flying-sheep:
@ivirshup The test failures are a bug exposed by the fixture refactoring. The tests were relying on `adata['uns']['pos']` being left over from a previous test run. Can you help me fix it?

ivirshup:
@flying-sheep, I see you're doing some updates here.

Are you doing anything to narrow the scope of the PR as requested last go-around?

flying-sheep:
I answered that back then already.

- https://github.com/scverse/scanpy/pull/2235#discussion_r850302669
- https://github.com/scverse/scanpy/pull/2235#discussion_r850304636

This PR has exactly the scope necessary to separate test utils and tests.

The only thing that I can think of to add to those answers is that the repeated code for all the data fixtures is necessary to make editors understand them. A more dynamic way to make all those fixtures breaks ctrl/cmd-clicking fixtures.

scanpy/testing/_pytest/fixtures/__init__.py
```diff
@@ -7,6 +7,38 @@
...
+
+__all__ = [
+    'array_type',
+    'float_dtype',
+    '_pbmc3k',
+    '_pbmc3k_normalized',
+    '_pbmc3k_processed',
+    '_pbmc68k_reduced',
+    '_krumsiek11',
+    '_paul15',
```
ivirshup: Why export these?

scanpy/testing/_pytest/fixtures/__init__.py
```diff
@@ -7,6 +7,38 @@
...
+    'pbmc3k',
+    'pbmc3k_normalized',
+    'pbmc3k_processed',
+    'pbmc68k_reduced',
+    'krumsiek11',
+    'paul15',
```
ivirshup: I prefer these as functions over fixtures. I don't think I see a ton of advantage to them being fixtures, but do think the idiom they introduce:

```python
def test_*(dataset):
    adata = dataset
    ...
```

Isn't great, and could be error prone, since it creates two variables pointing to the same object.

ivirshup: Similar to https://github.com/scverse/scanpy/pull/2235/files#r849662166

scanpy/testing/_pytest/marks.py
```diff
@@ -0,0 +1,20 @@
...
+    return pytest.mark.skipif(not find_spec(mod), reason=reason)
+
+
+needs_leidenalg = make_skip_mark("leidenalg")
+needs_louvain = make_skip_mark("louvain")
+needs_skmisc = make_skip_mark('skmisc', 'scikit-misc')
+needs_scanorama = make_skip_mark("scanorama")
+needs_scrublet = make_skip_mark("scrublet")
+needs_fa2 = make_skip_mark("fa2")
+needs_igraph = make_skip_mark("igraph", "python-igraph")
```
ivirshup: Do we need to define a mark for each package? Could we instead have one mark that we pass the package name to?

Also some documentation would be great.

scanpy/tests/test_paga.py
```diff
@@ -1,61 +1,63 @@
...
-        (
-            "master_paga_compare_continuous",
-            partial(sc.pl.paga_compare, color='CST3', legend_fontsize=5),
-        ),
-        (
-            "master_paga_compare_pca",
-            partial(sc.pl.paga_compare, basis='X_pca', legend_fontweight='normal'),
-        ),
-    ],
+    [pytest.param(p.id, *p.values, id=p.id, marks=p.marks) for p in paga_plot_tests],
```
ivirshup: Why?

scanpy/testing/_helpers.py
ivirshup: So, if making the cached datasets into fixtures is reverted, this can be as well?

I'm not sure I like the cached datasets as fixtures (mentioned in other comments).

pyproject.toml
```diff
@@ -150,8 +150,12 @@ exclude = [
 ]
 
 [tool.pytest.ini_options]
-python_files = "test_*.py"
-testpaths = "scanpy/tests/"
```
ivirshup, line 154: This prevented tests being found in `build`, what do we get from removing it?

pyproject.toml
```diff
@@ -150,8 +150,12 @@ exclude = [
...
+addopts = [
+    "--import-mode=importlib",
+    "-pscanpy.testing._pytest",
+]
+norecursedirs = ["docs"]
+xfail_strict = true
```
ivirshup: Could you explain these changes?

In particular, why they are needed for moving test utilities outside of the tests.

ivirshup (changes requested):

scanpy/testing/_pytest/fixtures/__init__.py
flying-sheep: So the `from .fixtures import *` in the parent module imports all of them. All used fixtures need to be in scope, which includes these, and having them all in the `scanpy.testing._pytest` module means that we can use that module as a pytest plugin.

Furthermore, the leading `_` means “use with care”, not “never use”. If you e.g. want to create your own session-scoped fixture, you can use these like here:

https://github.com/scverse/scanpy/blob/5ce5de70470484ca3e38a13f578b7d509d0b2171/scanpy/tests/test_paga.py#L17-L27

scanpy/testing/_pytest/marks.py
flying-sheep: They’re defined like this to be reused. They need to be reused because some module names don’t match the distribution/PyPI names.

What kind of documentation do you mean? A page in the “Contributing / Tests” section explaining what marks and fixtures we have? If you mean a docstring or comment, then I’d say that I think the skip marks are pretty self explanatory.

scanpy/tests/test_paga.py
flying-sheep: so there is an association between a parametrized test name (e.g. `scanpy/tests/test_paga.py::test_paga_plots[continuous_obs]`) and the file name (e.g. `master_paga_continuous_obs.png`)

otherwise `test_paga_plots[func4]` fails and you have to count the list items to know which one that was

scanpy/testing/_pytest/marks.py
ivirshup: > They’re defined like this to be reused.

Sure, but I would prefer something like: `@skip_if_uninstalled("scanorama")`. It's easier to see how to extend that to new cases.

------

A docstring saying why we're not using `importorskip` would be good.

There is also already a doc page for more general notes: https://scanpy.readthedocs.io/en/stable/dev/testing.html

scanpy/testing/_pytest/fixtures/__init__.py
flying-sheep: I disagree, fixtures are perfect for this exact use case. Implementing our own in-memory dataset caching method to get the same performance and ergonomics with functions is reinventing the wheel when there’s a perfectly good wheel factory already built.

We can find another way to get rid of that idiom if it bothers you. I’d also be OK with fixtures that resolve to a function instead of importing

pyproject.toml
flying-sheep: Simplicity. `build` is no longer a thing since we got rid of setuptools.

But we could replace those two lines with `testpaths = "scanpy"` to mirror how we do things in anndata, and enable doctests in the same way.

scanpy/tests/test_paga.py
ivirshup: Can you show me where this happens?

If I collect the tests on master, I see the `test_id` showing up in the collected name:

<details>
<summary> </summary>

```python
$ pytest -k test_paga -v
=================================================== test session starts ====================================================
platform darwin -- Python 3.9.16, pytest-7.2.0, pluggy-1.0.0 -- /usr/local/opt/python@3.9/bin/python3.9
cachedir: .pytest_cache
rootdir: /Users/isaac/github/scanpy, configfile: pyproject.toml, testpaths: scanpy/tests/
plugins: xdist-2.5.0, memray-1.4.0, assume-2.4.3, nunit-1.0.0, sphinx-0.5.0, profiling-1.7.0, anyio-3.6.2, forked-1.3.0, cov-3.0.0
collected 977 items / 965 deselected / 7 skipped / 12 selected                                                             

scanpy/tests/test_paga.py::test_paga_plots[master_paga-paga] PASSED                                                  [  8%]
scanpy/tests/test_paga.py::test_paga_plots[master_paga_continuous-func1] PASSED                                      [ 16%]
scanpy/tests/test_paga.py::test_paga_plots[master_paga_continuous_obs-func2] PASSED                                  [ 25%]
scanpy/tests/test_paga.py::test_paga_plots[master_paga_continuous_multiple-func3] PASSED                             [ 33%]
scanpy/tests/test_paga.py::test_paga_plots[master_paga_compare-func4] PASSED                                         [ 41%]
scanpy/tests/test_paga.py::test_paga_plots[master_paga_compare_continuous-func5] PASSED                              [ 50%]
scanpy/tests/test_paga.py::test_paga_plots[master_paga_compare_pca-func6] PASSED                                     [ 58%]
scanpy/tests/test_paga.py::test_paga_pie PASSED                                                                      [ 66%]
scanpy/tests/test_paga.py::test_paga_path PASSED                                                                     [ 75%]
scanpy/tests/test_paga.py::test_paga_compare PASSED                                                                  [ 83%]
scanpy/tests/test_paga.py::test_paga_positions_reproducible PASSED                                                   [ 91%]
```

</details>

pyproject.toml
flying-sheep: - `"--import-mode=importlib"` makes sure we don’t mix `pytest` idioms with imports from test directories. That gets ugly. This PR is there to fix that, that option is there to enforce the separation. It‘s recommended, we use it in our template, …
- `"-pscanpy.testing._pytest"` as described in https://github.com/scverse/scanpy/pull/2235#discussion_r1234145378, this module is a pytest plugin and contains pytest customizations we use. It replaces the use of a top-level `conftest`
- `norecursedirs = ["docs"]` is the inversion of `testpaths`: blocklist instead of allowlist. I’m happy to change that one to `testpaths = ['scanpy']`
- `xfail_strict = true` once tests we expect to fail suddenly work, we need to check why.

scanpy/testing/_helpers.py
flying-sheep: Sure, if we go that route. But I’m pretty sure fixtures are the better way.

scanpy/testing/_pytest/marks.py
flying-sheep: And `skip_if_uninstalled` uses a hardcoded list of all modules and refuses to work with others so we can make sure the error message names the correct thing? Sure that’d also work. I guess I don’t see the advantage of that compared with a bunch of marks, though.

> A docstring saying why we're not using `importorskip` would be good.

makes sense, done.

flying-sheep: Why do you think a `skip_if_uninstalled` is better than the individual marks?

scanpy/tests/test_paga.py
flying-sheep: Hm, you’re right, they’re both there. Will revert

ivirshup:
Could the paga fix be a separate PR, or was it somehow connected to all this?

flying-sheep:
It’s connected because the paga tests don’t `copy` the objects and therefore run sequentially, but I can extract it.

ivirshup:
Extracting it would be great.

What triggered the test to fail on other branches, I'd assume a matplotlib update?

flying-sheep:
I think so too. fixed in https://github.com/scverse/scanpy/pull/2521

scanpy/testing/_helpers.py
flying-sheep: OK, fixturized this too, much better now

scanpy/testing/_pytest/fixtures/__init__.py
flying-sheep: yeah, I just simplified the code talked about in https://github.com/scverse/scanpy/pull/2235#discussion_r849659638 by making it into fixtures. definitely better. We can reorg them or make them return functions, but they should definitely be fixtures. I just realized we can do

```py
@pytest.fixture
def krumsiek11(_krumsiek11) -> Callable[[], AnnData]:
    return _krumsiek11.copy

def test_something(krumsiek11):
    adata = krumsiek11()
    ...
```

flying-sheep: I’m not 100% sure if that’s better, see e.g.:

https://github.com/scverse/scanpy/blob/6bac462b37b97504ce5e43ef8f875cc55358b8f7/scanpy/tests/test_pca.py#L115-L123

Here we reuse the same variable carefully anyway. With fixtures returning functions, that’d look like

```py
def test_pca_chunked(pbmc3k_normalized): 
    pbmc_full = pbmc3k_normalized()
    pbmc = pbmc_full[::6].copy() 
    pbmc.X = pbmc.X.astype(np.float64) 
    chunked = sc.pp.pca(pbmc_full, chunked=True, copy=True) 
    default = sc.pp.pca(pbmc_full, copy=True) 
```

or so, confusing the reader with multiple names for the same thing

flying-sheep: Let’s discuss this in https://github.com/scverse/scanpy/pull/2522

scanpy/testing/_pytest/marks.py
flying-sheep: OK, so it’s impossible to use [`importlib.metadata.packages_distributions`](https://docs.python.org/3/library/importlib.metadata.html#mapping-import-to-distribution-packages) since we’re dealing with things that *aren’t* installed. So unless we want our tests to talk to the PyPI API, we should hardcode those.

flying-sheep:
OK, everything done as we agreed on yesterday:

- only parametrized datasets are fixtures
- the skip mark is now a single function.

Issue #2254, opened 2022-05-12
Error when repeating the tutorial for diffusion map in v1.9.1 scanpy


mmyoung:
- [ ] I have checked that this issue has not already been reported.
- [ ] I have confirmed this bug exists on the latest version of scanpy.
- [ ] (optional) I have confirmed this bug exists on the master branch of scanpy.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

### This is the code I'm using which was copied from the tutorial:

```python
import scanpy as sc
adata = sc.datasets.paul15()
sc.pp.log1p(adata)
sc.pp.neighbors(adata, n_neighbors=20, use_rep='X', method='gauss')
sc.tl.diffmap(adata)
sc.tl.dpt(adata, n_branchings=1, n_dcs=10)
sc.pl.diffmap(adata, color=['dpt_pseudotime', 'dpt_groups', 'paul15_clusters'])
```
The results I got are:
![image](https://user-images.githubusercontent.com/33322882/168080076-f6767970-0b3b-4623-ab25-e9e38201e6ef.png)

Which is totally different than in the tutorial:
![image](https://user-images.githubusercontent.com/33322882/168081015-cddf77eb-7b83-4802-8890-3b0d87a5755d.png)

Can anyone run into this problem ever? Thanks for your help.

#### Versions

<details>
-----

anndata     0.8.0

scanpy      1.9.1

-----

PIL                         8.4.0

backcall                    0.2.0

beta_ufunc                  NA

binom_ufunc                 NA

bottleneck                  1.3.2

cffi                        1.15.0

colorama                    0.4.4

cycler                      0.10.0

cython_runtime              NA

dateutil                    2.8.2

debugpy                     1.5.1

decorator                   4.4.2

defusedxml                  0.6.0

google                      NA

h5py                        3.6.0

ipykernel                   6.4.1

ipython_genutils            0.2.0

jedi                        0.18.1

joblib                      0.17.0

kiwisolver                  1.3.1

llvmlite                    0.38.0

...

Python 3.9.7 (default, Sep 16 2021, 13:09:58) [GCC 7.5.0]

Linux-3.16.0-11-amd64-x86_64-with-glibc2.19

-----

Session information updated at 2022-05-12 14:59

</details>

Koncopd:
[...]

Issue #2227, opened 2022-04-06
Bug in diffusion map plotting in 1.9.1?


cbravo93:
- [ x] I have checked that this issue has not already been reported.
- [ x] I have confirmed this bug exists on the latest version of scanpy.
- [ ] (optional) I have confirmed this bug exists on the master branch of scanpy.

---

Hi!

When running `sc.pl.diffmap` with 1.8.2 in my object I get a nice diffusion map plot:
<img width="268" alt="image" src="https://user-images.githubusercontent.com/19205518/162031248-89a4da34-8f1d-48b2-9b86-e04cf881a3c0.png">
However, when using the same function with 1.9.1 in exactly the same object (with exactly the same 'X_diffmap') I get an awful plot:
<img width="275" alt="image" src="https://user-images.githubusercontent.com/19205518/162031764-12a839d3-7022-48a7-b4d2-d64b5d50b10a.png">

Any idea what is going on? This only affects plotting (also tried using sc.pl.embedding), the dpt function gives the same results for example.

Thanks!

C

mmyoung:
I am also facing the same issue here, have you got any solutions yet? Thanks.

cbravo93:
I just reverted to scanpy 1.8.2 for now . C

Koncopd:
@cbravo93 @mmyoung a quick fix is to do
`adata.obsm["X_diffmap_"] = adata.obsm["X_diffmap"][:, 1:] `
`sc.pl.embedding(adata, "diffmap_")`

RainboWu:
Thanks for @Koncopd 's advice

PR #2255, opened 2022-05-12
Fix diffmap plotting

Koncopd:
Fixes
https://github.com/scverse/scanpy/issues/2254
https://github.com/scverse/scanpy/issues/2227

Before there was a hard coded offset for the diffmap basis (https://github.com/scverse/scanpy/blob/8736fc3b02a6af3b7214481c2cf78a3f8f5e06bd/scanpy/plotting/_tools/scatterplots.py#L1077) and later it disappeared. This restores the offset for diffmap, and also provides an option in `embedding` to select basis columns before `components`, `dimensions `etc.


Zethson:
@Koncopd can you get the CI to pass, please? Happy to review then.

Koncopd:
I am not sure, what happens with CI. I don't think it is because of the changes in this PR.

ivirshup:
@Koncopd, there was a bugged release of pip. I think this should work now that there has been a bugfix

Koncopd:
Ok, i will merge and check, thx.

Koncopd:
@giovp Could you check why the visium test fails? I don't think it is related to this PR.

giovp:
ah it seems a really minor tolerance thingy, try to increase it to 16 should be fine
```
save_and_compare_images = image_comparer(ROOT, FIGS, tol=15)
```

ivirshup (changes requested):
@Koncopd, got CI working and can review.

It looks like what happened is I forgot to add the dimension handling code the the diffmap wrapper.

The idea here is that any diffmap specific code should go in the diffmap wrapper function, not the embedding function. I think this can largely be solved by setting default arguments for diffmap (POC in #2280).

dburkhardt:
@ivirshup is this the accepted solution for https://github.com/scverse/scanpy/pull/2280?

Just trying to figure out if we can close the PR and fix the bug.

PR #2272, opened 2022-06-14
Mask for pca, normalize_pearson_residuals_pca, scatterplot, scale

chelseabright96:
mask parameter added to pca method in _pca.py 
test_pca_mask added to test_pca.py
Deprecation warning on use_highly_variable parameter added to test_deprecations.py


scanpy/preprocessing/_pca.py
```diff
@@ -230,9 +277,10 @@ def pca(
         adata.uns['pca'] = {}
         adata.uns['pca']['params'] = {
             'zero_center': zero_center,
+            'mask': use_existing_mask if use_existing_mask is not None else None,
             'use_highly_variable': use_highly_variable,
         }
-        if use_highly_variable:
+        if mask is not None:
```
ivirshup, line 300: Looks like the mask needs to be used here

scanpy/preprocessing/_pca.py
```diff
@@ -122,25 +128,66 @@ def pca(
...
+        adata = data
+    else:
+        adata = AnnData(data, dtype=data.dtype)
+
+    # Check for use_highly_varible
     if use_highly_variable is True and 'highly_variable' not in adata.var.keys():
+        warn(
+            "Argument `use_highly_variable` is deprecated, consider using a mask argument.",
+            FutureWarning,
+        )
```
ivirshup: Instead of having this in multiple places, I would do something like:

```python
# Backwards compat for use_highly_variable
if use_highly_variable is not None:
    warn(
        ...,  # use mask warning
        FutureWarning
    )
if use_highly_variable and mask is not None:
    raise ValueError(
        ... # these arguments are incompatible 
    )
elif use_highly_variable or (use_highly_variable is None):
    mask = "highly_variable"
```

and then only use the mask argument from here down. This would remove the `ValueError` below (`"Did not find adata.var[\'highly_variable\']..."`)

scanpy/preprocessing/_pca.py
```diff
@@ -22,6 +22,8 @@ def pca(
     random_state: AnyRandom = 0,
     return_info: bool = False,
     use_highly_variable: Optional[bool] = None,
+    use_existing_mask: Optional[str] = None,
+    mask: Optional[np.ndarray] = None,
```
ivirshup: As discussed, I think these can be collapsed into a single argument. I've expanded on this in the body of the review.

scanpy/preprocessing/_pca.py
```diff
@@ -122,25 +128,66 @@ def pca(
...
+    if mask is not None:
+        if len(mask) != len(adata.var):
+            raise ValueError(
+                'The mask you want to use does not fit the dimensions of your data '
+            )
+        adata_comp = adata[:, mask]
+        logg.info('     on specified mask')
+        if use_existing_mask is None:
+            adata.var['mask_used_for_PCA'] = mask
+            use_existing_mask = 'mask_used_for_PCA'
```
ivirshup: I wouldn't add this to the anndata object. Users can just do this themselves if they wanted.

scanpy/tests/test_deprecations.py
```diff
@@ -17,3 +17,10 @@ def test_deprecate_multicore_tsne():
 
     with pytest.warns(UserWarning, match="Falling back to scikit-learn"):
         sc.tl.tsne(pbmc, use_fast_tsne=True)
+
+
+def test_deprecate_use_highly_variable_genes():
+    pbmc = pbmc68k_reduced()
+
+    with pytest.warns(FutureWarning, match="is deprecated"):
+        sc.pp.pca(pbmc, use_highly_variable=True)
```
ivirshup: Could this be more specific? E.g. have the match contain "use_highly_variable"

Wouldn't want to catch some other deprecation warning.

scanpy/tests/test_pca.py
```diff
@@ -153,3 +157,42 @@ def test_pca_n_pcs(pbmc3k_normalized):
...
+        sc.pp.pca(adata, use_highly_variable=True)
+
+    # check warning on mask length
+    mask = np.random.choice([True, False], pbmc.shape[1] + 1)
+    with pytest.raises(ValueError):
+        sc.pp.pca(pbmc, mask=mask, copy=True, dtype=float_dtype)
+
+    # Test if pca result is equal when given mask vs. given use_highly_variable=True
+    mask = pbmc.var['highly_variable']
+    adata_mask = sc.pp.pca(pbmc, mask=mask, copy=True, dtype=float_dtype)
```
ivirshup: It would be good to have tests for masks under different keys, since this could work if the array wasn't actually passed. Instead, maybe something like:

```python
pbmc = datasets.pbmc3k_processed.raw.to_adata()
mask = np.random.choice([True, False], pbmc.shape[1] + 1)
pbmc_w_mask = pbmc.copy()
pbmc_w_mask.var["mask"] = mask

sc.pp.pca(pbmc, mask=mask)
sc.pp.pca(pbmc_w_mask, mask="mask")
```

Then compare the values of `pbmc` and `pbmc_w_mask`.

ivirshup (dismissed):
Thanks for getting this together!

# Questions

If we want the default to be using `highly_variable` if it's present, how do we let people not use a mask if `"highly_variable"` is in `var`?

I'm thinking the default may need to be the `_Empty` singleton, so we can explicitly check if `mask=None` was passed.

# Requested changes

## Collapse mask arguments

Instead of having two mask related arguments, you can collapse them into one:

```
    mask: Union[np.ndarray, str, None] = None,
```

Which is validated via some function, kinda like:

```python
def check_mask(
    adata: AnnData,
    mask: Union[str, np.ndarray],
    axis: int = 0,
) -> np.ndarray:  # Could also be a series, but should be one or the other
    """
    Validate mask argument

    Params
    ------
    adata
    mask
        The mask. Either an appropriatley sized boolean array, or name of a column which will be used to mask.
    axis
        The axis being masked
    """
    if isinstance(mask, str):
        annot = ("obs", "var")[axis]
        mask_array = annot[mask].values
    else:
        if len(mask) != adata.shape[axis]:
            raise ValueError(
                ... # Shapes don't match
            )
        mask_array = mask

    if not pd.api.types.is_bool_dtype(mask_array):
        raise ValueError(
            ... # mask array must be boolean, was {whatever dtype is was}
        )

    return mask_array
```

## Tests

I think a number of the tests may have broken from this PR since the logic for the masks isn't quite right. A good way to help debugging this is to split up the cases from the mask test into multiple ones. E.g., one that checks for the default behavior.

## Extra data files

Some extra data files got added. Could you remove those?

## Co-authorship

@tothmarcella should be a co-author on this PR, right? Could you make a commit with her listed as a co-author? [Here're some docs on how to do that.](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)

scanpy/preprocessing/_pca.py
```diff
@@ -142,52 +141,26 @@ def pca(
...
         )
+    if use_highly_variable and mask is not None:
         raise ValueError(
-            'Did not find adata.var[\'highly_variable\']. '
-            'Either your data already only consists of highly-variable genes '
-            'or consider running `pp.highly_variable_genes` first.'
+            'These arguments are incompatible.'
+            'Use_highly_variable can be called through mask="highly_variable"'
         )
+    elif use_highly_variable or (use_highly_variable is not None):
```
ivirshup: This check simplifies to `use_highly_variable is not None`, and would return `True` if `use_highly_variable=False`

tothmarcella:
> # Questions
> If we want the default to be using `highly_variable` if it's present, how do we let people not use a mask if `"highly_variable"` is in `var`?
> 
> I'm thinking the default may need to be the `_Empty` singleton, so we can explicitly check if `mask=None` was passed.
> 
Hi, the option of not using a mask when `highly_variable` is in `var` is still missing. We understand the idea but could you please help with how to create a singleton (and where)?

scanpy/preprocessing/_simple.py
```diff
@@ -754,7 +755,9 @@ def scale(
         raise ValueError(f"`layer` argument inappropriate for value of type {type(X)}")
     if obsm is not None:
         raise ValueError(f"`obsm` argument inappropriate for value of type {type(X)}")
-    return scale_array(X, zero_center=zero_center, max_value=max_value, copy=copy)
+    return scale_array(
+        X, zero_center=zero_center, max_value=max_value, copy=copy, mask=mask
+    )
```
totmar101, line 764: Hi, I have a quick question about the scale function. If I understood correctly how `singledispatch` works, this part of the code would only run when the data given is not an `np.ndarray`, nor a `sparse matrix`, nor `anndata` (because of the later defined functions). For what situation would this main part run for which it would be beneficial? I tried and I can see for example for a string argument it would only fail inside the `scale_array` function. Is it useful if someone gives an array-type object so it would still compute the scaling?

flying-sheep: The main part runs for anything that *doesn’t* have a registered function, it’s the fallback case.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -324,6 +340,11 @@ def embedding(
                 vcenter_float,
                 norm_obj,
             )
+
+            # Map nan values to na_color
+            current_cmap = colormaps.get_cmap(color_map)
+            current_cmap.set_bad(color=na_color)
```
flying-sheep: This seems to serve no purpose …

flying-sheep:
Possible TODO:

- normalize_pearson_residuals_pca

@ivirshup I reverted the change in a6290ee9e0d1baf0e3483118aa552b6f6dcf02c0 where you changed

```diff
-X_pca = np.zeros((X.shape[0], n_comps), X.dtype)
+X_pca = np.zeros((adata_comp.shape[0], n_comps), adata.X.dtype)
```

the commit message is “Fix up pca tests”, but that change doesn’t seem to impact tests and it takes properties from several different object without reasoning.

scanpy/plotting/_tools/scatterplots.py
flying-sheep: removed

PR #2274, opened 2022-06-15
Fix tests

ivirshup:
Some tests are broken, here are some fixes for that

* Visium reference test was just wrong, dots should have been transparent.
* igraph 0.9.11 has different results for `"fr"` layout, updated tests accordingly (https://github.com/igraph/python-igraph/issues/545)
* 32 bit PCA computation is now accurate within 32 bit rounding, so tests checking for random seed changes have been disabled


ivirshup:
Some clustering results are changing, but I can't reproduce locally 😢.

So I'm going to chuck a bandaid on it, since I expect default clustering results to change when we use igraph leiden.

PR #2283, opened 2022-06-20
Fix rank genes groups when log base is missing

ivirshup:


scanpy/tools/_rank_genes_groups.py
```diff
@@ -90,7 +90,10 @@ def __init__(
         comp_pts=False,
     ):
 
-        if 'log1p' in adata.uns_keys() and adata.uns['log1p']['base'] is not None:
+        if (
+            'log1p' in adata.uns_keys()
```
DriesSchaumont: I think this check should also take into account the 'layer' argument because this will only work if you want to use `.X`

flying-sheep:
superseded by #2546

PR #2295, opened 2022-07-18
pheno_jaccard_ig to be a csr matrix

dshefter:


Zethson:
Hi,

could you please provide more details? What issues did you run into?

dshefter:
Hi,

The issues I was mostly running into were that when saving the anndata
variable as a h5ad file, 'pheno_jaccard_ig' was not compatible with this
action. So, I had to either remove pheno_jaccard_ig from the anndata object
and then save it as h5ad or convert it to a sparse matrix. This also
happened with a few other functions I tried on the anndata object, and I
kept getting the error "this function is not compatible with COO matrix
format", always talking about pheno_jaccard_ig. Therefore, since a sparse
matrix object does not have any problems with the functions I was running
on adata, changing pheno_jaccard_ig to a sparse matrix from the start makes
sense to circumvent any of those issues I was getting before.

I hope this makes sense.
Thank you,
Deena Shefter

scanpy/external/tl/_phenograph.py
```diff
@@ -243,7 +244,7 @@ def phenograph(
     if copy:
         return communities, graph, Q
     else:
-        adata.obsp[ig_key] = graph
+        adata.obsp[ig_key] = sp.sparse.csr_matrix(graph)
```
adamgayoso:
```suggestion
        adata.obsp[ig_key] = graph.tocsr()
```

adamgayoso (changes requested):

PR #2296, opened 2022-07-19
Stephen/spaceranger2.0

stephenwilliams22:
This PR updates scanpy to accommodate the latest changes made in spaceranger 2.0 which will break the released version of scanpy. Will provide backwards compatibility to pre 2.0 releases.


Zethson:
Thank you! Maybe also relevant for @giovp ?

stephenwilliams22:
@Zethson just wanted to see if there were any issues with merging this PR.

scanpy/readwrite.py
```diff
@@ -410,7 +415,10 @@ def read_visium(
         }
 
         # read coordinates
-        positions = pd.read_csv(files['tissue_positions_file'], header=None)
+        positions = pd.read_csv(
+            files['tissue_positions_file'],
+            header=1 if tissue_positions_file.name == "tissue_positions.csv" else None,
+        )
         positions.columns = [
```
giovp, line 422: There is one missing edit (which we fixed in squidpy https://github.com/scverse/squidpy/blob/fb069ded7515e8a1386224d32344a8657cbecd2e/squidpy/read/_read.py#L78-L89 ) that is, the order of the columns is different.

giovp (changes requested):
hi @stephenwilliams22 , thanks for this PR, only one minor thing.

scanpy/readwrite.py
stephenwilliams22: @giovp this was actually a typo in the spaceranger documentation early on which has been fixed for some time. The order of the column names has always been `barcode,in_tissue,array_row,array_col,pxl_row_in_fullres,pxl_col_in_fullres`  no matter the version of spaceranger.

stephenwilliams22: Very sorry for the mixup! This should be updated in squidpy as well.

stephenwilliams22:
@giovp looking at this again, it seems you drop the mixed up columns anyway (https://github.com/scverse/squidpy/blob/fb069ded7515e8a1386224d32344a8657cbecd2e/squidpy/read/_read.py#L100) so I'd suggest having one code path with the correct column names. Please let me know if you need more clarity and help getting this over the hump for release.

giovp:
> @giovp looking at this again, it seems you drop the mixed up columns anyway ([scverse/squidpy@`fb069de`/squidpy/read/_read.py#L100](https://github.com/scverse/squidpy/blob/fb069ded7515e8a1386224d32344a8657cbecd2e/squidpy/read/_read.py#L100)) so I'd suggest having one code path with the correct column names. Please let me know if you need more clarity and help getting this over the hump for release.

Hi @stephenwilliams22 ,

we drop it from obs since we don't want to save spatial coordinates there, cause they should be only in obsm. 

So the only changes between space rangers version is the name of the tissue position file correct? but the order of the columns remains the same?

ivirshup:
@giovp, should this change happen in scanpy or squidpy?

giovp:
> @giovp, should this change happen in scanpy or squidpy?

I would make it happen in both and deprecate this function from the next release (as well as all the other spatial).

ivirshup:
Has this happened in squidpy already?

giovp:
> Has this happened in squidpy already?

yes, need one further clarification from @stephenwilliams22 and then will resolve also in squidpy.

stephenwilliams22:
> > @giovp looking at this again, it seems you drop the mixed up columns anyway ([scverse/squidpy@`fb069de`/squidpy/read/_read.py#L100](https://github.com/scverse/squidpy/blob/fb069ded7515e8a1386224d32344a8657cbecd2e/squidpy/read/_read.py#L100)) so I'd suggest having one code path with the correct column names. Please let me know if you need more clarity and help getting this over the hump for release.
> 
> Hi @stephenwilliams22 ,
> 
> we drop it from obs since we don't want to save spatial coordinates there, cause they should be only in obsm.
> 
> So the only changes between space rangers version is the name of the tissue position file correct? but the order of the columns remains the same?

Yes, this is correct.

stephenwilliams22:
@giovp do we have an update on merging this yet? Anything left on my end?

ivirshup:
@giovp, @stephenwilliams22, @Zethson

This PR seems to have broken a number of tests. I believe due to the change in `pixel_row`/ `pixel_col` order, resulting in a number of test plots now being transposed.

If this was always wrong... surely usage would have caught that, so we must be correcting for this somewhere else.

What's the correction here? And is it something that should be done quickly, or should I revert this PR until we can figure it out?

giovp:
from discussing with @stephenwilliams22 this made me think there shouldn't be changes with column names yet it seems like it is, 

> > So the only changes between space rangers version is the name of the tissue position file correct? but the order of the columns remains the same?
> 
> Yes, this is correct.

we actually might have had similar reports in squidpy as well, let me ask @LLehren for support, will probably have to fix it in squidpy side as well.

giovp:
for reference https://github.com/scverse/squidpy/issues/599

ivirshup:
Alright, I'm going to revert for now then

PR #2183, opened 2022-03-18
Inherit layer argument in _highly_variable_genes_single_batch when using sc.pp.highly_variable_genes

Ngort:
Quite a simple addition meant to fix a bug when one works with specific layers.

PR #2302, opened 2022-07-29
Fix using custom layer with highly_variable_genes

DriesSchaumont:
This PR takes off where https://github.com/scverse/scanpy/pull/2183 went stale.


Zethson (approved):
LGTM, thank you!

PR #2326, opened 2022-09-14
fix broken references in the Tutorials section

bpkwee:


Zethson:
@ivirshup is this the way to go? Think that we want to stick with the citations, no?

adamgayoso:
we should change to sphinx contrib bib extension

ivirshup:
We should switch to the sphinx bibtex thing, but IIRC I ran into difficulty here since there wasn't a standardized format for the current reference set.

PR #2355, opened 2022-10-19
add scalex[Xiong22 Nature Communications] to scanpy.external.pp

jsxlei:
Hi,
we want to add SCALEX, an online integration method that integrate different single-cell experiments including scRNA-seq and scATAC-seq, which also enable accurate projecting new incoming datasets onto the existing cell space. This method is published on Nature Communications. SCALEX is developed based-on scanpy. We appreciate and will be honored to contribute to the scanpy community.

Thank you!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -102,6 +102,7 @@ intersphinx_mapping = dict(
     seaborn=('https://seaborn.pydata.org/', None),
     sklearn=('https://scikit-learn.org/stable/', None),
     scanpy_tutorials=(scanpy_tutorials_url, None),
+    scalex=('https://scalex.readthedocs.io/en/latest/', None),
 )
 
 
--- a/docs/ecosystem.md
+++ b/docs/ecosystem.md
@@ -94,6 +94,7 @@ Interactive manifold viewers.
 ### Data integration
 
 - [scanaroma](https://github.com/brianhie/scanorama) {small}`MIT`
+- [scalex](https://github.com/jsxlei/SCALEX) {small}`Tsinghua`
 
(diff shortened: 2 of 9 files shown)


Zethson:
Hi,

all things that are not transcriptomics should ideally live in our sister packages such as https://github.com/scverse/muon . If you remove all things related to ATAC-seq in this PR we could consider it, but honestly I think that this might better live outside of scanpy external and much rather in the [scverse ecosystem](https://scverse.org/packages/#ecosystem)

jsxlei:
> Hi,
> 
> all things that are not transcriptomics should ideally live in our sister packages such as https://github.com/scverse/muon . If you remove all things related to ATAC-seq in this PR we could consider it, but honestly I think that this might better live outside of scanpy external and much rather in the [scverse ecosystem](https://scverse.org/packages/#ecosystem)

Hi,
Thank you for your reply. I could remove the ATAC in the PR, actually, this commit [7f74d8c](https://github.com/scverse/scanpy/pull/2355/commits/7f74d8c47005dd630f691ab5926095f0ff277ce8) is the version without ATAC. Please let me know if this does not work, I will commit another PR.
I think scalex is very suitable to be included in pp.external since it was developed based-on scanpy system, which could provide more choices for scanpy users to do single-cell integration.

Zethson:
So we're really not accepting any packages into scanpy.external anymore and will deprecate external soon. We've now updated our documentation to reflect this. However, we'd be very happy to welcome your package in the scverse ecosystem -> https://scverse.org/packages/#ecosystem

I'm sorry that you put in all this work but then get denied by us like this :(

PR #2405, opened 2023-02-03
Fix some typos in scanpy.tl.paga

yuanx749:
Found some typos in the docstring. Also searched them in other places.


Zethson:
Thanks!

PR #2407, opened 2023-02-06
Fixing mnn

smautner:
n_jobs =1 is not enough to disable multiprocessing in mnn, see here: 
https://github.com/chriscainx/mnnpy/blob/2097dec30c193f036c5ed7e1c3d1e3a6270e102b/mnnpy/utils.py#L34

so i fixed it. works for me :)


Zethson (approved):
Great, thanks!

PR #2400, opened 2023-01-17
Update external page

ivirshup:
Update policy for external tools and point to scverse ecosystem packages instead.

PR #2220, opened 2022-04-05
Doc overhaul

adamgayoso:
This is a version of Sphix-book-based scanpy docs.

**What's lost?**
- Docsearch
- Edit on github button (was weird anyway as no other package does that)

**What's gained?**
- Aesthetics, easier to read
- Git submodule for tutorials
- sphinx.ext.viewcode so each function has a source button (effectively replaces edit on github button)

**What else changed?**
- I removed the release latest stuff, not sure what value it was adding
- API docs are split logically

View the rendered docs here:

https://icb-scanpy--2220.com.readthedocs.build/en/2220/index.html

What's left to do:

- [x] Update tutorials index page to use nbgallery feature of nbsphinx
- [x] Make home index page use a grid of cards, two columns
- [x] Add to contributing guide how to manage the git submodule (really just occasionally do `git submodule update --remote`)
- [x] Annotate CSS overrides
- [x] Exchange sphinx ext viewcode for linkcode (probably in scanpydoc)

PR #2415, opened 2023-02-15
Revert "Update external page"

adamgayoso:
Reverts scverse/scanpy#2400


adamgayoso:
We need to do this for https://github.com/scverse/scanpy/pull/2220

I added this change into that PR

PR #2447, opened 2023-03-09
Increase ubuntu version & python version for azure CI

### LINKED ISSUE #2446: Azure removes support for Ubuntu 18
https://github.com/scverse/scanpy/blob/master/.azure-pipelines.yml#L14

https://learn.microsoft.com/en-us/answers/questions/1181262/what-happens-to-azure-vms-with-ubuntu-18-04-lts-af

and our CI already cries about it. Let's up this to a more recent version, maybe even 22.04?

While we're at it, I'd also remove 3.7 from the CI because it'll lose support very soon. We could add one of the more recent Python versions instead.

Zethson:
Fixes #2446 

Python 3.7 is EOL in 3 months https://endoflife.date/python and Pandas hasn't supported it for a loooong time already.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -11,11 +11,11 @@ variables:
 jobs:
 - job: PyTest
   pool:
-    vmImage: 'ubuntu-18.04'
+    vmImage: 'ubuntu-22.04'
   strategy:
     matrix:
-      Python37:
-        python.version: '3.7'
+      Python38:
+        python.version: '3.8'
       Python39: {}
       minimal_tests:
         TEST_EXTRA: 'test-min'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ requires = [
 [project]
 name = "scanpy"
 description = "Single-Cell Analysis in Python."
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 license = {file = "LICENSE"}
 authors = [
     {name = "Alex Wolf"},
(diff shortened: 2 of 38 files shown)


ivirshup:
Happy to bump the ubuntu version. Would also like this to happen on the backport branch.

I want to keep testing against the lowest python version we support. In theory we could jump to 3.9 as the minimum version. But I think numpy is still at 3.8, so that's reasonable too. Would you want to make this change here, or split to a separate PR?

ivirshup (commented):
Happy to bump the ubuntu version. Would also like this to happen on the backport branch.

I want to keep testing against the lowest python version we support. [In theory we could jump to 3.9 as the minimum version](https://scientific-python.org/specs/spec-0000/#description). But I think numpy is still at 3.8, so that's reasonable too. Would you want to make this change here, or split to a separate PR?

Zethson:
I personally think that we have enough users that are still on 3.8 so I'd keep it at 3.8?

Are you now trying to say that the lowest version that we're supporting is 3.7 or 3.9?

If we agree on 3.8 being the minimum I'd just keep the PR as it is and merge.

ivirshup:
I'm saying either:

* Don't change the CI python version
* Or: Change the minimum python version for the project to 3.8 as well

https://github.com/scverse/scanpy/blob/0692ef9ea30335b95f7e7f9aab7be856469d9f35/pyproject.toml#L13

Zethson:
Got it, thanks. I'll increase the minimum Python version to 3.8. I'll also check which compatibility code we can remove later. I'll ask for another review then.

flying-sheep:
@meeseeksdev backport to 1.9.x

PR #2424, opened 2023-02-20
updated read_visium() to read in spaceranger 2.0 files

LLehner:
Can now read in spaceranger 2.0 files where the tissue position file name is different and a header is included. Code adapated from squidpy's [read.visium()](https://github.com/scverse/squidpy/blob/main/squidpy/read/_read.py#L25) in response to [scverse/squidpy#599](https://github.com/scverse/squidpy/issues/599) and [scverse/scanpy#2296](https://github.com/scverse/scanpy/pull/2296#issuecomment-1273309536).

PR #2458, opened 2023-03-29
Read visium v2

johnyaku:
The "outs" from `spaceranger count` v2 differ from v1. Specifically, `tissue_positions_list.csv` has been renamed `tissue_positions.csv` and now has headers. Fortunately the column names exactly match those used by `scanpy`. See https://support.10xgenomics.com/spatial-gene-expression/software/pipelines/latest/output/spatial (search for `tissue_positions.csv`).

This PR adjusts `sc.read_visium()` so that it can load the outputs of `spaceranger count` v2.

The API / method signature is unchanged, so no changes to the documentation are required. The version is inferred from the presence or otherwise of the old `tissue_positions_list.csv` (implying v1) and then the files in `spatial/` are handled accordingly, including differential handling of headers / column names.

@linhuawang
@sopvdl
@TaopengWang

### DIFF
--- a/scanpy/readwrite.py
+++ b/scanpy/readwrite.py
@@ -369,7 +369,7 @@ def read_visium(
 
     adata.uns["spatial"][library_id] = dict()
 
-    # spaceranger count v1 uses "spatial_tissue_positions_list.csv" (no headers) 
+    # spaceranger count v1 uses "spatial_tissue_positions_list.csv" (no headers)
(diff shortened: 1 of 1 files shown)


LLehner:
This PR fixes the same issue as [PR#2424](https://github.com/scverse/scanpy/pull/2424), which will be merged.

giovp:
closing this as it was fixed in #2424  thanks @johnyaku for the PR nonetheless!

PR #2460, opened 2023-04-02
Coloring by boolean variables 

### LINKED ISSUE #1646: Support coloring by boolean variables

---

```python
import scanpy as sc
adata = sc.datasets.blobs()
sc.pp.pca(adata)

adata.obs['boolean'] = True

sc.pl.pca(adata, color='boolean')
```

```pytb
... storing 'blobs' as categorical

<ipython-input-11-1415b8dea7b8> in <module>
      5 adata.obs['boolean'] = True
      6 
----> 7 sc.pl.pca(adata, color='boolean')

/opt/conda/lib/python3.7/site-packages/scanpy/plotting/_tools/scatterplots.py in pca(adata, annotate_var_explained, show, return_fig, save, **kwargs)
    727     if not annotate_var_explained:
    728         return embedding(
--> 729             adata, 'pca', show=show, return_fig=return_fig, save=save, **kwargs
    730         )
    731     else:

/opt/conda/lib/python3.7/site-packages/scanpy/plotting/_tools/scatterplots.py in embedding(adata, basis, color, gene_symbols, use_raw, sort_order, edges, edges_width, edges_color, neighbors_key, arrows, arrows_kwds, groups, components, layer, projection, scale_factor, color_map, cmap, palette, na_color, na_in_legend, size, frameon, legend_fontsize, legend_fontweight, legend_loc, legend_fontoutline, vmax, vmin, add_outline, outline_width, outline_color, ncols, hspace, wspace, title, show, save, ax, return_fig, **kwargs)
    257         if sort_order is True and value_to_plot is not None and categorical is False:
    258             # Higher values plotted on top, null values on bottom
--> 259             order = np.argsort(-color_vector, kind="stable")[::-1]
    260         elif sort_order and categorical:
    261             # Null points go on bottom

TypeError: The numpy boolean negative, the `-` operator, is not supported, use the `~` operator or the logical_not function instead.
```

#### Versions

<details>

scanpy==1.7.0 anndata==0.7.5 umap==0.5.1 numpy==1.20.0 scipy==1.6.0 pandas==1.2.1 scikit-learn==0.24.1 statsmodels==0.12.2 python-igraph==0.8.3 louvain==0.7.0 leidenalg==0.8.3

</details>

namsaraeva:
Fixes #1646 

Now supports coloring by boolean variables such as `True` and `False`.

**Tasks to complete:** 

- [x] Add test


namsaraeva:
@Zethson do we really need a test here?

Zethson:
> @Zethson do we really need a test here?

Yes!

namsaraeva:
@ivirshup coloring by boolean values `(True, False)` is now possible:

1) The solution is based on casting the boolean columns to string columns, so that they can be colored in a categorical way. Actual columns in the anndata object are not modified.

2) I was thinking about the case where True is 1 and False is 0. Current behaviour: colorbar is plotted, since 0 and 1 are treated like continuous variables. Does it make sense to handle this case in scanpy?

3) Tests fail due to new pandas version, they do not fail locally though. 

What do you think?

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -434,7 +436,7 @@ def embedding(
             path_effect = None
 
         # Adding legends
-        if categorical:
+        if categorical or color_vector.dtype == bool:  # for categorical and boolean
```
Zethson:
```suggestion
        if categorical or color_vector.dtype == bool:
```

That's what the code already says. Adding some reasoning above would be more interested. But think this is fine without any comments.

Zethson (approved):
Thank you!

I'd still appreciate a look from @ivirshup and his thoughts.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -274,7 +274,9 @@ def embedding(
         order = slice(None)
         if sort_order is True and value_to_plot is not None and categorical is False:
             # Higher values plotted on top, null values on bottom
-            order = np.argsort(-color_vector, kind="stable")[::-1]
+            order = np.argsort(~pd.isnull(color_vector.astype(str)), kind="stable")[
```
ivirshup: Why as string? Surely this would break for integer values?

scanpy/tests/test_plotting.py
```diff
@@ -1218,6 +1218,15 @@ def test_scatter_raw(tmp_path):
     assert "Error" in comp, "Plots should change depending on use_raw."
 
 
+def test_binary_scatter(image_comparer):
+    save_and_compare_images = image_comparer(ROOT, FIGS, tol=15)
+    blobs = sc.datasets.blobs()
+    blobs.obs['binary'] = np.tile([True, False], 320)
+    sc.pp.pca(blobs)
```
ivirshup: In order to make tests faster, instead of needing to compute a PCA, could you directly provide the coordinates for the embedding?

ivirshup (dismissed):
Thanks for the PR on this. I've left a couple comments on the code, but had two larger questions.

## Code structure

This makes changes in a number of places to make booleans treated like categorical variables. Could you instead add a check around this code:

```python
        color_source_vector = _get_color_source_vector(
            adata,
            value_to_plot,
            layer=layer,
            use_raw=use_raw,
            gene_symbols=gene_symbols,
            groups=groups,
        )
```

that converts a boolean to a categorical?

Then a case could be added to the order check which checks if it's a categorical of booleans. 

### Colors

What do you think of a different set of default colors for booleans? Maybe red blue?

scanpy/plotting/_tools/scatterplots.py
namsaraeva: @ivirshup fixed it in my new commit, it did not influence much, but the order of values was broken (higher values were not on top but mixed with others). Now it should be fine.

scanpy/tests/test_plotting.py
namsaraeva: @ivirshup I don't quite get it, sorry. What should I exactly compare to pre-computed PCA and how it will test if the plotting is right?

flying-sheep: you could just define your data as as list-of-lists with 4 or so datapoints instead of the dataset.

flying-sheep (changes requested):
please simplify the test and we’re good to go!

flying-sheep (approved):

namsaraeva:
@flying-sheep Thank you very much for the review and finishing the PR!

flying-sheep:
No problem! did you address @ivirshup’s comment about code organization here? https://github.com/scverse/scanpy/pull/2460#pullrequestreview-1399566618

Then it’s ready

namsaraeva:
@flying-sheep yes I fixed it. The coloring stuff was not changed though. 

IMO I don't think it is necessary to change default blue and orange for the boolean plotting, but what do you think?

flying-sheep:
I agree!

racng:
Thanks for the fix! Can we include this in the next scanpy release?

Zethson:
It will be included in the next release, don't worry :)

PR #2469, opened 2023-04-12
PCA fix random_state

KalinNonchev:
Hello scanpy,

I noticed some inconsistencies in the [scanpy pca function](https://scanpy.readthedocs.io/en/stable/generated/scanpy.pp.pca.html) even when fixing the random seed. I looked at the pca code and it looks like the random seed in the PCA initialization is not set. I fixed it.

Best,


KalinNonchev:
pre-commit.ci autofix

Zethson:
Thanks!

PR #2484, opened 2023-05-08
Fix generated path

adamgayoso:
The previous docs PRs moved the generated docs path, which breaks old urls. This fixes that.

This also adds responsive handling of the grid on the main index page.


ivirshup (approved):
LGTM.

PR #2503, opened 2023-06-07
Convert issue report templates to forms

flying-sheep:
That makes it easier for people to fill it out.

Since GitHub doesn’t provide previews, I added them here: https://github.com/flying-sheep/playground/issues/new/choose


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,59 @@
...
+  validations:
+    required: false
+- type: textarea
+  id: versions
+  attributes:
+    label: Versions
+    description: Paste the output of scanpy.logging.print_versions() leaving a blank line after the details tag
+    render: markdown
+    value: |
+      <details>
```
Zethson, line 53: Think we could add a placeholder here

.github/ISSUE_TEMPLATE/enhancement-request.yml
```diff
@@ -0,0 +1,24 @@
...
+body:
+- type: dropdown
+  id: kind
+  attributes:
+    label: What kind of feature would you like to request?
+    options:
+    - 'Additional function parameters / changed functionality / changed defaults?'
+    - 'New analysis tool: A simple analysis tool you have been using and are missing in `sc.tools`?'
+    - 'New plotting function: A kind of plot you would like to seein `sc.pl`?'
+    - 'External tools: Do you know an existing package that should go into `sc.external.*`?'
```
Zethson: We've decided to NOT add more stuff into external for too many reasons (we can talk about it if you don't agree).
```suggestion
```

Zethson (approved):

.github/ISSUE_TEMPLATE/bug-report.yml
flying-sheep: A placeholder is the grey text that vanishes once you focus the field, `value` stays.

Since there’s no way to express “add `<details>` before the user content and `</details>` after”, this is the best we can do

.github/ISSUE_TEMPLATE/enhancement-request.yml
flying-sheep: Good catch. I just converted the templat faithfully and didn’t look at the options here much

Issue #2506, opened 2023-06-07
`compute_neighbors_umap` does not use `scanpy._settings.ScanpyConfig.n_jobs`


mumichae:
### Please make sure these conditions are met

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the latest version of scanpy.
- [X] (optional) I have confirmed this bug exists on the master branch of scanpy.

### What happened?

When trying to compute neighbors with default settings, I'm getting errors in slurm jobs, as the code tries to allocate all available cores on a node. I tried to control the number of cores available by setting `scanpy._settings.ScanpyConfig.n_jobs`, however I found that this setting does not get used in the offending function `compute_neighbors_umap`, where the following gets called.

https://github.com/scverse/scanpy/blob/9cab8cfa4033d3f47a36c7bb816b2c9fae5cfdc6/scanpy/neighbors/__init__.py#L314-L322

The `nearest_neighbors` function from the `umap` provides the parameter `n_jobs`, which is set to -1 (i.e. all cores) by default.

https://github.com/lmcinnes/umap/blob/e0cff20b392192f2b62fcc1843c9cb9ca2c9ee27/umap/umap_.py#L255-L266

However, as a user I would expect the number of cores to be defined by the ScanpyConfig.
I assume this could also cause the kernel dying unintentionally in other work setups when users aren't aware of the number of cores the function is requesting.

### Minimal code sample

```python
import scanpy as sc
sc._settings.ScanpyConfig(n_jobs=1)

ad = sc.read('quite_large_h5ad_file')
sc.pp.pca(ad)
sc.pp.neighbors(ad)
```

### Error output

```pytb
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
File ~/tools/mambaforge/envs/scanpy/lib/python3.9/multiprocessing/pool.py:212, in Pool.__init__(self, processes, initializer, initargs, maxtasksperchild, context)
    211 try:
--> 212     self._repopulate_pool()
    213 except Exception:

File ~/tools/mambaforge/envs/scanpy/lib/python3.9/multiprocessing/pool.py:303, in Pool._repopulate_pool(self)
    302 def _repopulate_pool(self):
--> 303     return self._repopulate_pool_static(self._ctx, self.Process,
    304                                         self._processes,
    305                                         self._pool, self._inqueue,
    306                                         self._outqueue, self._initializer,
    307                                         self._initargs,
    308                                         self._maxtasksperchild,
[...]

PR #2507, opened 2023-06-07
Passing ScanpyConfig.settings.n_jobs to the `umap.nearest_neighbors` function

mumichae:
addresses #2506


flying-sheep (approved):
Good idea. If we want to make this configurable via keyword argument as well, we can do this in another PR

PR #2508, opened 2023-06-10
Advertise Python 3.11 compatibility in package metadata

Zaf4:
Upon Numba compatibility with python3.11, scanpy is now compatible with python3.11


flying-sheep:
Thanks!

flying-sheep (approved):

Issue #2444, opened 2023-03-09
[Proposal] Integrate Marsilea to visualize AnnData


Mr-Milk:
- [x] New plotting function: A kind of plot you would like to seein `sc.pl`?
- [x] External tools: Do you know an existing package that should go into `sc.external.*`?

Hi, thanks for developing such a fantastic framework for analyzing single-cell data, it really helps a lot in my research. I'm the developer of [Heatgraphy](https://github.com/Heatgraphy/heatgraphy), which is a python package to visualize multi-dimensional data using the x-layout system. You may think of it as python's version of complexHeatmap, but Heatgraphy can do much more.

Personally, I think Heatgraphy can help visualize the AnnData intuitively. Here is [an example](https://heatgraphy.readthedocs.io/en/latest/auto_examples/plot_pbmc3k.html) of visualization of the PBMC3K dataset using Heatgraphy. The structure of AnnData fits the structure of this visualization quite well. We can plot the `.X` as the heatmap, and other attributes in `.obs`/`.var` or `.obsm`/`.varm` as side plots.

![](https://heatgraphy.readthedocs.io/en/latest/_images/sphx_glr_plot_pbmc3k_001_2_0x.png)

Therefore, I propose adding a new API using Heatgraphy to help visualize AnnData.

If adding to `sc.pl`, the API may look like this:

```python
sc.pl.heatgraphy(adata, left=[("cell_type", "color"), ("cell_type", "label")], right=[("gene_name", "label")])
```

If added as an extra class, the API can be more flexible and offer much more customization. It may look like this:

```python
viz = AnnDataViz(adata)
viz.add_left(key="cell_type", plot="color", cmap="Set2")  # use a key from .obs and plot as color strip
viz.add_left(key="cell_type", plot="label")
viz.render()
```

It can also be applied to specific visualization for analysis.

Zethson:
This sounds great! Please go for it and submit a minimal draft PR. Then we can iterate over it

ivirshup:
Example of creating dot plot, which to my mind was a main missing feature. This should go great with `sc.get.aggregate`: https://github.com/Marsilea-viz/marsilea/issues/27#issuecomment-2022353426

Ideally, I think we could switch to marsilea for many plots for scanpy 2.0, replacing our existing BasePlot subclasses. This definitely requires more exploration first though.

[...]

PR #2512, opened 2023-06-15
Integration of Marsilea to create Heatmap from AnnData

Mr-Milk:
This PR is for #2444 

The project "heatgraphy" mentioned in the issue is renamed to "Marsilea", https://github.com/Marsilea-viz/marsilea

The previous link to the image mentioned in the issue is not available anymore. Please refer to [this link](https://marsilea.readthedocs.io/en/latest/auto_examples/plot_pbmc3k.html)

I added a minimum example in the `external` part. Here is a code example of using the plotting function

Remember to install Marsilea before proceeding:

```shell
pip install marsilea
```

```python
import scanpy as sc
adata.var["gene_names"] = adata.var.index

t = sc.external.pl.MarsileaHeatmap(adata, vmin=-2, vmax=2, center=0, label="Expression")
t.add_right("bulk_labels", "colors", size=.2, label="Cell Lables")
t.add_dendrogram("right", add_base=False, add_meta=True)
t.h_groupby("bulk_labels")
t.add_top("n_counts", "bar", show_value=False, label="Counts")

genes = ['CDK6', 'UQCRFS1', 'SH2D2A', 'CD63', 'CEP152',
         'SIAH2', 'LY86', 'HSPB1', 'POLR2G', 'IGBP1']

t.add_bottom("gene_names", "annot", mark=genes)
t.add_title("Expression Heatmap")
t.legend()
t.show()
```
If everything goes well, it should display something like this:

download

### DIFF
--- a/scanpy/external/_marsilea_heatmap.py
+++ b/scanpy/external/_marsilea_heatmap.py
@@ -1,0 +1,74 @@
+import marsilea as ma
+import marsilea.plotter as mp
+import numpy as np
+from anndata import AnnData
+from scipy.sparse import issparse
+from natsort import natsorted
+
+
+class MarsileaHeatmap:
+
+    def __init__(self,
+                 adata: AnnData,
+                 layer=None,
+                 **kwds):
+        if layer is None:
+            cluster_data = adata.X
+        else:
+            cluster_data = adata.layers[layer]
+
+        if issparse(cluster_data):
+            cluster_data = cluster_data.toarray()
+
+        self.adata = adata
+        self.heatmap = ma.Heatmap(cluster_data, **kwds)
+
+    PLOTTER_POOL = {
+        "bar": mp.Numbers,
+        "text": mp.Labels,
+        "annot": mp.AnnoLabels,
+        "colors": mp.Colors,
+    }
+
+    def _add_plot(self, slot, side, keys, plot, size=1, pad=.1, **kwds):
+        data = getattr(self.adata, slot)[keys]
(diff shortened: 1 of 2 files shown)


flying-sheep:
Hi, we’re currently planning scanpy 2.0.

We want to reduce the number of ways people can make heatmaps, but maybe marsilea would be a good base for the new version. We’ll keep an eye on it.

ivirshup:
Thanks for the PR! We've been thinking about refactoring this part of the package, and this looks like an interesting way to do it.

However, we're not accepting any additions to the `external` module anymore. Instead we are pointing people to the broader [scverse ecosystem](https://scverse.org/packages/#ecosystem).

We may be interested in using this as a direct dependency but may need to do some research into this first + request/ add a few features in `Marsilea` such as dot plots.

cc @grst

Mr-Milk:
Thanks for your interest! It would be cool if `Marsilea` could be part of `Scanpy`. 

The `Marsilea` is shipped with a wide range of plot options which already include the [dot plot](https://marsilea.readthedocs.io/en/latest/tutorial/heatmap.html#matrix-heatmap-with-sized-elements) (We call it Sized Heatmap). Here is a [list](https://marsilea.readthedocs.io/en/latest/api/plots.html) of all available plot options for your reference. 

Any further suggestions or requests to expand the plot options are welcome.

flying-sheep:
Thank you! I added this to our planning document. I’ll close the PR since as said, we’re not adding to external anymore

PR #2514, opened 2023-06-16
Switch to ruff

### LINKED ISSUE #2513: Switch to ruff

Other?

Like anndata, scanpy can benefit from the speed and consistency of ruff.

flying-sheep:
Fixes #2513

Compare https://github.com/scverse/anndata/pull/891 and https://github.com/scverse/anndata/commit/14e0cd4795730673e55d14639d562fbe7691965b (sorry for accidentally pushing to main)


pyproject.toml
```diff
@@ -162,9 +162,26 @@ source = ["scanpy"]
 omit = ["*/tests/*"]
 
 [tool.black]
-line-length = 88
-target-version = ["py38"]
 skip-string-normalization = true
-exclude = """
-/build/.*
-"""
```
flying-sheep, line 170: - `line-length = 88` is default
- `target-version` is inferred from `requires-python`
- `exclude` is derived from .gitignore

pyproject.toml
```diff
@@ -162,9 +162,26 @@ source = ["scanpy"]
...
+
+[tool.ruff]
```
flying-sheep, line 167: - no `exclude`, the default contains all we need
- no `line-length`, the default is 88
- don’t mention all the black-incompatible rules
- remove the “`F811` Redefinition of unused name” from per-file ignores, as it wasn’t triggered
- demote the “`E731` Do not assign a lambda expression” to a per-file ignore, as it’s only used there

scanpy/tests/conftest.py
```diff
@@ -9,7 +9,7 @@
 import pytest
 
 import scanpy
-
+from scanpy.tests.fixtures import array_type, float_dtype  # noqa: F401
```
flying-sheep, line 12: It’s better to import fixtures in the conftest file, or use plugins, than to import fixtures in the test file

flying-sheep:
pre-commit.ci autofix

Zethson (approved):

scanpy/tests/conftest.py
ivirshup: Why?

flying-sheep: Let’s meet up and we can go over pytest’s namespacing and so on (`conftest` files are like autoloaded plugins, …). I think that’s more effective than picking out one aspect here.

in short:

- editor support is there (cmd-click on a `test_*` function’s fixture parameter works)
- linters don’t understand it
- in pytest land we should use pytest idioms
- in #2235 we do that anyway

PR #2518, opened 2023-06-19
Check PRs for milestones

flying-sheep:
So we can’t forget adding them, thereby ensuring bugfixes make it into the next bugfix release (including CI fixes)


.github/workflows/check-pr.yml
```diff
@@ -0,0 +1,24 @@
...
+  pull_request:
+    branches:
+      - trunk
+    types:
+      - labeled
+      - unlabeled
+      - opened
+      - reopened
+      - synchronize
+      - edited
```
ivirshup: What's all this for? Could it just be PRs onto main?

flying-sheep: ignore the `trunk`, the `types` are all event types that can actually add a milestone:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

ivirshup: Why don't we need those for other actions?

flying-sheep: See the first Note in the link above:

> By default, a workflow only runs when a `pull_request` event's activity type is `opened`, `synchronize`, or `reopened`.

where `synchronize` means “A pull request's head branch was updated“.

That’s a set of events that’s perfect for running as a response to repo content changes.

But this workflow needs to run on metadata changes (labels and milestones) as well.

ivirshup: Issues have a `milestoned`, `demilestoned` event. Do you know if `labelled` covers that for PRs?

flying-sheep: There seems to be a discrepancy between the workflow-focused events page and the webhook-focused event payload page. the former doesn’t mention the `[de]milestoned` events while the latter does:

- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

  > grafik


- https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request

  > grafik

ivirshup: Maybe just check out the API? I think there is a JSON-schema for it somewhere.

flying-sheep: Here? https://github.com/SchemaStore/schemastore/blob/726fe2b26b20621a11e29f6f0de01c0979023c4c/src/schemas/json/github-workflow.json#L1352-L1377

I think the official spec is what I linked above.

I think webhooks are currently simply more powerful than the GH workflow action types: https://github.com/orgs/community/discussions/8674#discussioncomment-6219278

ivirshup: Have you tested this somewhere?

I would like to know if this gets triggered when a milestone is added. Or if adding a `milestoned` would work.

> I think the official spec is what I linked above.

Which link? I thought these were all docs, and generally haven't had the best experience with the accuracy of github docs.

flying-sheep: I’ll test it in my playground, good call.

> Or if adding a `milestoned` would work.

If so, it would be an undocumented feature and therefore not reliable.

As said: I’m pretty sure GitHub just shares the schema of workflow event payloads and webhook payloads. That doesn’t mean that the same list of action types can be used to trigger a workflow and a webhook.

ivirshup: Tbh, I've found the documented features pretty unreliable too.

ivirshup: Maybe submit a support request asking for clarification?

image

flying-sheep: OK, I checked: `milestoned` is triggered.

- [first run, no milestone](https://github.com/flying-sheep/playground/actions/runs/5313789551/jobs/9620154329#step:4:7):

  grafik

- [second run, milestoned event](https://github.com/flying-sheep/playground/actions/runs/5313797804/jobs/9620175670?pr=6)

  grafik

flying-sheep: haha we posted the same thing at the same time!

flying-sheep: There’s a request to update the docs: https://github.com/github/docs/issues/23909

ivirshup (commented):
Do we need to look at event types other than `milestoned`, `demilestoned`?

flying-sheep:
I added comments. The only events I’m not more or less sure about are “reopened” and ”edited”.

ivirshup (approved):
I would think we wouldn't need this for anything beyond the milestones

Maybe unless we're requiring checks pass on the last commit, though I'm not sure that would even apply here. Then maybe we'd add the defaults?

IDK, happy as is.

flying-sheep:
takes less than a second to run, so it doesn’t matter at all if it runs a few too many times.

PR #2521, opened 2023-06-20
Fix paga tests

flying-sheep:
In the sense that they no longer fail, not that they make sense. That’s what #2235 is for.

### DIFF
--- a/scanpy/tests/test_paga.py
+++ b/scanpy/tests/test_paga.py
@@ -24,7 +24,7 @@ needs_igraph = pytest.mark.skipif(
 def pbmc():
     pbmc = pbmc68k_reduced()
     sc.tl.paga(pbmc, groups='bulk_labels')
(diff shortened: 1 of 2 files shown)


scanpy/tests/test_paga.py
```diff
@@ -24,7 +24,7 @@
 def pbmc():
     pbmc = pbmc68k_reduced()
     sc.tl.paga(pbmc, groups='bulk_labels')
-    pbmc.obs['cool_feature'] = pbmc[:, 'CST3'].X.squeeze()
+    pbmc.obs['cool_feature'] = pbmc[:, 'CST3'].X.squeeze().copy()
```
flying-sheep, line 27: `pbmc[:, 'CST3'].X.squeeze()` returns an `ArrayView`, which ends up all `[nan, nan, …]`.

Tests fail because I assume matplotlib doesn’t like that anymore.

ivirshup (approved):

flying-sheep:
@meeseeksdev backport 1.9.x

PR #2536, opened 2023-06-27
Make `neighbors` work based on KNeighborsTransformer-compatible APIs

### LINKED ISSUE #2519: Add approximate nearest neighbor backend
Goals:

- flexibility in creation
- Neighbor querying API
- Harmonize storage of backends’ querying indices

	- use it in doublet detection and scanorama reimplementations

Current solution: `method` arg.

https://github.com/scverse/scanpy/blob/ed8e1401d39068782f2435d258b33fce4f7b4a9e/scanpy/neighbors/__init__.py#L32

- if `use_dense_distances`: [`sklearn.metrics.pairwise_distances`](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html)
- if `method == 'rapids'`: [`cuml.neighbors.NearestNeighbors`](https://docs.rapids.ai/api/cuml/stable/api/#cuml.neighbors.NearestNeighbors)
- otherwise: [`umap.umap_.nearest_neighbors`](https://umap-learn.readthedocs.io/en/latest/api.html#umap.umap_.nearest_neighbors)
- if `method == 'gauss'`: use umap distances, overwrite its connectivities

## Evaluating options

See [ann-benchmarks.com](https://ann-benchmarks.com/index.html#datasets).

Build time vs query time is not straightforward, see https://github.com/erikbern/ann-benchmarks/issues/207#issuecomment-1180389432 and https://github.com/erikbern/ann-benchmarks/issues/207#issuecomment-1180747770

> If, however, you are simply interested in knn-graph construction then you can get that from pynndescent in less time than even the index construction time (since the prepare phase isn't required, but is a non-trivial part of the index construction time).

Plots for index building are on individual dataset pages, like [glove-100-angular](https://ann-benchmarks.com/glove-100-angular_10_angular.html).

<details>
<summary>Used metrics</summary>

https://github.com/scverse/scanpy/blob/ed8e1401d39068782f2435d258b33fce4f7b4a9e/scanpy/neighbors/__init__.py#L35-L56

</details>

Interesting:

Name | Demo | wheels: Platforms | wheels: Python | Search speed | Index build
--- | --- | --- | --- | --- | ---
[NMSLIB](https://github.com/nmslib/nmslib) | [6 notebooks](https://github.com/nmslib/nmslib/blob/master/python_bindings/notebooks/README.md) | [Linux wheels](https://pypi.org/project/nmslib/2.1.1/#files) |[Outdated (3.9)](https://github.com/nmslib/nmslib/issues/529) | top | slow
[ScaNN](https://github.com/google-research/google-research/tree/master/scann#readme) | [notebook](https://github.com/google-research/google-research/blob/master/scann/docs/example.ipynb) | [All wheels](https://pypi.org/project/scann/1.2.9/#files) | [Slightly outdated (3.10)](https://github.com/google-research/google-research/issues/1590) | very fast | middling
[Faiss](https://github.com/facebookresearch/faiss) | [wiki](https://github.com/facebookresearch/faiss/wiki/Getting-started) | –* | –* | middling to very fast | faiss-ivf: top

*no official package on PyPI

flying-sheep:
[`sklearn.neighbors.KNeighborsTransformer`](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsTransformer.html) is a good abstraction for KNN calculations

We could follow the suggestion here, then finish this PR https://github.com/frankier/sklearn-ann/issues/8#issuecomment-1609553421

### For reviewing

- most functions were moved unchanged
- I extensively changed https://github.com/scverse/scanpy/blob/73915b4bd0dc84108df08ad83ec28e627a9a0e0d/scanpy/neighbors/__init__.py#L471
  - This uncovered very grown-over-time logic: https://github.com/scverse/scanpy/blob/73915b4bd0dc84108df08ad83ec28e627a9a0e0d/scanpy/neighbors/__init__.py#L522-L523  
    *Should we change that logic for more predictability at the expense of backwards compatibility*? Especially the “euclidean” condition makes not much sense IMHO
  -  Another piece of weird logic: *Where to allow`knn=False`*? It would make sense for e.g. the scipy knn transformer with `algorithm='kd_tree'` …  
    https://github.com/scverse/scanpy/blob/73915b4bd0dc84108df08ad83ec28e627a9a0e0d/scanpy/neighbors/__init__.py#L614-L617
  - I also changed `method` to only mean “connectivity method”. `transformer_cls` alone now determines how to calculate distances

### TODO

- [x] split of sklearn-ann part into own issue/PR
- [x] figure out what the `_more_tags` methods are 
- [x] allow specifying algorithm and/or backend
- [x] revert 75c6670, move connectivities code out of backends
- [x] switch our stuff to KNeighborsTransformer
- [x] unify selection: algorithm+backend, metric, connectivity (maybe separate out connectivity)
- [x] figure out how to do connectivities: can mode be changed after fit? *no, we just use umap connectivities as before*
- [x] check out where we have coverage
  - is there paga specific stuff? *not in the parts I changed*
  - gauss: dense matrix when knn=True (“build a symmetric mask”, …) *not covered, but also the logic shouldn’t have changed*
  - pre-computed in umap transformer; handling of disconnected components in umap (indices == -1) *replaced with our own implementation, see below*
- add tests
  - [x] pyknndescent (we already depend on it through umap)

Maybe in another PR?

- [ ] maybe store index in unified way?
- [ ] for umap: use `UMAP(precomputed_knn=...)` instead of `compute_connectivities`?
- [ ] unifiy transformer args, e.g. verbose

### The way it used to be

Our default use case was basically a thin wrapper around umap’s `nearest_neighbors` function, which in turn is a thin wrapper around PyNNDescent. We did some special casing around euclidean distance and small data sizes. That special casing reduced our test coverage: 

- we don’t actually test umap’s pynndescent codepath at all (just the fast `precomputed` path for small data)
- umap’s `precomputed` code does some weird things to its knn `indices` array, which we don’t test for: 
grafik

The logic was:

- if small data (<8000) and euclidean metric or knn==False, calculate `pairwise_distances`
  - if knn=True, then sparsify that matrix by pulling out KNN using `_get_indices_distances_from_dense_matrix` and converting that into a sparse one
- if not,  run `umap.nearest_neighbors`.
  - if even smaller data (<4000), calculate `pairwise_distances` like above and run `umap.nearest_neighbors` with `metric='precomputed'`. its internal logic then does the same as `_get_indices_distances_from_dense_matrix`
  - else run `umap.nearest_neighbors` directly, which simply runs PyNNDescent with `n_trees` and `n_iters` set with a heuristic.

### now the logic is

- if small data (for either definition of small), calculate `pairwise_distances`, sparsify on demand. (this covers both our previous optimization and the umap `'precomputed'` code path)
- else run a KNN transformer. if the old `'umap'` method was chosen, pick the same parameters mentioned aboveto the `PyNNDescent` transformer

---

Fixes #2519


flying-sheep:
@ivirshup apart from the API design, this is done. We’re waiting on sklearn-ann to get upstreamed, but we could also just finish it without the third party transformers (allowing people to specify their own).

What do you think?

scanpy/neighbors/__init__.py
```diff
@@ -1,71 +1,60 @@
...
-from .. import _utils
+from .. import _utils, settings
 from .._utils import _doc_params, AnyRandom, NeighborsView
 from ..tools._utils import _choose_representation, doc_use_rep, doc_n_pcs
-from .. import settings
+
+if sys.version_info >= (3, 9):
+    RPForestDict = Mapping[str, Mapping[str, np.ndarray]]
+else:
+    RPForestDict = Mapping
```
ivirshup, line 43: Shouldn't this be fine with `from __future__ import annotations`?

scanpy/neighbors/__init__.py
```diff
@@ -104,8 +93,23 @@ def neighbors(
...
+        Also accepts the following known options:
+
+        `None` (the default)
+            Behavior depends on data size.
+            For small data, we will calculate exact kNN, otherwise we use
+            :class:`~pynndescent.pynndescent_.PyNNDescentTransformer`
+        `'pynndescent'`
+            :class:`~pynndescent.pynndescent_.PyNNDescentTransformer`
+        `'rapids'`
+            A transformer based on :class:`cuml.neighbors.NearestNeighbors`.
```
ivirshup, line 125: I'm not sure I want to add more documentation saying this is supported if we plan to remove it.

Maybe we could remove the string input option as well?

scanpy/neighbors/__init__.py
```diff
@@ -1,71 +1,60 @@
...
 import scipy
 from anndata import AnnData
-from scipy.sparse import issparse, coo_matrix, csr_matrix
+from scipy.sparse import issparse, csr_matrix
 from sklearn.utils import check_random_state
-
+from pynndescent import NNDescent, PyNNDescentTransformer
+
+from . import _connectivity
+from ._enums import _Metric, _MetricFn, _Method, _KnownTransformer
```
ivirshup: Maybe this file should be called `_types`?

scanpy/neighbors/__init__.py
```diff
@@ -836,75 +589,48 @@ def compute_neighbors(
...
-            Dsq = self._distances.power(2)
-            indices, distances_sq = _get_indices_distances_from_sparse_matrix(
-                Dsq, self.n_neighbors
+    @staticmethod
+    def _handle_transform_args(
+        method: _Method | Literal['gauss'],
+        transformer_cls: type | _KnownTransformer | None,
+        *,
+        knn: bool,
+    ) -> tuple[_Method, type, bool]:
```
ivirshup: Could you document this method? A short description of behavior + what the output arguments mean – especially "shortcut".

scanpy/neighbors/__init__.py
```diff
@@ -768,66 +507,80 @@ def compute_neighbors(
...
         # do not use the cached rp_forest
         self._rp_forest = None
         self.n_neighbors = n_neighbors
         self.knn = knn
         X = _choose_representation(self._adata, use_rep=use_rep, n_pcs=n_pcs)
+
         # neighbor search
         use_dense_distances = (metric == 'euclidean' and X.shape[0] < 8192) or not knn
-        if use_dense_distances:
+        if shortcut and (use_dense_distances or X.shape[0] < 4096):
```
ivirshup: Could we use the `sklearn.KNeighborsTransformer(algorithm="brute")` for this? At least when `knn=False`?

scanpy/neighbors/_backends/rapids.py
```diff
@@ -0,0 +1,96 @@
...
+    'canberra',
+    'minkowski',
+    'chebyshev',
+    'jensenshannon',
+    'cosine',
+    'correlation',
+]
+
+
+CudaArrayLike = np.ndarray  # TODO
```
ivirshup: What is TODO?

scanpy/preprocessing/_pca.py
```diff
@@ -126,7 +128,10 @@ def pca(
     if data_is_AnnData:
         adata = data.copy() if copy else data
     else:
-        adata = AnnData(data, dtype=data.dtype)
+        if pkg_version('anndata') < version.parse('0.8.0rc1'):
+            adata = AnnData(data, dtype=data.dtype)
+        else:
+            adata = AnnData(data)
```
ivirshup, line 163: Not related right? I think fine, but would also be fine to make a small separate PR and just merge it.

scanpy/neighbors/__init__.py
```diff
@@ -735,26 +471,29 @@ def to_igraph(self):
...
+        *,
         use_rep: Optional[str] = None,
+        knn: bool = True,
         method: _Method = 'umap',
+        transformer_cls: type | _KnownTransformer | None = None,
+        transformer_kwds: Mapping[str, Any] = MappingProxyType({}),
+        metric: Union[_Metric, _MetricFn] = 'euclidean',
+        metric_kwds: Mapping[str, Any] = MappingProxyType({}),
         random_state: AnyRandom = 0,
         write_knn_indices: bool = False,
```
ivirshup: Is this ever used? Can we deprecate/ remove it?

scanpy/tests/test_neighbors.py
```diff
@@ -114,41 +114,78 @@ def neigh() -> Neighbors:
...
+    from pynndescent import NNDescent
+
+    # When trying to compress a too-small index, pynndescent complains
+    mocker.patch.object(NNDescent, 'compress_index', return_val=None)
+
+    neigh.compute_neighbors(n_neighbors, method=method)
+    np.testing.assert_allclose(neigh.distances.toarray(), distances_euclidean)
+
+
+@pytest.mark.parametrize(
```
ivirshup, line 133: Could we add a case passing in an actual transformer class? Maybe one of the sklearn.neighbors provided ones?

scanpy/tests/test_neighbors.py
```diff
@@ -114,41 +114,78 @@ def neigh() -> Neighbors:
...
+@pytest.mark.parametrize('method', ['umap', 'gauss'])
+def test_distances_euclidean(mocker, neigh, method):
+    """umap and gauss behave the same for distances.
+
+    They call pynndescent for large data.
+    """
+    from pynndescent import NNDescent
+
+    # When trying to compress a too-small index, pynndescent complains
+    mocker.patch.object(NNDescent, 'compress_index', return_val=None)
```
ivirshup, line 127: What does this actually do? Are we just catching a warning?

ivirshup (dismissed):
Looking pretty good! A few requests (+ a few more on the code)

* Can we add an example of using this to the documentation?
* The coverage looks a little low, can we try and get the diff coverage as high as possible?
* What do you think about using being able to pass in an instance of a class, rather than the class + kwargs?

scanpy/neighbors/__init__.py
flying-sheep: Sadly that only affects annotations, not expressions. With it activated, `var: typ = expr` is the same as `var: 'typ' = expr`, but `expr` still needs to evaluate successfully

scanpy/tests/test_neighbors.py
flying-sheep: it crashes without that.

scanpy/tests/test_neighbors.py
flying-sheep: sure!

scanpy/neighbors/__init__.py
flying-sheep: trying

scanpy/preprocessing/_pca.py
flying-sheep: right

scanpy/neighbors/_backends/rapids.py
flying-sheep: done

ivirshup (changes requested):
Looks pretty good. Just two remaining points from last review:

- [x] Can we add an example of using this to the documentation?
  - Could be under the "Usage" section? Maybe use a ball tree on low dim data?
- [x] What do you think about using being able to pass in an instance of a class, rather than the class + kwargs?
  - Just looking for a reasoning here.

And then:

- [x] Changelog entry

flying-sheep:
What “Usage” section? We have a “Usage principles” page and functions/classes can have an “Examples” section.

I’m pretty sure there was a very good reason for not passing an instance, but this was all so long ago that I forgot. I’ll see if it’s possible to maintain backwards compat while doing that!

ivirshup:
> What “Usage” section? We have a “Usage principles” page and functions/classes can have an “Examples” section.

Yes, examples

> I’m pretty sure there was a very good reason for not passing an instance, but this was all so long ago that I forgot. I’ll see if it’s possible to maintain backwards compat while doing that!

Maybe it was about passing in arguments to the function? But I do think it would be quite useful to pass in a class that gets `.fit` so you can keep a reference to the index.

flying-sheep:
I think it was just so convoluted that it seemed too hard to do while keeping backwards compatibility, but I managed it!

flying-sheep:
all done

ivirshup (dismissed):
For when you're back, I believe all that needs doing is:

- [x] Revert change to plotting example
- [x] Fix doc builds

flying-sheep:
@ivirshup you can approve PRs where the only change missing is fixing checks.

ivirshup (approved):

PR #2552, opened 2023-07-10
Add numfocus info

ivirshup:
Adding NumFOCUS info to README


Zethson (approved):
Under the assumption that you copied their required wording

PR #2563, opened 2023-07-19
Dask PCA support

selmanozleyen:
Hi,

Just wanted to start the PR. Passes the tests except one. Also need to deal with solver names since they don't correspond to anything dask uses.

Also refactored where the DaskArray mock class is.

Pinging @ivirshup


pyproject.toml
```diff
@@ -90,13 +90,14 @@ test-min = [
     "pytest>=4.4",
     "pytest-nunit",
     "profimp",
+    "scanpy[dask]",
+    "scanpy[dask-ml]",
```
flying-sheep: please undo.

the min tests are specifically there to test scanpy with no optional dependencies.

if they don’t run, you need to do one or more of the following:.

- add a `@needs('...')` to tests that require some dependency
- move an import from module level to the inside of a scanpy function so scanpy can be imported without that dependency installed

scanpy/preprocessing/_pca.py
```diff
@@ -52,7 +52,8 @@ def pca(
     zero_center
         If `True`, compute standard PCA from covariance matrix.
         If `False`, omit zero-centering variables
-        (uses :class:`~sklearn.decomposition.TruncatedSVD`),
+        (uses `~sklearn.decomposition.TruncatedSVD` or
+        `dask_ml.decomposition.TruncatedSVD`),
```
flying-sheep: these should be actual links

```suggestion
        (uses sklearn :class:`~sklearn.decomposition.TruncatedSVD` or
        dask-ml :class:`~dask_ml.decomposition.TruncatedSVD`),
```

scanpy/preprocessing/_pca.py
```diff
@@ -73,6 +74,11 @@ def pca(
         Efficient computation of the principal components of a sparse matrix
         currently only works with the `'arpack`' or `'lobpcg'` solvers.
 
+        If X is a dask array, then `'randomized'` will use the randomized algorithm
+        from dask-ml. Otherwise, `'auto'` will be given to `dask_ml.decomposition.PCA` and
+        `'tsqr'` will be given to `dask_ml.decomposition.TruncatedSVD`.
```
flying-sheep:
```suggestion
        from dask-ml. Otherwise, `'auto'` will be given to :class:`~dask_ml.decomposition.PCA` and
        `'tsqr'` will be given to :class:`~dask_ml.decomposition.TruncatedSVD`.
```

scanpy/preprocessing/_pca.py
```diff
@@ -151,19 +157,52 @@ def pca(
...
     logg.info(f'    with n_comps={n_comps}')
 
-    random_state = check_random_state(random_state)
-
     X = adata_comp.X
 
-    if chunked:
+    is_dask = isinstance(X, DaskArray)
+
+    random_state = check_random_state(random_state) if not is_dask else random_state
```
flying-sheep: why is that condition there?

scanpy/preprocessing/_pca.py
```diff
@@ -151,19 +157,52 @@ def pca(
...
+
+    if is_dask and not chunked:
+        # we will only let the user choose the randomized solver otherwise
+        # we will use our default
+        if svd_solver != 'randomized' and svd_solver != 'auto':
+            logg.debug('Ignoring svd_solver')
```
flying-sheep: Good idea to log it! Please add to the parameter documentation which values are valid when.

Maybe the default should be changed to `'auto'` or `None` if the actual default is different depending on the type of X?

scanpy/preprocessing/_pca.py
```diff
@@ -151,19 +157,52 @@ def pca(
...
 
-        X_pca = np.zeros((X.shape[0], n_comps), X.dtype)
+            pca_ = daskIncrementalPCA(n_components=n_comps)
+            X_pca = da.zeros((X.shape[0], n_comps), X.dtype)
+        else:
+            from sklearn.decomposition import IncrementalPCA
 
-        pca_ = IncrementalPCA(n_components=n_comps)
+            pca_ = IncrementalPCA(n_components=n_comps)
+            X_pca = np.zeros((X.shape[0], n_comps), X.dtype)
```
flying-sheep:
```suggestion
        if is_dask:
            from dask.array import zeros
            from dask_ml.decomposition import IncrementalPCA
        else:
            from numpy import zeros
            from sklearn.decomposition import IncrementalPCA

        pca_ = IncrementalPCA(n_components=n_comps)
        X_pca = zeros((X.shape[0], n_comps), X.dtype)
```

do the same with the `pca_ =` and `X_pca =` lines above

flying-sheep (changes requested):
looks straightforward enough!

1. please make sure the docs look right and you are using links. You probably have to add a line to `intersphinx_mapping` in `docs/conf.py` to be able to link to the dask-ml docs.
2. the logic seems straightforward, but the user should know when which `svd_solver` argument works, as that part is getting complicated.
3. you have a lot of duplicated lines that can just be moved to after their `if`/`else` block to deduplicate them

pyproject.toml
```diff
@@ -143,6 +144,7 @@ scrublet = ["scrublet"]  # Doublet detection
 # Acceleration
 rapids = ["cudf>=0.9", "cuml>=0.9", "cugraph>=0.9"]  # GPU accelerated calculation of neighbors
 dask = ["dask[array]!=2.17.0"]  # Use the Dask parallelization engine
+dask-ml = ["dask-ml"]  # Dask-ML for sklearn-like API
```
flying-sheep: This makes sure that our version constraint gets obeyed when installing `scanpy[dask-ml]`

```suggestion
dask-ml = ["dask-ml", "scanpy[dask]"]  # Dask-ML for sklearn-like API
```

scanpy/preprocessing/_pca.py
selmanozleyen: check_random_state returns a numpy RandomState object but dask expects an int

scanpy/preprocessing/_pca.py
selmanozleyen: I would agree, but since it was auto before, I thought there should've been a good reason for them to change it. But I can also change it again if you think it's a good idea

flying-sheep: - Our default value shouldn’t trigger a warning (even if it’s just a log message here). It’s OK to translate our default value to another value for some or all implementations
- It should be possible for the user to specify all possible values that implementations accept.

E.g. if `svd_solver='auto'` is accepted by some implementation, but we want that implementation to be called with `svd_solver='foo'` by default, our default should be e.g. `svd_solver=None` which then causes us to pass `svd_solver='foo'` to the implementation. If an user then sets `svd_solver='auto'`, that should be passed directly to any implementation that supports it.

scanpy/preprocessing/_pca.py
flying-sheep:
```suggestion
    if not is_dask:
        random_state = check_random_state(random_state)
    elif not isinstance(random_state, int):
        msg = f'random_state needs to be an int, not a {type(random_state).__name__} when passing a dask array'
        raise TypeError(msg)
```

scanpy/preprocessing/_pca.py
```diff
@@ -1,24 +1,24 @@
 from typing import Optional, Union
 
 import numpy as np
+from anndata import AnnData
 from scipy.sparse import issparse, spmatrix
 from scipy.sparse.linalg import LinearOperator, svds
 from sklearn.utils import check_array, check_random_state
 from sklearn.utils.extmath import svd_flip
```
flying-sheep:
```suggestion
from scipy.sparse import issparse, spmatrix
from scipy.sparse.linalg import LinearOperator, svds
from sklearn.utils import check_array, check_random_state
from sklearn.utils.extmath import svd_flip

from anndata import AnnData

```

scanpy/preprocessing/_normalization.py
```diff
@@ -6,18 +6,10 @@
...
-
-    class DaskArray:
-        pass
-
-
 from scanpy import logging as logg
 from scanpy._utils import view_to_actual
-
 from scanpy.get import _get_obs_rep, _set_obs_rep
+from .._compat import DaskArray
```
flying-sheep:
```suggestion

from .._compat import DaskArray
```

scanpy/preprocessing/_pca.py
```diff
@@ -47,32 +48,50 @@ def pca(
...
-          An alternative SciPy solver.
+            An alternative SciPy solver. Not available with dask arrays.
+        `'tsqr'`
+            Only available with *dask* arrays. "tsqr"
+            algorithm from Benson et. al. (2013).
 
-        .. versionchanged:: 1.4.5
-           Default value changed from `'auto'` to `'arpack'`.
+        .. versionchanged:: 1.9.3
+           Default value changed from `'arpack'` to None.
```
flying-sheep, line 91:
```suggestion
        .. versionchanged:: 1.9.3
           Default value changed from `'arpack'` to None.
        .. versionchanged:: 1.4.5
           Default value changed from `'auto'` to`'arpack'`.
```

scanpy/preprocessing/_pca.py
```diff
@@ -47,32 +48,52 @@ def pca(
...
+            An alternative SciPy solver. Not available with dask arrays.
+        `'tsqr'`
+            Only available with *dask* arrays. "tsqr"
+            algorithm from Benson et. al. (2013).
 
+        .. versionchanged:: 1.9.3
+           Default value changed from `'arpack'` to None.
         .. versionchanged:: 1.4.5
-           Default value changed from `'auto'` to `'arpack'`.
+           Default value changed from `'auto'` to`'arpack'`.
```
flying-sheep:
```suggestion
           Default value changed from `'auto'` to `'arpack'`.
```

scanpy/preprocessing/_pca.py
```diff
@@ -320,3 +376,46 @@ def rmatmat(x):
...
+    }
+
+    return _handle_x_args('sklearn', svd_solver, method, method2args, method2default)
+
+
+def _handle_x_args(lib, svd_solver, method, method2args, method2default):
+    changed = False
+
+    if svd_solver not in method2args[method]:
+        changed = True
```
flying-sheep:
```suggestion
    changed = svd_solver not in method2args[method]
    if changed:
```

scanpy/preprocessing/_pca.py
```diff
@@ -320,3 +376,44 @@ def rmatmat(x):
...
+
+    return _handle_x_args('sklearn', svd_solver, method, method2args, method2default)
+
+
+def _handle_x_args(lib, svd_solver, method, method2args, method2default):
+    changed = svd_solver not in method2args[method]
+    if changed:
+        svd_solver = method2default[method]
+
+    if changed:
```
flying-sheep:
```suggestion
```

scanpy/preprocessing/_pca.py
```diff
@@ -189,12 +238,12 @@ def pca(
     elif issparse(X) and zero_center:
         from sklearn.decomposition import PCA
 
-        if svd_solver == "auto":
-            svd_solver = "arpack"
-        if svd_solver not in {'lobpcg', 'arpack'}:
+        if svd_solver is None:
+            svd_solver = 'arpack'
```
flying-sheep: that’ll mean that `changed` will be set to true later, resulting in a warning when using the default.

as said, please make sure that leaving the default doesn’t result in a warning.

you’re almost there, you just need to leave this as `None` until the point where you can actually determine the final solver (in your handle_*args function)

flying-sheep (changes requested):

scanpy/preprocessing/_pca.py
```diff
@@ -320,3 +370,44 @@ def rmatmat(x):
...
+        'PCA': 'arpack',
+        'TruncatedSVD': 'randomized',
+        'PCA (with sparse input)': 'arpack',
+    }
+
+    return _handle_x_args('sklearn', svd_solver, method, method2args, method2default)
+
+
+def _handle_x_args(lib, svd_solver, method, method2args, method2default):
+    changed = svd_solver not in method2args[method]
```
flying-sheep: `None` is not in there, so you still throw a warning by default.

Please add tests that check

1. no warning is thrown when the default (None) is specified
2. a warning is thrown when an inappropriate `svd_solver` is selected

sth like

```suggestion
    changed = svd_solver is not None and svd_solver not in method2args[method]
```

flying-sheep (changes requested):
https://github.com/scverse/scanpy/blob/3c76af915d01da737b03391747129cbc650e54c3/scanpy/preprocessing/_pca.py#L191-L192

at this point, `svd_solver` is `None`, right? Please make this line work with both `None` and the explicit value

selmanozleyen:
@flying-sheep , haven't considered all combinations yet but wanted to check if this way is good for testing warnings.

flying-sheep:
Looks great! You can get rid of `expect_warning`, as you can just simply check `if expected_warning_message is not None` instead, otherwise pretty ideal!

flying-sheep:
You also need to fix the links. `dask.arrays.linalg.svd_compressed` should be `dask.arrays.linalg.svd_compressed`, then the docs will build.

scanpy/tests/test_pca.py
```diff
@@ -107,14 +102,13 @@ def test_pca_warnings(pca_params):
...
-        expected_warning_message,
+        expected_warning,
     ) = pca_params
     A = array_type(A_list).astype('float32')
     adata = AnnData(A)
 
-    if expect_warning:
+    if expected_warning is not None:
+        expected_warning_type, expected_warning_message = expected_warning
         with pytest.warns(expected_warning_type, match=expected_warning_message):
```
flying-sheep:
```suggestion
        expected_warning_type, expected_warning_pattern = expected_warning
        with pytest.warns(expected_warning_type, match=expected_warning_pattern):
```

flying-sheep (approved):
great! once the docs build (and you’re happy with it), that’s ready

PR #2569, opened 2023-07-21
Add check for release notes

flying-sheep:
\  
- [x] Closes N/A

\  
- [x] Release notes not necessary because: It’s a dev process PR

(made comments visible to show what PR authors see)

Current design is that the jobs fail, respectively:
- If there’s no milestone
- If the “Development process” label isn’t there and the “Release notes …” checkbox is not checked, and there are not edits to the relevant release notes file

This is how it looks like if that check fails:

grafik


.github/workflows/check-pr.yml
```diff
@@ -27,5 +27,17 @@ jobs:
     name: Check Milestone
     runs-on: ubuntu-latest
     steps:
-      - if: github.event.pull_request.milestone == null && contains( env.LABELS, 'no milestone' ) == false
-        run: exit 1
+      - name: Check if milestone or “no milestone” label is present
+        run: exit ${{ (github.event.pull_request.milestone != null || contains(env.LABELS, 'no milestone')) && '0' || '1'}}
```
ivirshup: This is much harder to read than the previous version, and I think requires more knowledge of github's expression language than is necessary for the task.

ivirshup (dismissed):
I like the idea of checking this, but also think a lot of small PRs don't need it. Mainly "bug fixes for code that hasn't been released yet" or "changes to docs". Also how does this work for 

I would like a way of overriding that wasn't `dev-process` specific. I do think making the PR template have a checklist is a fine way to start on this.

What about using the checklist on anndata and just checking that all the checkboxes in the PR description have been ticked?

.github/workflows/check-pr.yml
flying-sheep: fixed

flying-sheep:
> Also how does this work for

you seem to have sent this too early

> What about using the checklist on anndata and just checking that all the checkboxes in the PR description have been ticked?

That gives me an idea!

flying-sheep:
OK, this should work. The only issue is that if users check “No release notes necessary” while not checking another box, the “check-relnotes” job still runs.

flying-sheep:
OK, done, now wokrs with regexes

ivirshup (commented):
Just discussed in person but:

* I think we can manually check the test and issue closing ones. The test one should generally be covered by codecov. We've only really had repeated issues with release notes.

.github/pull_request_template.md
```diff
@@ -3,8 +3,7 @@ Thanks for opening a PR to scanpy!
 Please be sure to follow the guidelines in our contribution guide (https://scanpy.readthedocs.io/en/latest/dev/index.html) to familiarize yourself with our workflow and speed up review.
 -->
 
-<!-- Please check (“- [x]”) and fill in the following two boxes -->
+<!-- Please check (“- [x]”) and fill in the following box -->
 - [ ] Closes #
-- [ ] Tests included or not required because:
```
ivirshup: I would like to keep this line (because we often have to ask people for tests), but just don't think we actually need to check it. It would be fine if this line was deleted for some PRs.

Can't seem to suggest a change here, but would just keep it as is.

.github/pull_request_template.md
```diff
@@ -2,3 +2,8 @@
 Thanks for opening a PR to scanpy!
 Please be sure to follow the guidelines in our contribution guide (https://scanpy.readthedocs.io/en/latest/dev/index.html) to familiarize yourself with our workflow and speed up review.
 -->
+
+<!-- Please check (“- [x]”) and fill in the following box -->
+- [ ] Closes #
+<!-- Only check the following box if you did not include release nodes -->
+- [ ] Release notes not necessary because:
```
ivirshup:
```suggestion
<!-- Please check (“- [x]”) and fill in the following boxes -->
- [ ] Closes #
- [ ] Tests included or not required because:
<!-- Only check the following box if you did not include release notes -->
- [ ] Release notes not necessary because:
```

ivirshup (commented):
Would like to keep the "test" checkbox so that people see it is a requirement. I think reviewers already have the tools needed to see if they were added, but authors still need the nudge.

PR #2575, opened 2023-07-25
Simplify tests

flying-sheep:
- Even our `test-min` extra has `pytest-nunit`, so we don’t need a dummy version
- `--internet-tests` can just be replaced by the mark itself.


ivirshup (dismissed):
I'm very against using `addopts` for pytest options. I really don't like it when my interface to `pytest` changes based on the repository I'm working in. There are a few cases where this is okay, but I don't think selecting markers is one of them.

PR #2577, opened 2023-07-28
Mellon

katosh:
This implements `scanpy.external.tl.mellon` using [Mellon](https://github.com/settylab/mellon) to compute cell-state density.

### DIFF
--- a/docs/external/tools.md
+++ b/docs/external/tools.md
@@ -39,3 +39,12 @@
    tl.cyclone
 
 ```
+
+### Cell statistics
+
+```{eval-rst}
+.. autosummary::
+   :toctree: generated/
+
+   tl.mellon
+```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -107,6 +107,7 @@ test-full = [
     "scanpy[harmony]",
     "scanpy[scanorama]",
     "scanpy[scrublet]",
+    "scanpy[mellon]",
 ]
 doc = [
     "sphinx>=4.4,<5", # remove upper bound when theme supports dark mode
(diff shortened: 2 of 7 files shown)


Zethson:
Dear @katosh,

unfortunately, we do not accept contributions to external anymore and are deprecating it. However, we'd happily list mellon in the [scverse ecosystem](https://scverse.org/packages/#ecosystem) registry if all requirements are ticked off.

Thank you very much!

katosh:
@Zethson thank you for the consideration and explanation. I am not sure Mellon would pass the criteria since it does not depend on or explicitly use AnnData although we do recommend using AnnData: https://mellon.readthedocs.io/en/latest/notebooks/basic_tutorial.html
Additionally, it relies on [Palantir](https://github.com/dpeerlab/Palantir) which does also not qualify since it does not have a CI yet.
Do you think we should try making a PR to https://github.com/scverse/ecosystem-packages regardless?

Zethson:
> @Zethson thank you for the consideration and explanation. I am not sure Mellon would pass the criteria since it does not depend on or explicitly use AnnData although we do recommend using AnnData: https://mellon.readthedocs.io/en/latest/notebooks/basic_tutorial.html Additionally, it relies on [Palantir](https://github.com/dpeerlab/Palantir) which does also not qualify since it does not have a CI yet. Do you think we should try making a PR to https://github.com/scverse/ecosystem-packages regardless?

No, I'm afraid that these are hard criteria. However, I can highly recommend that you support AnnData first class :)

katosh:
AnnData a large dependency given that the only interaction is
```python
ad.obs["mellon_log_density"] = mellon.DensityEstimator().fit_predict(ad.obsm["DM_EigenVectors"])
```
I understand the criteria though.

PR #2589, opened 2023-08-03
Fixed wrong order for groups with logreg

Intron7:
Fixes a bug where the `groups` argument would return a wrongly sorted list for logreg in rank_gene_groups.

### DIFF
--- a/scanpy/tools/_rank_genes_groups.py
+++ b/scanpy/tools/_rank_genes_groups.py
@@ -341,12 +341,14 @@ class _RankGenes:
         clf = LogisticRegression(**kwds)
         clf.fit(X, self.grouping.cat.codes)
         scores_all = clf.coef_
-        for igroup, _ in enumerate(self.groups_order):
+        unique_codes = np.unique(self.grouping.cat.codes)
+        for igroup, cat in enumerate(self.groups_order):
             if len(self.groups_order) <= 2:  # binary logistic regression
                 scores = scores_all[0]
             else:
-                scores = scores_all[igroup]
-
+                cat = np.where(self.grouping.cat.categories == cat)[0][0]
+                cat = np.where(unique_codes == cat)[0][0]
+                scores = scores_all[cat]
             yield igroup, scores, None
 
             if len(self.groups_order) <= 2:
--- a/scanpy/tools/_rank_genes_groups.py
+++ b/scanpy/tools/_rank_genes_groups.py
@@ -349,9 +349,9 @@ class _RankGenes:
             else:
                 # cat code is index of cat value in .categories
                 cat_code: int = np.argmax(self.grouping.cat.categories == cat)
-                # index of scores array is index of cat code in array of existing codes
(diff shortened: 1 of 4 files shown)


ivirshup (dismissed):
@Intron7, were you able to come up with a test for this?

Also could it get a release note?

Intron7:
I added a test and a release node

scanpy/tools/_rank_genes_groups.py
```diff
@@ -341,12 +341,14 @@ def logreg(self, **kwds):
...
-        for igroup, _ in enumerate(self.groups_order):
+        unique_codes = np.unique(self.grouping.cat.codes)
+        for igroup, cat in enumerate(self.groups_order):
             if len(self.groups_order) <= 2:  # binary logistic regression
                 scores = scores_all[0]
             else:
-                scores = scores_all[igroup]
-
+                cat = np.where(self.grouping.cat.categories == cat)[0][0]
+                cat = np.where(unique_codes == cat)[0][0]
```
flying-sheep: can you please walk us through the logic here?

what’s the goal, what does each step do?

Intron7: Ok, so the idea is to match the groups to the codes for the logistic regression. Let's say we have the Groups ["2", "3", "1"]. And the categories are named ["0","1","3","2"]. Then, the grouping codes that go into the logistic regression are [1,2,3]. The groups remain ["2", "3", "1"]. The resulting indexes for the coefs_ are [0,1,2]. So, we have to rearrange everything to keep the Groups in the correct order while yielding the correct score. However, I could have used better variable names.

>`cat = np.where(self.grouping.cat.categories == cat)[0][0]`

This looks for the code of the category.

> `cat = np.where(unique_codes == cat)[0][0]`

This looks for the position within the unqiue categories for the regression.

flying-sheep: OK, to simplify:

1. `np.where(haystack == needle)[0][0]` is just `np.argmax(haystack == needle)` (or `list(haystack).index(needle)`), so the ugly `[0][0]` can be dropped.
2. `np.unique(a_categorical.codes)` is just `np.arange(len(a_categorical.categories))` (it sorts its output array), and `list(a_range).index(n) == n`, so the second step is superfluous.

I simplified the code without changing its semantics.

If you intended for it to have different semantics (e.g. you assumed that `np.unique` doesn’t sort the array it returns), your test doesn’t test for that, and your code didn’t do that.

Intron7: There is one issue with this solution. `self.grouping.cat.categories` still has all categories. So we get an error. I updated the test to account for this behaviour. The previous solution does account for this.

flying-sheep: Great! I added some comments, now it’s ready.

flying-sheep (approved):

flying-sheep:
@Intron7 FYI: the release notes were in the wrong file. this is in milestone 1.9.4, so they go in the 1.9.4 file

Issue #2234, opened 2022-04-13
add `mask_obs`/`mask_var` arguments where appropriate


ivirshup:
I think we should introduce a standardized “mask” argument to scanpy functions. This would be a boolean array (or reference to a boolean array in `obs`/ `var`) which masks out certain data entries.

This can be thought of as a generalization of how highly variable genes is handled. As an example:

```python
sc.pp.pca(adata, use_highly_variable=True)
```

Would be equivalent to:

```python
sc.pp.pca(adata, mask="highly_variable")
# or
sc.pp.pca(adata, mask=adata.obs["highly_variable"])
```

One of the big advantages of making this more widespread is that tasks which previously required using `.raw` or creating new anndata objects will be much easier

Some uses for this change:

### Plotting

A big one is plotting. Right now if you want to show gene expression for a subset of cells, you have to manually work with the Matplotlib Axes:

```python
ax = sc.pl.umap(pbmc, show=False)
sc.pl.umap(
    pbmc[pbmc.obs["louvain"].isin(['CD4 T cells', 'B cells', 'CD8 T cells',])],
    color="LDHB",
    ax=ax,
)
```

If a user could provide a mask, this could be reduced, and would make plotting more than one value possible:

```python
sc.pl.umap(
    pbmc,
    color=['LDHB', 'LYZ', 'CD79A’],
    mask=pbmc.obs["louvain"].isin(['CD4 T cells', 'B cells', 'CD8 T cells’,]),
)
```

### Other uses

This has come up before in a few contexts:

* Performing normalization on just some variables https://github.com/scverse/scanpy/issues/2142#issuecomment-1046729522
* Selecting a subset of variables for DE tests: https://github.com/scverse/scanpy/issues/1744
    * See also https://github.com/scverse/scanpy/issues/748
* Changing use_raw https://github.com/scverse/scanpy/issues/1798#issuecomment-819998988

## Implementation

I think this could fit quite well into the `sc.get` getter/ validation functions (https://github.com/scverse/scanpy/issues/828#issuecomment-560072919).

flying-sheep:
@Intron7 said he had experience with this and it’s a really good way to do things fast with dask etc.

- works well in dask
- scale is a good example how to do it
- can’t do in 3rd party like PCA

flying-sheep:
@Intron7 still waiting for your comment here!

Intron7:
[...]

PR #2590, opened 2023-08-03
(feat): Aggregation via group-by in `sc.get`

### LINKED ISSUE #556: AnnData split-apply-combine
`anndata` should get a split apply combine framework. This is something we’ve wanted for a while, but this issue is just establishing a concrete place to track it.

### API

There are a number of split-apply-combine frameworks/ API's out there. Which do we want to emulate, and what are the most important features?

* Access to intermediate elements (e.g. `for k, sub_adata in adata.groupby(obs="celltype")`)
* Efficient reductions and combination
* Nested groupby, group over multiple dimensions
* Lagged/ windowed grouping (e.g. groupby neighbors of a cell, moving average over time)

### What is the correct default return type of a reduction? Can we be flexible about this?

* `AnnData` would be nice for consistency, but do we want to keep all that metadata?
    * Also, how would we handle reductions over non-`X` data? E.g. mean point in PCA space for a cluster
* DataFrames would let us keep labels, but should they be used as "labelled arrays"?
* `xarray.DataArrays` make sense for just labelled arrays

### Reductions and sparse data

How can we make reductions efficient for sparse matrices? One of the most annoying problems when writing code for scanpy and anndata is getting operations to work on both dense and sparse arrays, since sparse arrays have a different API. If reductions and combinatons are included here, then we'll have to be sure sparse arrays work, which could be a lot of extra code. My personal preference for addressing this is to get better support for sparse data with an array API in upstream libraries like pydata/sparse.

## Current and previous attempts/ discussion

* https://github.com/ivirshup/mantis#group-by
* https://github.com/rafaqz/DimensionalData.jl/issues/3#issuecomment-590061903
* https://github.com/theislab/anndata/pull/350#issuecomment-607033414
* https://github.com/theislab/scanpy/pull/1390 (@gokceneraslan)
* #527 (@vals)

Ping @jbloom22, who has also done some work on this.

## Other implementations

* [xarray](http://xarray.pydata.org/en/stable/groupby.html)
  * Modeled on pandas, but has multidimensional groupby
* [pandas](https://pandas.pydata.org/docs/user_guide/groupby.html)
* hail
  * [MatrixTable groupby](https://hail.is/docs/0.2/overview/matrix_table.html#group-by)
  * [Aggregations](https://hail.is/docs/0.2/guides/agg.html#)
* [SplitApplyCombine.jl](https://github.com/JuliaData/SplitApplyCombine.jl)

ilan-gold:
Modified (for `scanpy`) version of https://github.com/scverse/anndata/pull/564.  Fixes https://github.com/scverse/anndata/issues/556

Big points of change:
1. No more tuple-indices and related functionality (i.e., scoring pairwise)
2. Allow for `obs` and `var` group-by +`varm`, `obsm`, `layers` as options for data to aggregate
3. Output is `AnnData` object instead of `DataFrame`
4. `scanpy`-style public API

## TODO (by @ivirshup):

Necessary:

- [x] Docs
- [x] Aggregate along other axis
- [x] Keep grouping cols in result
- [x] Reconsider API for non-anndata version (maybe return a dict of arrays?)
- [ ] Decide on naming convention for `"nonzero"` variations, should this be `"nonzero_count"` so it's a little like `"nanmean"`

Optional, can do later:

- [ ] Weighted (although.... Idk, maybe can skip. Does "weights" affect "count_nonzero"?)
- [ ] Option for keeping around unseen groups, probably needs `fill_value` argument for those values
- [x] Support for `obsm`, `varm`
- [ ] Directly pass Series to groupby
- [ ] More aggregation functions (mean_nonzero, min, max, std, `nan*` variations)
- [ ] Mask argument
- [ ] Dask support


ilan-gold:
Ok @ivirshup I think we're ready to go here.  Thanks for the guidance!

scanpy/get/__init__.py
```diff
@@ -3,3 +3,5 @@
 
 # Private
 from .get import _get_obs_rep, _set_obs_rep
+
+from .aggregated import aggregated, aggregated_from_array
```
ivirshup:
```suggestion
from ._aggregated import aggregated
```

* I am curious about how it works to overwrite the field in the namespace
* I don't think we need to export the `aggregated`

scanpy/get/aggregated.py
```diff
@@ -0,0 +1,401 @@
...
+        df[key] = df[key].cat.remove_unused_categories()
+    return df.groupby(key).first()[_superset_columns(df, key)]
+
+
+@singledispatch
+def aggregated(
+    adata: AnnData,
+    by: str,
+    how: Literal['count', 'mean', 'sum', 'count_mean_var'] = 'count_mean_var',
+    groupby_df_key: Literal['obs', 'var'] = 'obs',
```
ivirshup:
```suggestion
    *,
    groupby_df_key: Literal['obs', 'var'] = 'obs',
```

I'd like to limit how many of these can be positional.

scanpy/get/aggregated.py
```diff
@@ -0,0 +1,401 @@
...
+        df = df[df[key].isin(key_set)]
+    if df[key].dtype.name == 'category':
+        df[key] = df[key].cat.remove_unused_categories()
+    return df.groupby(key).first()[_superset_columns(df, key)]
+
+
+@singledispatch
+def aggregated(
+    adata: AnnData,
+    by: str,
```
ivirshup:
```suggestion
    by: str,
```

I would like to be able to groupby multiple keys, e.g. "donor + cell type".

scanpy/get/aggregated.py
```diff
@@ -0,0 +1,401 @@
...
+        key (str): Column to be grouped on.
+        key_set (List[str]): values in the `key` column to keep before groupby.
+
+    Returns:
+        pd.DataFrame: Grouped-by Dataframe.
+    """
+    df = df.copy()
+    if key_set is not None:
+        df = df[df[key].isin(key_set)]
+    if df[key].dtype.name == 'category':
```
ivirshup:
```suggestion
    if pd.api.types.is_categorical_dtype(df[key]):
```

scanpy/get/aggregated.py
```diff
@@ -0,0 +1,401 @@
...
+    weight_key: Optional[str] = None,
+    key_set: Optional[Iterable[str]] = None,
+    dof: int = 1,
+    layer: Optional[str] = None,
+    obsm: Optional[str] = None,
+    varm: Optional[str] = None,
+) -> AnnData:
+    """Aggregate data based on one of the columns of one of the axes (`obs` or `var`).  If none of `layer`, `obsm`, or `varm` are passed in, `X` will be used for aggregation data.
+
+    Args:
```
ivirshup: This isn't rendering correctly, probably because it's not numpy doc style.

image

Also the types get auto populated from the annotations.

scanpy/get/aggregated.py
```diff
@@ -0,0 +1,401 @@
...
+        adata (AnnData): AnnData to be aggregated.
+        by (str): Key of the column to be grouped-by.
+        how (Literal[&#39;count&#39;, &#39;mean&#39;, &#39;sum&#39;, &#39;count_mean_var&#39;], optional):  How to aggregate. Defaults to 'count_mean_var'.
```
ivirshup: I think I would go with something like `func` instead of `how`. This is going off of:

* https://github.com/ml31415/numpy-groupies
* https://flox.readthedocs.io/en/latest/aggregations.html

I would also like to be able to calculate multiple functions in on pass, so like:

```python
AggType = Literal["mean", "var", "sum", "count"]

func: AggType | list[AggType]
```

But this is a bit more work.

scanpy/get/aggregated.py
```diff
@@ -0,0 +1,401 @@
...
+        groupby_df_key (Literal[&#39;obs&#39;, &#39;var&#39;], optional): Axis on which to find group by column. Defaults to 'obs'.
```
ivirshup: I would go for `dim`, similar to `xarray`. I think `axis: int` is a little more common within anndata/ scanpy though.

scanpy/get/aggregated.py
```diff
@@ -0,0 +1,401 @@
...
+        weight_key (Optional[str], optional): Key of the `groupby_df_key` containing weights for a weighted sum aggregation. Defaults to None.
+        key_set (Optional[Iterable[str]], optional): Subset of groupby_df_key on which to filter. Defaults to None.
```
ivirshup: Could we do a `mask` argument for this instead? Like:

* https://github.com/scverse/scanpy/issues/2234

scanpy/get/aggregated.py
```diff
@@ -0,0 +1,401 @@
...
+        dof (int, optional): Degrees of freedom for variance. Defaults to 1.
```
ivirshup: I'm wondering if we can remove this. I think this would only ever apply to `var`, and the value should almost always be 1, maybe sometimes 0. I would even consider just say `var` and `var_ddof0`. I think I've seen other packages do this.

scanpy/get/aggregated.py
ilan-gold: I was under the impression that this was what we wanted to pare back? Or you would just instead of creating a multi-index, create some sort of concatenated representation?

scanpy/get/aggregated.py
ilan-gold: I think `dim` makes more sense.  The `axis` argument does not appear to be really used with `AnnData` objects directly, but always when working with `X` (judging by a `ctrl+f` of the `scanpy` repo).

scanpy/get/aggregated.py
ilan-gold: This seems similar to `key_set`.  Perhaps a superset of the functionality offered by `key_set`.  Perhaps we should remove that in favor of this?

scanpy/get/aggregated.py
ilan-gold: Are you saying you would have two different arguments, `var` and `var_ddof0`?  Why not just have one?

scanpy/get/aggregated.py
ilan-gold: Have a look!

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+    elif obsm is not None:
+        data = adata.obsm[obsm]
+        write_to_xxxm = True
+    elif layer is not None:
+        data = adata.layers[layer]
+        if dim == 'var':
+            data = data.T
+    elif (
+        dim == 'var'
+    ):  # i.e., all of `varm`, `obsm`, `layers` are None so we use `X` which must be transposed
```
flying-sheep:
```suggestion
    elif dim == 'var':
        # i.e., all of `varm`, `obsm`, `layers` are None so we use `X` which must be transposed
```

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+    Returns
+    -------
+    Aggregated :class:`~anndata.AnnData`.
+    """
+    data = adata.X
+    write_to_xxxm = None
+    if varm is not None:
+        data = adata.varm[varm]
+        write_to_xxxm = True  # the data will have to be transposed so this is accurate
+    elif obsm is not None:
```
flying-sheep: what if `varm` and `obsm` is not `None`? there should be some validation that only one is set if that’s what the API supports.

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+    data_dict = {
+        'layers': {},
+        'X': None,
+        'obsm': {},
+    }
+    func_set = func
+    write_key = 'obsm' if write_to_xxxm else 'layers'
+    if not isinstance(func, list):
+        func_set = [func]
+    func_set = set(func_set)
```
flying-sheep: Parameter signatures shouldn’t contain concrete container classes (except for `tuple`), and neither should you use them in typechecks. Be lenient in what you accept (in this case a string or an iterable of strings) and strict in what you return (`list[X]` is fine as a return type). Therefore the parameter type should be `Iterable[AggType] | ArgType` here and you can do that simply like this:

```suggestion
    funcs = set([func] if isinstance(func, str) else func)
    write_key = 'obsm' if write_to_xxxm else 'layers'
```

afterwards you should validate that all entries of `funcs` are handled, e.g. by doing

```py
if unknown := funcs - set(get_args(AggType)):
    raise ValueError(f'… {unknown} …')
``` 

(see [`typing.get_args`](https://docs.python.org/3/library/typing.html#typing.get_args))

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+    columns = []
+    groupy_key_codes = df[groupby_key].astype('category')
+    for key in df:
+        if key != groupby_key:
+            key_codes = df[key].astype('category')
+            if all(
+                [
+                    key_codes[groupy_key_codes == group_key_code].nunique() == 1
+                    for group_key_code in groupy_key_codes
+                ]
```
flying-sheep: Don’t create an unnecessary list here. If you leave off the brackets, you pass a generator object to the function that lazily calculates values one by one. Since `all` exits early when it encounters `False`, the following code therefore does less work when that happens (and is formatted prettier).

```suggestion
                key_codes[groupy_key_codes == group_key_code].nunique() == 1
                for group_key_code in groupy_key_codes
```

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+        if self._key_set is not None:
+            keys, key_index, df_index, weight_value = _filter_indices(
+                self._key_set, keys, key_index, df_index, weight_value
+            )
+        self._key_index = key_index  # passed to count and count_mean_var to avoid re-extracting in the latter
+        return keys, key_index, df_index, weight_value
+
+
+def _power(X, power):
+    return X ** power if isinstance(X, np.ndarray) else X.power(power)
```
flying-sheep: why is this necessary? spmatrix supports `**` just like ndarray.

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+                    remap[j] = i
+                keys = [keys[j] for j in keep]
+                key_index = np.array(
+                    [remap[i] for i in key_index[mask]], dtype=np.int64
+                )
+                df_index = df_index[mask]
+                if weight_value is not None:
+                    weight_value = weight_value[mask]
+            return keys, key_index, df_index, weight_value
+
```
flying-sheep:
```suggestion
    def _filter_indices(keys, key_index, df_index, weight_value=None):
        keep = [i for i, k in enumerate(keys) if k in set(self._key_set)]
        if len(keep) == 0:
            raise ValueError("No keys in key_set found in adata.obs[key].")
        elif len(keep) < len(keys):
            mask = np.in1d(key_index, keep)
            remap = np.zeros(len(keys), dtype=np.int64)
            for i, j in enumerate(keep):
                remap[j] = i
            keys = [keys[j] for j in keep]
            key_index = np.array(
                [remap[i] for i in key_index[mask]], dtype=np.int64
            )
            df_index = df_index[mask]
            if weight_value is not None:
                weight_value = weight_value[mask]
        return keys, key_index, df_index, weight_value

    def _extract_indices(self):
```

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+        keys, key_index = np.unique(_ndarray_from_seq(key_value), return_inverse=True)
+        df_index = np.arange(len(key_index))
+        if self._weight is None:
+            weight_value = None
+        else:
+            weight_value = self._weight.values[df_index]
+        if self._key_set is not None:
+            keys, key_index, df_index, weight_value = _filter_indices(
+                self._key_set, keys, key_index, df_index, weight_value
+            )
```
flying-sheep:
```suggestion
            keys, key_index, df_index, weight_value = self._filter_indices(
                keys, key_index, df_index, weight_value
            )
```

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+            weight_value = np.ones(len(key_index))
+        A = coo_matrix(
+            (weight_value, (key_index, df_index)),
+            shape=(len(keys), self._data.shape[0]),
+        )
+        if normalize:
+            n_row = A.shape[0]
+            row_sums = np.asarray(A.sum(axis=1))
+            D = dia_matrix(((row_sums.T**-1), [0]), shape=(n_row, n_row))
+            A = D * A
```
flying-sheep: here and afterwards you perform operations with this coo_matrix. I’m a bit rusty, but isn’t it better to convert these to some more efficient format before doing things?

Or is there another reason why you leave it as coo_matrix?

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+            A_unweighted, _ = Aggregate(
+                groupby=self._groupby,
+                data=self._data,
+                weight=self._weight,
+                key_set=self._key_set,
+            )._sparse_aggregator()
+            mean_unweighted = utils.asarray(A_unweighted * self._data)
+            sq_mean = 2 * mean_ * mean_unweighted + mean_unweighted**2
+        var_ = mean_sq - sq_mean
+        precision = 2 << (42 if self._data.dtype == np.float64 else 20)
```
flying-sheep: please explain magic values in a comment.

scanpy/tests/test_aggregated.py
```diff
@@ -0,0 +1,348 @@
...
+            [11, 8],  # c
+            [12, 6],  # d
+        ],
+        dtype=np.float32,
+    )
+    data_dense = X
+    if dim == 'obs':
+        data_sparse_mat_dict = {data_key: {'test': csr_matrix(X)}}
+        adata_sparse = ad.AnnData(
+            **{'obs': df_groupby, 'var': df_base, **data_sparse_mat_dict}
```
flying-sheep: please do this everywhere:

```suggestion
            obs=df_groupby, var=df_base, **data_sparse_mat_dict
```

scanpy/tests/test_aggregated.py
```diff
@@ -0,0 +1,317 @@
...
+            *[[0, -2], [1, 13], [2, 1]],  # v
+            *[[3, 12], [4, 2]],  # w
+            *[[5, 11], [6, 3], [7, 10]],  # a
+            *[[8, 4], [9, 9]],  # b
+            *[[10, 5], [11, 8]],  # c
+            [12, 6],  # d
+        ],
+        dtype=np.float32,
+    )
+    data_dense = X
```
flying-sheep: a lot of this looks identical to the test data below. please deduplicate so I can see at a glance if the test datasets are identical or if not, where the differences are.

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,437 @@
...
+        )
+        if normalize:
+            n_row = A.shape[0]
+            row_sums = np.asarray(A.sum(axis=1))
+            D = dia_matrix(((row_sums.T**-1), [0]), shape=(n_row, n_row))
+            A = D * A
+        return A, keys
+
+    def _extract_indices(self):
+        def _filter_indices(key_set, keys, key_index, df_index, weight_value=None):
```
flying-sheep: please explain on a high level what these do (in comments or docstring)

flying-sheep (changes requested):
Is singledispatch the right tool here? Seems like you have a public function operating on `AnnData` and a private function operating on arrays/matrices. Maybe it should just be that.

I didn’t check the tests yet, since they contain too much duplication. Please fix that, then I can take a look.

scanpy/tests/test_aggregated.py
ilan-gold: I will need to think harder about this.  I did make quite an attempt to have one parametrized test but it did not go well.  There are just so many subtle differences in using different locations e.g., `obsm/test_data` vs. `X` that it was getting too confusing with the conditionals.  Perhaps I should just rewrite the tests completely from where they were when I found the PR.

scanpy/get/_aggregated.py
ilan-gold: I don't know.  I blackboxed some of the code that (I thought) was already reviewed and didn't requiring changing (basically the math), but it seems that more is required of it.  I'll update this and address the other comments on the old code.

scanpy/get/_aggregated.py
ilan-gold: Ditto for old code.  Will fix.

scanpy/get/_aggregated.py
ilan-gold: Ditto for old code.  Will fix.

scanpy/get/_aggregated.py
ilan-gold: Ditto for old code.  Will review this and add docstrings.

scanpy/get/_aggregated.py
ilan-gold: This was the sort of thing I did not touch from the old PR because I kind of assumed that it was already reviewed.    I'll need to look at this.

scanpy/get/_aggregated.py
ilan-gold: Ditto for old code.  Will add (if I understand).

ilan-gold (commented):
@flying-sheep Thanks for the review.   It looks like I'll need to review the code I didn't touch closer from after porting the PR.  I'll try to figure out something for the tests as well.

scanpy/tests/test_aggregated.py
ilan-gold: Got it.

flying-sheep:
> It looks like I'll need to review the code I didn't touch closer from after porting the PR. 

if you add something like `TODO: document (unchanged from original code)` to the relevant parts, that’s also OK. It doesn’t need to be your responsibility to figure out what that code does and document it.

scanpy/tests/test_aggregated.py
flying-sheep: don’t overthink it. I’m mostly talking about the fact that those lines seem absolutely identical:

<table>
<tr>
 <td>

https://github.com/scverse/scanpy/blob/cdbc2288098830a040f9c4712a617db7933b9dfb/scanpy/tests/test_aggregated.py#L23-L54

 <td>

https://github.com/scverse/scanpy/blob/cdbc2288098830a040f9c4712a617db7933b9dfb/scanpy/tests/test_aggregated.py#L192-L223

</table>

ilan-gold: Ok, I see what you mean now then.  These can certainly just be fixtures.

ilan-gold:
Great to know @flying-sheep!

ilan-gold:
@flying-sheep Do you want me to resolve your comments as I addressed them or do you prefer to do that yourself?  I've seen it done both ways.  Once I know, I'll either resolve or not and then re-request your review.  The only thing I'm personally still curious about is if you think the tests are "too" duplicated still, but there might be other new things/poorly-fixed old things to look at.

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,492 @@
...
+    obs_var_dict = {'obs': _df_grouped(groupby_df, by, key_set), 'var': no_groupby_df}
+    data_dict = {
+        'layers': {},
+        'X': None,
+        'obsm': {},
+    }
+    write_key = 'obsm' if write_to_xxxm else 'layers'
+    funcs = set([func] if isinstance(func, str) else func)
+    if unknown := funcs - set(get_args(AggType)):
+        raise ValueError(f'… {unknown} …')
```
flying-sheep, line 460: This needs an actual error message haha

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,488 @@
...
+        if self._key_set is not None:
+            keys, key_index, df_index, weight_value = self._filter_indices(
+                keys, key_index, df_index, weight_value
+            )
+        self._key_index = key_index  # passed to count and count_mean_var to avoid re-extracting in the latter
+        return keys, key_index, df_index, weight_value
+
+
+def _power(X: Array, power: Union[float, int]) -> Array:
+    """Generate elementwise power of a matrix.  Needed for sparse matrices because they do not support ** so the `.power` function is used.
```
flying-sheep: As I already said: I’m pretty sure that’s a lie, sparse matrices *do* support `**`. Why do you think they don’t?

```python
>>> import scipy.sparse as sp
>>> print(sp.csr_matrix([[1, 0], [.5, .2]]) ** 2)
  (0, 0)        1.0
  (1, 1)        0.04000000000000001
  (1, 0)        0.6

File: scanpy/get/_aggregated.py [hunk 24]
```diff
@@ -0,0 +1,488 @@
...
+    obs_var_dict = {'obs': _df_grouped(groupby_df, by, key_set), 'var': no_groupby_df}
+    data_dict = {
+        'layers': {},
+        'X': None,
+        'obsm': {},
+    }
+    write_key = 'obsm' if write_to_xxxm else 'layers'
+    funcs = set([func] if isinstance(func, str) else func)
+    if unknown := funcs - set(get_args(AggType)):
+        raise ValueError(f'… {unknown} …')
```
flying-sheep: This needs an actual error message haha

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,486 @@
...
+        weight_key=weight_key,
+        key_set=key_set,
+        func=func,
+        dof=dof,
+    )
+
+
+@aggregated.register(np.ndarray)
+@aggregated.register(spmatrix)
+def aggregated_from_array(
```
flying-sheep: As said:

> Is singledispatch the right tool here? Seems like you have a public function operating on AnnData and a private function operating on arrays/matrices. Maybe it should just be that.

I.e. `aggregated` can be a regular function, calling the helper function `aggregated_from_array` with no singledispatch going on.

They have different signatures and only one of them is supposed to be called by users, right?

flying-sheep (changes requested):
Shaping up very well! you’re right about the tests, there really is quite some copy and pasted code …

I’ll go over the minor parts but the tests should really be deduplicated and probably split up. Seems like each of the two test functions contain the same sequence of basically independent tests.

scanpy/tests/test_aggregated.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+
+    # superset columns can be kept but not subsets
+    assert 'key_superset' in getattr(stats_sparse, dim)
+    assert 'key_subset' not in getattr(stats_sparse, dim)
+
+    assert np.allclose(
+        getattr(stats_sparse, dim)['count'],
+        getattr(stats_sparse, dim)['count'],
+    )
```
flying-sheep: tautology

scanpy/tests/test_aggregated.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+@pytest.mark.parametrize('dim', ['obs', 'var'])
+def test_groupby_X(dim, df_base, df_groupby, X):
+    if dim == 'obs':
+        adata_sparse = ad.AnnData(obs=df_groupby, var=df_base, X=csr_matrix(X))
+        adata_dense = ad.AnnData(obs=df_groupby, var=df_base, X=X)
+    else:
+        adata_sparse = ad.AnnData(obs=df_base, var=df_groupby, X=X.T)
+        adata_dense = ad.AnnData(obs=df_base, var=df_groupby, X=csr_matrix(X).T)
```
flying-sheep: those are flipped, the `adata_sparse` should contain the sparse matrix

scanpy/tests/test_aggregated.py
```diff
@@ -0,0 +1,273 @@
...
+    )
+
+    # superset columns can be kept but not subsets
+    assert 'key_superset' in getattr(stats_sparse, dim)
+    assert 'key_subset' not in getattr(stats_sparse, dim)
+
+    assert np.allclose(
+        getattr(stats_sparse, dim)['count'],
+        getattr(stats_sparse, dim)['count'],
+    )
```
flying-sheep: tautology again

flying-sheep (commented):
@ilan-gold FYI: I fixed those problems

flying-sheep:
pre-commit.ci autofix

scanpy/get/_aggregated.py
ilan-gold: @flying-sheep right, I forgot to address this point.  I actually do not have a strong feeling as to what this needs to be.  I think @ivirshup suggested the `singledispatch` so he can weigh in.

ilan-gold:
@flying-sheep I tried dedup-ing!

ilan-gold:
I will clean up a few other things, but you're welcome to have a look at that now if you'd like.

scanpy/get/_aggregated.py
ilan-gold: I will update the comment, but it is true for non-square matrices.

scanpy/get/_aggregated.py
ilan-gold: Should be done.

flying-sheep (approved):
Beautiful! Checking the tests I see this:

```
scanpy/tests/test_aggregated.py: 18 warnings
  /opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/scanpy/get/_aggregated.py:154: RuntimeWarning: divide by zero encountered in divide
    var_ *= (count_ / (count_ - dof))[:, np.newaxis]

scanpy/tests/test_aggregated.py: 18 warnings
  /opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/site-packages/scanpy/get/_aggregated.py:154: RuntimeWarning: invalid value encountered in multiply
    var_ *= (count_ / (count_ - dof))[:, np.newaxis]
```

is that expected?

You might be able to run `pytest -Werror scanpy/tests/test_aggregated.py` to see where they happen.

ilan-gold:
@flying-sheep This is part of the "blackboxed" code I took from the old PR.  However, I think this only occurs when you are trying to calculate the variance on a n-sample group with n-degrees of freedom (so, for example, 1 dof with 1 sample point).  What do you think should be done in this case?  I think perhaps a check for `dof` and `count_` being equal should suffice but I am not sure what should be done in that case.  Throw a warning ourselves?

flying-sheep:
Yeah, maybe skipping the broken calculation and instead throwing our own, more expressive warning is a good idea. But it’s late, I’m happy to finish this up on Thursday

scanpy/get/_aggregated.py
flying-sheep: Ah, the issue is matrix interface vs array interface, got it.

scanpy/get/_aggregated.py
flying-sheep: OK, pinging him!

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,486 @@
...
+        mean_sq = utils.asarray(A @ _power(self.data, 2))
+        if self.weight is None:
+            sq_mean = mean_**2
+        else:
+            A_unweighted, _ = Aggregate(
+                groupby=self.groupby,
+                data=self.data,
+                weight=self.weight,  # TODO: why pass weights when creating unweighted A?
+                key_set=self.key_set,
+            )._sparse_aggregator()
```
flying-sheep: So the difference to above is that it doesn’t call `_sparse_aggregator` with `normalize=True`. So this should probably be simply:

```suggestion
            A_unweighted, _ = self._sparse_aggregator()
```

scanpy/get/_aggregated.py
```diff
@@ -28,7 +30,6 @@ class Indices(NamedTuple):
     weight_value: pd.Series | Array | None
 
 
-@dataclass
```
flying-sheep: Stick with this. You can use `default_factory`, `InitVar` and/or `__post_init__` to do whatever you want, no need to go away from data classes.

flying-sheep (changes requested):
Also make sure if you add `axis: Literal[0, 1]` that you also add `dim: Literal["obs", "var"]` as an aternative

ivirshup:
This is getting pretty close to done. I've updated the comment at the top with a todo list. All the "extras" can be moved to an issue on merge. 

Let me know if any of those aren't actually optional.

I'm currently inclined to drop support for `weight`. Checking older commits, it doesn't seem to actually work for the variance calculation (returns all zeros), and isn't really something we support elsewhere. It would also make adding things like "median" significantly more complicated in the future.

flying-sheep (changes requested):
Please either revert 0aef147e7a4bef0244a099f1e19ff51d034a9c01 or allow both as in #2771, or wait for https://github.com/scverse/anndata/issues/1245

also as I asked before: why go away from dataclasses?

ivirshup:
> also as I asked before: why go away from dataclasses?

I don't think that switching away from data classes removed any meaningful functionality here, but having to use `default_factory`, `InitVar`, and/or `__post_init__` would add more complexity.

I don't think that there being some internal data classes is important here, especially since it's not user visible and may change at any time anyways. I have a few ideas for ways to change the implementation to add more methods, none of which are compatible with `Aggregate` being a data class.

* One path forward just removes the class entirely, since it doesn't do much now
* The other uses a number of cached properties, which I don't think make a ton of sense to use with dataclasses

Is there some functionality the data class was adding that I'm missing?

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,382 @@
...
+    return X**power if isinstance(X, np.ndarray) else X.power(power)
+
+
+@singledispatch
+def aggregate(
+    adata: AnnData,
+    by: str | list[str],
+    func: AggType | Iterable[AggType],
+    *,
+    axis: Literal[0, 1] | None = None,
```
flying-sheep: Since we have a consensus now:

```suggestion
    axis: Literal["obs", 0, "var", 1] | None = None,
```

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,382 @@
...
+        var: 'n_cells'
+        layers: 'mean', 'count_nonzero'
+
+    Note that this filters out any combination of groups that wasn't present in the original data.
+    """
+    if axis not in [0, 1, None]:
+        raise ValueError(f"axis must be one of 0 or 1, was '{axis}'")
+    # TODO replace with get helper
+    if mask is not None:
+        mask = _check_mask(adata, mask, ["obs", "var"][axis])
```
flying-sheep:
```suggestion
    axis, axis_name = _handle_axis(axis)
    if mask is not None:
        mask = _check_mask(adata, mask, axis_name)
```

scanpy/get/_aggregated.py
```diff
@@ -0,0 +1,382 @@
...
+        data = adata.obsm[obsm]
+    elif layer is not None:
+        data = adata.layers[layer]
+        if axis == 1:
+            data = data.T
+    elif axis == 1:
+        # i.e., all of `varm`, `obsm`, `layers` are None so we use `X` which must be transposed
+        data = data.T
+
+    dim_df = getattr(adata, ["obs", "var"][axis])
```
flying-sheep:
```suggestion
    dim_df = getattr(adata, axis_name)
```

flying-sheep (changes requested):
Regarding data classes, as said in person: I tend to use them when

- they basically just mean I can delete the `__init__` function and maybe add a `= field(...)`
- instances of the class actually end up as dict keys, being compared or so

The former is almost the case, at the time I wrote it, it was the case.

Let’s get the consensus `axis` design into this!

flying-sheep (approved):

flying-sheep:
`sparse_indicator` doesn’t have its `weights` branches hit at all, maybe we should remove that? Or will this be used at some point?

ivirshup:
> sparse_indicator doesn’t have its weights branches hit at all, maybe we should remove that? Or will this be used at some point?

I think it will be used at some point, but also happy to remove.

I think parameterizing `test_aggregate_axis_specification` is overkill for what the test does.

flying-sheep:
> I think it will be used at some point, but also happy to remove.

OK, good to know! Then this PR is fine as far as I’m concerned.

PR #2595, opened 2023-08-04
Make helpers work with dask

flying-sheep:
Preparation for more thorough testing of dask support

- [x] move `DaskArray` to `_compat` to have it available everywhere
- [x] test everything using `array_type` or similar with the same fixture
- [x] Add Dask support to `is_constant` (which uses above fixture)

the code in [scanpy/metrics/_common.py](https://github.com/scverse/scanpy/pull/2595/files#diff-caabb542dafdc95621693d71cb6a514af1457c05926438e307d3f8107bf91401) has been moved from [scanpy/metrics/_gearys_c.py](https://github.com/scverse/scanpy/pull/2595/files#diff-1ff58f43272e7b1451de7df9813a0d20aba57f55b23d38b2d46e309d75c2879b) without changes.

the only real meat this has is in [scanpy/_utils/compute/is_constant.py](https://github.com/scverse/scanpy/pull/2595/files#diff-e2c27335c5bfbbfb5ae9ac042c09411bfaca7d22340c257974d5d2bc68aea677):

https://github.com/scverse/scanpy/blob/f4f4185709b438e9cfe56db806a449634e214756/scanpy/_utils/compute/is_constant.py#L127-L134

it’s not a great implementation since it concatenates everything along an axis, but that makes it trivially correct.

### DIFF
--- a/scanpy/_utils/compute/is_constant.py
+++ b/scanpy/_utils/compute/is_constant.py
@@ -5,6 +5,8 @@ import numpy as np
 from numba import njit
 from scipy import sparse
 
+from ..._compat import DaskArray
+
 
 @singledispatch
 def is_constant(a, axis=None) -> np.ndarray:
@@ -41,6 +43,7 @@ def is_constant(a, axis=None) -> np.ndarray:
 
 
 @is_constant.register(np.ndarray)
+@is_constant.register(DaskArray)
 def _(a, axis=None):
     # Should eventually support nd, not now.
     if axis is None:
--- a/scanpy/_utils/compute/is_constant.py
+++ b/scanpy/_utils/compute/is_constant.py
@@ -127,7 +127,7 @@ def _(a: sparse.csc_matrix, axis: int | None = None) -> bool | NDArray[np.bool_]
 @is_constant.register(DaskArray)
 def _(a: DaskArray, axis: int | None = None) -> bool | NDArray[np.bool_]:
     if axis is None:
-        v = a[tuple(0 for _ in range(a.ndim))]
+        v = a[tuple(0 for _ in range(a.ndim))].compute()
         return (a == v).all()
     _check_axis_supported(axis)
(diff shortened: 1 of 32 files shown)


scanpy/_utils/compute/is_constant.py
```diff
@@ -1,13 +1,25 @@
...
+from numpy.typing import NDArray
 from numba import njit
 from scipy import sparse
 
+from ..._compat import DaskArray
+
+
+def _check_axis_supported(axis: Any) -> None:
+    if not isinstance(axis, Integral):
+        raise TypeError("axis must be integer or None.")
```
Zethson: Just below you're testing whether the axis is 1 or 0.

Maybe simplify this by just raising a ValueError if the axis is not in `(None, 0, 1)`?

scanpy/tests/test_utils.py
```diff
@@ -64,3 +62,32 @@ def test_is_constant(array_type):
...
+        pytest.param(0, [True, True, False, False], id='0'),
+        pytest.param(1, [False, False, True, True, False, True], id='1'),
+    ],
+)
+@pytest.mark.parametrize('block_type', [np.array, csr_matrix])
+def test_is_constant_dask(axis, expected, block_type):
+    import dask.array as da
+
+    if (axis is None) and block_type is csr_matrix:
+        pytest.skip('dask has weak suppoer for scipy sparse matrices')
```
Zethson:
```suggestion
        pytest.skip('Dask has weak support for scipy sparse matrices')
```

Zethson (approved):
LGTM to my novice eyes

scanpy/_utils/compute/is_constant.py
flying-sheep: I clarified that different kinds of errors are supposed to be thrown and made sure the message matches the check.

PR #2601, opened 2023-08-07
fixed `get.rank_genes_groups_df` with logreg

Intron7:
`get.rank_genes_groups_df` didn't work when the method was "logreg". Now it works as intended


Zethson (approved):
Thanks!

PR #2605, opened 2023-08-08
run doctests

flying-sheep:
This PR adds the `--doctest-modules` flag to pytest and makes changes necessary to allow the doctests to run.

Those changes include:

- Fixing verbosity
- using [pytest-doctestplus](https://github.com/scientific-python/pytest-doctestplus) to skip doctests entirely or when their dependencies are absent
- fixing broken doctests (the bulk of the changes)


ivirshup:
Converted to a draft while it's blocked

flying-sheep:
It’s unblocked again apparently!

flying-sheep:
@ivirshup any way to force Azure to clear its cache or use a different runner? The “invalid instruction” error here probably comes from using a binary wheel compiled for a newer CPU.

/edit: wow, 9 attempts. Maybe just dropping Python 3.8 will get us there.

Zethson:
Scientific Python even deprecated Python 3.9 this month. Kill 3.8 please...

flying-sheep:
For sure, but it won’t help with this issue, I had to merge https://github.com/scverse/scanpy/pull/2687 for that.

scanpy/testing/_pytest/marks.py
```diff
@@ -1,40 +1,54 @@
...
+    louvain='louvain',
+    skmisc='scikit-misc',
+    fa2='fa2',
+    igraph='igraph',
+    dask='dask',
+    dask_ml='dask-ml',
+    zarr='zarr',
+    zappy='zappy',
+    pybiomart='pybiomart',
+    gprofiler='gprofiler-official',
```
ivirshup: We generally use `"`

scanpy/tests/conftest.py
```diff
@@ -5,31 +5,18 @@
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-import matplotlib as mpl
-
-mpl.use('agg')
-from matplotlib import pyplot
-from matplotlib.testing.compare import compare_images, make_test_filename
```
ivirshup: Why import these inside of functions?

ivirshup (commented):
Is this ready for review? I think it mostly looks good.

> drop 3.8

If our dependencies have dropped support we can too.

---

Have you thought about doctestplus in the context of this PR?

scanpy/testing/_pytest/marks.py
flying-sheep: No we don’t. Scanpy predominately uses `'`, I checked by trying both and letting git count the diff lines. I don’t have the exact statistics in mind but it’s something like 3/4 or 2/3 `'`. In our last conversation about this, you said you’d prefer going the path most code uses, therefore this change.

I don’t care, I just want us to make a formatter unify this ASAP. (We can use Ruff for `'` if we decide to go this way)

scanpy/tests/conftest.py
flying-sheep: Because one should never do global state changes in modules.

I would prefer to never import `pyplot` in library code at all since it does that.

scanpy/testing/_pytest/marks.py
ivirshup: https://github.com/scverse/scanpy/pull/2701

flying-sheep: Great! Let’s merge this one first, then format it so it’s less work.

flying-sheep:
> If our dependencies have dropped support we can too.

You mean like anndata? :laughing: 

> Is this ready for review? I think it mostly looks good.

yeah!

flying-sheep:
@ivirshup ping

flying-sheep:
TODO: check out doctest-plus, if it helps enough, switch to it, then merge

flying-sheep:
OK, doctest-plus can’t even handle relative imports, and doesn’t respect `--import-mode`.

It needs quite some work before it works for us.

PR #2610, opened 2023-08-11
Improve support for n_pcs if bigger than settings.N_PCS

Intron7:
This should prevent an unexpected behaviour when `n_pcs` is bigger than `settings.N_PCS`

Fixes some parts of the doc strings for neighbours and tsne

### DIFF
--- a/scanpy/tools/_utils.py
+++ b/scanpy/tools/_utils.py
@@ -10,7 +10,7 @@ doc_use_rep = """\
 use_rep
     Use the indicated representation. `'X'` or any key for `.obsm` is valid.
     If `None`, the representation is chosen automatically:
-    For `.n_vars` < :attr:`scanpy.settings.N_PCS`(default: 50), `.X` is used, otherwise 'X_pca' is used.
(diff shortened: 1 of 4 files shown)


scanpy/tools/_utils.py
```diff
@@ -26,6 +26,7 @@ def _choose_representation(adata, use_rep=None, n_pcs=None, silent=False):
         settings.verbosity = 1
     if use_rep is None and n_pcs == 0:  # backwards compat for specifying `.X`
         use_rep = 'X'
+    print(settings.N_PCS, type(settings.N_PCS))
```
Zethson:
```suggestion
```

Zethson (approved):
Thanks!

1. Release note please.
2. Why do we have `N_PCS` in the settings? Wasn't aware of that actually

Intron7:
> 1. Release note please.

I'll do so

> 2. Why do we have `N_PCS` in the settings? Wasn't aware of that actually

I don't know I just came across this while working on RSC and thought the behavior was not what I expected. I hope that this is more in line with what the user intended.

PR #2612, opened 2023-08-14
changed  hvg with PR  to work with numba

Intron7:
The pearson residuals implementation for hvg is currently very slow and memory inefficient. I switched it to a numba kernel for csc and dense F-continous arrays.

It's based on the cuda-kernel in [rapids-singlecell](https://github.com/scverse/rapids_singlecell/blob/main/src/rapids_singlecell/cunnData_funcs/_hvg.py#L14-L258). For 90000 cells we go from 24 seconds to less than 5 with the new implementation. For smaller datasets we don't see a speedup.

### DIFF
--- a/scanpy/experimental/pp/_highly_variable_genes.py
+++ b/scanpy/experimental/pp/_highly_variable_genes.py
@@ -230,6 +230,7 @@ def _highly_variable_pearson_residuals(
             sums_cells = np.array(X_batch.sum(axis=1)).ravel()
             sum_total = np.sum(sums_genes).squeeze()
             X_batch = X_batch.tocsc()
+            X_batch.eliminate_zeros()
             residual_gene_var = calculate_res_sparse(
                 X_batch.indptr,
                 X_batch.indices,
--- a/scanpy/experimental/pp/_highly_variable_genes.py
+++ b/scanpy/experimental/pp/_highly_variable_genes.py
@@ -1,4 +1,3 @@
-from multiprocessing.sharedctypes import Value
 import warnings
 from typing import Optional, Literal
 
@@ -17,48 +16,18 @@ from scanpy.get import _get_obs_rep
 from scanpy._utils import _doc_params
 from scanpy.preprocessing._utils import _get_mean_var
 from scanpy.preprocessing._distributed import materialize_as_ndarray
-from scanpy.preprocessing._simple import filter_genes
 from scanpy.experimental._docs import (
     doc_adata,
     doc_dist_params,
     doc_genes_batch_chunk,
     doc_check_values,
     doc_layer,
-    doc_copy,
     doc_inplace,
 )
 
 
-@nb.njit
-def clac_clipped_res_sparse(
-    value,
-    sum_total,
-    sums_gene,
-    sums_cell,
-    theta,
-    clip,
-) -> np.float64:
-    mu = sums_gene * sums_cell / sum_total
-    mu_sum = value - mu
-    pre_res = mu_sum / sqrt(mu + mu * mu / theta)
-    res = np.float64(min(max(pre_res, -clip), clip))
-    return res
-
-
-@nb.njit
-def get_value(cell, sparse_idx, index, stop_idx, data) -> np.float64:
-    """
-    This function navigates the sparsity of the CSC (Compressed Sparse Column) matrix,
-    returning the value at the specified cell location if it exists, or zero otherwise.
-    """
-    if sparse_idx < stop_idx and index[sparse_idx] == cell:
-        return data[sparse_idx]
-    else:
-        return np.float64(0.0)
-
-
 @nb.njit(parallel=True)
-def calculate_res_sparse(
+def _calculate_res_sparse(
     indptr: NDArray[np.integer],
     index: NDArray[np.integer],
     data: NDArray[np.float64],
@@ -71,6 +40,23 @@ def calculate_res_sparse(
     n_genes: int,
     n_cells: int,
 ) -> NDArray[np.float64]:
+    def get_value(cell: int, sparse_idx: int, stop_idx: int) -> np.float64:
+        """
+        This function navigates the sparsity of the CSC (Compressed Sparse Column) matrix,
+        returning the value at the specified cell location if it exists, or zero otherwise.
+        """
+        if sparse_idx < stop_idx and index[sparse_idx] == cell:
+            return data[sparse_idx]
+        else:
+            return np.float64(0.0)
+
+    def clac_clipped_res_sparse(gene: int, cell: int, value: np.float64) -> np.float64:
+        mu = sums_genes[gene] * sums_cells[cell] / sum_total
+        mu_sum = value - mu
+        pre_res = mu_sum / sqrt(mu + mu * mu / theta)
+        res = np.float64(min(max(pre_res, -clip), clip))
+        return res
+
     residuals = np.zeros(n_genes, dtype=np.float64)
     for gene in nb.prange(n_genes):
         start_idx = indptr[gene]
@@ -80,15 +66,8 @@ def calculate_res_sparse(
         var_sum = np.float64(0.0)
         sum_clipped_res = np.float64(0.0)
         for cell in range(n_cells):
-            value = get_value(cell, sparse_idx, index, stop_idx, data)
-            clipped_res = clac_clipped_res_sparse(
-                value=value,
-                sum_total=sum_total,
-                sums_gene=sums_genes[gene],
-                sums_cell=sums_cells[cell],
-                theta=theta,
-                clip=clip,
-            )
+            value = get_value(cell, sparse_idx, stop_idx)
+            clipped_res = clac_clipped_res_sparse(gene, cell, value)
             if value > 0:
                 sparse_idx += 1
             sum_clipped_res += clipped_res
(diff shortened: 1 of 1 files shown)


scanpy/experimental/pp/_highly_variable_genes.py
```diff
@@ -105,24 +162,41 @@ def _highly_variable_pearson_residuals(
         if clip < 0:
             raise ValueError("Pearson residuals require `clip>=0` or `clip=None`.")
 
+        residual_gene_var = np.zeros((X_batch.shape[1]), dtype=np.float64)
```
flying-sheep: Please avoid mutating parameters. You can allocate this from within the functions, no?

scanpy/experimental/pp/_highly_variable_genes.py
```diff
@@ -27,6 +28,86 @@
...
+    data,
+    sums_genes,
+    sums_cells,
+    residuals,
+    sum_total,
+    clip,
+    theta,
+    n_genes,
+    n_cells,
+):
```
flying-sheep:
```suggestion
def calculate_res_sparse(
    indptr,
    index,
    data,
    *,
    sums_genes,
    sums_cells,
    sum_total,
    clip,
    theta,
    n_genes,
    n_cells,
) -> NDArray[np.float64]:
```

scanpy/experimental/pp/_highly_variable_genes.py
```diff
@@ -27,6 +28,86 @@
...
+            diff = clipped_res - mean_clipped_res
+            var_sum += diff * diff
+
+        residuals[gene] = var_sum / n_cells
+
+
+@nb.njit(parallel=True)
+def calculate_res_dense(
+    matrix, sums_genes, sums_cells, residuals, sum_total, clip, theta, n_genes, n_cells
+):
```
flying-sheep:
```suggestion
def calculate_res_dense(
    matrix, *, sums_genes, sums_cells, sum_total, clip, theta, n_genes, n_cells
) -> NDArray[np.float64]:
```

scanpy/experimental/pp/_highly_variable_genes.py
```diff
@@ -27,6 +28,86 @@
...
+    for gene in nb.prange(n_genes):
+        sum_clipped_res = np.float64(0.0)
+        for cell in range(n_cells):
+            mu = sums_genes[gene] * sums_cells[cell] / sum_total
+            value = matrix[cell, gene]
+
+            mu_sum = value - mu
+            pre_res = mu_sum / sqrt(mu + mu * mu / theta)
+            clipped_res = min(max(pre_res, -clip), clip)
```
flying-sheep: deduplicate with second copy

scanpy/experimental/pp/_highly_variable_genes.py
```diff
@@ -27,6 +28,86 @@
...
+        sum_clipped_res = np.float64(0.0)
+        for cell in range(n_cells):
+            mu = sums_genes[gene] * sums_cells[cell] / sum_total
+            value = np.float64(0.0)
+            if sparse_idx < stop_idx and index[sparse_idx] == cell:
+                value = data[sparse_idx]
+                sparse_idx += 1
+            mu_sum = value - mu
+            pre_res = mu_sum / sqrt(mu + mu * mu / theta)
+            clipped_res = min(max(pre_res, -clip), clip)
```
flying-sheep: deduplicate with second copy.

Also could you please explain the line `if sparse_idx < stop_idx and index[sparse_idx] == cell`? I’m a bit rusty with how sparse matrix layout works, and additionally it seems weird to me that the `mu_sum = … clipped_res = …` block is run even if that block is skipped.

flying-sheep (changes requested):
I’d like to see some deduplication and maintainability changes.

Please never create functions that don’t have a `*` after the 3rd parameter at the latest (unless the function is called `rgba` I guess).

Numba sadly has weird behavior around keyword-only arguments, but even if it ignores them (allowing to specify arguments positionally that shouldn’t), you should still call the function as if everything was working correctly. https://github.com/numba/numba/issues/5655

```py
>>> @numba.njit()
... def test(x, *, y):
...     return x + y 
>>> test(1, 2)
3  # no! bad numba! that should be a TypeError
```

scanpy/experimental/pp/_highly_variable_genes.py
Intron7: I'm ok with the * but. This function updates residuals inplace it doesn't return.

scanpy/experimental/pp/_highly_variable_genes.py
Intron7: We need this twice since we need the variance of the residuse. So the first pass calculates the mean and the 2nd the variance

scanpy/experimental/pp/_highly_variable_genes.py
Intron7: Ok so I don't know if you can do this more elegant. Since genes cell combination also have a residue when the count is 0. We still have to acount for them. I came up with this logic to run over all combination and check if this combination has a count.

scanpy/experimental/pp/_highly_variable_genes.py
Intron7: I think you are right with this. I was still thinking in cuda kernel terms and there you have to create the array outside of the function

scanpy/experimental/pp/_highly_variable_genes.py
flying-sheep: I’m aware, I’m talking about creating a function to abstract this away. Note the exact lines that I commented on: They are identical and should be replaced with a function call.

Don’t copy&paste a multi-line block of identical code

scanpy/experimental/pp/_highly_variable_genes.py
flying-sheep: you still need to update the call site. As said, numba mangles the signature and doesn’t actually enforce keyword-only. But the whole point is to call the functions with most parameters specified by keyword.

flying-sheep (changes requested):
see unresolved conversations

flying-sheep:
Perf measurements for the use case of running the HVG tests on my machine (not very accurate, and not very reminiscent of how users use it).

Tests get a bit slower, real world gets faster.

- scanpy master:

  ```console
  $ git switch master
  $ perf stat -r 10 -B hatch run +py=3.11 test:run -n0 scanpy/tests/test_highly_variable_genes.py
	 Performance counter stats for 'hatch run +py=3.11 test:run -n0 scanpy/tests/test_highly_variable_genes.py' (10 runs):
	
	         71.915,07 msec task-clock:u                     #   14,035 CPUs utilized               ( +-  9,53% )
	                 0      context-switches:u               #    0,000 /sec
	                 0      cpu-migrations:u                 #    0,000 /sec
	         1.168.035      page-faults:u                    #   29,496 K/sec                       ( +-  9,58% )
	   191.815.791.770      cycles:u                         #    4,844 GHz                         ( +-  9,53% )  (83,37%)
	    10.610.492.234      stalled-cycles-frontend:u        #   10,05% frontend cycles idle        ( +-  9,44% )  (83,34%)
	    59.853.476.395      stalled-cycles-backend:u         #   56,69% backend cycles idle         ( +-  9,56% )  (83,32%)
	   257.750.810.841      instructions:u                   #    2,44  insn per cycle
	                                                         #    0,13  stalled cycles per insn     ( +-  9,57% )  (83,33%)
	    45.773.330.764      branches:u                       #    1,156 G/sec                       ( +-  9,58% )  (83,33%)
	     1.147.567.613      branch-misses:u                  #    4,56% of all branches             ( +-  9,54% )  (83,37%)
	
	            5,1241 +- 0,0242 seconds time elapsed  ( +-  0,47% )
  ```

- this PR:

  ```console
  $ git switch hvg_PR_numba
  $ perf stat -r 10 -B hatch run +py=3.11 test:run -n0 scanpy/tests/test_highly_variable_genes.py
	 Performance counter stats for 'hatch run +py=3.11 test:run -n0 scanpy/tests/test_highly_variable_genes.py' (10 runs):
	
	        113.085,21 msec task-clock:u                     #   15,789 CPUs utilized               ( +-  9,56% )
	                 0      context-switches:u               #    0,000 /sec
	                 0      cpu-migrations:u                 #    0,000 /sec
	         1.636.606      page-faults:u                    #   26,373 K/sec                       ( +-  9,55% )
	   310.410.832.165      cycles:u                         #    5,002 GHz                         ( +-  9,55% )  (83,35%)
	    14.117.222.045      stalled-cycles-frontend:u        #    8,30% frontend cycles idle        ( +-  9,46% )  (83,38%)
	    75.813.970.243      stalled-cycles-backend:u         #   44,56% backend cycles idle         ( +-  9,57% )  (83,35%)
	   373.047.679.552      instructions:u                   #    2,19  insn per cycle
	                                                         #    0,11  stalled cycles per insn     ( +-  9,57% )  (83,34%)
	    67.830.590.839      branches:u                       #    1,093 G/sec                       ( +-  9,58% )  (83,35%)
	     1.702.825.180      branch-misses:u                  #    4,56% of all branches             ( +-  9,56% )  (83,28%)
	
	            7,1623 +- 0,0560 seconds time elapsed  ( +-  0,78% )
  ```

flying-sheep (approved):

PR #2615, opened 2023-08-14
Fix key name in documentation

padix-key:
This PR fixes a typo in the docstring of `scanpy.external.pp.scrublet()`: The `obs` column is actually called `predicted_doublet` instead of `predicted_doublets`.


flying-sheep (approved):
Thank you!

PR #2621, opened 2023-08-17
Handle Dask arrays in some utilities

flying-sheep:
This PR

- makes `_utils.check_nonnegative_integers` work with Dask arrays and with that also `rank_genes_groups`
- Adds type hints that I added while re-familiarizing myself with some APIs
- adds a function pair `lazy_{and,or}` that make it possible to delay checks with dask:

```py
>>> def fail():
...     raise AssertionError()
>>> # rhs can be a function or dask array
>>> lazy_and(False, fail)
False
>>> lazy_and(False, da.array(True).map_blocks(lambda _: fail(), meta=np.bool_(True)))
False
>>> # lhs can be a function or dask array for nested use
>>> # when not nested, a lhs function will be called eagerly like in `a() and b`
>>> lazy_and(False, lazy_and(fail, _))
False
>>> # will not create a recursive dask array
>>> lazy_and(da.array(True), da.array(False)).compute()
False
>>> # will complain on invalid use
>>> lazy_and(True, lambda: da.array(...))
'AssertionError: Use lazy_*(_, da.array(...)) instead of lazy_*(_, lambda: da.array(...)).'
```


ivirshup (commented):
What is the goal of this PR?

To me, I think of dask support as:

* Making a computation function work all the way through with dask arrays, returning a lazy result
* Allow us to scale out a computation, e.g. show that this works on larger data.

I'm wondering if "returns a result for dask arrays" is a level of support we want (for non-plotting functions). WDYT?

flying-sheep:
I don’t think that’s possible without changing what kinds of checks we do. E.g. `check_nonnegative_integers` checks if every single value in the data is a whole number.

But sure, it’s more useful to make sure no `n_obs × ?` array ever gets converted into a in-memory array if dask is used. The question is “make everything work with dask now, make sure everything is fast later” or “make a few functions work and go fast now, handle other functions later”

ivirshup:
> The question is “make everything work with dask now, make sure everything is fast later” or “make a few functions work and go fast now, handle other functions later”

My worry with this is that if we do ever make it so the computation is delayed it's a behavior change for dask arrays. E.g. the result changes from immediate to delayed and now functions may have to call `.compute` to get a result.

flying-sheep:
OK, makes sense. I’ll keep working on this then.

ivirshup:
FWIW, I do think this is one of the more difficult cases to make work well with dask.

I would suspect we'd end up wanting something like: https://flox.readthedocs.io/en/latest/ here.

Then I'd do a groupby on the groups, calculate the sum and squared sum for each, do a another level of aggregation for `"rest"` categories, then derive mean and variance from there.

scanpy/_utils/__init__.py
```diff
@@ -483,24 +486,56 @@ def update_params(
...
+    return np
+
+
+def check_nonnegative_integers(X: _FlatSupportedArray) -> bool:
     """Checks values of X to ensure it is count data"""
     from numbers import Integral
 
-    data = X if isinstance(X, np.ndarray) else X.data
+    ufuncs = get_ufuncs(X)
+    data = _get_values(X)
```
ivirshup: Until we have other cases, can this be a branch? I feel like it's a lot of indirection for this branch.

```suggestion
    # get (possibly nonzero) values as (possibly flat) array
    data = X if not sparse.issparse(x) else X.data
```

scanpy/_utils/__init__.py
```diff
@@ -483,24 +486,56 @@ def update_params(
 # --------------------------------------------------------------------------------
 
 
-def check_nonnegative_integers(X: Union[np.ndarray, sparse.spmatrix]):
+_FlatSupportedArray = Union[np.ndarray, sparse.spmatrix, DaskArray]
```
ivirshup: I don't really understand the name. Why `Flat`?

scanpy/preprocessing/_utils.py
```diff
@@ -1,14 +1,17 @@
...
-    if sparse.issparse(X):
+def _get_mean_var(X: _FlatSupportedArray, *, axis=0):
+    if sparse.issparse(X):  # TODO: sparse-in-dask?
         mean, var = sparse_mean_variance_axis(X, axis=axis)
     else:
-        mean = np.mean(X, axis=axis, dtype=np.float64)
-        mean_sq = np.multiply(X, X).mean(axis=axis, dtype=np.float64)
+        ufuncs = get_ufuncs(X)
+        mean = ufuncs.mean(X, axis=axis, dtype=np.float64)
+        mean_sq = ufuncs.multiply(X, X).mean(axis=axis, dtype=np.float64)
```
ivirshup: Instead of calling this `ufuncs`, can we pretend dask does array-api instead? I think that'll be closer to the truth, and be closer to a common pattern for interoperable packages.

```suggestion
        xp = get_array_module(X)
        mean = xp.mean(X, axis=axis, dtype=np.float64)
        mean_sq = xp.multiply(X, X).mean(axis=axis, dtype=np.float64)
```

Similar can be done above if we switch from `signbit` to `sign` (https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.sign.html)

ivirshup: Alternatively we can avoid all that and do:

```suggestion
        mean = X.mean(axis=axis, dtype=np.float64)
        mean_sq = (X * X).mean(axis=axis, dtype=np.float64)
```

scanpy/preprocessing/_utils.py
```diff
@@ -1,14 +1,17 @@
...
 import numba
+from scipy import sparse
+
+from .._utils import _FlatSupportedArray, get_ufuncs
 
 
-def _get_mean_var(X, *, axis=0):
-    if sparse.issparse(X):
+def _get_mean_var(X: _FlatSupportedArray, *, axis=0):
+    if sparse.issparse(X):  # TODO: sparse-in-dask?
```
ivirshup: I feel like `sparse-in-dask` is such a large TODO that it may not be worth mentioning here specifically.

scanpy/tests/test_utils.py
```diff
@@ -28,20 +28,20 @@ def test_descend_classes_and_funcs():
...
+        pytest.param(lambda: np.random.normal(size=(100, 100)), False, id='normal'),
+    ],
+)
+def test_check_nonnegative_integers(array_type, mk_add, expected):
+    X = array_type(
+        np.random.poisson(size=(100, 100)).astype(np.float64) + (add := mk_add())
+    )
+    assert check_nonnegative_integers(X) is expected
+    if isinstance(add, np.ndarray):
+        assert check_nonnegative_integers(-X) is False
```
ivirshup:
```suggestion
@pytest.mark.parametrize(
    ('array_value', 'expected'),
    [
        pytest.param(np.random.poisson(size=(100, 100)).astype(np.float64), True, id='poisson-float64'),
        pytest.param(np.random.poisson(size=(100, 100)).astype(np.uint32), True, id='poisson-uint32')
        pytest.param(np.random.normal(size=(100, 100)), False, id='normal'),
        pytest.param(np.array([[0, 0, 0,], [0, -1, 0]. [0, 0, 0]]), False, id="middle"),
    ],
)
def test_check_nonnegative_integers(array_type, array_value, expected):
    X = array_type(array_value)

    assert check_nonnegative_integers(X) is expected
```

ivirshup (changes requested):
I've left a bunch of comments, but there is one big question I have.

Right now, `check_nonnegative_integers` called on a dask array, returns a `bool`. Should it return a real value or a task? I would have thought it would wait on a downstream call to `.compute` to actually do the check.

scanpy/_utils/__init__.py
flying-sheep: good point, this is no longer true

scanpy/preprocessing/_utils.py
flying-sheep: sadly not because `spmatrix * spmatrix` is a matrix multiplication

flying-sheep: ah, nice, because it’s in the non-sparse branch

scanpy/_utils/__init__.py
```diff
@@ -497,38 +497,29 @@ def get_ufuncs(data: np.ndarray | DaskArray):
...
+    # get (possibly nonzero) values as (possibly flat) array
+    data: np.ndarray | DaskArray = X.data if sparse.issparse(X) else X
+    # return bool or lazy dask array resolving to one
+    return (
+        # Check no negatives
+        (~ufuncs.signbit(data).any())
+        & (
+            # Check all are integers
+            issubclass(data.dtype.type, Integral)
+            |
```
ivirshup: Won't this execute every check?

```python
In [1]: def foo():
   ...:     print("hi")
   ...:     return True
   ...: 

In [2]: True and foo()
hi
Out[2]: True

In [3]: True & foo()
hi
Out[3]: True
```

I think I'd prefer to keep as is so we exit as fast as possible.

Similarly, I wonder if we should just do a `da.map_blocks(check_nonnegative_integers).all()` to [avoid a large graph](https://docs.dask.org/en/stable/best-practices.html#avoid-very-large-graphs) for the dask case.

ivirshup (dismissed):

scanpy/_utils/__init__.py
flying-sheep: Ah good thing we’re doing this together. You’re right, I forgot that bitwise or doesn’t early exit even when used on bools. So basically we don’t have the support of Python’s syntax anymore and need to do this manually:

```py
lazy_and(
    ~ufuncs.signbit(data).any(),
    lazy_or(
        issubclass(data.dtype.type, Integral),  # eagerly executed
        ~((data % 1) != 0).any(),
    ),
)
```

and then implement the combinators for dask and bools.

flying-sheep: OK, I did it! And it’s not even super ugly to write.

flying-sheep:
OK, so the graph in `test_check_nonnegative_integers`, generated via

```py
import dask
dask.visualize(rv, engine="cytoscape", filename=request.node.name)
```


```mermaid
flowchart LR

4342680003182880388["(0, 0)"]
4363265869429493506((signbit))
4342680003182880388 --> 4363265869429493506

550394025789815978["(0, 1)"]
5167392774578066548((signbit))
550394025789815978 --> 5167392774578066548

3764517430824237394["(1, 0)"]
5336730508589856979((signbit))
3764517430824237394 --> 5336730508589856979

2743123325277761031["(1, 1)"]
2513425685193572888((signbit))
2743123325277761031 --> 2513425685193572888

284808496767994156["(0, 0)"]
4363265869429493506 --> 284808496767994156

6263727941369393084((any))
284808496767994156 --> 6263727941369393084

6222832259334004269["(0, 1)"]
5167392774578066548 --> 6222832259334004269

7256567839680908872((any))
6222832259334004269 --> 7256567839680908872

8881403918513157720["(1, 0)"]
5336730508589856979 --> 8881403918513157720

5898621639535744825((any))
8881403918513157720 --> 5898621639535744825

2373763162411159295["(1, 1)"]
2513425685193572888 --> 2373763162411159295

1659302467096852217((any))
2373763162411159295 --> 1659302467096852217

7195453449900658805["(0, 0)"]
6263727941369393084 --> 7195453449900658805

7976077601232067203((any-\naggregate))
7195453449900658805 --> 7976077601232067203

687812693798660380["(0, 1)"]
7256567839680908872 --> 687812693798660380
687812693798660380 --> 7976077601232067203

3901936098833081796["(1, 0)"]
5898621639535744825 --> 3901936098833081796
3901936098833081796 --> 7976077601232067203

8795010127805778162["(1, 1)"]
1659302467096852217 --> 8795010127805778162
8795010127805778162 --> 7976077601232067203

1203378416021505679["()"]
7976077601232067203 --> 1203378416021505679
9179805111332178500((invert))

1203378416021505679 --> 9179805111332178500
5169565091578776769["()"]
9179805111332178500 --> 5169565091578776769
814146044537405006((and))
5169565091578776769 --> 814146044537405006

1050532709569538834["()"]
814146044537405006 --> 1050532709569538834
```

I *am* of course using `map_blocks`. If we really wanted, I assume we could still replace sequences of two operations like

```mermaid
flowchart LR

step0["(0, 0)"] --> op0((signbit)) --> step1["(0, 0)"] --> op1((any)) --> step2["(0, 0)"]
```

with individual operations, but I’m not sure if that’s worth the code readability problems.

Smells of premature optimization.

<details>
<summary>mean_var graph</summary>


```mermaid
flowchart LR

step000["(0, 0)"] --> op000((mean_\nchunk)) --> step001["(0, 0)"] --> op00((mean_agg-\naggregate)) --> step00["0"]
step100["(1, 0)"] --> op100((mean_\nchunk)) --> step101["(1, 0)"] --> op00

step010["(0, 1)"] --> op010((mean_\nchunk)) --> step011["(0, 1)"] --> op10((mean_agg-\naggregate)) --> step10["1"]
step110["(1, 1)"] --> op110((mean_\nchunk)) --> step111["(1, 1)"] --> op10
```

</details>

Issue #2495, opened 2023-05-28
Request: Ability to subsample `AnnData` in backed mode


nleroy917:
## Feature type

- [x] Additional function parameters / changed functionality / changed defaults?
- [ ] New analysis tool: A simple analysis tool you have been using and are missing in `sc.tools`?
- [ ] New plotting function: A kind of plot you would like to seein `sc.pl`?
- [ ] External tools: Do you know an existing package that should go into `sc.external.*`?
- [ ] Other?

## Request

The `scanpy.pp.subsample` function has been really useful for iteration, experimentation, and tutorials. However, I've found that sometimes I don't want to (or simply can't) read the entire `AnnData` object into memory before subsampling. As such, it would be nice to instantiate an `AnnData` object in backed mode, then subsample that directly.

### Current behavior:

```python
import scanpy as sc

adata = sc.read_h5ad("matrix/scatlas.h5ad", backed="r")
adata_sample = sc.pp.subsample(adata, fraction=0.01, copy=True)
```
yields:
```
ValueError: To copy an AnnData object in backed mode, pass a filename: `.copy(filename='myfilename.h5ad')`. To load the object into memory, use `.to_memory()`.
```
### Workaround:
A workaround I've found is by creating a boolean array and using boolean indexing to grab a random subset of rows from the `AnnData` object:

```python
import numpy as np

FRACTION = 0.01

np.random.seed(0)
random_bool = np.random.choice(
    [True, False], size=adata.shape[0], p=[FRACTION, 1 - FRACTION]
)

adata_sample = adata[random_bool, :]
adata_sample_mem = adata_sample.to_memory()
```

It's easy enough but would be nice to have within `scanpy` itself!

eroell:
Hi,

thanks for the input and the nice description.

Agree that this would be nice to have within scanpy itself.

And agree, as you’ve shown in your example, that this should be the case when setting `copy=True`.

I have set up a draft PR for the moment, with a suggestion where the error message especially persists when `copy=False`; I’d suggest to keep it this way:

- Overwriting the backed file seems not to be the expected behaviour to me
- Writing a new file for backing would occur in a rather hidden manner, confusing the user or even unexpectedly further fill the disk at worst over time

eroell:
With the merged PR which addresses the issue raised, this probably could be closed?
@ivirshup

eroell:
We will close the issue for now, as based on the provided information and the discussion so far, it seems that the question has been addressed and hopefully resolved :)

[...]

PR #2624, opened 2023-08-18
allow sc.pp.subsample in backed mode when returning a copy

eroell:
Adresses issue #2495.

Suggested change to allow `sc.pp.subsample` in backed mode of AnnData, when `copy=True` in `sc.pp.subsample`

```
import scanpy as sc

adata = sc.read_h5ad("matrix/scatlas.h5ad", backed="r")
adata_sample = sc.pp.subsample(adata, fraction=0.01, copy=True)
```

Tagging @ivirshup.

### DIFF
--- a/scanpy/tests/test_preprocessing.py
+++ b/scanpy/tests/test_preprocessing.py
@@ -123,7 +123,7 @@ def test_subsample_copy_backed(tmp_path):
     adata_d.filename = filename
     # This should not throw an error
     assert sc.pp.subsample(adata_d, n_obs=40, copy=True).shape == (40, 10)
-    assert np.allclose(
+    np.testing.assert_array_equal(
         sc.pp.subsample(adata_m, n_obs=40, copy=True).X,
         sc.pp.subsample(adata_d, n_obs=40, copy=True).X,
     )
--- a/scanpy/tests/test_preprocessing.py
+++ b/scanpy/tests/test_preprocessing.py
@@ -127,6 +127,8 @@ def test_subsample_copy_backed(tmp_path):
         sc.pp.subsample(adata_m, n_obs=40, copy=True).X,
         sc.pp.subsample(adata_d, n_obs=40, copy=True).X,
     )
+    with pytest.raises(NotImplementedError):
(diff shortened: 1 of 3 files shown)


nleroy917:
Appreciate the PR - let me know if I can be of any assistance.

ivirshup (commented):
Thanks for opening this!

What happens if the object is backed, but `copy=False`?

eroell:
When the object is backed, but `copy=False`, the ValueError, which before occured for both `copy=False` and `copy=True`, is shown:
`ValueError: To copy an AnnData object in backed mode, pass a filename: '.copy(filename='myfilename.h5ad')'. To load the object into memory, use '.to_memory()'.`

ivirshup:
Could you throw a more informative error message for `copy=False`? Maybe:

`NotImplementedError("Inplace subsampling is not implemented for backed objects")`

?

scanpy/tests/test_preprocessing.py
```diff
@@ -115,6 +115,22 @@ def test_subsample_copy():
...
 
+def test_subsample_copy_backed(tmp_path):
+    A = np.random.rand(200, 10).astype(np.float32)
+    adata_m = AnnData(A.copy())
+    adata_d = AnnData(A.copy())
+    filename = tmp_path / 'test.h5ad'
+    adata_d.filename = filename
+    # This should not throw an error
+    assert sc.pp.subsample(adata_d, n_obs=40, copy=True).shape == (40, 10)
+    assert np.allclose(
```
ivirshup:
```suggestion
    np.testing.assert_array_equal(
```

* The np.testing functions can give better error messages
* I think these should be equal, not just close

ivirshup (changes requested):
One minor change, I would invert the loops. I think it makes more sense to consider the `isbacked` cases together. So:

```python
if data.isbacked:
    if copy:
        return data[obs_indices].to_memory()
    else:
        raise NotImplementedError(...)
else:
    if copy:
        ...
```

Otherwise looking good! One minor change in the tests.

ivirshup (approved):
Looks great, thanks!

Issue #2587, opened 2023-08-02
Unexpected behavior of `sc.pp.neighbors`


almaan:
### Please make sure these conditions are met

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the latest version of scanpy.
- [ ] (optional) I have confirmed this bug exists on the master branch of scanpy.

### What happened?

Hej!

Thanks for maintaining such a great package! This issue relates another [issue](https://github.com/scverse/squidpy/issues/735) posted (by me) in the `squidpy` repo, but I think it might be worth bringing up here as well. 

The issue in question is how the  `sc.pp.neighbors` function returns an inconsistent number of neighbors even when `knn=True`. In the [documentation](https://scanpy.readthedocs.io/en/stable/generated/scanpy.pp.neighbors.html) of `sc.pp.neighbors` it's stated that :

>n_neighbors: The size of local neighborhood (in terms of number of neighboring data points) used for manifold approximation. Larger values result in more global views of the manifold, while smaller values result in more local data being preserved. In general values should be in the range 2 to 100. If knn is True, number of nearest neighbors to be searched. If knn is False, a Gaussian kernel width is set to the distance of the n_neighbors neighbor.

and
> knn: If True, use a hard threshold to restrict the number of neighbors to n_neighbors, that is, consider a knn graph. Otherwise, use a Gaussian Kernel to assign low weights to neighbors more distant than the n_neighbors nearest neighbor.

as well as 
> Connectivities: Weighted adjacency matrix of the neighborhood graph of data points. Weights should be interpreted as connectivities.

Hence I would expect that the number of non-zero elements in `adata.obsp['connectivities']` in an object to which `sc.pp.neighbors(adata, n_neighbors = k, knn = True)` have been applied, would sum to `k` for each row. However, when inspecting these results, it is not true. The number of non-zero elements in a row varies between both higher as well as lower values than the specified `n_neighbors` (obviously, sometimes it's also the expected `n_neighbors` value).

Perhaps I'm misunderstanding something, but this behavior is somewhat counterintuitive to me and not what I expect; happy to be corrected though!

/Alma

### Minimal code sample

```python
# Import packages

import scanpy as sc
import anndata as ad
import numpy as np

# set random seed
np.random.seed(42)

# create dummy data

adata = ad.AnnData(shape=(1000,1))
adata.obsm['rep'] = np.random.random(size = (1000,2))

# get spatial connectivities
k = 10
[...]

PR #2640, opened 2023-08-24
Highlight distance matrix in docs of sc.pp.neighbors

eroell:
A suggestion to highlight that the nearest neighbors distance matrix is one of two outputs of this method. Addressing #2587.


flying-sheep:
Thanks!

flying-sheep (approved):

PR #2646, opened 2023-08-31
Improve code samples in `neighbors` docs

yfarjoun:
added some quotes to make the code copy-able in the docs


flying-sheep:
thanks! let’s see if re-running the flaky tests makes CI pass

PR #2647, opened 2023-09-01
correct scanorama_integrate docs

fbnrst:
I just found a small mistake in the documentation of `scanorama_integrate`:
**kwargs are passed to assemble, not integrate


flying-sheep:
thanks! the tests here are a bit flaky it seems, and have nothing to do with your change

Zethson (approved):

PR #2648, opened 2023-09-03
Fix and deprecate RAPIDS implementations in favor of rapids_singlecell

Intron7:
Hotfix for rapids nn:

* fixes bug with `random_state` for `RapidsKNNTransformer.__init__`
* now return distance matrix with `self.nn.kneighbors_graph(X_contiguous, mode="distance")`


flying-sheep:
thanks! seems like I need to fix something about that failing pbmc3k notebook first though.

flying-sheep (approved):

PR #2657, opened 2023-09-08
Add note to the groups arg in rank_genes_groups()

lazappi:
It was unclear if the `groups` argument in `rank_genes_groups()` is equivalent to subsetting when `reference='rest'`(i.e. only the selected groups are tested against each other) or if only those groups are tested but all groups are still used as the reference.

This PR adds a note explaining the current behaviour (option 2).


lazappi:
Any hints on the CI? It just says `run exit 1`.

Zethson:
@flying-sheep Yup, I think we need to at least print a better failure message than just returning -1 for the milestone check. It's not a good experience for people that are not familiar with this.

@lazappi thank you!

PR #2684, opened 2023-10-14
DensMAP support

AlePur:
I added an optional boolean parameter to scanpy.tl.umap for DensMAP. Since umap 0.5, ``simplicial_set_embedding`` takes ``densmap`` and ``densmap_kwds`` parameters, and tools/_umap is using this function, so DensMAP support is easily implemented.

### DIFF
--- a/scanpy/tools/_umap.py
+++ b/scanpy/tools/_umap.py
@@ -102,7 +102,13 @@ def umap(
     copy
         Return a copy instead of writing to adata.
     method
-        Use the original 'umap' implementation, or 'rapids' (experimental, GPU only)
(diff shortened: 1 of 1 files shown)


flying-sheep:
I see, [densmap](https://umap-learn.readthedocs.io/en/latest/densmap_demo.html)

Hmm, I think that `method='densmap'` and `method_kwds={...}` would be a better API for us (which would then be translated into `densmap=True, densmap_kwds=method_kwds`)

This also needs tests and a release note.

Also we probably should just remove the umap 0.4 compatibility code, what do you think @ivirshup?

PR #2695, opened 2023-10-17
Switch to hatch and Python 3.9 minimum

flying-sheep:
Combines the non-linter changes from https://github.com/scverse/anndata/pull/1001 and https://github.com/scverse/anndata/pull/1125

The linter changes like https://github.com/scverse/anndata/pull/1125/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711 can be done later too


Zethson (approved):
LGTM although I got triggered with the replacement of a relative file link to `LICENSE` and a hard coded license :) But I don't expect this to (ever) change.

flying-sheep:
> replacement of a relative file link to LICENSE and a hard coded license

Ah, but that’s not what that is. It’s a switch to [PEP 639](https://peps.python.org/pep-0639/), which says:

> If the license-files key is not present and not explicitly marked as dynamic, tools MUST assume a default value of the following:
> 
> ```toml
> license-files.globs = ["LICEN[CS]E*", "COPYING*", "NOTICE*", "AUTHORS*"]
> ```

So nothing is lost, the file still gets included.

PR #2696, opened 2023-10-17
Handle Dask arrays in some utilities

flying-sheep:
See #2621


ivirshup (changes requested):
I find that an implementation (using `master` branch) that looks like:

```python
X_dask_f32.map_blocks(
    check_nonnegative_integers,
    dtype=bool,
    drop_axis=(0, 1)
).compute()
```

Has a few strong advantages to this approach:

* It's about 4x faster for a the case `X = da.array(rng.poisson(size=(20_000, 10_000)), chunks=(1000, 10_000))` (similar speed to in-memory)
* It works when each chunk of the dask array is a sparse matrix

As mentioned in the last PR, this is also the approach that the docs for dask.array seems to suggest. I think we should go with this approach here.

<details>
<summary> Code for benchmarking </summary>

Setup

```python
import numpy as np, anndata as ad, h5py
from scipy import sparse

rng = np.random.default_rng()

X = rng.poisson(size=(20_000, 10_000))
X_dense_f32 = X.astype(np.float32)
X_sparse_f32 = sparse.csr_matrix(X_dense_f32)


with h5py.File("arrays.h5", "w") as f:
    g = f["/"]
    ad.experimental.write_elem(g, "X_dense_f32", X_dense_f32)
    ad.experimental.write_elem(g, "X_sparse_f32", X_sparse_f32)
```

Benchmarking

```python
import scanpy as sc
import anndata as ad
import h5py
from scanpy._utils import check_nonnegative_integers
from scipy import sparse
import dask.array as da

with h5py.File("arrays.h5") as f:
    X_dense_f32 = ad.experimental.read_elem(f["X_dense_f32"])
    X_sparse_f32 = ad.experimental.read_elem(f["X_sparse_f32"])


X_dask_f32 = da.from_array(X_dense_f32, chunks=(1000, 10_000))
```

```python
%timeit X_dask_f32.map_blocks(check_nonnegative_integers, dtype=bool, drop_axis=(0, 1)).compute()
```

Testing that it works for sparse arrays:

```python
(
    X_dask_f32
    .map_blocks(sparse.csr_matrix)
    .map_blocks(check_nonnegative_integers, dtype=bool, drop_axis=(0, 1))
    .compute()
)
```

</details>

I would note that neither case seems to spend that much time doing computation in parallel, which is a little curious.

The docs for the map_blocks function also recommends using `da.reduction` here, though I believe that would take more rewriting and haven't checked it yet.

flying-sheep:
Makes sense. I think we need a good sparse story before we think about supporting sparse-in-dask.

flying-sheep:
I went over all the places where we use the `array_type` fixture and thought about your idea to use `@pytest.mark.parametrize` and I came around to it for this case:

For **unfinished** features, it’s great. Everwhere we can’t say “we fully support this” and gradually build in support, we should use it.

It has its disadvantages:

- `@pytest.mark.parametrize("array_type", ARRAY_TYPES)` is so long that in practice, it’s hard to see the difference to something like this: `@pytest.mark.parametrize("array_type", ARRAY_TYPES_XYZ)`


	E.g. I don’t like seeing
	
	```py
	@pytest.mark.parametrize("array_type", ARRAY_TYPES)
	@pytest.mark.parametrize("dtype", ["float32", "int64"])
	```

	4 times in `test_normalize_total`. If the 3rd test had a different list of values in one of the params, it would be near impossible to see.

- Fixtures can depend on other fixtures, but can’t easily have a parameter matrix without that. (`pytest.fixture(params=...)` only accepts a single list of parameters, we’d have to manually use `product` in there for a matrix)

  That’s why I didn’t go away from a fixture in `test_pca.py`

I therefore propose that we use `@pytest.mark.parametrize` for

- things that aren’t heavily reused
- things we don’t fully support

and fixtures for everything where there’s ~3 or more test functions using the same list of parameter values.

ivirshup:
> @pytest.mark.parametrize("array_type", ARRAY_TYPES) is so long that in practice, it’s hard to see the difference to something like this: @pytest.mark.parametrize("array_type", ARRAY_TYPES_XYZ)

Does doing something like: `ARRAY_TYPES + [DaskArray]` help here? I think this is nice and explicit, plus super obvious how to modify.

scanpy/_utils/__init__.py
```diff
@@ -483,8 +486,41 @@ def update_params(
 # --------------------------------------------------------------------------------
 
 
-def check_nonnegative_integers(X: Union[np.ndarray, sparse.spmatrix]):
+_DataMatrix = Union[sparse.csr_matrix, sparse.csc_matrix]
```
ivirshup: Maybe this variable could have `Sparse` in the name?

scanpy/tests/test_pca.py
```diff
@@ -72,27 +77,30 @@ def pca_params(array_type, svd_solver_type, zero_center):
     expected_warning = None
     svd_solver = None
     if svd_solver_type is not None:
-        if array_type is as_dense_dask_array:
+        # TODO: are these right for sparse?
+        if array_type in {as_dense_dask_array, as_sparse_dask_array}:
```
ivirshup, line 81: Why is this changing?

scanpy/tests/test_preprocessing.py
```diff
@@ -46,6 +47,22 @@ def test_log1p_rep(count_matrix_format, base, dtype):
     check_rep_results(sc.pp.log1p, X, base=base)
 
 
+# TODO: Add support for sparse-in-dask
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_SUPPORTED)
+def test_mean_var(array_type):
+    pbmc = pbmc3k()
+    pbmc.var_names_make_unique()
```
ivirshup:
```suggestion
```

Not needed, no?

ivirshup (commented):
I think the code changes look pretty good (a few comments).

The `MAP_ARRAY_TYPES` may be overkill, but fine with it for now.

scanpy/tests/test_pca.py
flying-sheep: some parts of scanpy’s tests were using our own coercion, other parts anndata’s. We should be consistent unless we have a reason not to be.

ivirshup: Oh, I actually meant "why change from list to set" but I did totally miss that more meaningful change.

I'm a little surprised that these things don't seem to be throwing errors for sparse dask arrays. I will take a look locally.

flying-sheep: OK, so I’ll check if this just coerces to numpy array

ivirshup: Dense dask arrays do appear to work. It looks like we will use `dask-ml` here.

ivirshup: Sparse dask arrays don't seem to work. Shouldn't we be seeing an error from that?

flying-sheep: Perfect! This file uses `ARRAY_TYPES_SUPPORTED`, which marks sparse dask arrays as strict `xfail`. D.h. we’re testing that in-memory arrays and dense dask arrays work, and tests that sparse dask arrays throw an error.

Maybe this array should have a better name. Any suggestions?

ivirshup: `as_sparse_dask_array`? I think it's fine. Or which array?

How do we start allowing sparse dask arrays if the parameter in `ARRAY_TYPES_SUPPORTED` marks them as xfail? Should we add the xfail here?

flying-sheep: >Or which array?

`ARRAY_TYPES_SUPPORTED`

>How do we start allowing sparse dask arrays

For one test:
```diff
-ARRAY_TYPES_SUPPORTED
+ARRAY_TYPES
```

Once everything supports sparse dask arrays, we remove `ARRAY_TYPES_SUPPORTED` including its xfail marks

ivirshup: For both things, fine for now, but okay to change later.

ivirshup (approved):

Issue #2331, opened 2022-09-16
deprecate/remove read spatial and plotting spatial from scanpy.


giovp:
I think it makes sense to deprecate and then remove `sc.read_visium`. and `sc.pl.spatial` since I don't plan to maintain it here anymore and updated versions are present in Squidpy. What do you think @ivirshup ?

Zethson:
I would like to second this. I think that it makes sense that we don't duplicate functionality and responsibility of our core tools. People shouldn't be looking for spatial stuff in scanpy nor be confused if they suddenly see spatial tooling in the scanpy docs. We should always point them to the awesome squidpy.

linhuawang:
I just had issues using `sc.pl.spatial` after reading spaceranger2.0.0 outputs using `sc.read_visium`. Although the issue is resolved. I am wondering if these functions are not maintained anymore. I am not familiar with Squidpy. But it seems to use these functions in its package. I would very much appreciate any clarification here.

giovp:
hi @linhuawang ,

yes I'd suggest you to use squidpy for plotting, unfortunately the spaceranger 2.0 reading is not in main yet (will be by end of the week).

PR #2702, opened 2023-10-20
Wrap legacy APIs

### LINKED ISSUE #464: Transition away from positional APIs
We never wanted APIs that can be used with more that ~2 positional parameters. We should go to keyword-only-parameters.

This can be done via [legacy-api-wrap](https://github.com/flying-sheep/legacy-api-wrap)

flying-sheep:
Part of [scanpy 2.0](https://github.com/orgs/scverse/projects/18/views/1)

- [ ] Closes #

I implemented a Ruff check (PLR0917) for this, but setting `max-positional-args = 3` would make this massive PR even larger, so I opted to do it in a separate one.

## Reviewers

Your main job is to check if the position of the `*` makes sense for each exported function (i.e. the ones with the `@legacy_api` decorator). I tried my best to base it on internal usage of each API, but one placement or the other might be to early.

The only real logic changes are in `scanpy/tests/test_package_structure.py`. This PR:

- makes public APIs with more than 5 parameters keyword-only without breaking backwards compatibility (for now)
- makes private APIs with more than 5 parameters keyword-only
- checks that we don’t internally use the old positional APIs using a warning filter with `action='error'`
- checks that APIs use our convention for the `copy` parameter (`adata` as first param, returns `adata` type or None`)
- manually checked that APIs use our convention `filename: Path | str`/`path: Path | str`

## Follow-up changes

- [ ] Set `max-positional-args = 3`
- [ ] make sure that no new public API gets introduced without being included in the `api_module_names` list
- [ ] enforce convention `show`, `return_fig`, `ax`
- [ ] enforce convention for `random_seed: AnyRandom`
- [x] https://github.com/scverse/scanpy/issues/2331


scanpy/external/pp/_bbknn.py
```diff
@@ -70,18 +73,17 @@ def bbknn(
...
-        'rogerstanimoto', 'russellrao', 'sokalsneath', 'sokalmichener', 'yule'])
+        'linfty', 'linf', 'minkowski', 'seuclidean', 'standardised_euclidean', 'wminkowski', ...])
 
         KDTree supports members of :class:`sklearn.neighbors.KDTree`’s ``valid_metrics`` list, or parameterised
         :class:`~sklearn.metrics.DistanceMetric` objects:
 
+        >>> import sklearn.neighbors
         >>> sklearn.neighbors.KDTree.valid_metrics
-        ['p', 'chebyshev', 'cityblock', 'minkowski', 'infinity', 'l2', 'euclidean', 'manhattan', 'l1']
+        ['euclidean', 'l2', 'minkowski', 'p', 'manhattan', 'cityblock', 'l1', 'chebyshev', 'infinity']
```
ilan-gold: Why did this change?

scanpy/external/pp/_hashsolo.py
```diff
@@ -261,15 +264,17 @@ def _calculate_bayes_rule(data, priors, number_of_noise_barcodes):
...
 def hashsolo(
-    adata: anndata.AnnData,
-    cell_hashing_columns: list,
-    priors: list = [0.01, 0.8, 0.19],
-    pre_existing_clusters: str = None,
-    number_of_noise_barcodes: int = None,
+    adata: AnnData,
+    cell_hashing_columns: Sequence[str],
+    *,
+    priors: tuple[float, float, float] = (0.01, 0.8, 0.19),
```
ilan-gold, line 275: Was the type annotation here wrong?

ilan-gold: Or is this just because lists and tuples are somewhat polymorphic?

scanpy/external/tl/_phenograph.py
```diff
@@ -18,10 +19,29 @@
...
+    "resolution_parameter",
+    "n_iterations",
+    "use_weights",
+    "seed",
+    "copy",
+)
 @doctest_needs("phenograph")
 def phenograph(
-    adata: AnnData | np.ndarray | spmatrix,
+    data: AnnData | np.ndarray | spmatrix,
```
ilan-gold, line 44: Isn't this a breaking a change?

scanpy/external/tl/_phenograph.py
```diff
@@ -246,9 +265,10 @@ def phenograph(
...
     if copy:
         return communities, graph, Q
-    else:
-        adata.obsp[ig_key] = graph.tocsr()
-        if comm_key:
-            adata.obs[comm_key] = pd.Categorical(communities)
-        if Q:
-            adata.uns[q_key] = Q
+
+    assert isinstance(data, AnnData)
```
ilan-gold: I get the impulse to test this here, but seeing this in this PR makes me think something might be broken that is not stated here and is somehow being fixed by this.

scanpy/neighbors/__init__.py
```diff
@@ -345,6 +346,7 @@ class Neighbors:
     def __init__(
         self,
         adata: AnnData,
+        *,
         n_dcs: int | None = None,
         neighbors_key: str | None = None,
     ):
```
ilan-gold, line 355: This is a public API, no? https://scanpy.readthedocs.io/en/stable/generated/scanpy.Neighbors.html Can we not wrap this?

scanpy/plotting/_anndata.py
```diff
@@ -881,24 +933,24 @@ def violin(
...
-            return g
-        elif len(axs) == 1:
-            return axs[0]
-        else:
-            return axs
+    if show:
+        return None
+    if multi_panel and groupby is None and len(ys) == 1:
+        return g
+    return axs[0] if len(axs) == 1 else axs
```
ilan-gold: nit: I would prefer to just see another if block.  don't even need else because it's a guard condition

scanpy/preprocessing/_simple.py
```diff
@@ -289,15 +300,15 @@ def filter_genes(
 
 @singledispatch
 def log1p(
-    X: AnnData | np.ndarray | spmatrix,
+    data: AnnData | np.ndarray | spmatrix,
```
ilan-gold, line 311: Again, seems like a breaking change.

scanpy/tools/_utils_clustering.py
```diff
@@ -1,18 +1,41 @@
...
+
 
 def rename_groups(
-    adata, key_added, restrict_key, restrict_categories, restrict_indices, groups
-):
-    key_added = restrict_key + "_R" if key_added is None else key_added
+    adata: AnnData,
+    restrict_key: str,
+    *,
+    key_added: str | None,
```
ilan-gold, line 19: Did you mean to change the order here?  `key_added` is now after `restrict_key`

scanpy/tests/test_package_structure.py
```diff
@@ -1,23 +1,54 @@
...
+    "sc.tl",
+    "sc.pl",
+    "sc.experimental.pp",
+    "sc.external.pp",
+    "sc.external.tl",
+    "sc.external.pl",
+    "sc.external.exporting",
+    "sc.get",
+    "sc.logging",
+    # "sc.neighbors",  # Not documented
```
ilan-gold, line 35: Why is this here?

scanpy/preprocessing/_simple.py
ilan-gold: I think you need to do this because this is now a part of the testing suite, but I am not sure this is such a good idea with breaking changes, unless I am missing something.  I have definitely used `X=` syntax before for no other reason than it exists

scanpy/tests/test_package_structure.py
```diff
@@ -60,3 +96,99 @@ def test_function_headers(f):
...
+
+
+copy_sigs: defaultdict[str, ExpectedSig | None] = defaultdict(
+    lambda: ExpectedSig(first_name="adata", copy_default=False, return_ann=None)
+)
+# full exceptions
+copy_sigs["sc.external.tl.phenograph"] = None  # external
+copy_sigs["sc.pp.filter_genes_dispersion"] = None  # deprecated
+copy_sigs["sc.pp.filter_cells"] = None  # unclear `inplace` situation
+copy_sigs["sc.pp.filter_genes"] = None  # unclear `inplace` situation
```
ilan-gold, line 130: What do you mean by "unclear `inplace` situation?"

ilan-gold (commented):
Nothing blocking, but I am pretty worried about the breaking changes.  This isn't really my call though.

scanpy/external/pp/_bbknn.py
flying-sheep: It changed before this PR. `bbknn` isn’t actually tested, I just saw that things changed when running tests locally.

scanpy/external/pp/_hashsolo.py
flying-sheep: `list`, `set`, and `dict` are almost never correct in a *parameter* type annotation (use them in return types). One should use `collections.abc.*` instead to express the interface the function actually supports here.

In this particular case, a correct type annotation is impossible: “`collections.abc.Sequence[float]` of length 3” can’t be expressed, so options capturing part of the truth are `collections.abc.Sequence[float]` and `tuple[float, float, float]`.

It’s common practice to pass short fixed-length sequences as tuples, so using the type I did encourages that.

scanpy/external/tl/_phenograph.py
flying-sheep: Technically yes. Practically, people rather use too many positional parameters than too many positional ones.

Maybe adding a `/` after `data` to enforce positional-only parameter specification would be nice in some cases?

scanpy/external/tl/_phenograph.py
flying-sheep: I use asserts like this just for type assertions. But good point, I broke the logic here, and since we don’t test external APIs, this would break.

scanpy/tools/_utils_clustering.py
flying-sheep: yes, I did, because it’s derived. I updated all call sites.

scanpy/tests/test_package_structure.py
flying-sheep: The `filter_cells` and `filter_genes` functions for some reason say that `copy` is deprecated despite us deciding that `inplace` should be deprecated, so that’s confusing.

scanpy/tests/test_package_structure.py
flying-sheep: This module is not part of the public API despite having a name not starting with `_`, so I mention that here in case people go over the public submodules and wonder.

scanpy/external/tl/_phenograph.py
flying-sheep: see other one.

scanpy/preprocessing/_simple.py
flying-sheep: Hmm, I guess we can add a helper for those cases. (add the old name as kw-only arg and throw deprecation warning if specified that way)

scanpy/tests/test_package_structure.py
ilan-gold: Got it.

scanpy/external/pp/_hashsolo.py
ilan-gold: Great, wanted to understand best practices yeah

scanpy/preprocessing/_simple.py
ilan-gold: Sounds good!

PR #2703, opened 2023-10-23
Move scrublet out of `external`

### LINKED ISSUE #173: Doublet filtering function
Hi,

I tried the [`DoubletDetection`](https://github.com/JonathanShor/DoubletDetection) Python library on my data and got some interesting result. As it can be run directly on a numpy array of count matrix (`adata.X`), I thought it would be an interesting feature for `scanpy`. 

```
clf = doubletdetection.BoostClassifier() 
doublets = clf.fit(adata.X).predict()
adata.obs['doublet'] = pd.Categorical(doublets.astype(bool))
```

flying-sheep:
- [x] Closes #173

Docs:
- https://icb-scanpy--2703.com.readthedocs.build/en/2703/release-notes/index.html#version-1-10
- https://icb-scanpy--2703.com.readthedocs.build/en/2703/api/generated/scanpy.pp.scrublet.html
- https://icb-scanpy--2703.com.readthedocs.build/en/2703/api/generated/scanpy.pp.scrublet_simulate_doublets.html
- https://icb-scanpy--2703.com.readthedocs.build/en/2703/api/generated/scanpy.pl.scrublet_score_distribution.html

### How to review this PR

I made tests quantitative before this PR, so note that the only change that modified tests is 42143d88a0d499130fac8e5ca60eef0c19163734.

In that PR, I make it so there are no longer any duplicate simulated doublets being created. This is necessary to be able to support any neighborhood detection algorithm. I also feel like it makes more sense. This is the only algorithmic change to upstream. Please use your own judgement to check if this makes sense to you.

### TODO:

 - [x] remove unused utils (plotting, preprocessing)
 - [ ] figure out what remaining utils to replace with ours
   - [x] PCA/SVD: https://github.com/scverse/scanpy/blob/bf5f1f9343f5729df6f90f7c68363682022e0480/scanpy/preprocessing/_scrublet/__init__.py#L415-L417
   - [x] mean_center, normalize_variance, zscore: small enough to be left alone I think
   - [ ] get_knn_graph: no need to have multiple implementations here, but our current implementation automatically calculates connectivities, which this doesn’t need https://github.com/scverse/scanpy/pull/2723
 - [ ] refactor so the class API matches the way we use it 
 - [x] switch to dataclass for centralized attr defs
 - [x] use non-deprecated random state style
 - [x] use our logging instead of print statements


flying-sheep:
TODO: add quantitative tests

flying-sheep:
some preliminary stats about the three versions: https://gist.github.com/flying-sheep/ecd20674bdcd1839d0bb7c16f67794cf#file-scrublet-bench-ipynb

scanpy/preprocessing/_scrublet/pipeline.py
```diff
@@ -0,0 +1,79 @@
...
+if TYPE_CHECKING:
+    from ..._utils import AnyRandom
+    from .core import Scrublet
+
+
+def mean_center(self: Scrublet) -> None:
+    gene_means = self._counts_obs_norm.mean(0)
+    self._counts_obs_norm = sparse.csc_matrix(self._counts_obs_norm - gene_means)
+    if self._counts_sim_norm is not None:
+        self._counts_sim_norm = sparse.csc_matrix(self._counts_sim_norm - gene_means)
```
ilan-gold, line 19:
```suggestion
def mean_center(scrublet: Scrublet) -> None:
    gene_means = scrublet._counts_obs_norm.mean(0)
    scrublet._counts_obs_norm = sparse.csc_matrix(scrublet._counts_obs_norm - gene_means)
    if scrublet._counts_sim_norm is not None:
        scrublet._counts_sim_norm = sparse.csc_matrix(scrublet._counts_sim_norm - gene_means)
```
Although, I think it would make sense to make these class methods.  I mean `self` is already in the name.  And this would allow arguments to be passed into them a bit easier or things set.  Also there are loads of "private" attributes here being used, which seems like a red flag for the design.

ilan-gold: This seems like the sort of API we want to avoid in scanpy 2.0 to be honest.  I don't know if this is out of scope here, maybe a follow up PR is needed, but an alternative to making the `Scrublet` class all-encompassing would be a `ScrubletPipeline` class.

ilan-gold: I do understand the need/desire though to just make as few changes as possible to the original `scrublets` package, though.  @ivirshup should weigh in.

scanpy/preprocessing/_scrublet/__init__.py
```diff
@@ -563,27 +537,21 @@ def scrublet_simulate_doublets(
...
-    scrub = sl.Scrublet(X, random_state=random_seed)
+    scrub = Scrublet(X, random_state=random_seed)
 
     scrub.simulate_doublets(
         sim_doublet_ratio=sim_doublet_ratio,
         synthetic_doublet_umi_subsampling=synthetic_doublet_umi_subsampling,
     )
 
-    adata_sim = AnnData(scrub._E_sim)
+    adata_sim = AnnData(scrub._counts_sim)
```
ilan-gold, line 554: I am a bit of a sucker for `_xxx` as semantically indicating that they are private, but that is personal preference and I don't know what scanpy does in general off hand.  Again, the desire to make this PR small and minimal probably overrides this desire as there are other places where variable names did not change.  I do think then a refactor follow up would be good.

ilan-gold: I won't comment on every instance of these sorts of little variable things because it's out of scope.  But there are also these trailing underscores which I don't get.  Again, refactor is in order

scanpy/preprocessing/_scrublet/core.py
```diff
@@ -0,0 +1,466 @@
...
+    sim_doublet_ratio: float = 2.0
+    n_neighbors: InitVar[int | None] = None
+    expected_doublet_rate: float = 0.1
+    stdev_doublet_rate: float = 0.02
+    random_state: InitVar[AnyRandom] = 0
+
+    # private fields
+
+    _n_neighbors: int = field(init=False, repr=False)
+    _random_state: RandomState = field(init=False, repr=False)
```
ilan-gold, line 87: But this is used in non-private contexts? https://github.com/scverse/scanpy/pull/2703/files#diff-a42cf1237551ce149c338478446ae890441dc3871397ee52fd5af061fa25ba3eR436-R438

ilan-gold: This is true of other things as well, but in this case this one used to be public and making/using the private namespace created more diff-bloat actually.

scanpy/preprocessing/_scrublet/__init__.py
```diff
@@ -429,38 +410,31 @@ def _scrublet_call_doublets(
...
-        sl.pipeline_pca(
-            scrub, n_prin_comps=n_prin_comps, random_state=scrub.random_state
-        )
+        pipeline.pca(scrub, n_prin_comps=n_prin_comps, random_state=scrub._random_state)
     else:
         logg.info("Embedding transcriptomes using Truncated SVD...")
-        sl.pipeline_truncated_svd(
-            scrub, n_prin_comps=n_prin_comps, random_state=scrub.random_state
+        pipeline.truncated_svd(
+            scrub, n_prin_comps=n_prin_comps, random_state=scrub._random_state
```
ilan-gold, line 437: See other comment, but I don't really get why `self.random_state` can't be assigned in `__post_init__`

scanpy/preprocessing/_scrublet/core.py
```diff
@@ -0,0 +1,466 @@
...
+        else:
+            self.sim_doublet_ratio = sim_doublet_ratio
+
+        n_obs = self._counts_obs.shape[0]
+        n_sim = int(n_obs * sim_doublet_ratio)
+
+        pair_ix = sample_comb((n_obs, n_obs), n_sim, random_state=self._random_state)
+
+        E1 = cast(sparse.csc_matrix, self._counts_obs[pair_ix[:, 0], :])
+        E2 = cast(sparse.csc_matrix, self._counts_obs[pair_ix[:, 1], :])
```
ilan-gold, line 235: Again, not sure if the goal was copy as much and refactor later or not, but these are very unhelpful variable names.

ilan-gold: That being said, this one _is_ internal so won't contribute to the code-diff.  It might be worth handling internal changes here, but also don't mind if it's done later.

ilan-gold (commented):
In general I have lots of comments, but a lot of the code is copy-paste and I think that this approach is not a bad idea.  I think a follow-up refactor is needed.  I also am not a fan of the original design of editing things in place.  So I would be happy to review a follow up PR but I think that can be done separately.

scanpy/preprocessing/_scrublet/pipeline.py
flying-sheep: Yeah, I would personally do things quite differently, but this PR is mainly about getting the tests right and replacing the parts that scanpy already has.

scanpy/preprocessing/_scrublet/__init__.py
flying-sheep: Yeah, the parts I changed were mostly to help myself understand. This example is unifying the naming scheme for some related class field names.

flying-sheep: > But there are also these trailing underscores which I don't get. 

https://scikit-learn.org/dev/developers/develop.html#estimated-attributes

scanpy/preprocessing/_scrublet/core.py
flying-sheep: You’re right, this is not very clear. the `Scrublet` class isn’t public at all. I guess this means “fields that would be only for use in the library and not by users if the Scrublet class was a public API”

scanpy/preprocessing/_scrublet/__init__.py
flying-sheep: Because `field(converter=...)` only comes in Python 3.13: https://peps.python.org/pep-0712/

scanpy/preprocessing/_scrublet/core.py
flying-sheep: > Again, not sure if the goal was copy as much and refactor later or not

yes, it is

> but these are very unhelpful variable names.

I agree so much. Worst part of working with mathematicians.

scanpy/preprocessing/_scrublet/__init__.py
ilan-gold: Nice!

scanpy/preprocessing/_scrublet/__init__.py
ilan-gold: Ok, I guess there's no way around this then.  We want to make `random_state` public via `__init__` but then need to use it again in a different form.   Unfortunate.  I guess, though, `_random_state` is more of a generator though?  Could call it `rng`?  But fine leaving this

scanpy/preprocessing/_scrublet/__init__.py
```diff
@@ -138,9 +141,9 @@ def scrublet(
         `doublet_scores_sim_` histogram and/or based on co-localization of
         predicted doublets in a 2-D embedding.
     verbose
-        If True, print progress updates.
+        If True, log progress updates.
```
ilan-gold:
```suggestion
        If :data:`True`, log progress updates.
```

scanpy/preprocessing/_scrublet/__init__.py
```diff
@@ -292,41 +289,31 @@ def _run_scrublet(ad_obs, ad_sim=None):
 
     logg.info("    Scrublet finished", time=start)
 
-    if copy:
-        return adata
-    else:
-        return None
+    return adata if copy else None
```
ilan-gold, line 292:
```suggestion
    if copy:
        return adata
    return None
```
Not a huge fan of ternary in return statement

scanpy/preprocessing/_scrublet/__init__.py
```diff
@@ -382,7 +369,7 @@ def _scrublet_call_doublets(
     random_state
         Initial state for doublet simulation and nearest neighbors.
     verbose
-        If True, print progress updates.
+        If True, log progress updates.
```
ilan-gold:
```suggestion
        If :data:`True`, log progress updates.
```

ilan-gold (approved):

scanpy/preprocessing/_scrublet/__init__.py
flying-sheep: why? It’s much more concise and very readable, especially if all three expressions are a single word

ilan-gold: I guess the readable aspect is up for debate haha, but I approved so you absolutely do not have to do this.

scanpy/preprocessing/_scrublet/pipeline.py
ilan-gold: @flying-sheep This was what I asked for Isaac's opinion, but I think we decided that this PR would be restricted in scope so as to not include refactoring.  I don't think we need an opinion then.

flying-sheep: OK, great!

PR #2714, opened 2023-10-25
Add deprecated_func decorator

### LINKED ISSUE #2505: Tell people they’re using deprecated functions
#2500 probably happened because the user didn’t see that the function was deprecated.

We should maybe use a `@deprecated` decorator to both add the deprecation notice to the docs and also throw a `FutureWarning`.

uzolotikov:
Can you give a dict of deprecated functions and there new analogues? 
Then we can add f"Use {new_func[deprecated_func]} instead." for raise FutureWarning. 
 
 
 
- [x] Closes #2505  
Didn't add decorator to any function(don't know which are deprecated) 
 
Too small issue

### DIFF
--- a/scanpy/_utils/__init__.py
+++ b/scanpy/_utils/__init__.py
@@ -122,6 +122,13 @@ def deprecated_arg_names(arg_mapping: Mapping[str, str]):
     return decorator
 
 
+def deprecated_func(func: Callable[..., Any]) -> None:
+    """
+    Decorator that marks function as deprecated. It will raise FutureWarning for user.
+    """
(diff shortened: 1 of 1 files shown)


uzolotikov:
I think this pull request should be labeled as `/label no milestone`. Can someone take a look?

flying-sheep:
I think we should rather go with an existing decorator like https://github.com/tantale/deprecated

A PR should modify all deprecated functions and docstrings to rely on the decorator instead.

flying-sheep:
I’m tackling this now in https://github.com/scverse/scanpy/pull/3380

Thanks for starting this!

PR #2723, opened 2023-10-31
Use our neighbors for scrublet

flying-sheep:


scanpy/neighbors/__init__.py
```diff
@@ -584,7 +588,7 @@ def compute_neighbors(
             self._connectivities = _connectivity.gauss(
                 self._distances, self.n_neighbors, knn=self.knn
             )
-        else:
+        elif method is not None:
             msg = f"{method!r} should have been coerced in _handle_transform_args"
             raise AssertionError(msg)
         logg.debug("computed connectivities", time=start_connect)
```
ilan-gold: It seems like with `method=None` we actually did not compute connectivities, so shouldn't warn?

flying-sheep: With `methods=None`, we threw a `ValueError` before, no?

ilan-gold: I think it was the above `AssertionError`, but either way, this comment was more directed at the log message.  If we don't compute the connectivities, we probably shouldn't log that we do.  Sorry if I'm misunderstanding this.

flying-sheep: ah, got it!

PR #2742, opened 2023-11-09
Harmonization of Return Section in Documentation

### LINKED ISSUE #2276: quick sc.tl.louvain documentation fix
- [x] Other?
Quick documentation fix, under sc.tl.louvain:
https://scanpy.readthedocs.io/en/stable/generated/scanpy.tl.louvain.html
The return info only includes `adata.obs['louvain']` and doesn't include `adata.uns['louvain']['params']`

### LINKED ISSUE #2741: Documentation return fields are heterogeneous

Other?

The return fields in the documentation are heterogeneous.

E.g. for methods of the same output type scheme, such as [sc.pp.louvain](https://scanpy.readthedocs.io/en/latest/generated/scanpy.tl.louvain.html) and [sc.pp.leiden](https://scanpy.readthedocs.io/en/latest/generated/scanpy.tl.leiden.html), the return fields in the documentation are different.

Harmonizing return fields would make navigation through the different tools more clear.

One such style could be (tagging @flying-sheep here) to use prose containing a definition list, similar to

> Returns `None` if `copy=False`, otherwise an `AnnData` object. Sets the following:
>
> `.obs['foo']` : `NDArray[np.float64]`
> > Information on foo
>
>  `.obs['bar']` : `NDArray[np.float64]`
> > Information on bar

eroell:
- [x] Closes #2741, Closes #2276
- [x] Tests included or not required because: Only docstring changes

- [ ] Release notes not necessary because: Added

**Notes**

- [x]  How does `sc.pp.combat` return `None`? A: when `inplace=True`: it never returns an `AnnData` object though. Hence removed `AnnData` from the return type list.

- `sc.pp.normalize` has both `inplace` and `copy` so did not force to harmonize with others
- `sc.pp.pca` allows adata and array/sparsematrix input, so did not force to harmonize with others
- `sc.pp.filter_cells`,  `sc.pp.filter_genes` , `sc.pp.subsample` acts on data in different manner (changing dims), so did not force to harmonize with others
- `sc.pp.log1p` , `sc.pp.sqrt` seem to be understandable enough without bloating this up

### DIFF
--- a/scanpy/tools/_tsne.py
+++ b/scanpy/tools/_tsne.py
@@ -76,7 +76,7 @@ def tsne(
     -------
     Returns `None` if `copy=False`, else returns an `AnnData` object. Sets the following field:
 
-    `adata.obsm['X_tsne']` (:class:`numpy.ndarray`, dtype `float`)
+    `adata.obsm['X_tsne']` : :class:`numpy.ndarray` (dtype `float`)
         tSNE coordinates of data.
     """
     import sklearn
--- a/scanpy/tools/_tsne.py
+++ b/scanpy/tools/_tsne.py
@@ -74,9 +74,9 @@ def tsne(
 
     Returns
     -------
-    Depending on `copy`, returns or updates `adata` with the following fields.
+    Returns `None` if `copy=False`, else returns an `AnnData` object. Sets the following field:
 
-    **X_tsne** : `np.ndarray` (`adata.obsm`, dtype `float`)
+    `adata.obsm['X_tsne']` (:class:`numpy.ndarray`, dtype `float`)
         tSNE coordinates of data.
     """
     import sklearn
--- a/scanpy/tools/_tsne.py
+++ b/scanpy/tools/_tsne.py
@@ -76,8 +76,11 @@ def tsne(
     -------
     Returns `None` if `copy=False`, else returns an `AnnData` object. Sets the following field:
 
-    `adata.obsm['X_tsne']` : :class:`numpy.ndarray` (dtype `float`)
+    `adata.obsm['X_tsne']` : :class:`numpy.ndarray` | :class: `` (dtype `float`)
         tSNE coordinates of data.
+    `adata.uns['tsne']` : :class:`dict`
+        tSNE parameters.
+
     """
     import sklearn
 
--- a/scanpy/tools/_tsne.py
+++ b/scanpy/tools/_tsne.py
@@ -74,9 +74,9 @@ def tsne(
 
     Returns
     -------
-    Returns `None` if `copy=False`, else returns an `AnnData` object. Sets the following field:
+    Returns `None` if `copy=False`, else returns an `AnnData` object. Sets the following fields:
 
-    `adata.obsm['X_tsne']` : :class:`numpy.ndarray` | :class: `` (dtype `float`)
+    `adata.obsm['X_tsne']` : :class:`numpy.ndarray` (dtype `float`)
         tSNE coordinates of data.
     `adata.uns['tsne']` : :class:`dict`
         tSNE parameters.
--- a/docs/release-notes/1.10.0.md
+++ b/docs/release-notes/1.10.0.md
@@ -15,6 +15,7 @@
 ```{rubric} Docs
 ```
 * Fixed a lot of broken usage examples {pr}`2605` {smaller}`P Angerer`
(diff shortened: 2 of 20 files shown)


flying-sheep:
Pretty great! I switched it to use classifier syntax for the types like parameter docs, otherwise I saw nothing I think needs changing that isn’t in your notes

```restructuredtext
field : type
    description
```

flying-sheep (changes requested):
looks great! once the notes are addressed and you added a release note, this is good to go.

eroell:
Reviewed harmonization paying attention to some more details.

What do you think of using the `|` separator to describe `adata.X | adata.layers[layer]` e.g. [here](https://icb-scanpy--2742.com.readthedocs.build/en/2742/generated/scanpy.pp.regress_out.html)?

Some things causing some sort of heterogeneity and are NOT taken care of here:
- the inconsistent and mixed use of `inplace` and `copy` (effort: lot of work)
- some inconsistent use of `key_added` & flavours thereof, which affect the return section (effort: medium amount of work)

What is also not taken care of here:
- Other small things, such as [ingest](https://scanpy.readthedocs.io/en/latest/generated/scanpy.tl.ingest.html) not having a `return_joint` argument although this is mentioned in its doc. Might raise smaller issues in the future for these specific things rather than bloating this purpose-driven PR up?

scanpy/tools/_tsne.py
```diff
@@ -76,8 +76,11 @@ def tsne(
     -------
     Returns `None` if `copy=False`, else returns an `AnnData` object. Sets the following field:
 
-    `adata.obsm['X_tsne']` : :class:`numpy.ndarray` (dtype `float`)
+    `adata.obsm['X_tsne']` : :class:`numpy.ndarray` | :class: `` (dtype `float`)
```
flying-sheep: seems like you missed something here, otherwise beautiful!

flying-sheep:
> What do you think of using the | separator to describe adata.X | adata.layers[layer] e.g. [here](https://icb-scanpy--2742.com.readthedocs.build/en/2742/generated/scanpy.pp.regress_out.html)?

works for me!

> the inconsistent and mixed use of inplace and copy (effort: lot of work)

yeah, we definitely need to do this … in a dedicated PR

>  Might raise smaller issues in the future for these specific things rather than bloating this purpose-driven PR up?

Yeah! please add them here: https://github.com/orgs/scverse/projects/18/views/1

scanpy/tools/_tsne.py
eroell: Yes, think this should be gone after [303c385](https://github.com/scverse/scanpy/pull/2742/commits/303c385077d67b88440114684c6dc212b834f31d).

eroell:
An idea on the Python3.11 PyTest fail/whether its related to changes?

flying-sheep:
No, those “invalid instruction” errors pop up sometimes. I think they’re caused by some dependency being compiled for an instruction set that not all GitHub runners support. Ways to deal with it:

1. just restart until it works (annoying, but not much work)
2. figure out broken dependency, then
   1. if the wheel on PyPI is broken, raise an issue upstream
   2. if we compile it in the runner ourselves, set a compile flag to make it only use instructions that are compatible with all runners (i.e. not `-m arch=native` but select [an older architecture](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels))

flying-sheep (approved):

flying-sheep:
Hm. Are all functions that you edit in 1.9.6? Then we could set the milestone to 1.9.7.

If you’re not sure, it’s safer to set it to 1.10.0

eroell:
> Hm. Are all functions that you edit in 1.9.6? Then we could set the milestone to 1.9.7.
> 
> If you’re not sure, it’s safer to set it to 1.10.0

I think `sc.pp.normalize_per_cell` is not in the API section of the doc of 1.9.6 (intentionally?).
Other functions I modified are listed there as far as I checked.... You have `pp` or `tl` methods in mind which could conflict here?
Not 100% sure I don’t miss anything, so can add 1.10.0 👍

flying-sheep:
@eroell you [marked this as ready for review yesterday](https://github.com/scverse/scanpy/pull/2742#event-10978748914), but the title still says “draft”. Is this ready to merge?

You don’t need to add tags to the title, that’s what labels and thinks like draft status for PR are for.

eroell:
Yes ready for review, forgot the draft in the title.

flying-sheep:
Awesome, thank you so much!

PR #2743, opened 2023-11-09
fix: read header of "tissue_positions.csv"

### LINKED ISSUE #2746: The compatibility with SpaceRanger v2.0 or later of `read_visium` function

- [X] (optional) I have confirmed this bug exists on the master branch of scanpy.

### What happened?

When reading the `tissue_positions.csv` file generated by SpaceRanger v2.0 or later, `read_visium` reads the header from the second row (with `header=1`), while it should read from the first row instead (with `header=0`).

### Minimal code sample

```python
positions = pd.read_csv(
            files['tissue_positions_file'],
            header=0 if tissue_positions_file.name == "tissue_positions.csv" else None,
            index_col=0,
        )
```

### Error output

_No response_

### Versions

<details>

```
anndata     0.10.2
scanpy      1.10.0.dev157+g9b7f6032
```

</details>

senbaikang:
- [x] Closes  #2746
- [x] Tests included or not required because: this is a minor fix

- [x] Release notes not necessary because: the related code has not been released by scanpy

### DIFF
--- a/scanpy/readwrite.py
+++ b/scanpy/readwrite.py
@@ -460,7 +460,7 @@ def read_visium(
         # read coordinates
         positions = pd.read_csv(
             files["tissue_positions_file"],
-            header=1 if tissue_positions_file.name == "tissue_positions.csv" else None,
+            header=0 if tissue_positions_file.name == "tissue_positions.csv" else None,
             index_col=0,
         )
         positions.columns = [


flying-sheep:
Please provide context, ideally link an issue that describes the problem. Then in here, describe why your change fixes it.

senbaikang:
Hi @flying-sheep, thank you for the reply. I have created an issue #2746 and described the problem there. I find that this issue has been [mentioned](https://github.com/scverse/scanpy/pull/2424#issuecomment-1607082327) before, but for some reason it was not fixed right away.

flying-sheep:
Great! And indeed, it hasn’t been released yet, so if you don’t want one, we don’t need a release note.

PR #2756, opened 2023-11-16
Improve neighbors shortcut

flying-sheep:
- [x] Closes N/A
- [x] Tests included

- [x] Release notes not necessary because: Modifies unreleased features

I made the code and comments/docs more explicit about when the nearest-neighbor-is-cell-itself comes into play and when it doesn’t, and tested a bunch of our utilities.

@mumichae this might make it fast for your code
You can install this branch using

```bash
pip install -U 'scanpy@git+https://github.com/scverse/scanpy.git@improve-neighbors-shortcut'
```

|   | Before | After |
| --------- |--------|--------|
| small data | master | shortcut |
| bigger data | master | shortcut|

### DIFF
--- a/scanpy/_utils/compute/is_constant.py
+++ b/scanpy/_utils/compute/is_constant.py
@@ -80,7 +80,7 @@ def is_constant(
 def _(a: NDArray, axis: Literal[0, 1] | None = None) -> bool | NDArray[np.bool_]:
     # Should eventually support nd, not now.
     if axis is None:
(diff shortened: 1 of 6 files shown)


flying-sheep:
@Intron7 this was surprisingly hard to get right. Unfortunately, there are now a few more checks and some `hstack`ing.

Do those tank the performance?

/edit: I benchmarked some, this is better than what we had before

PR #2757, opened 2023-11-16
pp.highly_variable_genes handles subset and inplace consistently

### LINKED ISSUE #1867: sc.pp.highly_variable does not always handle `subset` and `inplace` arguments consistently
Hey,
while writing tests for #1715 I noted the following behavior:

`output = sc.pp.highly_variable(adata,inplace=True,subset=False,n_top_genes=100)`
--> Returns nothing :heavy_check_mark: 
--> `adata.var` fields are updated but shape stays the same :heavy_check_mark: 

`output = sc.pp.highly_variable(adata,inplace=True,subset=True,n_top_genes=100)`
--> Returns nothing  :heavy_check_mark: 
--> `adata.var` fields are updated and shape is subsetted correctly :heavy_check_mark: 

`output = sc.pp.highly_variable(adata,inplace=False,subset=False,n_top_genes=100)`
--> output is a dataframe with the original number of genes as rows :heavy_check_mark: 
--> `adata` is unchanged :heavy_check_mark: 

`output = sc.pp.highly_variable(adata,inplace=False,subset=True,n_top_genes=100)`
--> Returns nothing :x: 
--> `adata` shape is changed an `var` fields are updated :x: 

I think the last case is unexpected, assuming that `inplace=False` should protect `adata` from changes and results in an output. I think these parts of code are the cause (depending on the flavor):

https://github.com/theislab/scanpy/blob/4dd8de9e355ce8d59009c15522670a3d0970462c/scanpy/preprocessing/_highly_variable_genes.py#L148-L174

https://github.com/theislab/scanpy/blob/4dd8de9e355ce8d59009c15522670a3d0970462c/scanpy/preprocessing/_highly_variable_genes.py#L528-L553

To fix this, one could change `if subset or inplace:` to `if inplace:`. Then one would need to add another `if subet:` in the `else` block, like so (half-pseudocode):

```
    if inplace:        
        
       #update adata
        
        if batch_key is not None:
           #drop batch related keys
        if subset:
            adata._inplace_subset_var(df['highly_variable'].values)
    else:
        if batch_key is None:
            #drop batch related keys
        if subset:            
            df=df.iloc[df.highly_variable.values,:]
            
        return df
```

I can make a quick PR if this is the intended behavior. :slightly_smiling_face: 
best, jan.

eroell:
- [X] Fixes #1867
- [X] Tests included or not required because: New tests included which catch the failure mode described in #1867. Current implementation fails these.

- [ ] Release notes not necessary because: Added entry in release notes.

Addresses issue #1867 with a fix as outlined by @jlause and tests which catch the failure mode detected and nicely demonstrated by @jlause.


scanpy/tests/test_highly_variable_genes.py
```diff
@@ -514,3 +514,52 @@ def test_cellranger_n_top_genes_warning():
...
+        else:
+            assert len(adata.var) == n_genes
+
+    else:
+        assert output_df is not None
+
+        if subset:
+            assert len(output_df) == 10
+        else:
+            assert len(output_df) == n_genes
```
flying-sheep:
```suggestion
    assert (output_df is None) == inplace
    assert len(adata.var if inplace else output_df) == 10 if subset else n_genes
```

flying-sheep (approved):

scanpy/tests/test_highly_variable_genes.py
```diff
@@ -514,3 +514,39 @@ def test_cellranger_n_top_genes_warning():
...
+@pytest.mark.parametrize("flavor", ["seurat", "cell_ranger"])
+@pytest.mark.parametrize("subset", [True, False])
+@pytest.mark.parametrize("inplace", [True, False])
+def test_highly_variable_genes_subset_inplace_consistency(
+    flavor,
+    subset,
+    inplace,
+):
+    if flavor == "seurat" or flavor == "cell_ranger":
+        adata = pbmc3k()
```
eroell: Since for this test shape is most relevant, will try to use a minimalistic dataset (which does not clash with the highly_variable_genes flavors)

eroell: got it I think

PR #2761, opened 2023-11-23
Add all ruff checks we use in anndata

### LINKED ISSUE #2762: Add lints and import sorting

Other?

Just like in scverse/anndata#1125.

After https://github.com/scverse/anndata/issues/1241, we can do a companion PR here too

flying-sheep:
- [x] Closes #2762
- [x] Tests included

- [x] Release notes not necessary because: dev process change

All changes were automatic, except for:

- removing unused imports or replacing them with `__all__`. Much more uncontroversial than AnnData as scanpy’s public modules were more well defined from the start. There were no ambiguous cases except for `sc.get` re-exporting `"_check_mask"`, `"_get_obs_rep"`, and `"_set_obs_rep"`. But since those aren’t documented, we can decide over their fate at a later date.
- Fixing circular imports like `sc.{pp,tl}.pca`. I only needed to create a `neighbors/_doc.py` file for shared neighbors/tools doc parts, and put the `pca` import in `tools` in `__getattr__` (supported since 3.7)
- replacing some `with open(p) as f: x = json.loads(f.read())`s with `x = json.loads(Path(p).read_text())`

All in all surprisingly easy


docs/api/tools.md
```diff
@@ -17,7 +17,7 @@ Any transformation of the data matrix that is not *preprocessing*. In contrast t
    :nosignatures:
    :toctree: ../generated/
 
-   tl.pca
+   pp.pca
```
flying-sheep, line 20: I took the liberty to make this more Zen.

It’s the same function, it should be only in one place. (It’s still importable from the other, we can add a deprecation if we want, or just leave it.)

The only places that used to use it from `tl` are externals, I harmonized all that.

PR #2769, opened 2023-11-28
obs. vs. obs for all grouped plots using `BasePlot`

yugeji:
- [x] Closes [#1876](https://github.com/scverse/scanpy/issues/1876) and reimplements stale [PR #2055](https://github.com/scverse/scanpy/pull/2055) zhangguy:dotplot2d
- [ ] Empty group currently breaks
- [ ] dendrogram doesn't work with all plots
- [ ] I don't think anything would break if you just removed var_name reordering from the dendrogram entirely
- [ ] add tests using `groupby_cols` params
- [ ] Docs!


scanpy/plotting/_stacked_violin.py
```diff
@@ -461,38 +475,66 @@ def _make_rows_of_violinplots(
...
         # violin plot will not distinguish those genes
-        _matrix.columns = [f"{x}_{idx}" for idx, x in enumerate(_matrix.columns)]
-
-        # transform the  dataframe into a dataframe having three columns:
+        # _matrix.columns = [f"{x}_{idx}" for idx, x in enumerate(_matrix.columns)] # added warning as this row was commented out
+        if _matrix.columns.value_counts().max() > 1:
+            warn(
+                "Duplicate gene or condition labels present."
+                "Results might be unexpected."
+            )
```
flying-sheep, line 483: @yugeji commenting this out doesn’t work with newer pandas versions anymore. Does your branch handle duplicated columns? Then we can just move this into the `if len(self.groupby_cols) < 1:` branch

Issue #1244, opened 2020-05-22
Import error when old version of tqdm is installed


grst:
When an old version of tqdm ([< 4.29.1](https://github.com/nteract/papermill/issues/287)) is installed, scanpy cannot be imported. 

```bash
conda create -n test_scanpy_tqdm scanpy=1.5.1 tqdm=4.19.4
conda activate test_scanpy_tqdm
python -c "import scanpy as sc" 
```

```pytb
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/sturm/anaconda3/envs/test_scanpy_tqdm/lib/python3.8/site-packages/scanpy/__init__.py", line 39, in <module>
    from . import datasets, logging, queries, external, get
  File "/home/sturm/anaconda3/envs/test_scanpy_tqdm/lib/python3.8/site-packages/scanpy/datasets/__init__.py", line 3, in <module>
    from ._datasets import (
  File "/home/sturm/anaconda3/envs/test_scanpy_tqdm/lib/python3.8/site-packages/scanpy/datasets/_datasets.py", line 8, in <module>
    from tqdm.auto import tqdm
ModuleNotFoundError: No module named 'tqdm.auto'
```

### Suggested solution
Either 
 * merge #1130 (which has the additional benefit that the [ipywidgets extension does not have to be installed when running in jupyterlab](https://stackoverflow.com/questions/53247985/tqdm-4-28-1-in-jupyter-notebook-intprogress-not-found-please-update-jupyter-an)) or 
 * require a minimal version of `tqdm>=4.29.1`.

PR #2771, opened 2023-11-28
Generalize tl.dendrogram function to either axis

### LINKED ISSUE #1549: sc.tl.dendrogram doesn't use var_names
Latest on pip at least scanpy-1.6.0
---

I'm using the sc.pl.dendrogram multiple times different lists of genes on my dataset (incrementing number of highly variable genes basically). The outputted dendrogram is alway the same (I guess it's taking into account all the genes because it's using something like 32go of ram....)

```python
hvegene_sets = [sc.pp.highly_variable_genes(adata, inplace=False, subset=False, n_top_genes=nhvg)["highly_variable"] for nhvg in [500,1000,2000, 3000,4000, 5000]]
```
then
```python
[sum(hvgene) for hvgene in hvegene_sets]
```
outputs:
[499, 1000, 1999, 2999, 4000, 4999] (so i have my different genesets)

then 
```python
dendro1 = sc.tl.dendrogram(adata,                   
                   var_names=adata.var_names[hvegene_sets[1]].values, 
                   optimal_ordering=True,
                   cor_method="spearman", linkage_method="complete", inplace=False,
                   groupby="Annotation")
dendro2 = sc.tl.dendrogram(adata,                   
                   var_names=adata.var_names[hvegene_sets[5]].values, 
                   optimal_ordering=True,
                   cor_method="spearman", linkage_method="complete", inplace=False,
                   groupby="Annotation")
[dendro1[key] ==dendro2[key] for key in dendro1.keys()] 
```
outputs: 
```code
[array([[ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True],
        [ True,  True,  True,  True]]),
 True,
 True,
 True,
 True,
 True,
 True,
 True,
 array([[ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True],
        [ True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True,  True,  True,  True,  True,  True,  True,  True,
          True,  True]])]
```
At first I was creating all dendrograms in a list comprehension and it did the same. 
I also directly inputted a list of my own and I obtained the same result....
I guess dendrogram don't detect the genes.

When running functions such as 
```python 
## Testing with creating the dendro manually

def do_corr_mat(adata,  var_names, groupby, method = "spearman") :
    categories, obs_tidy = _prepare_dataframe(adata, var_names=var_names, groupby=groupby)
    mean_df = obs_tidy.groupby(level=0).mean()
    
    return mean_df.T.corr(method=method)

def do_dendro(corr_matrix, method="ward") :
    z_var = linkage(corr_matrix, method=linkage)
    return dendrogram(z_var, labels=mean_df.index)
``` 
Everything works fine ! 

Thanks by advance, 
C

#### Versions

1.6.0

</details>

dburkhardt:
- [X] Closes #1549
- [X] Tests included or not required because:
  - Added regression test for subsetting var_names
  - Added test for when groupby is None


scanpy/tools/_dendrogram.py
```diff
@@ -103,49 +112,70 @@ def dendrogram(
...
-
-        rep_df.set_index(categorical, inplace=True)
-        categories = rep_df.index.categories
+            categorical = adata.obs[groupby[0]]
+            if len(groupby) > 1:
+                for group in groupby[1:]:
+                    # create new category by merging the given groupby categories
+                    categorical = (
+                        categorical.astype(str) + "_" + adata.obs[group].astype(str)
+                    ).astype("category")
```
ilan-gold:
```suggestion
            if len(groupby) == 1:
                categorical = adata.obs[groupby[0]]
            else:
                categorical = adata.obs[groupby].apply("_".join, axis=1).astype("category")
```

scanpy/tools/_dendrogram.py
```diff
@@ -103,49 +112,70 @@ def dendrogram(
...
-        groupby = [groupby]
-    for group in groupby:
-        if group not in adata.obs_keys():
-            raise ValueError(
-                "groupby has to be a valid observation. "
-                f"Given value: {group}, valid observations: {adata.obs_keys()}"
+    if dim is None and axis is None:
+        dim = "obs"
+    axis, dim = _resolve_dim(axis=axis, dim=dim)
+    alt_axis, alt_dim = _resolve_dim(axis=1 - axis)
```
ilan-gold: This variable is not used.

scanpy/tools/_dendrogram.py
```diff
@@ -103,49 +112,70 @@ def dendrogram(
...
+        else:
+            rep_df = pd.DataFrame(
+                _choose_representation(
+                    adata[:, var_names], use_rep=use_rep, n_pcs=n_pcs
+                )
+            )
+        categories = rep_df.axes[axis]
 
     import scipy.cluster.hierarchy as sch
     from scipy.spatial import distance
```
ilan-gold: Not sure if there is done all over scanpy, but wouldn't wrapping this in a try-catch block with a useful error message be good?

scanpy/tools/_dendrogram.py
```diff
@@ -60,6 +63,12 @@ def dendrogram(
...
         Annotated data matrix
+    groupby
+        Optional, the obs column(s) to use to group observations. Default is None.
+    dim
+        Dimension (obs or var) along which to calculate the dedrogram. Pass either `dim` or `axis` but not both.
+    axis
+        Axis (0 or 1) along which to calculate the dendrogram. Pass either `dim` or `axis` but not both.
     {n_pcs}
     {use_rep}
     var_names
```
ilan-gold, line 82: Is it overkill to make this also generalize i.e., you can `groupby` on either axis and you can subset (i.e., what `var_names` does now) on the opposite axis?

ilan-gold: I will say this part kind of threw me given the title of the PR.  I see what what is going on now, but would be curious your thoughts @flying-sheep.  I would expect to do identical operations on either axis but again, maybe overkill

scanpy/tools/_dendrogram.py
ilan-gold: nit and unrelated to this PR, but I think this should go first in the function.  The function can't run without them and the `groupby` call can be expensive.  It happens before this, so no point wasting people's time if they somehow don't have these packages installed (I understand that they are dependencies so that shouldn't happen, but still).

scanpy/tools/_dendrogram.py
ilan-gold: Sorry this looks like old code, but this seems cleaner and also seems to work.

flying-sheep:
pre-commit.ci autofix

scanpy/tools/_dendrogram.py
flying-sheep: > Not sure if there is done all over scanpy, but wouldn't wrapping this in a try-catch block with a useful error message be good?
> I understand that they are dependencies so that shouldn't happen, but still

Why still? Of course all hard dependencies (like scipy) are expected to be installed properly. No need to handle this in any special way.

The inline import is probably for performance reasons.

ilan-gold: You're right, there really isn't a "still" here.  That makes sense

flying-sheep:
OK, updated this so it follows the decision implemented in #1244

flying-sheep (changes requested):
OK, I checked this out and did some more stylistic changes myself.

The big things for me are:

1. Documentation:

   > The obs column(s) to use to group observations. Default is None.
   
   What does `None` mean?
   
2. The corresponding `.pl` function: Examples look like this:

   ```py
   >>> sc.tl.dendrogram(adata, groupby='bulk_labels')
   >>> sc.pl.dendrogram(adata, groupby='bulk_labels')
   ```

   and there’s no guidance or example for leaving out `pl`.
   If people try this, it will throw an error:

   ```py
   >>> sc.tl.dendrogram(adata)
   >>> sc.pl.dendrogram(adata)
   ```

   So will this

   ```py
   >>> sc.tl.dendrogram(adata, axis='var')
   >>> sc.pl.dendrogram(adata, axis='var')
   ```
   
   I think these things should work, or at least be very well documented

dburkhardt:
@flying-sheep the first is easy to fix. For 2. , it's not clear to me what you're asking for here. It's been a while since I worked on this. Am I supposed to import this function in the `__init__.py` for `tl` and `pl`?

flying-sheep:
Hi Dan, about 1.: I’m asking you what the semantic meaning is :smile: 

about 2.: there are two functions called `dendrogram`, and they have compatible signatures. Each computed dendrogram can be plotted. So what I’m saying is that the plotting version hasn’t been adapted.

Also an important question: in `tl.dendrogram`, we call `_choose_representation`, which will compute a PCA for the .obs axis. When specifying `axis='var'`, should it compute a PCA for the `var` axis instead?

PR #2772, opened 2023-11-30
Faster filtering on sparse matrices

gjhuizing:
- [ ] Closes # _no existing issue_
- [ ] Tests included or not required because: _No new tests_

- [ ] Release notes not necessary because: _I did not write release notes_

Hi :)

I am proposing a change that speeds up `filter_cells` (x1000 speedup) and  `filter_genes` (x2 speedup) for CSR sparse matrices. On my personal machine for 1M cells, `sc.pp.filter_cells(adata, min_genes=xx)` runs in 1ms instead of 10s currently. The speedup should be even stronger on sparser modalities like ATAC.

In spirit, this simply replaces `np.sum(X > 0, axis=axis)` with `X.getnnz(axis=axis)`, which is much more efficient. But the axis argument in `getnnz` in `csr_array` may be deprecated. I think it should still be fine with `csr_matrix`, but since I don't know for sure I manually implemented it for the CSR case as in https://github.com/scipy/scipy/issues/19405 .

What do you think?

Regarding `getnnz`: Of course it would be nicer to be able to write `.getnnz(axis=axis)`, which extends beyond CSR to other sparse matrices. Can we assume that we're getting sparse matrices and not sparse arrays ?

Pinging @dschult from the Scipy issue liked above, who mentioned: 

> I'm pretty sure that a reasonable and commonly occuring use-case would be enough to make the developers include this feature somehow.

(edited because I confused `csr_array` and `csr_matrix`)

### DIFF
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -148,10 +148,13 @@ def filter_cells(
     X = data  # proceed with processing the data matrix
     min_number = min_counts if min_genes is None else min_genes
     max_number = max_counts if max_genes is None else max_genes
-    number_per_cell = np.sum(
-        X if min_genes is None and max_genes is None else X > 0, axis=1
-    )
-    if issparse(X):
+    if min_genes is None and max_genes is None:
+        number_per_cell = np.sum(X, axis=1)
+    elif isspmatrix_csr(X):
+        number_per_cell = np.diff(X.indptr)
+    else:
+        number_per_cell = np.sum(X > 0, axis=1)
+    if issparse(X) and not isspmatrix_csr(X):
         number_per_cell = number_per_cell.A1
     if min_number is not None:
         cell_subset = number_per_cell >= min_number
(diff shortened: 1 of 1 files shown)


gjhuizing:
This made me realize `.getnnz` is used elsewhere in the code, opened an issue https://github.com/scverse/scanpy/issues/2773

scanpy/preprocessing/_simple.py
```diff
@@ -148,11 +148,16 @@ def filter_cells(
...
-        X if min_genes is None and max_genes is None else X > 0, axis=1
-    )
-    if issparse(X):
-        number_per_cell = number_per_cell.A1
+    if min_genes is None and max_genes is None:
+        number_per_cell = np.sum(X, axis=1)
+        if issparse(X):
+            number_per_cell = number_per_cell.A1
+    elif isspmatrix_csr(X):
+        number_per_cell = np.diff(X.indptr)
```
dschult, line 156: I believe that this will work for `csr_array` (and `csr_matrix` of course) if you change this `elif` clause:
```suggestion
    elif issparse(X) and X.format == 'csr':
        number_per_cell = np.diff(X.indptr)
```

scanpy/preprocessing/_simple.py
```diff
@@ -260,11 +265,16 @@ def filter_genes(
...
-    number_per_gene = np.sum(
-        X if min_cells is None and max_cells is None else X > 0, axis=0
-    )
-    if issparse(X):
-        number_per_gene = number_per_gene.A1
+    if min_cells is None and max_cells is None:
+        number_per_gene = np.sum(X, axis=0)
+        if issparse(X):
+            number_per_gene = number_per_gene.A1
+    elif isspmatrix_csr(X):
```
dschult, line 272: similarly here:
```suggestion
    elif issparse(X) and X.format == 'csr':
```

dschult (commented):
I believe to make this work for `csr_array`, you just need to avoid `isspmatrix_csr` (which rules out the `*_array` objects). more below.

PR #2776, opened 2023-12-05
Enable scrublet tests

flying-sheep:
- [x] Closes N/A

- [x] Release notes not necessary because: dev changes

This PR contains more quantitative scrublet tests.

It should be merged prior to #2703 to make sure our changes don’t change how scrublet works.


flying-sheep:
OK, since Isaac has no time, I guess we add more tests in a follow-up PR if necessary

Issue #2088, opened 2021-12-18
`flavor="seurat_v3"` HVG aggregation over batches not intended 


adamgayoso:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of scanpy.
- [x] (optional) I have confirmed this bug exists on the master branch of scanpy.

In the test of the flavor, the sorting over batches is correct with respect to how Seurat actually does it:

https://github.com/theislab/scanpy/blob/83f90141fd18943a1795772d3d39f4e9eefd65c3/scanpy/tests/test_highly_variable_genes.py#L138-L142

First sort by number of batches it's called a HVG and then second break ties by the median rank.

In the actual function, it's the reverse:

https://github.com/theislab/scanpy/blob/83f90141fd18943a1795772d3d39f4e9eefd65c3/scanpy/preprocessing/_highly_variable_genes.py#L136-L143

I propose we put in the fix (which is currently being tested with high accuracy). The simplest solution would be to call it a new flavor. Ideally this would get in as soon as possible as this is a high volume function from what I understand.

ivirshup:
Might be of interest to @gokceneraslan @jlause

jlause:
I think this is related to #1733 (which I never followed up on, sorry about that!)

adamgayoso:
I suppose we need to maintain backwards compatibility, so I still opt for the new flavor plus a userwarning on the old flavor

adamgayoso:
Any ideas on how to handle this? It would be great to at least be consistent with what Seurat does (even if we haven't come to a conclusion on the ideal to gather across batches)

adamgayoso:
@Zethson @ivirshup can we make this change? It's disappointing that this has been around for so long as users might not realize there's an inconsistency.

flying-sheep:
OK, let’s close this in favor of #1733 and update that one with the newest information if necessary

Issue #1733, opened 2021-03-10
sc.pp.highly_variable_genes with flavor='seurat_v3' and batch_key: conflict with documentation


jlause:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of scanpy.
- [x] (optional) I have confirmed this bug exists on the master branch of scanpy.

---

Hey, I've noticed another potential problem within the `seurat_v3` flavor of `sc.pp.highly_variable_genes()`. The documentation of the `batch_key` argument says on how the genes are ranked

>For all flavors, genes are first sorted by how many batches they are a HVG. For dispersion-based flavors ties are broken by normalized dispersion. If `flavor = 'seurat_v3'`, ties are broken by the median (across batches) rank based on within-batch normalized variance.

However, when genes are sorted after computing everything, the `seurat_v3` method sorts first by the median ranks and then by how many batches a gene is highly variable (contrary to what the docstring says):

https://github.com/theislab/scanpy/blob/da66e15eee137b51868a6fb85603cfca9de557a7/scanpy/preprocessing/_highly_variable_genes.py#L139-L144

For comparison, the other flavors sort the other way around (as the docstring says):

https://github.com/theislab/scanpy/blob/da66e15eee137b51868a6fb85603cfca9de557a7/scanpy/preprocessing/_highly_variable_genes.py#L505-L510

Not sure which sorting would be correct here to make `seurat_v3` match the original behavior in Seurat, but I found these comments by @adamgayoso in #1204 that mention that when using `batch_key` in the current implementation, results don't match Seurat.. maybe the sorting order is the problem here?

https://github.com/theislab/scanpy/pull/1204#issuecomment-645700601
https://github.com/theislab/scanpy/pull/1204#issuecomment-663879113

Here is a piece of code that shows the top5 genes after `seurat_v3` gene selection and sorting by `highly_variable_nbatches` (which should then all be HVG, but they are not)

<details>

```python
import numpy as np
import scanpy as sc
import anndata 

import sys
sys.path.append("scanpy/preprocessing")
from _utils import _get_mean_var

np.random.seed(42)
adata = anndata.AnnData(np.random.randint(0,5,(100,100)))
adata.obs['batch'] = np.random.randint(0,5,(100))
adata.obs['batch'] = adata.obs['batch'].astype('category')
n_top_genes = 50

adata = adata.copy()
[...]

Issue #2780, opened 2023-12-15
`highly_variable_genes(flavor='seurat')` results differ from Seurat’s HVG results


carversh:
### What kind of feature would you like to request?

Additional function parameters / changed functionality / changed defaults?

### Please describe your wishes

If these methods are supposed to be similar, and I am normalizing and taking the log of the count matrices in both cases to produce the same log-normalized count matrix, then why am I getting super different results?

flying-sheep:
@giovp @adamgayoso you had thoughts on this, right?

adamgayoso:
I only know about the issue with how genes are aggregated over batches with `flavor='seurat_v3'`. See ~~#2088~~ #1733.

I vaguely remember older issues with comparisons of the cellranger and original seurat flavors

flying-sheep:
I’m talking about [this private thread](https://scverse.zulipchat.com/#narrow/stream/315186-core-team/topic/Lior.2C.20again/near/386542660).

giovp:
@flying-sheep I think in that thread we were discussing about differential expression and not feature selection. I _think_ some of those pints were addressed during the hackathon.

Regarding inconsistency, I think there has been some work here https://github.com/scverse/scanpy/pull/1732 but admittedly there were still discrepancies. 

More generally, I think it would be useful to understand how much is the discrepancy there is right now, @carversh would you be able to provide some stats?

carversh:
Yes, I'll send an example in a bit, recovered variable genes seem wildly discrepant. I can get to this tomorrow! Thanks for your quick response

carversh:
Hi everyone, here is the way I extracted the top 500 most variable gees in seurat: 
![Screen Shot 2023-12-22 at 11 58 39 AM](https://github.com/scverse/scanpy/assets/65792233/5b59b6d1-696e-4695-a7df-ad8f74d6412f)

carversh:
Here is the way I extracted the most variable genes in scanpy: 
![Screen Shot 2023-12-22 at 11 59 13 AM](https://github.com/scverse/scanpy/assets/65792233/97be237d-eff8-4910-b858-163d301c2bf6)

carversh:
And out of the 500 genes identified, only 138 overlap between the sets: 
![Screen Shot 2023-12-22 at 11 59 35 AM](https://github.com/scverse/scanpy/assets/65792233/7d2459b2-a167-45cb-8107-9433aae6aac4)

carversh:
[...]

PR #2792, opened 2023-12-22
`sc.pp.highly_variable_genes(adata, flavor="seurat_v3")` consistency improvement and implementation discussion

eroell:
- [x] Adresses #2088 and adresses #1733

- [x] Tests included or not required because: They are required and some suggested
- [x] Release notes
- [x] Doc update - depending on feedback here
- [x] Doc update - guidance scanpy vs seurat

**Context**
As discussed in issues #2088 and #1733, `sc.pp.highly_variable_genes(adata, flavor="seurat_v3", batch_key=SOME_KEY)` potentially differs in the implementation of how HVGs are ranked from its Seurat counterpart:
- either by sorting by number-of-batches-in-which-genes-are-highly-variable and then breaking ties with median-rank-in-batches (this is described in [Stuart et al. 2019](https://www.cell.com/cell/pdf/S0092-8674(19)30559-8.pdf), and implemented in Seurat's [`SelectIntegrationFeatures`](https://satijalab.org/seurat/reference/selectintegrationfeatures)*).
- OR by sorting first by median-rank-in-batches and breaking ties with number-of-batches-in-which-genes-are-highly-variable (this is how `"seurat_v3"` in scanpy is currently implemented)
 
causing quite some discrepancy in the results.

*I am not an R expert, so this might not be correct: Digging into the code of `SelectIntegrationFeatures`, I suspect the genes _above_ a treshold level of batches in which they are HVGs are [ordered by their median rank](https://github.com/satijalab/seurat/blob/41d19a8a55350bff444340d6ae7d7e03417d4173/R/integration.R#L2988), in contrary to the textual description in Stuart et al.; and only the genes displaying this threshold of number of batches in which they are highly variable are ranked by their median rank - to decide which are kept as highly variable. This would have an effect on the ordering of the very top genes, but NOT on the actual genes which are selected by `SelectIntegrationFeatures`.

**Note**
All of this does not affect the fairly good match, up to potentially numerics, between `sc.pp.highly_variable_genes(adata, flavor="seurat_v3", batch_key=None)` and Seurat's `FindVariableFeatures` with `selection.method = 'vst'`  introduced in Stuart et al.
If it helps to avoid confusion between the two: `FindVariableFeatures` is called within `SelectIntegrationFeatures`, on each batch separately.

**Technical additions here**
This PR suggests to solve this by introducing a new flavor. Either

-`seurat_v3_paper` This fixes to exactly what @jlause noticed and @adamgayoso pinpointed in #1733.
OR
-`seurat_v3_implementation` This matches more closely the suspected Seurat implementation I mentioned above.

They select the same genes. Leaning towards favoring the style of `seurat_v3_paper`.

Better naming suggestions more than welcome.

**Examples**
- Good when no `batch_key` used:
```py
import numpy as np
import pandas as pd
import scanpy as sc

# load exactly the data from seurat tutorial

pbmc = sc.datasets.pbmc3k()
# use the exact filterin from Seurat tutorial
sc.pp.filter_cells(pbmc, min_genes=200) # this doesnt do anything btw
sc.pp.filter_genes(pbmc, min_cells=3)

print(pbmc)

# default settings in scanpy are the same as for Seurat
sc.pp.highly_variable_genes(pbmc, flavor="seurat_v3")

# this has been prepared in the R script "scanpy/scanpy/tests/_scripts/seurat_extract_hvg_v3.R" (adapted from https://satijalab.org/seurat/articles/pbmc3k_tutorial)
pbmc3k_tutorial_FindVariableGenes_seurat = pd.read_csv("scanpy/scanpy/scanpy/tests/_scripts/seurat_hvg_v3.csv.gz", index_col=0)

# This is used to order and rank the hvg when no batch information used
assert np.allclose(
    pbmc3k_tutorial_FindVariableGenes_seurat["variance.standardized"],
    pbmc.var["variances_norm"],
)

# Another quantity reported by both
assert np.allclose(
    pbmc3k_tutorial_FindVariableGenes_seurat["mean"],
    pbmc.var["means"],
)

# Another quantity reported by both
assert np.allclose(
    pbmc3k_tutorial_FindVariableGenes_seurat["variance"],
    pbmc.var["variances"]),
)
```

- Discrepancy when `batch_key` is used: The flavor `'seurat_v3'` shows 764/2000 genes overlap with Seurat. The new suggested flavors `'seurat_v3_paper'` and `'seurat_v3_implementation'` show a 1990/2000 selected genes overlap with Seurat. The 10 non-matching genes contain 1 gene renamed by Seurat and therefore not found matching, some corner cases where numerics might have turned narrow outcomes, so actually 1991/2000.

```py
pmbc = sc.datasets.pbmc3k()
# use the exact filterin from Seurat tutorial
sc.pp.filter_cells(pbmc, min_genes=200) # this doesnt do anything btw
sc.pp.filter_genes(pbmc, min_cells=3)

# introduce a dummy "technical covariate"; this is used in Seurat's SelectIntegrationFeatures
pbmc.obs["dummy_tech"] = "source_1"
pbmc.obs.loc[pbmc.obs.index[500:1000], "dummy_tech"] = "source_2"
pbmc.obs.loc[pbmc.obs.index[1000:1500], "dummy_tech"] = "source_3"
pbmc.obs.loc[pbmc.obs.index[1500:2000], "dummy_tech"] = "source_4"
pbmc.obs.loc[pbmc.obs.index[2000:], "dummy_tech"] = "source_5"

# default settings in scanpy are the same as for Seurat
seurat_v3_hvg = sc.pp.highly_variable_genes(pbmc, flavor="seurat_v3", batch_key="dummy_tech", inplace=False)
seurat_v3_paper_hvg = sc.pp.highly_variable_genes(pbmc, flavor="seurat_v3_paper", batch_key="dummy_tech", inplace=False)
seurat_v3_implementation_hvg = sc.pp.highly_variable_genes(pbmc, flavor="seurat_v3_implementation", batch_key="dummy_tech", inplace=False)

# this has been prepared in the R script "scanpy/scanpy/tests/_scripts/seurat_extract_hvg_v3.R" (adapted from https://satijalab.org/seurat/articles/pbmc3k_tutorial)
pbmc3k_tutorial_FindVariableGenes_seurat_batch = pd.read_csv("scanpy/scanpy/tests/_scripts/seurat_hvg_v3_batch.csv", index_col=0)

seu = pd.Index(pbmc3k_tutorial_FindVariableGenes_seurat_batch["x"].values)
```

Matching genes `'seurat_v3'` and `FindVariableGenes`
```
len(seu.intersection(seurat_v3_hvg[seurat_v3_hvg.highly_variable].index))
```
`764`

Matching genes `'seurat_v3_paper'` and `FindVariableGenes`
```
len(seu.intersection(seurat_v3_paper_hvg[seurat_v3_paper_hvg.highly_variable].index))
```
`1990`

Matching genes `'seurat_v3_implementation'` and `FindVariableGenes`
```
len(seu.intersection(seurat_v3_implementation_hvg[seurat_v3_implementation_hvg.highly_variable].index))
```
`1990`


scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -455,12 +525,17 @@ def highly_variable_genes(
             "pass `inplace=False` if you want to return a `pd.DataFrame`."
         )
 
-    if flavor == "seurat_v3":
+    if (
+        flavor == "seurat_v3"
+        or flavor == "seurat_v3_paper"
+        or flavor == "seurat_v3_implementation"
+    ):
```
flying-sheep:
```suggestion
    if flavor in {"seurat_v3", "seurat_v3_paper", "seurat_v3_implementation"}:
```

Fun fact (not relevant to this usage): this is not only elegant, it’s also the fastest option, since Python has an optimizer which will extract the set literal to an anonymous module-level frozen set at import time.

flying-sheep (commented):
Thank you for the detective work!

I think before we commit, we should talk with the Seurat people and ask them if fixing their implementation so it matches the paper is in the cards.

scanpy/preprocessing/_highly_variable_genes.py
eroell: Nice forgot to revisit this before the commit thx!

ivirshup:
I think this looks pretty good.

One thing we had discussed was moving out the merge logic for multiple batches from being specified by `flavor` to being specified by a different argument, maybe `merge_flavor` or `batch_flavor`. Have you thought about/ looked into this?

flying-sheep:
@ivirshup last Tuesday, we agreed that we’ll proceed with this once we know if Seurat considers this a bug and will fix it, see https://github.com/scverse/scanpy/issues/2780#issuecomment-1892109193

I think if they change how Seurat works, there’s no need to add a bug-for-bug compatible version for their former implementation.

ivirshup:
I think this is orthogonal to that. The idea with having a separate argument for how we merge the results from different batches would mean factoring out the merge logic from each flavor and having it be a stand alone operation. It would also change the API here, since we wouldn't be adding a new `flavor`.

eroell:
> One thing we had discussed was moving out the merge logic for multiple batches from being specified by `flavor` to being specified by a different argument, maybe `merge_flavor` or `batch_flavor`. Have you thought about/ looked into this?

I see what you mean. Not yet looked into that here, indeed. Slight risk of increasing confusion potential for users?

flying-sheep:
I think the two best ways to go forward:

- Either the Seurat people agree their implementation’s order is a bug and we switch to the paper order. Then we don’t _necessarily_ need to add any way to configure it, just to follow suit. But we could add a way to configure it to support different possible orderings
- Or they decide that it isn’t, in which case we should add that way to configure things

I think it makes more sense to encode orthogonal choices into orthogonal options. If it makes sense to offer a common set of orderings for all flavors, it should definitely be a separate option.

E.g. if the following makes sense, then it should for sure be multiple options:

```python
for flavor, order in product(
    ('seurat_v3', 'seurat'),
    ('rank', 'batches'),
):
    hvg(…, flavor=flavor, order=order)
```

(`order`, `'rank'`, and `'batches'` are ad-hoc names, not necessarily good ones)

eroell:
What is there in the latest update:
- The simple adoption, at the heart of this PR, that `flavor=seurat_v3_paper` matches Seurat better when using `batch_key`.
- The `flavor=seurat_v3` remains untouched, hence not a breaking change.
- The doc is more detailed now

What is not there:
- Refactoring of single vs multi batch. Reason: While this effort will enhance code maintenance, it may quickly require almost the entire _highly_variable_genes.py to be touched. Suggest to do this thorough & separately?
- orthogonality of flavor and ordering. Reason: I think this is very hard to understand and match against other methods for users. 

    > If it makes sense to offer a common set of orderings for all flavors, it should definitely be a separate option.

    Does it make sense? There isn't benchmarking literature I know, and the flavors don't offer a decoupled ordering choice themselves. From user issues, I experience the consistency with other tools to be the primary concern.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -80,6 +81,7 @@ def _highly_variable_genes_seurat_v3(
         X_batch = X[batch_info == b]
 
         mean, var = _get_mean_var(X_batch)
+
```
flying-sheep:
```suggestion
```

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -133,24 +135,29 @@ def _highly_variable_genes_seurat_v3(
...
-        .sort_values(
-            ["highly_variable_rank", "highly_variable_nbatches"],
-            ascending=[True, False],
-            na_position="last",
-        )
+        df[sort_cols]
+        .sort_values(sort_cols, ascending=sort_ascending, na_position="last")
         .index
     )
+
```
flying-sheep:
```suggestion
```

flying-sheep:
OK! Please add a release note and we’re good to go I think.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -471,7 +471,7 @@ def highly_variable_genes(
             "pass `inplace=False` if you want to return a `pd.DataFrame`."
         )
 
-    if flavor in ["seurat_v3", "seurat_v3_paper"]:
+    if flavor in {"seurat_v3", "seurat_v3_paper"}:
```
flying-sheep: please always use sets for things like this

eroell:
> OK! Please add a release note and we’re good to go I think.

Added note, under features

flying-sheep:
Thank you for the work!

PR #2803, opened 2024-01-11
(feat): add `n_components` parameter to tSNE

### LINKED ISSUE #460: Additional parameters in some functions - new functions

ch1ru:
Added the ```n_components``` parameter in the tsne function, similar to the one in umap and updated docstring.

- [x] Closes #460
- [ ] Tests included or not required because: Not included but can add tests if necessary.

- [x] Release notes not necessary because: Only minor change.


flying-sheep:
Thanks!

Issue #888, opened 2019-10-24
n_bins not respected in highly_variable_genes(..., flavour='cell_ranger')


flying-sheep:
This code creates the range `[10, 15, …, 100]`, which is 19 values. Taking those percentiles together with -inf and +inf creates 20 bins (from 21 bin borders), understood.

https://github.com/theislab/scanpy/blob/e6e08e51d63c78581bb9c86fe6e302b80baef623/scanpy/preprocessing/_highly_variable_genes.py#L90

But 1. why start at 10% and 2. why is n_bins ignored and there are always 20 bins created?

@Koncopd @falexwolf you introduced this in #330, is this what cellranger does?

davidhbrann:
We also tried to address this in #572 and #624 but neither of those PRs have been merged since they're not fully backwards compatible.

flying-sheep:
Well, we *really* need to get that release out of the door.

But on the other hand, as @gokceneraslan said the current behavior is a bug, so we can change it now!

Gökçen, do you still plan on consolidating those PRs?

flying-sheep:
5 years later, this is still not fixed … where is the decision recorded that they shouldn’t be merged?

PR #2809, opened 2024-01-15
Support Dask in highly_variable_genes

### LINKED ISSUE #2777: Dask support for pp.highly_variable_genes (default "seurat" flavor)
One of the most frequently requested features for Dask support, and required for basic preprocessing recipes.
See also:
- #2807 
- #2808

### LINKED ISSUE #2807: Dask support for pp.highly_variable_genes ("cellranger" flavor)
After (or with) https://github.com/scverse/scanpy/issues/2777, we should also support the “cellranger” flavor
See also:
- #2808 

Blocked by
- https://github.com/dask/dask/issues/10853

flying-sheep:
- Depends on  #2814

- [x] Closes #2777, closes #2807

TODO:

- [x] batched

<table>
<thead>
<tr>
 <th scope=row>

`flavor=`

 <th>

`"seurat"`

 <th>

`"cell_ranger"`

<tbody>
<tr>
 <th scope=row>

`n_top_genes=n`

 <td>

- [x] &zwnj;

 <td>

- [x] (https://github.com/dask/dask/issues/10853)

<tr>
 <th scope=row>

`{min,max}_{disp,mean}`

 <td>

- [x] &zwnj;

 <td>

- [x] &zwnj;

</table>


scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -486,14 +487,16 @@ def highly_variable_genes(
...
         gene_list = adata.var_names
         for batch in batches:
             adata_subset = adata[adata.obs[batch_key] == batch]
 
             # Filter to genes that are in the dataset
             with settings.verbosity.override(Verbosity.error):
-                filt = filter_genes(adata_subset, min_cells=1, inplace=False)[0]
+                filt, _ = filter_genes(
+                    _get_obs_rep(adata_subset, layer=layer), min_cells=1, inplace=False
+                )
```
flying-sheep: this has been broken the whole time by operating on X instead of the selected layer?

flying-sheep:
Hmm, I found several weirdnesses in hvg:

1. as mentioned [above](#discussion_r1453628289), this line has always ignored the `layer` parameter (fixed in this PR): https://github.com/scverse/scanpy/blob/86b85ee2f4e8acfc9db3ce4cfff6e905d96a59eb/scanpy/preprocessing/_highly_variable_genes.py#L496
2. How does the binning in `flavour='cell_ranger'` work? (see #888, where I asked @Koncopd and @falexwolf how to address that)
3. This line has survived since the very first commit, but what does it do? it sorts the view in-place? https://github.com/scverse/scanpy/blob/86b85ee2f4e8acfc9db3ce4cfff6e905d96a59eb/scanpy/preprocessing/_highly_variable_genes.py#L276-L278

ivirshup (changes requested):
Initial comment before looking too deep into the code, I'm not seeing a test that checks that the result of the dask version is correct.

Could we add a test checking that the result is equivalent between dask and in-memory versions?

ivirshup:
Error when trying this where `X` is a dask array with sparse chunks:

```python
result = sc.pp.highly_variable_genes(adata, inplace=False)
```

<details>
<summary> Traceback </summary>

```pytb
File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/array/reductions.py:363, in partial_reduce(func, x, split_every, keepdims, dtype, name, reduced_meta)
    362 try:
--> 363     meta = func(reduced_meta, computing_meta=True)
    364 # no meta keyword argument exists for func, and it isn't required

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/array/reductions.py:685, in mean_combine(pairs, sum, numel, dtype, axis, computing_meta, **kwargs)
    684 ns = deepmap(lambda pair: pair["n"], pairs) if not computing_meta else pairs
--> 685 n = _concatenate2(ns, axes=axis).sum(axis=axis, **kwargs)
    687 if computing_meta:

TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'

During handling of the above exception, another exception occurred:

Cell In[19], line 1
----> 1 result = sc.pp.highly_variable_genes(adata, inplace=False)
      2 result

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/legacy_api_wrap/__init__.py:80, in legacy_api.<locals>.wrapper.<locals>.fn_compatible(*args_all, **kw)
     77 @wraps(fn)
     78 def fn_compatible(*args_all: P.args, **kw: P.kwargs) -> R:
     79     if len(args_all) <= n_positional:
---> 80         return fn(*args_all, **kw)
     82     args_pos: P.args
     83     args_pos, args_rest = args_all[:n_positional], args_all[n_positional:]

File /mnt/workspace/repos/scanpy/scanpy/preprocessing/_highly_variable_genes.py:702, in highly_variable_genes(***failed resolving arguments***)
    699 del min_disp, max_disp, min_mean, max_mean, n_top_genes
    701 if batch_key is None:
--> 702     df = _highly_variable_genes_single_batch(
    703         adata, layer=layer, cutoff=cutoff, n_bins=n_bins, flavor=flavor
    704     )
    705 else:
    706     df = _highly_variable_genes_batched(
    707         adata, batch_key, layer=layer, cutoff=cutoff, n_bins=n_bins, flavor=flavor
    708     )

File /mnt/workspace/repos/scanpy/scanpy/preprocessing/_highly_variable_genes.py:274, in _highly_variable_genes_single_batch(adata, layer, cutoff, n_bins, flavor)
    271     else:
    272         X = np.expm1(X)
--> 274 mean, var = _get_mean_var(X)
    275 # now actually compute the dispersion
    276 mean[mean == 0] = 1e-12  # set entries equal to zero to small value

File /mnt/workspace/repos/scanpy/scanpy/preprocessing/_utils.py:21, in _get_mean_var(X, axis)
     19     mean, var = sparse_mean_variance_axis(X, axis=axis)
     20 else:
---> 21     mean = X.mean(axis=axis, dtype=np.float64)
     22     mean_sq = elem_mul(X, X).mean(axis=axis, dtype=np.float64)
     23     var = mean_sq - mean**2

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/array/core.py:2605, in Array.mean(self, axis, dtype, keepdims, split_every, out)
   2595 """Returns the average of the array elements along given axis.
   2596 
   2597 Refer to :func:`dask.array.mean` for full documentation.
   (...)
   2601 dask.array.mean : equivalent function
   2602 """
   2603 from dask.array.reductions import mean
-> 2605 return mean(
   2606     self,
   2607     axis=axis,
   2608     dtype=dtype,
   2609     keepdims=keepdims,
   2610     split_every=split_every,
   2611     out=out,
   2612 )

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/array/reductions.py:719, in mean(a, axis, dtype, keepdims, split_every, out)
    717 else:
    718     dt = getattr(np.mean(np.zeros(shape=(1,), dtype=a.dtype)), "dtype", object)
--> 719 return reduction(
    720     a,
    721     mean_chunk,
    722     mean_agg,
    723     axis=axis,
    724     keepdims=keepdims,
    725     dtype=dt,
    726     split_every=split_every,
    727     combine=mean_combine,
    728     out=out,
    729     concatenate=False,
    730 )

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/array/reductions.py:235, in reduction(x, chunk, aggregate, axis, keepdims, dtype, split_every, combine, name, out, concatenate, output_size, meta, weights)
    232 else:
    233     reduced_meta = None
--> 235 result = _tree_reduce(
    236     tmp,
    237     aggregate,
    238     axis,
    239     keepdims,
    240     dtype,
    241     split_every,
    242     combine,
    243     name=name,
    244     concatenate=concatenate,
    245     reduced_meta=reduced_meta,
    246 )
    247 if keepdims and output_size != 1:
    248     result._chunks = tuple(
    249         (output_size,) if i in axis else c for i, c in enumerate(tmp.chunks)
    250     )

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/array/reductions.py:291, in _tree_reduce(x, aggregate, axis, keepdims, dtype, split_every, combine, name, concatenate, reduced_meta)
    289     func = compose(func, partial(_concatenate2, axes=sorted(axis)))
    290 for _ in range(depth - 1):
--> 291     x = partial_reduce(
    292         func,
    293         x,
    294         split_every,
    295         True,
    296         dtype=dtype,
    297         name=(name or funcname(combine or aggregate)) + "-partial",
    298         reduced_meta=reduced_meta,
    299     )
    300 func = partial(aggregate, axis=axis, keepdims=keepdims)
    301 if concatenate:

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/array/reductions.py:367, in partial_reduce(func, x, split_every, keepdims, dtype, name, reduced_meta)
    365 except TypeError:
    366     try:
--> 367         meta = func(reduced_meta)
    368     except ValueError as e:
    369         # min/max functions have no identity, don't apply function to meta
    370         if "zero-size array to reduction operation" in str(e):

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/array/reductions.py:684, in mean_combine(pairs, sum, numel, dtype, axis, computing_meta, **kwargs)
    681 if not isinstance(pairs, list):
    682     pairs = [pairs]
--> 684 ns = deepmap(lambda pair: pair["n"], pairs) if not computing_meta else pairs
    685 n = _concatenate2(ns, axes=axis).sum(axis=axis, **kwargs)
    687 if computing_meta:

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/utils.py:153, in deepmap(func, *seqs)
    142 """Apply function inside nested lists
    143 
    144 >>> inc = lambda x: x + 1
   (...)
    150 [[11, 22], [33, 44]]
    151 """
    152 if isinstance(seqs[0], (list, Iterator)):
--> 153     return [deepmap(func, *items) for items in zip(*seqs)]
    154 else:
    155     return func(*seqs)

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/utils.py:153, in <listcomp>(.0)
    142 """Apply function inside nested lists
    143 
    144 >>> inc = lambda x: x + 1
   (...)
    150 [[11, 22], [33, 44]]
    151 """
    152 if isinstance(seqs[0], (list, Iterator)):
--> 153     return [deepmap(func, *items) for items in zip(*seqs)]
    154 else:
    155     return func(*seqs)

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/utils.py:155, in deepmap(func, *seqs)
    153     return [deepmap(func, *items) for items in zip(*seqs)]
    154 else:
--> 155     return func(*seqs)

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/dask/array/reductions.py:684, in mean_combine.<locals>.<lambda>(pair)
    681 if not isinstance(pairs, list):
    682     pairs = [pairs]
--> 684 ns = deepmap(lambda pair: pair["n"], pairs) if not computing_meta else pairs
    685 n = _concatenate2(ns, axes=axis).sum(axis=axis, **kwargs)
    687 if computing_meta:

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/scipy/sparse/_index.py:44, in IndexMixin.__getitem__(self, key)
     43 def __getitem__(self, key):
---> 44     row, col = self._validate_indices(key)
     46     # Dispatch to specialized methods.
     47     if isinstance(row, INT_TYPES):

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/scipy/sparse/_index.py:156, in IndexMixin._validate_indices(self, key)
    154         row += M
    155 elif not isinstance(row, slice):
--> 156     row = self._asindices(row, M)
    158 if isintlike(col):
    159     col = int(col)

File /mnt/workspace/mambaforge/envs/scanpy-dev/lib/python3.11/site-packages/scipy/sparse/_index.py:180, in IndexMixin._asindices(self, idx, length)
    177     raise IndexError('invalid index') from e
    179 if x.ndim not in (1, 2):
--> 180     raise IndexError('Index dimension must be 1 or 2')
    182 if x.size == 0:
    183     return x

IndexError: Index dimension must be 1 or 2
```

</details>

flying-sheep:
Are we committing to support sparse-in-dask?

I’m defaulting to `ARRAY_TYPES_SUPPORTED`, which marks sparse as xfail. That’s how we treat other dask-capable utils so far. You can see the errors with

```console
$ pytest -vv --runxfail scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_{subset_inplace_consistency,no_inplace} -k sparse
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_subset_inplace_consistency[copy-full-dask_array_sparse-cell_ranger] - TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_subset_inplace_consistency[copy-subset-dask_array_sparse-cell_ranger] - TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_subset_inplace_consistency[inplace-subset-dask_array_sparse-cell_ranger] - TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_subset_inplace_consistency[inplace-full-dask_array_sparse-cell_ranger] - TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_subset_inplace_consistency[copy-full-dask_array_sparse-seurat] - TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_subset_inplace_consistency[inplace-subset-dask_array_sparse-seurat] - TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_subset_inplace_consistency[copy-subset-dask_array_sparse-seurat] - TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_subset_inplace_consistency[inplace-full-dask_array_sparse-seurat] - TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_no_inplace[dask_array_sparse-batched] - TypeError: _cs_matrix.sum() got an unexpected keyword argument 'keepdims'
FAILED scanpy/tests/test_highly_variable_genes.py::test_highly_variable_genes_no_inplace[dask_array_sparse-single] - IndexError: Index dimension must be 1 or 2
```

flying-sheep:
OK, so neither `_get_mean_var` nor `filter_genes` support sparse-in-dask.

Therefore I’d consider adding support for that out-of-scope for this PR

flying-sheep:
OK, this should be mostly it. Maybe some cleanup, but no major changes.

Test failures are all the server for `ebi_expression_atlas` breaking.

scanpy/preprocessing/_distributed.py
```diff
@@ -47,3 +84,82 @@ def materialize_as_ndarray(
...
+
+@contextmanager
+def suppress_pandas_warning() -> Generator[None, None, None]:
+    import warnings
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore", r"The default of observed=False", category=FutureWarning
+        )
+        yield
```
ivirshup: Is this necessary?

I think this is pretty similar in functionality to `dask.compute` and new versions of `dask.dataframe` don't seem to throw this warning. I don't get this warning when running this locally after commenting out the suppression.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -201,108 +260,275 @@ def _highly_variable_genes_single_batch(
...
-    df["means"] = mean
-    df["dispersions"] = dispersion
+    df: pd.DataFrame | DaskDataFrame
+    if isinstance(data, DaskArray):
+        import dask.array as da
+        import dask.dataframe as dd
+
+        df = dd.from_dask_array(
+            da.vstack((mean, dispersion)).T, columns=["means", "dispersions"]
+        )
```
```suggestion
    if isinstance(data, DaskArray):
        import dask

        df = pd.DataFrame(dict(zip(["means", "dispersions"], dask.compute(mean, dispersion))))
```

Re-my previous thought "I think this will be faster if we use pandas instead of dask.dataframe", with this change I see about a 4x speed up. Note that presumably the most computationally expensive part of this whole thing is the mean and dispersion calculation, which happens the same in both cases.

Here is the script I used to test: https://gist.github.com/ivirshup/ecf131b6ba8b094c04bd4a275ad7a804

And here is the branch with my code: https://github.com/scverse/scanpy/tree/dask-hvg-w-pandas

I haven't dug too deeply in yet so:

* I don't know the effect of using batches
* I've only looked at Seurat
* I haven't looked at the effect of using different chunk sizes

One red flag I noticed while playing around with this was that the dask dataframe returned with `inplace=False` had ~1000 high level graph nodes when I used a chunk size of 1000 cells. That seems like it could be incurring a lot of overhead. In addition, the majority of time to calculate actually came before this dask dataframe was returned, due to an intermediate computation.

Surprisingly, I was seeing increasing times to compute as I increased chunk size – but haven't run this enough to be that confident in this result.

ivirshup (changes requested):

scanpy/preprocessing/_distributed.py
flying-sheep: If they fixed it, I can probably remove this, since it isn’t used in too many spots anymore

flying-sheep:
With that change, we also support cell_ranger in this PR

flying-sheep:
OK, I added a branch for the Dask DF version: https://github.com/scverse/scanpy/tree/hvg-ddf

Regarding delayed, are you’re thinking this or is there a fancier way to do it?

```py
if is_dask:
	dfs = [dask.delayed(single_batch)(batch) for batch in batches]
	df = dask.delayed(pd.concat)(dfs).compute()
else:
	df = pd.concat(single_batch(batch) for batch in batches)
```

/edit: let’s postpone it for a different PR

ivirshup (commented):
> This line has survived since the very first commit, but what does it do? it sorts the view in-place?

Missed this question. Yes, it sorts inplace in reverse order.

---

Big question, after removing the dask dataframe stuff there is still a large amount of changes to the dataframe code here. Could you give and overview of what changed and why? It'll help direct what I'm looking for in review.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -181,15 +188,54 @@ def _highly_variable_genes_seurat_v3(
...
+            if p.name in cutoffs
+        }
+        if {k: v for k, v in locals().items() if k in cutoffs} != defaults:
+            logg.info("If you pass `n_top_genes`, all cutoffs are ignored.")
+        return n_top_genes
+
+    def in_bounds(
+        self,
+        mean: NDArray[np.float64] | DaskArray,
+        dispersion_norm: NDArray[np.float64] | DaskArray,
```
```suggestion
        mean: NDArray[np.floating] | DaskArray,
        dispersion_norm: NDArray[np.floating] | DaskArray,
```

This would work on a broader range of types, no?

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -201,108 +247,221 @@ def _highly_variable_genes_single_batch(
     A DataFrame that contains the columns
     `highly_variable`, `means`, `dispersions`, and `dispersions_norm`.
     """
-    X = adata.layers[layer] if layer is not None else adata.X
+    data = _get_obs_rep(adata, layer=layer)
```
ivirshup: Was there a technical reason to rename the variable? It seems like it's causing a lot of churn, but unclear if there is added value

scanpy/tests/test_highly_variable_genes.py
```diff
@@ -556,12 +566,23 @@ def test_cellranger_n_top_genes_warning():
         sc.pp.highly_variable_genes(adata, n_top_genes=1000, flavor="cell_ranger")
 
 
+mark_no_cell_ranger = pytest.mark.xfail(
+    reason="See https://github.com/dask/dask/issues/10853"
+)
```
ivirshup:
```suggestion
```

I believe this is unused

flying-sheep:
> there is still a large amount of changes to the dataframe code here

Not really changes: it’s almost all refactoring, because the code was spaghetti with quite some duplication.

I’m doing nothing more than

1. I introduce helper functions so code gets more readable, e.g. a clean `disp_cut_off = _nth_highest(dispersion_norm, n_top_genes)` instead of a large inline code block that has to be decyphered line by line to figure out that it finds the nth highest value.

    This is especially necessary for the huge main pile of spaghetti that used to be the `if flavor == "seurat":`/`elif flavor == "cell_ranger":` branches. I simply put their contents into a `_get_mean_bins` helper and two helpers `_stats_seurat` and `_stats_cell_ranger` (while deduplicating shared code)
2. Making sure pandas indices match up while removing `.to_numpy()`. That way instead of having `.to_numpy()` potentially copy and and convert data in extension arrays, the series are just used directly. Not to mention that three `.to_numpy()` per line make things hard to read.
3. refactor the 5 cutoff parameters into one value `cutoff` for clarity, less inline code, and better type information (we either have either `n_top_genes: int` or a `_Cutoffs` instance, never both. This way, the type system knows)

and that’s it. <ins>potentially</ins> faster, much more maintainable, and almost dask-compatible.

After my changes, it should be easier to further refactor things so the seurat_v3 flavor is integrated into the overall structure instead of doing its own thing.

ivirshup:
> Faster

Is it faster? Could you demonstrate this?

flying-sheep:
Probably not, since we probably only have numpy dtypes here. But that’s irrelevant, since I described more than one reason to do it.

It would be faster if we used that code on some extension dtype `pd.Series`, so it’s generally good practice to not use `.to_numpy()` by default.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -201,108 +247,221 @@ def _highly_variable_genes_single_batch(
...
-            n_top_genes = dispersion_norm.size
-        disp_cut_off = dispersion_norm[n_top_genes - 1]
-        gene_subset = np.nan_to_num(df["dispersions_norm"].to_numpy()) >= disp_cut_off
+
+    return pd.cut(means, bins=bins)
+
+
+def _stats_seurat(mean_bins: pd.Series, disp_grouped: SeriesGroupBy) -> pd.DataFrame:
+    """Compute mean and std dev per bin."""
+    disp_bin_stats = disp_grouped.agg(avg="mean", dev=partial(np.std, ddof=1))
```
```suggestion
    disp_bin_stats = disp_grouped.agg(avg="mean", dev="std")
```

This should be equivalent and use the pandas method, which could be optimized by pandas. Maybe faster (but I'm not 100% convinced) in a quick check:

```python
In [16]: %timeit res1 = df.groupby("cat").agg(["mean", "std"])
1.48 ms ± 46.6 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [14]: %timeit res2 = df.groupby("cat").agg(["mean", partial(np.std, ddof=1)])
1.77 ms ± 56.9 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -201,108 +247,221 @@ def _highly_variable_genes_single_batch(
...
-            f"normalized dispersion cutoff of {disp_cut_off}"
+            f"Gene indices {gen_indices} fell into a single bin: their "
+            "normalized dispersion was set to 1.\n    "
+            "Decreasing `n_bins` will likely avoid this effect."
         )
-    else:
+        disp_bin_stats["dev"].loc[one_gene_per_bin] = disp_bin_stats["avg"].loc[
+            one_gene_per_bin
+        ]
+        disp_bin_stats["avg"].loc[one_gene_per_bin] = 0
```
ivirshup: This throws a warning:

```
[/mnt/workspace/repos/scanpy/scanpy/preprocessing/_highly_variable_genes.py:324](https://vscode-remote+ssh-002dremote-002b172-002e17-002e6-002e114.vscode-resource.vscode-cdn.net/mnt/workspace/repos/scanpy/scanpy/preprocessing/_highly_variable_genes.py:324): FutureWarning: ChainedAssignmentError: behaviour will change in pandas 3.0!
You are setting values through chained assignment. Currently this works in certain cases, but when using Copy-on-Write (which will become the default behaviour in pandas 3.0) this will never work to update the original DataFrame or Series, because the intermediate object on which we are setting values will behave as a copy.
A typical example is when you are setting values in a column of a DataFrame, like:

df["col"][row_indexer] = value

Use `df.loc[row_indexer, "col"] = values` instead, to perform the assignment in a single step and ensure this keeps updating the original `df`.

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
```


```suggestion
        disp_bin_stats.loc["dev", one_gene_per_bin] = disp_bin_stats["avg"].loc[
            one_gene_per_bin
        ]
        disp_bin_stats.loc["avg", one_gene_per_bin] = 0
```


But also:

```python
        disp_bin_stats.loc["dev", one_gene_per_bin] = disp_bin_stats["avg"].loc[
            one_gene_per_bin
        ]
```

Is this code right? Setting the deviances to be the values of the means?

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -201,108 +247,221 @@ def _highly_variable_genes_single_batch(
...
+        # MAD calculation raises the warning: "Mean of empty slice"
+        warnings.simplefilter("ignore", category=RuntimeWarning)
+        disp_bin_stats = disp_grouped.agg(avg="median", dev=mad)
+    return _unbin(disp_bin_stats, mean_bins)
+
+
+def _unbin(df: pd.DataFrame, mean_bins: pd.Series) -> pd.DataFrame:
+    df = df.loc[mean_bins]
+    df["gene"] = mean_bins.index
+    df.set_index("gene", inplace=True)
```
```suggestion
    df.index = mean_bins.index
```

Is this equivalent? Especially after the previous line?

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -201,108 +247,221 @@ def _highly_variable_genes_single_batch(
...
+    df["dispersions_norm"] = (df["dispersions"] - disp_stats["avg"]) / disp_stats["dev"]
+    df["highly_variable"] = _subset_genes(
+        adata,
+        mean=mean,
+        dispersion_norm=df["dispersions_norm"].to_numpy(),
+        cutoff=cutoff,
+    )
+
+    df.set_index(adata.var_names, inplace=True)
+    df.index.name = "gene"
```
ivirshup: I don't think we should be setting this

```suggestion
```

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -201,108 +247,221 @@ def _highly_variable_genes_single_batch(
...
+
     # all of the following quantities are "per-gene" here
-    df = pd.DataFrame()
-    df["means"] = mean
-    df["dispersions"] = dispersion
+    df = pd.DataFrame(
+        dict(zip(["means", "dispersions"], materialize_as_ndarray((mean, dispersion))))
+    )
+    df["mean_bin"] = _get_mean_bins(df["means"], flavor, n_bins)
+    disp_grouped = df.groupby("mean_bin", observed=True)["dispersions"]
```
ivirshup: I think `df` should be passed into the `stats` functions and `disp_grouped` should be created within those methods.

It's a line of duplication, but I think it makes the code much easier to read, since all that code relies on a very particular relationship between these two values. Also the `disp_grouped` is directly derived from `df` so we can easily get to fewer arguments. 

Alternatively, you could write docstrings for the internal functions that explain the assumptions about how the arguments are related. But I think it's easier/ clearer/ shorter to just pass the dataframe.

scanpy/preprocessing/_highly_variable_genes.py
ivirshup: Is there a reason you are using `avg` and `dev` instead of the more standard `mean`, `std`?

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -201,108 +247,221 @@ def _highly_variable_genes_single_batch(
...
+    mean_bins: pd.Series, disp_grouped: SeriesGroupBy
+) -> pd.DataFrame:
+    """Compute median and median absolute dev per bin."""
+    from statsmodels.robust import mad
+
+    with warnings.catch_warnings():
+        # MAD calculation raises the warning: "Mean of empty slice"
+        warnings.simplefilter("ignore", category=RuntimeWarning)
+        disp_bin_stats = disp_grouped.agg(avg="median", dev=mad)
+    return _unbin(disp_bin_stats, mean_bins)
```
```suggestion
    return disp_bin_stats.loc[mean_bins].set_index(mean_bins.index)
```

This could be an easier to follow way to do this. It could also be easier to follow if argument naming more consistently followed the variable naming convention.

docs/release-notes/1.10.0.md
```diff
@@ -14,6 +14,7 @@
 * {func}`scanpy.pp.pca`, {func}`scanpy.pp.scale`, {func}`scanpy.pl.embedding`, and {func}`scanpy.experimental.pp.normalize_pearson_residuals_pca`
   now support a `mask` parameter {pr}`2272` {smaller}`C Bright, T Marcella, & P Angerer`
 * {func}`scanpy.tl.rank_genes_groups` no longer warns that it's default was changed from t-test_overestim_var to t-test {pr}`2798` {smaller}`L Heumos`
+* {func}`scanpy.pp.highly_variable_genes` supports dask for the default `seurat` flavor and partially for the `cell_ranger` flavor {pr}`2809` {smaller}`P Angerer`
```
ivirshup: Is it still true that it's only partially for the `"cell_ranger"` flavor?

ivirshup (changes requested):
Alright, taken a look.

I wasn't really expecting this feature PR to also include such a large refactor, and I'm still not 100% convinced the behaviour here is exactly the same as before. I have done a few tests, which have been okay, but I haven't tried much parameterization. I'm ~80% convinced the results should be the same.

I would note that the dataframe returned when `inplace=False` has a different index than it did previously.

Apart from the comments, can we get a regression test for `"cell_ranger"` (e.g. generate results with an older version)? I don't think we have one in the test suite.

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: true. In practice, there are sometimes problems with numpy’s overloading when one uses `floating` instead of concrete types, but here it should be fine.

flying-sheep:
> I wasn't really expecting this feature PR to also include such a large refactor

It would have been necessary for the Dask Dataframe version. Now I 1. did the work and 2. improved readability, so it would be counter productive to undo it.

> I'm still not 100% convinced the behaviour here is exactly the same as before. I have done a few tests, which have been okay, but I haven't tried much parameterization. I'm ~80% convinced the results should be the same.

If you have any specific things in mind, you should probably make a PR that adds tests for the properties you think we should preserve. We can then merge that one, update this one, and see if it actually breaks something.

I can’t check for speculative differences if I have no idea where those could be.

> I would note that the dataframe returned when inplace=False has a different index than it did previously.

Yup, now it actually matches instead of discarding the original Index and replacing it with a RangeIndex for no reason.

> Apart from the comments, can we get a regression test for "cell_ranger" (e.g. generate results with an older version)? I don't think we have one in the test suite.

Sure! That’s a concrete thing I can do. I’ll do that on thursday, I did the rest of what you asked today

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: This abstracts over the flavors. they have each a metric for an average and a deviation.

So I made both helpers return a compatible data frame

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: It doesn’t, the changes are entirely localized to the next block. The reason is that it’s not X, it’s `X` or a layer.

scanpy/tests/test_highly_variable_genes.py
flying-sheep: good catch

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: I didn‘t realize that pandas has a different default. Sure!

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: No idea if it’s right, but it is identical to before the refactor.

I did the COW thing, good call!

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: yeah, this is a remnant of the things I had to do to make the recalcitrant Dask DataFrames behave

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: sure, we can do that in `drop_index` instead

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: Or even better, just get rid of all the duplication by using a `_get_disp_stats` function that only receives the data frame

scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: done, in the aforementioned `_get_disp_stats` helper

docs/release-notes/1.10.0.md
flying-sheep: good call, full support!

scanpy/preprocessing/_highly_variable_genes.py
ivirshup: I don't know that I find it confusing that `X` is a variable that could refer to `adata.X` or something in `adata.layers`. I do think that `X` is a nicer variable name than `data` since it carries more assumptions around it (like the shape) which hold for this function, and fits with the scikit-learn convention. `data` is very generic.

scanpy/preprocessing/_highly_variable_genes.py
ivirshup: That is much better!

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -307,45 +298,46 @@ def _get_mean_bins(
...
+        disp_bin_stats = disp_grouped.agg(avg="mean", dev="std")
+        _postprocess_seurat(disp_bin_stats, df["mean_bin"])
+    elif flavor == "cell_ranger":
+        disp_bin_stats = disp_grouped.agg(avg="median", dev=_mad)
+    else:
+        raise ValueError('`flavor` needs to be "seurat" or "cell_ranger"')
+    return disp_bin_stats.loc[df["mean_bin"]].set_index(df.index)
+
+
+def _postprocess_seurat(disp_bin_stats: pd.DataFrame, mean_bin: pd.Series) -> None:
```
```suggestion
        _postprocess_dispersions_seurat(disp_bin_stats, df["mean_bin"])
    elif flavor == "cell_ranger":
        disp_bin_stats = disp_grouped.agg(avg="median", dev=_mad)
    else:
        raise ValueError('`flavor` needs to be "seurat" or "cell_ranger"')
    return disp_bin_stats.loc[df["mean_bin"]].set_index(df.index)


def _postprocess_dispersions_seurat(disp_bin_stats: pd.DataFrame, mean_bin: pd.Series) -> None:
```

Suggestion for a more descriptive name

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -201,108 +245,214 @@ def _highly_variable_genes_single_batch(
...
+    # actually do the normalization
+    df["dispersions_norm"] = (df["dispersions"] - disp_stats["avg"]) / disp_stats["dev"]
+    df["highly_variable"] = _subset_genes(
+        adata,
+        mean=mean,
+        dispersion_norm=df["dispersions_norm"].to_numpy(),
+        cutoff=cutoff,
+    )
+
+    df.set_index(adata.var_names, inplace=True)
```
ivirshup: Any reason why not:

```suggestion
    df.index = adata.var_names
```

Also I believe `set_index(..., inplace=True)` is set to be deprecated:

* https://github.com/pandas-dev/pandas/issues/56507

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -181,15 +186,54 @@ def _highly_variable_genes_seurat_v3(
...
+            return cls(min_disp, max_disp, min_mean, max_mean)
+
+        cutoffs = {"min_disp", "max_disp", "min_mean", "max_mean"}
+        defaults = {
+            p.name: p.default
+            for p in signature(highly_variable_genes).parameters.values()
+            if p.name in cutoffs
+        }
+        if {k: v for k, v in locals().items() if k in cutoffs} != defaults:
+            logg.info("If you pass `n_top_genes`, all cutoffs are ignored.")
```
ivirshup: This line has no coverage, but probably should (as we should be checking this behaviour).

I would also be up for changing this to being a real warning, since we should probably be louder that we are ignoring arguments the user passed.

ivirshup (changes requested):
> > Apart from the comments, can we get a regression test for "cell_ranger" (e.g. generate results with an older version)? I don't think we have one in the test suite.
>
> Sure! That’s a concrete thing I can do. I’ll do that on thursday, I did the rest of what you asked today

I don't think this has happened yet, could we add this?

----

There are two more lines which aren't covered, but I believe they should be unreachable (both just ValueError that the arg should be "cell_ranger" or "seurat") so it's fine.

----

I'm a little concerned about changing the return for `inplace=False`, in case anyone was relying on that. Not too concerned, but what do you think about tracking that for 2.0?

ivirshup:
Sorry for the accidental push, trying to see if we can get inline code coverage on github:

* https://github.com/scverse/scanpy/pull/2847

flying-sheep:
> can we get a regression test for "cell_ranger" (e.g. generate results with an older version)? I don't think we have one in the test suite.

I’ll do that today if you can give me more info. Usually “regression test” refers to testing specific properties that were broken in a bug and subsequently fixed. What properties exactly are you looking for? Why `cell_ranger` and not `seurat`? Are you implying that we do that already for seurat?

/edit: done in https://github.com/scverse/scanpy/pull/2851

> There are two more lines which aren't covered, but I believe they should be unreachable (both just ValueError that the arg should be "cell_ranger" or "seurat") so it's fine.

yeah, lines like that are more defensive coding. I add them even to internal code to force us to look at everything instead of having a `else:  # flavor == "cell_ranger"` branch or so

> I'm a little concerned about changing the return for inplace=False, in case anyone was relying on that.

You mean the fact that the index makes the dataframe now actually useful? I can’t think of a way in which this breaks things in a way that isn’t immediately obvious and welcome. Of course, code can be infinitely weird, but can you think of a scenario?

scanpy/tests/conftest.py
```diff
@@ -39,6 +39,15 @@ def close_logs_on_teardown(request):
     request.addfinalizer(clear_loggers)
 
 
+@pytest.fixture(autouse=True)
+def _caplog_adapter(caplog: pytest.LogCaptureFixture):
+    import scanpy as sc
+
+    sc.settings._root_logger.addHandler(caplog.handler)
+    yield
+    sc.settings._root_logger.removeHandler(caplog.handler)
```
flying-sheep: This was put in for an older request for the `sc.logging.info()` call that was replaced with a `warnings.warn`. It allows us to actually use `caplog` for scanpy logs.

I can extract it if you want. Don’t let it distract you from approving :wink:

scanpy/preprocessing/_highly_variable_genes.py
ivirshup: I bet we didn't notice since the nonzero values are likely pretty similar, if not the same, for common use cases

ivirshup:
> You mean the fact that the index makes the dataframe now actually useful? I can’t think of a way in which this breaks things in a way that isn’t immediately obvious and welcome. Of course, code can be infinitely weird, but can you think of a scenario?

I get that it is more useful, but any code that was accessing it with `.loc` especially if it was relying on unique indices could run into a problem. It's minor, I'd be fine to leave it. Just we may need to revert.

ivirshup (approved):

flying-sheep:
We have documented the columns returned by the data frame, not the index, so I’d say the documented behavior stayed the same

PR #2815, opened 2024-01-16
(feat): `igraph` leiden implementation now included as an option in `sc.tl.leiden`

### LINKED ISSUE #1053: Leiden now included in python-igraph
The Leiden algorithm is now [included](https://igraph.org/python/doc/igraph.Graph-class.html#community_leiden) in the latest release of `python-igraph`, version 0.8.0. I believe this alleviates the need to depend on the `leidenalg` packages. The Leiden algorithm provided in `python-igraph` is substantially faster than the `leidenalg` package. It is simpler though, providing fewer options, but I believe the more extensive options of the `leidenalg` package are not necessarily needed for the purposes of `scanpy`. We  provide binary wheels on PyPI and binaries for conda are available from the conda-forge channel, also for Windows.

ilan-gold:
- [x] Closes #1053


ilan-gold:
TODOs:

1. Figure out why some tests are passing when they shouldn't (hence why I pushed the branch, curious about CI).  UPDATE: `tol` for `matplotlib.testing.compare.compare_images` is too high for a sparse-ish plot like `rank_genes_groups`.  This is somewhat worrying so will need to be amended.  Other than that, changed plotting outputs make sense so this should be resolved.
2. Check with scanpy tutorials to see what needs to be changed there as well, if anything (if needed, the two PRs should be merged in tandem).  The following use leiden in some capacity:
    a. https://scanpy-tutorials.readthedocs.io/en/latest/pbmc3k.html
    b. https://scanpy-tutorials.readthedocs.io/en/latest/plotting/core.html
    c. https://scanpy-tutorials.readthedocs.io/en/latest/spatial/basic-analysis.html
    d. https://scanpy-tutorials.readthedocs.io/en/latest/spatial/integration-scanorama.html
3. Do a large dataset test - check NMI for accuracy of the new default against the old one, check speed to confirm what we're doing makes sense (although this was covered, it seems, in #1053), and scalability

ilan-gold:
Failing `violin_2` does not fail locally, so that's bizarre.

scanpy/tests/notebooks/test_pbmc3k.py
```diff
@@ -115,32 +116,32 @@ def test_pbmc3k(image_comparer):
...
     # save_and_compare_images('rank_genes_groups_4')
 
-    if adata[adata.obs["leiden"] == "4", "CST3"].X.mean() < 1:
+    if adata[adata.obs["leiden"] == "3", "CST3"].X.mean() < 1:
         (  # switch clusters
+            adata.obs["leiden"][adata.obs["leiden"] == "3"],
             adata.obs["leiden"][adata.obs["leiden"] == "4"],
-            adata.obs["leiden"][adata.obs["leiden"] == "5"],
-        ) = ("5", "4")
+        ) = ("4", "3")
```
ilan-gold: The 4-5 swap is now a 3-4 swap, should it happen

ilan-gold: image

image

scanpy/tools/_leiden.py
```diff
@@ -143,13 +147,19 @@ def leiden(
     # as this allows for the accounting of a None resolution
     # (in the case of a partition variant that doesn't take it on input)
     if use_weights:
-        partition_kwargs["weights"] = np.array(g.es["weight"]).astype(np.float64)
-    partition_kwargs["n_iterations"] = n_iterations
-    partition_kwargs["seed"] = random_state
+        clustering_args["weights"] = "weight" if use_igraph else np.array(g.es["weight"]).astype(np.float64)
+    clustering_args["n_iterations"] = n_iterations
+    if not use_igraph:
+        clustering_args["seed"] = random_state
```
ilan-gold, line 153: I think both of these should be added to `uns` as outputs, `weights` in the `params` part of the dict and `use_igraph` at the top levels

```python
adata['leiden']
# { 'params': { ... 'use_weights': True }, 'use_igraph': False }

File: scanpy/tests/notebooks/test_pbmc3k.py [hunk 3]
```diff
@@ -29,6 +29,7 @@
 @needs.leidenalg
 def test_pbmc3k(image_comparer):
     save_and_compare_images = partial(image_comparer, ROOT, tol=20)
+    save_and_compare_images_rank_genes = partial(image_comparer, ROOT, tol=10) # 20 is too high for such sparse plots
```
ilan-gold, line 32: Do we think this could be a bug elsewhere, tolerance too low for picking up errors?  I can certainly open an issue for it.

scanpy/tests/notebooks/test_pbmc3k.py
```diff
@@ -29,6 +29,9 @@
 @needs.leidenalg
 def test_pbmc3k(image_comparer):
     save_and_compare_images = partial(image_comparer, ROOT, tol=20)
+    save_and_compare_images_rank_genes = partial(
+        image_comparer, ROOT, tol=10
+    )  # 20 is too high for such sparse plots
```
ilan-gold: Do we think this could be a bug elsewhere, tolerance too low for picking up errors?  I can certainly open an issue for it.

scanpy/tools/_leiden.py
```diff
@@ -134,22 +137,38 @@ def leiden(
...
     # appending to whatever the user provided. It needs to be this way
     # as this allows for the accounting of a None resolution
     # (in the case of a partition variant that doesn't take it on input)
     if use_weights:
-        partition_kwargs["weights"] = np.array(g.es["weight"]).astype(np.float64)
-    partition_kwargs["n_iterations"] = n_iterations
-    partition_kwargs["seed"] = random_state
+        clustering_args["weights"] = (
+            "weight" if use_igraph else np.array(g.es["weight"]).astype(np.float64)
+        )
```
ilan-gold: I think both of these should be added to `uns` as outputs, `weights` in the `params` part of the dict and `use_igraph` at the top levels

```python
adata['leiden']
# { 'params': { ... 'use_weights': True }, 'use_igraph': False }

**Comment by ilan-gold** [comment]
Failing test looks similar to what happens when I run out of memory locally: https://dev.azure.com/scverse/scanpy/_build/results?buildId=5329&view=logs&j=cb4d9293-b492-5d67-02b0-e6a595893958&t=a42f7c48-3122-5ff2-641c-e8a7971de511&l=86

and now this again only on the CI: https://github.com/scverse/scanpy/pull/2815#issuecomment-1894530944

**Comment by ilan-gold** [comment]
So it seems the pbmc3k leiden results are _different_ on azure for some reason...first actual, then expected
actual
vs. 
expected

**Comment by flying-sheep** [comment]
> Failing test looks similar to what happens when I run out of memory locally

I’ve mostly seen these “illegal instruction” errors in a case where something is run on the wrong CPU architecture (e.g. compiled for a newer architecture than supported on that specific runner)

File: scanpy/tests/notebooks/test_pbmc3k.py [hunk 6]
```diff
@@ -28,11 +28,11 @@
...
 def test_pbmc3k(image_comparer):
+    np.random.seed(
+        0
+    )  # ensure violin plots and other non-determinstic plots have deterministic behavior
     save_and_compare_images = partial(image_comparer, ROOT, tol=20)
-
-    adata = sc.read(
-        "./data/pbmc3k_raw.h5ad", backup_url="https://falexwolf.de/data/pbmc3k_raw.h5ad"
-    )
+    adata = sc.datasets.pbmc3k()
```
ilan-gold, line 34: No reason not to do this

scanpy/tools/_leiden.py
```diff
@@ -134,22 +144,39 @@ def leiden(
...
+        igraph.set_random_number_generator(random)
     if resolution is not None:
-        partition_kwargs["resolution_parameter"] = resolution
+        clustering_args[
+            f"resolution{'_parameter' if not use_igraph else ''}"
+        ] = resolution
     # clustering proper
-    part = leidenalg.find_partition(g, partition_type, **partition_kwargs)
+    if use_igraph:
+        part = g.community_leiden(objective_function="modularity", **clustering_args)
```
ilan-gold: `objective_function` as argument to the scanpy `leiden` function explicitly?

scanpy/tests/notebooks/test_pbmc3k.py
```diff
@@ -132,23 +160,22 @@ def test_pbmc3k(image_comparer):
     if adata[adata.obs["leiden"] == "4", "CST3"].X.mean() < 1:
         (  # switch clusters
             adata.obs["leiden"][adata.obs["leiden"] == "4"],
-            adata.obs["leiden"][adata.obs["leiden"] == "5"],
-        ) = ("5", "4")
+            adata.obs["leiden"][adata.obs["leiden"] == "3"],
+        ) = ("3", "4")
```
ilan-gold: 3 (new) <-> 5 (old)
Screenshot 2024-01-24 at 12 39 49
Screenshot 2024-01-24 at 12 39 03

flying-sheep: if you manually order them in the first place, why reorder here?

scanpy/tools/_leiden.py
```diff
@@ -96,9 +98,11 @@ def leiden(
         `obsp` and `neighbors_key` at the same time.
     copy
         Whether to copy `adata` or modify it inplace.
-    **partition_kwargs
-        Any further arguments to pass to `~leidenalg.find_partition`
-        (which in turn passes arguments to the `partition_type`).
+    use_leidenalg
+        Whether or not to use the `leidenalg` leiden implementation.
+    **clustering_args
+        Any further arguments to pass to `~leidenalg.find_partition` (which in turn passes arguments to the `partition_type`)
```
flying-sheep: Please fix this link (``:func:`leidenalg.find_partition` ``) and add a link to ``:meth:`igraph.Graph.community_leiden` ``.

scanpy/tools/_leiden.py
```diff
@@ -134,22 +146,41 @@ def leiden(
...
+    if use_leidenalg:
+        clustering_args["seed"] = random_state
+    else:
+        random.seed(random_state)
+        igraph.set_random_number_generator(random)
     if resolution is not None:
-        partition_kwargs["resolution_parameter"] = resolution
+        clustering_args[
+            f"resolution{'_parameter' if use_leidenalg else ''}"
+        ] = resolution
```
flying-sheep: [The igraph docs also say `resolution_parameter`](https://igraph.org/python/doc/api/igraph.Graph.html#community_leiden)

```suggestion
        clustering_args["resolution_parameter"] = resolution
```

scanpy/tools/_leiden.py
flying-sheep: or `clustering_args.get("objective_function", "modularity")`.

The way it is right now, it’ll throw an error when people try to use it. (“duplicate kwargs specified” or so)

flying-sheep (changes requested):
Looks pretty straightforward, nice!

Maybe `use_leidenalg` should be replaced with a parameter taking `Literal['igraph', 'leidenalg']`.

IDK how to name it. `method` is usually for the algorithm, I think (which in this case is `leiden`). There should be something about naming in the scanpy 2.0 board: https://github.com/orgs/scverse/projects/18/views/1

Let’s talk in the standup about this.

ilan-gold:
@ivirshup We have now way of distinguishing between `sc.leiden(backend="igraph", partition_type=leidenalg.RBConfigurationVertexPartition)` and `sc.leiden(partition_type=leidenalg.RBConfigurationVertexPartition)` and I think we would only want the override in the second case.  I think we should not do overriding then.  The first is really a `ValueError` imo

ilan-gold:
> IDK how to name it. method is usually for the algorithm, I think (which in this case is leiden). There should be something about naming in the scanpy 2.0 board: https://github.com/orgs/scverse/projects/18/views/1

I went with `backend`.  You're saying we should add an issue?

flying-sheep:
I’ll quickly go over all our APIs and check what related conventions we have.

I don’t want to have too many parameter names for the same set of ideas.

flying-sheep:
OK, done, please check out https://github.com/scverse/scanpy/issues/2828

The sister API `sc.pp.louvain` uses `flavor` for this exact use case. It also says about the `'vtraag'` implementation: “Much more powerful than 'igraph', and the default”. I’m not sure why we consider it that much more “powerful” …

scanpy/_utils/__init__.py
```diff
@@ -45,7 +47,39 @@ def __repr__(self) -> str:
 _empty = Empty.token
 
 # e.g. https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html
-AnyRandom = Union[int, random.RandomState, None]  # maybe in the future random.Generator
+AnyRandom = Union[
+    int, np.random.RandomState, None
+]  # maybe in the future random.Generator
```
flying-sheep:
```suggestion
# maybe in the future random.Generator
AnyRandom = Union[int, np.random.RandomState, None]
```

scanpy/_utils/__init__.py
```diff
@@ -20,10 +22,10 @@
 import numpy as np
 from anndata import AnnData
 from anndata import __version__ as anndata_version
-from numpy import random
 from numpy.typing import NDArray
 from packaging import version
 from scipy import sparse
+from sklearn.dummy import check_random_state
```
flying-sheep: Good find! I’d just have written our own util, this is better. But isn’t [this](https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html) the exported location of it?

```suggestion
from sklearn.utils import check_random_state
```

scanpy/tests/test_clustering.py
```diff
@@ -13,10 +14,75 @@ def adata_neighbors():
 
 
 @needs.leidenalg
-def test_leiden_basic(adata_neighbors):
-    sc.tl.leiden(adata_neighbors)
+@needs.igraph
+@pytest.mark.parametrize("flavor", ["igraph", "leidenalg", "leidenalg"])
```
flying-sheep: Why is `leidenalg` duplicated here?

```suggestion
@pytest.mark.parametrize(
    "flavor",
    [
        pytest.param("igraph", marks=needs.igraph),
        pytest.param("leidenalg", marks=needs.leidenalg),
    ],
)
```

An alternative is to reuse the param list here and below:

```suggestion
FLAVORS = [
    pytest.param("igraph", marks=needs.igraph),
    pytest.param("leidenalg", marks=needs.leidenalg),
]

@pytest.mark.parametrize("flavor", FLAVORS)
```

scanpy/tests/test_clustering.py
```diff
@@ -13,10 +14,75 @@ def adata_neighbors():
...
+        adata_neighbors,
+        flavor=flavor,
+        resolution=resolution,
+        n_iterations=n_iterations,
+    )
+    assert adata_neighbors.uns["leiden"]["params"]["resolution"] == resolution
+    assert adata_neighbors.uns["leiden"]["params"]["n_iterations"] == n_iterations
+
+
+@pytest.mark.parametrize("flavor", ["igraph", "leidenalg", "leidenalg"])
```
flying-sheep:
```suggestion
@pytest.mark.parametrize(
    "flavor",
    [
        pytest.param("igraph", marks=needs.igraph),
        pytest.param("leidenalg", marks=needs.leidenalg),
    ],
)
```

or 

```suggestion
@pytest.mark.parametrize("flavor", FLAVORS)
```

scanpy/tools/_leiden.py
```diff
@@ -134,22 +138,40 @@ def leiden(
...
-        partition_kwargs["resolution_parameter"] = resolution
+        clustering_args["resolution_parameter"] = resolution
     # clustering proper
-    part = leidenalg.find_partition(g, partition_type, **partition_kwargs)
+    if not flavor == "leidenalg":
+        objective_function = clustering_args.pop("objective_function", "modularity")
+        with _utils.set_igraph_random_state(random_state):
+            part = g.community_leiden(
+                objective_function=objective_function, **clustering_args
+            )
```
flying-sheep: This would work too, no?

```suggestion
        clustering_args.setdefault("objective_function", "modularity")
        with _utils.set_igraph_random_state(random_state):
            part = g.community_leiden(**clustering_args)
```

flying-sheep (changes requested):
Looking great! Some nitpicks, otherwise this is good to go!

scanpy/tests/notebooks/test_pbmc3k.py
ilan-gold: I didn't want to touch whatever this is doing since it's based on mean expression.

flying-sheep (approved):

scanpy/tools/_leiden.py
```diff
@@ -96,9 +97,11 @@ def leiden(
...
     copy
         Whether to copy `adata` or modify it inplace.
-    **partition_kwargs
-        Any further arguments to pass to `~leidenalg.find_partition`
-        (which in turn passes arguments to the `partition_type`).
+    flavor
+        Which package's implementation to use.
+    **clustering_args
+        Any further arguments to pass to :func:`~leidenalg.find_partition` (which in turn passes arguments to the `partition_type`)
+        or :method:`igraph.Graph.community_leiden` from `igraph`.
```
flying-sheep:
```suggestion
        or :meth:`igraph.Graph.community_leiden` from `igraph`.
```

flying-sheep:
OK, to help debugging, I extracted the data and checked if it’s all we need. And it is.

```python
import seaborn as sns
import pandas as pd

data = pd.read_parquet("violin-mess.parquet")
palette = {
    "CD4 T cells": "#1f77b4", "CD8 T cells": "#ff7f0e", "B cells": "#2ca02c",
    "FCGR3A+ Monocytes": "#d62728", "NK cells": "#9467bd", "CD14+ Monocytes": "#8c564b",
    "Dendritic cells": "#e377c2", "Megakaryocytes": "#7f7f7f",
}
kwargs = {
    "x": "leiden", "y": "CST3", "order": None, "orient": "vertical", "density_norm": "width",
    "hue": "leiden", "palette": palette, "cut": 0, "inner": None,
}

sns.violinplot(data=data, **kwargs)
```

[violin-mess.parquet.gz](https://github.com/scverse/scanpy/files/14310251/violin-mess.parquet)

flying-sheep:
Hm, weird, I get the same results on the master branch

flying-sheep:
The data is actually different, but when I plot both, they both get super slim:

grafik

grafik

flying-sheep:
Aha, the data is actually different. With current versions of pandas and seaborn: 

https://gist.github.com/flying-sheep/1e7d96ac9d367a645fc487947077081b

ilan-gold:
@flying-sheep That the data is different is not wildly surprising because the cluster assignment is different.

ilan-gold:
I am definitely surprised about the magnitude of difference, though.

flying-sheep:
look at how messed up the positioning of these violins is. They aren’t even centered on their axis ticks. Maybe the fact that one cluster is completely missing in your version is responsible.

scanpy/tests/test_clustering.py
```diff
@@ -61,17 +61,14 @@ def test_leiden_random_state(adata_neighbors, flavor):
...
         directed=is_leiden_alg,
         n_iterations=n_iterations,
     )
     assert (adata_1.obs["leiden"] == adata_1_again.obs["leiden"]).all()
     # This random state produces different categories so can't check the arrays against each other.
-    assert (
-        adata_2.obs["leiden"].cat.categories
-        != adata_1_again.obs["leiden"].cat.categories
-    ).any()
+    assert (adata_2.obs["leiden"] != adata_1_again.obs["leiden"]).any()
```
flying-sheep: comment seems no longer valid

scanpy/tools/_leiden.py
```diff
@@ -134,22 +142,38 @@ def leiden(
             adjacency=adjacency,
         )
     # convert it to igraph
+    if not flavor == "leidenalg" and directed:
```
ivirshup:
```suggestion
    if flavor == "igraph" and directed:
```

scanpy/tools/_leiden.py
```diff
@@ -112,13 +116,17 @@ def leiden(
...
+    if flavor == "leidenalg":
+        try:
+            import leidenalg
+
+            msg = 'Use of leidenalg is discouraged and will be deprecated in the future.  Please use `flavor="igraph"` `n_iterations=2` to achieve similar results.  `directed` must also be `False` to work with `igraph`\'s implementation.'
+            warnings.warn(msg, FutureWarning)
+        except ImportError:
+            raise ImportError(
+                "Please install the leiden algorithm: `conda install -c conda-forge leidenalg` or `pip3 install leidenalg`."
+            )
```
ivirshup:
```suggestion
    if flavor == "leidenalg":
        try:
            import leidenalg

            msg = 'Use of leidenalg is discouraged and will be deprecated in the future.  Please use `flavor="igraph"` `n_iterations=2` to achieve similar results.  `directed` must also be `False` to work with `igraph`\'s implementation.'
            warnings.warn(msg, FutureWarning)
        except ImportError:
            raise ImportError(
                "Please install the leiden algorithm: `conda install -c conda-forge leidenalg` or `pip3 install leidenalg`."
            )
    elif flavor == "igraph":
        # Make sure igraph is installed
        ...
    else:
        raise ValueError(f"flavor must be either 'igraph' or 'leidenalg', but '{flavor}' was passed")
```

Similar handling for igraph and making sure the `flavor` argument is valid.

scanpy/tools/_leiden.py
```diff
@@ -134,22 +142,38 @@ def leiden(
...
+        )
+    clustering_args["n_iterations"] = n_iterations
+    if flavor == "leidenalg":
+        clustering_args["seed"] = random_state
     if resolution is not None:
-        partition_kwargs["resolution_parameter"] = resolution
+        clustering_args["resolution_parameter"] = resolution
     # clustering proper
-    part = leidenalg.find_partition(g, partition_type, **partition_kwargs)
+    if not flavor == "leidenalg":
```
ivirshup:
```suggestion
    if flavor == "igraph":
```

ivirshup (commented):
There are a lot of branches here like `if flavor == "leidenalg"`, I think it could be cleaner to just have one higher level branch for the two cases. 

------

I think we can change the default for `directed` to be `None` then dynamically set the value based on which flavor is passed.

We could do the same with `n_iterations`, though I'm a little on the fence about this.

---

A number of plots are being changed that don't seem to be very different. Mostly things pre-clustering, like pca and qc metrics. I'm going to check if these are actually different enough to trigger test failures.

ilan-gold:
@ivirshup You are right about some of the pre-processing plots.  I should have created a separate branch for those.  I simply think hte outputs have changed and we never caught it but would be curious to see what you get.  It could be my M1

scanpy/tools/_leiden.py
ilan-gold: We don't check this elsewhere.  `igraph` is a direct dependency of `scanpy`.  We can raise a `ValueError` though.

ilan-gold: Nevermind, it's not.  But we didn't check this before.  I'll add a check then.

ilan-gold:
@ivirshup I simplified the conditionals a bit and there are only two sets now.  One to check for various `{Value/Import}Error`s and another to do the `clustering_kwargs` building.  I think this is cleaner and faster since no code will run that doesn't have to.  I didn't really see a way to do it with only one set of conditionals without code duplication.  There's some code that's just common to both, but that shouldn't be run in the case of one of the  `{Value/Import}Error`s .

ivirshup:
About the preprocessing plots:

`scanpy/tests/notebooks/_images_pbmc3k/filter_genes_dispersion/expected.png`

* Text and some points are shifted slightly. I'm not totally sure whether any points are actually in a different place

`scanpy/tests/notebooks/_images_pbmc3k/highest_expr_genes/expected.png`
`scanpy/tests/notebooks/_images_pbmc3k/pca/expected.png`
`scanpy/tests/notebooks/_images_pbmc3k/pca_variance_ratio/expected.png`
`scanpy/tests/notebooks/_images_pbmc3k/scatter_2/expected.png`

* Axis text shifted slightly
* Can probably be reverted if the tests still pass

`scanpy/tests/notebooks/_images_pbmc3k/scatter_1/expected.png`

* y axis moved

ilan-gold:
@ivirshup I will revert those, and hopefully tests pass

ivirshup (commented):
Looks good! But I don't think that you addressed these:

> I think we can change the default for directed to be None then dynamically set the value based on which flavor is passed.
>
> We could do the same with n_iterations, though I'm a little on the fence about this.

I think I would definitely change `directed`, and add a note in the docs that setting `n_iterations=2` will be faster and is the default for the underlying packages.

ilan-gold:
Ah, you're right.  Will change.

ivirshup (approved):
Great, looks good!

alam-shahul:
Hi there @ilan-gold @ivirshup, I just wanted to get some clarification on something. Suppose I have the following bit of code:

```
sc.pp.neighbors(dataset, random_state=0, n_neighbors=40)
sc.tl.leiden(dataset, resolution=1.0, random_state=0)
```

Should it return the same results with `scanpy==1.9.3` and `scanpy=1.10.0`? From the discussion, it seems that these two should both use the `leidenalg` algorithm with `n_iterations=-1` unless other arguments are passed in. But I am finding that the results are different with different package versions.

ilan-gold:
@alam-shahul can you create an issue with a reproducer?

alam-shahul:
@ilan-gold Sorry for the delay, here you go: #3422

PR #2818, opened 2024-01-18
Revert "Use version helper everywhere"

ivirshup:
Reverts scverse/scanpy#2817 I would like a chance to review and actually come to a conclusion here

cc: @flying-sheep


flying-sheep:
🤷 sure, but I don’t think it makes a difference

PR #2843, opened 2024-02-08
Prevent oversubscription during tests

ivirshup:
A number of multithreaded functions and libraries we use default to `os.cpu_count()` number of threads. This is a problem when multiple processes are running in parallel, as is the case when using pytest-xdist.

This oversubscription can lead to an increase in test time when multiple workers are used.

This PR limits how many threads most libraries use via `threadpoolctl`, and scales this to the number of workers available on the host.

I personally see improvements of ~10x when running with this setting on a server with 16 cores, using `-n auto`.

### DIFF
--- a/scanpy/testing/_pytest/__init__.py
+++ b/scanpy/testing/_pytest/__init__.py
@@ -39,6 +39,20 @@ def _global_test_context(request: pytest.FixtureRequest) -> Generator[None, None
     plt.close("all")
 
 
+@pytest.fixture(autouse=True, scope="session")
+def limit_cpus():
+    import os
+
+    import threadpoolctl
+
+    if "PYTEST_XDIST_WORKER_COUNT" in os.environ:
+        n_workers = int(os.environ["PYTEST_XDIST_WORKER_COUNT"])
+        max_threads = os.cpu_count() // n_workers
+
+        with threadpoolctl.threadpool_limits(limits=max_threads):
+            yield
+
+
 def pytest_addoption(parser: pytest.Parser) -> None:
     parser.addoption(
         "--internet-tests",
--- a/scanpy/testing/_pytest/__init__.py
+++ b/scanpy/testing/_pytest/__init__.py
@@ -50,7 +50,7 @@ def limit_multithreading():
     if (n_workers := os.environ.get("PYTEST_XDIST_WORKER_COUNT")) is not None:
         import threadpoolctl
 
-        max_threads = os.cpu_count() // n_workers
+        max_threads = max(os.cpu_count() // n_workers, 1)
 
         with threadpoolctl.threadpool_limits(limits=max_threads):
             yield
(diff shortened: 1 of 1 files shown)


ivirshup:
~~But it seems to throw an error if I combine it with `-k`:~~

```
ValueError: limit_multithreading did not yield a value
```

Fixed

scanpy/testing/_pytest/__init__.py
```diff
@@ -39,6 +40,23 @@ def _global_test_context(request: pytest.FixtureRequest) -> Generator[None, None
...
+    running at once.
+    """
+    if "PYTEST_XDIST_WORKER_COUNT" in os.environ:
+        import threadpoolctl
+
+        n_workers = int(os.environ["PYTEST_XDIST_WORKER_COUNT"])
+        max_threads = os.cpu_count() // n_workers
+
+        with threadpoolctl.threadpool_limits(limits=max_threads):
+            yield
```
ivirshup, line 56:
```suggestion
            yield
    else:
        yield
```

scanpy/testing/_pytest/__init__.py
```diff
@@ -39,6 +40,25 @@ def _global_test_context(request: pytest.FixtureRequest) -> Generator[None, None
...
+def limit_multithreading():
+    """Limit number of threads used per worker when using pytest-xdist.
+
+    Prevents oversubscription of the CPU when multiple tests with parallel code are
+    running at once.
+    """
+    if "PYTEST_XDIST_WORKER_COUNT" in os.environ:
+        import threadpoolctl
+
+        n_workers = int(os.environ["PYTEST_XDIST_WORKER_COUNT"])
```
flying-sheep:
```suggestion
    if (n_workers := os.environ.get("PYTEST_XDIST_WORKER_COUNT")) is not None:
        import threadpoolctl

```

flying-sheep (approved):
That’s great! I don’t see much change in the test suite duration on Azure, should we expect any change?

Also, what happens if `max_threads` becomes 0 e.g. if the number of workers is 2 and the CPU count is 1?

ivirshup:
> That’s great! I don’t see much change in the test suite duration on Azure, should we expect any change?

I wouldn't expect a change on CI, we're not using pytest-xdist here plus I'm not sure if we have enough CPUs to bother.

Mainly for dev machine testing

scanpy/testing/_pytest/__init__.py
```diff
@@ -39,6 +40,24 @@ def _global_test_context(request: pytest.FixtureRequest) -> Generator[None, None
...
+def limit_multithreading():
+    """Limit number of threads used per worker when using pytest-xdist.
+
+    Prevents oversubscription of the CPU when multiple tests with parallel code are
+    running at once.
+    """
+    if (n_workers := os.environ.get("PYTEST_XDIST_WORKER_COUNT")) is not None:
+        import threadpoolctl
+
+        max_threads = os.cpu_count() // n_workers
```
ivirshup:
```suggestion
        max_threads = max(os.cpu_count() // n_workers, 1)
```

flying-sheep:
> we're not using pytest-xdist here

Ah. you do install it in the minimum-tests PR though. I guess that means that you intended the min-deps install script for local CI as well.

anyway, LGTM!

PR #2844, opened 2024-02-08
Fix some plot warnings

### LINKED ISSUE #2839: Fix plotting warnings raised in tutorial notebooks

- [X] (optional) I have confirmed this bug exists on the master branch of scanpy.

### What happened?

Many different calls in scanpy emit warnings that are currently suppressed by our testing framework (I think).  

### Minimal code sample

I discovered this unrelatedly by editing the notebooks, see for example: https://github.com/scverse/scanpy-tutorials/blob/master/spatial/integration-scanorama.ipynb

@flying-sheep mentioned that the scanpy tests filter out warnings and indeed you can reproduce these by e.g.,:
```sh
pytest -W error::FutureWarning -n auto scanpy/tests/test_plotting.py
```

### Error output

- [x] `…/scanpy/plotting/_tools/scatterplots.py:401:`

  > UserWarning: No data for colormapping provided via 'c'. Parameters 'cmap' will be ignored

- [x] `…/scanpy/plotting/_tools/__init__.py:1269:`

   > FutureWarning: The `scale` parameter has been renamed and will be removed in v0.15.0. Pass `density_norm='width'` for the same effect.  
  > `_ax = sns.violinplot(`

- [x] `…/scanpy/preprocessing/_simple.py:274:`

  > ImplicitModificationWarning: Trying to modify attribute `.var` of view, initializing view as actual.
  > `adata.var["n_cells"] = number`

- [x] `…/scanpy/plotting/_stacked_violin.py:503: FutureWarning:`

  > Passing `palette` without assigning `hue` is deprecated and will be removed in v0.14.0. Assign the `x` variable to `hue` and set `legend=False` for the same effect. 
  > `row_ax = sns.violinplot(`

### Versions

<details>

```
-----
anndata     0.10.4
scanpy      1.10.0.dev191+gf7f5d5c6
-----
IPython             8.20.0
PIL                 10.2.0
asciitree           NA
asttokens           NA
cffi                1.16.0
cloudpickle         3.0.0
cycler              0.12.1
cython_runtime      NA
dask                2024.1.0
dateutil            2.8.2
decorator           5.1.1
defusedxml          0.7.1
executing           2.0.1
fasteners           0.19
h5py                3.10.0
igraph              0.10.8
iniconfig           NA
jedi                0.19.1
jinja2              3.1.3
joblib              1.3.2
kiwisolver          1.4.5
legacy_api_wrap     NA
leidenalg           0.10.1
llvmlite            0.41.1
markupsafe          2.1.3
matplotlib          3.8.2
mpl_toolkits        NA
msgpack             1.0.7
natsort             8.4.0
numba               0.58.1
numcodecs           0.12.1
numpy               1.26.3
packaging           23.2
pandas              2.1.4
parso               0.8.3
pexpect             4.9.0
pluggy              1.3.0
prompt_toolkit      3.0.43
psutil              5.9.7
ptyprocess          0.7.0
pure_eval           0.2.2
py                  NA
pygments            2.17.2
pyparsing           3.1.1
pytest              7.4.4
pytz                2023.3.post1
scipy               1.11.4
session_info        1.0.0
setuptools          68.2.2
setuptools_scm      NA
sitecustomize       NA
six                 1.16.0
sklearn             1.3.2
sparse              0.15.1
stack_data          0.6.3
tblib               3.0.0
texttable           1.7.0
threadpoolctl       3.2.0
tlz                 0.12.0
toolz               0.12.0
traitlets           5.14.1
typing_extensions   NA
wcwidth             0.2.13
yaml                6.0.1
zarr                2.16.1
zipp                NA
-----
Python 3.11.6 (main, Nov  2 2023, 04:39:43) [Clang 14.0.3 (clang-1403.0.22.14.1)]
macOS-13.6.1-arm64-arm-64bit
-----
Session information updated at 2024-01-30 16:40
```

</details>

flying-sheep:
- [x] Closes #2839

How to review:

1. https://github.com/scverse/scanpy/compare/585f58c9e4dd82dd7809a831538c4e230b008818...60804430db089f1887085e537ad946b7f691a8b4
2. 5710a1de4f091db607d76b790676b56f988638f8
3. https://github.com/scverse/scanpy/compare/5710a1de4f091db607d76b790676b56f988638f8...2cbe106328bcb9585c13def47ac3e1bf9629e448

### DIFF
--- a/scanpy/plotting/_stacked_violin.py
+++ b/scanpy/plotting/_stacked_violin.py
@@ -390,9 +390,7 @@ class StackedViolin(BasePlot):
         if self.are_axes_swapped:
             _color_df = _color_df.T
 
-        cmap = plt.get_cmap(self.kwds.get("cmap", self.cmap))
-        if "cmap" in self.kwds:
-            del self.kwds["cmap"]
+        cmap = plt.get_cmap(self.kwds.pop("cmap", self.cmap))
         normalize = check_colornorm(
             self.vboundnorm.vmin,
             self.vboundnorm.vmax,
@@ -441,9 +439,7 @@ class StackedViolin(BasePlot):
     ):
         import seaborn as sns  # Slow import, only import if called
 
-        row_palette = self.kwds.get("color", self.row_palette)
-        if "color" in self.kwds:
-            del self.kwds["color"]
+        row_palette = self.kwds.pop("color", self.row_palette)
         if row_palette is not None:
             if is_color_like(row_palette):
                 row_colors = [row_palette] * _color_df.shape[0]
@@ -525,10 +521,12 @@ class StackedViolin(BasePlot):
             row_ax = sns.violinplot(
                 x=x,
                 y="values",
-                hue=x,
                 data=_df,
                 orient="vertical",
                 ax=row_ax,
+                # use a single `color`` if row_colors[idx] is defined
+                # else use the palette
+                hue=None if palette_colors is None else x,
                 palette=palette_colors,
                 color=row_colors[idx],
                 **self.kwds,
--- a/scanpy/plotting/_stacked_violin.py
+++ b/scanpy/plotting/_stacked_violin.py
@@ -140,6 +140,10 @@ class StackedViolin(BasePlot):
 
     def __getattribute__(self, name: str) -> object:
         """Called unconditionally when accessing an instance attribute"""
+        # If the user has set the deprecated version on the class,
+        # and our code accesses the new version from the instance,
+        # return the user-specified version instead and warn.
+        # This is done because class properties are hard to do.
         if name == "DEFAULT_DENSITY_NORM" and hasattr(self, "DEFAULT_SCALE"):
             msg = "Don’t set DEFAULT_SCALE, use DEFAULT_DENSITY_NORM instead"
             warnings.warn(msg, FutureWarning)
--- a/scanpy/plotting/_stacked_violin.py
+++ b/scanpy/plotting/_stacked_violin.py
@@ -298,7 +298,7 @@ class StackedViolin(BasePlot):
             (see :func:`~seaborn.color_palette`).
             Alternatively, a single color name or hex value can be passed,
             e.g. `'red'` or `'#cc33ff'`.
-        scale
+        density_norm
             The method used to scale the width of each violin.
             If 'width' (the default), each violin will have the same width.
             If 'area', each violin will have the same area.
--- a/scanpy/plotting/_stacked_violin.py
+++ b/scanpy/plotting/_stacked_violin.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import warnings
 from typing import TYPE_CHECKING, Literal
 
 import numpy as np
@@ -13,7 +14,13 @@ from .._settings import settings
 from .._utils import _doc_params
 from ._baseplot_class import BasePlot, _VarNames, doc_common_groupby_plot_args
 from ._docs import doc_common_plot_args, doc_show_save_ax, doc_vboundnorm
-from ._utils import _AxesSubplot, check_colornorm, make_grid_spec, savefig_or_show
+from ._utils import (
+    _AxesSubplot,
+    _deprecated_scale,
+    check_colornorm,
+    make_grid_spec,
+    savefig_or_show,
+)
 
 if TYPE_CHECKING:
     from collections.abc import (
@@ -52,7 +59,7 @@ class StackedViolin(BasePlot):
         Order in which to show the categories. Note: if `dendrogram=True`
         the categories order will be given by the dendrogram and `order`
         will be ignored.
-    scale
+    density_norm
         The method used to scale the width of each violin.
         If 'width' (the default), each violin will have the same width.
         If 'area', each violin will have the same area.
@@ -108,7 +115,7 @@ class StackedViolin(BasePlot):
     DEFAULT_JITTER_SIZE = 1
     DEFAULT_LINE_WIDTH = 0.2
     DEFAULT_ROW_PALETTE = None
-    DEFAULT_SCALE = "width"
+    DEFAULT_DENSITY_NORM: Literal["area", "count", "width"] = "width"
     DEFAULT_PLOT_YTICKLABELS = False
     DEFAULT_YLIM = None
     DEFAULT_PLOT_X_PADDING = 0.5  # a unit is the distance between two x-axis ticks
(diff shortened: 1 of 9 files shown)


scanpy/tests/test_plotting.py
```diff
@@ -1685,3 +1685,10 @@ def test_string_mask(tmp_path, check_same_image):
     plt.close()
 
     check_same_image(p1, p2, tol=1)
+
+
+def test_violin_scale_warning(monkeypatch):
+    adata = pbmc3k_processed()
+    monkeypatch.setattr(sc.pl.StackedViolin, "DEFAULT_SCALE", "count", raising=False)
+    with pytest.warns(FutureWarning, match="Don’t set DEFAULT_SCALE"):
+        sc.pl.StackedViolin(adata, adata.var_names[:3], groupby="louvain")
```
ilan-gold, line 1694: This is just spitballing, but do you think it's worth adding a test so that these functions don't raise __any__ warnings (of a certain kind, perhaps)?  Just a loose idea, but would it be easier to permit certain warnings than to make a test every time we come across a warning that should be squashed?

scanpy/plotting/_stacked_violin.py
```diff
@@ -131,6 +138,14 @@ class StackedViolin(BasePlot):
     # None will draw unadorned violins.
     DEFAULT_INNER = None
 
+    def __getattribute__(self, name: str) -> object:
+        """Called unconditionally when accessing an instance attribute"""
+        if name == "DEFAULT_DENSITY_NORM" and hasattr(self, "DEFAULT_SCALE"):
```
ilan-gold, line 147:
```suggestion
        if (name == "DEFAULT_DENSITY_NORM" and hasattr(self, "DEFAULT_SCALE")) or name == 'DEFAULT_SCALE':
```

Is there any scenario in which people are using this publicly? May want to warn on that as well.

scanpy/plotting/_stacked_violin.py
```diff
@@ -248,11 +263,13 @@ def style(
         jitter_size: int | None = DEFAULT_JITTER_SIZE,
         linewidth: float | None = DEFAULT_LINE_WIDTH,
         row_palette: str | None = DEFAULT_ROW_PALETTE,
-        scale: Literal["area", "count", "width"] | None = DEFAULT_SCALE,
+        density_norm: Literal["area", "count", "width"] = DEFAULT_DENSITY_NORM,
```
ilan-gold, line 270: docstring below needed?

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -262,6 +262,7 @@ def embedding(
     #     color=gene2, components = [1, 2], color=gene2, components=[2,3],
     # ]
     for count, (value_to_plot, dims) in enumerate(zip(color, dimensions)):
+        kwargs = kwargs_global.copy()  # is potentially mutated for each plot
```
ilan-gold: Not a huge fan of `kwargs_global`.  Why not `kwargs_scatter = kwargs.copy()`?  The copy of `kwargs` are only used for the scatter plot and `kwargs_outline` is formed from this copy.

scanpy/tests/test_plotting.py
ilan-gold: For example, at the moment, I get 

```python
In [5]: sc.pl.rank_genes_groups_violin(adata, groups='0', n_genes=8, density_norm='count')
/Users/ilangold/Projects/Theis/scanpy/scanpy/plotting/_tools/__init__.py:1287: FutureWarning:

Setting a gradient palette using color= is deprecated and will be removed in v0.14.0. Set `palette='dark:black'` for the same effect.

  _ax = sns.stripplot(
/Users/ilangold/Projects/Theis/scanpy/scanpy/plotting/_tools/__init__.py:1303: UserWarning: set_ticklabels() should only be used with a fixed number of ticks, i.e. after set_ticks() or using a FixedLocator.
  _ax.set_xticklabels(new_gene_names, rotation="vertical")
```

flying-sheep: That’s the goal. Right now we have to fix the myriad of warnings in the plotting code one-by-one because nobody can review a PR that fixes all the warnings.

And for this PR here I chose to fix the ones you mentioned. Of course I can add that one, but I don’t want this to spiral out of control.

scanpy/plotting/_stacked_violin.py
flying-sheep: The problem is that this is a class attribute. So I can’t catch this when it’s being set on the class (that would involve a `property` on a metaclass, which is super out of the way and quite complicated)

I instead settled on the warning being thrown when the new attribute is read (which we always do from the instance) while the user has manually set the old one. Your suggested change wouldn’t do anything. But I’ll add a comment explaining this.

PS: setting that on the class isn’t a great UX, I know, but it’s what we have now. Let’s fix that when we overhaul plotting.

scanpy/plotting/_stacked_violin.py
flying-sheep: good call, done

scanpy/plotting/_tools/scatterplots.py
flying-sheep: OK!

scanpy/tests/test_plotting.py
ilan-gold: Got it.

scanpy/plotting/_tools/scatterplots.py
```diff
@@ -262,6 +262,7 @@ def embedding(
     #     color=gene2, components = [1, 2], color=gene2, components=[2,3],
     # ]
     for count, (value_to_plot, dims) in enumerate(zip(color, dimensions)):
+        kwargs = kwargs_scatter.copy()  # is potentially mutated for each plot
```
ilan-gold: Maybe I was not clear, but I mean to not have some object outside the loop.
```suggestion
        kwargs_scatter = kwargs.copy()  # is potentially mutated for each plot
```
And the replace all `kwargs` in the loop with `kwargs_scatter` since they are only used on `scatter` and then make `kwargs_outline` from `kwargs_scatter`

ilan-gold: Is there something I'm missing about this?  I think defining a variable to store something outside a loop is a little strange.

flying-sheep: No, that makes total sense. I basically did what you want, but less clean and with a smaller diff.

In the long run, we should just extract the loop body to a function and only pass a copy into it, then all is good.

ilan-gold (approved):

PR #2852, opened 2024-02-13
Ignore dask dataframe warning in tests

flying-sheep:
- [x] Closes N/A

- [x] Release notes not necessary because: dev change

### DIFF
--- a/scanpy/testing/_pytest/__init__.py
+++ b/scanpy/testing/_pytest/__init__.py
@@ -41,7 +41,7 @@ def _global_test_context(request: pytest.FixtureRequest) -> Generator[None, None
 
 
 @pytest.fixture(autouse=True, scope="session")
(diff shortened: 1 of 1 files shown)


scanpy/testing/_pytest/__init__.py
```diff
@@ -55,9 +55,26 @@ def limit_multithreading():
         max_threads = max(n_cpus // n_workers, 1)
 
         with threadpoolctl.threadpool_limits(limits=max_threads):
-            yield
+            yield max_threads
```
ivirshup, line 58: What's this about?

flying-sheep: Fixtures that don’t start with an underscore evaluate to things that aren’t `None`. I just made this one return a thing.

I can alternatively prefix its name with an underscore

Issue #2578, opened 2023-07-28
Dask support


flying-sheep:
### What kind of feature would you like to request?

Additional function parameters / changed functionality / changed defaults?

### Please describe your wishes

Meta issue tracking scanpy functions without dask support

Related #921

First we should test things more generically:

- [x] add global `array_type` fixture to use in all tests for features below
  - #2595

@ivirshup came up with places where we should prioritize dask support:

- [x] `normalize_total`
  - seems to work: scverse/scanpy#2466
  - has tests: scverse/scanpy#1663
- [x] `log1p`, `normalize_per_cell`, `filter_cells`/`*_genes`:
  - [x] accept Dask arrays: scverse/scanpy#283
  - [x] Test that they are kept as Dask arrays throughout: scverse/scanpy#2814
- [x] `sc.pp.pca`: scverse/scanpy#2563 scverse/scanpy#3263
- [x] `sc.pp.calculate_qc_metrics` #3307
- `sc.pp.highly_variable_genes`:
  - [x] scverse/scanpy#2777
  - [x] scverse/scanpy#2807
  - [x] scverse/scanpy#2808
  - [ ] experimental “pearson_residuals” flavor
- [ ] `sc.tl.rank_genes_groups`
  - [x] depends on #2621
- [x] `sc.pp.subsample`/`sc.pp.sample`: so simple it has always been supported
- [x] Aggregate: https://flox.readthedocs.io/en/latest/ #3659
  - [x] CSC: #3872

Later:

- [x] scale scverse/scanpy#2942
- [ ] nearest-neighbors: https://github.com/dask/dask-ml/issues/982 may need to implement ourselves...
- [ ] umap: implement ourselves?
- [ ] metrics: depednent on squidpy decisions but should be dask-ifiable
- [ ] scrublet?

mumichae:
For the highly_variable_genes dask support, are you also considering cases where the dask chunks are scipy sparse matrices?

flying-sheep:
Not currently, but since the scope of that PR got reduced, it shouldn’t be too much work.

jan-engelmann:
would be great to be able to use `sc.get.aggregate` with dask arrays!

PR #2856, opened 2024-02-15
(feat): pre-processing functions for `dask` with sparse chunks

### LINKED ISSUE #2764: Highly variable genes for sparse dataset in backed mode

### What happened?

The exception happened when try to run scanpy `highly_variable_genes` with sparse dataset loaded in backed mode

### Minimal code sample

```python
# read backed
adata = anndata.read_h5ad(file_path, backed='r')
X = adata.raw.X if adata.raw is not None else adata.X
# dataset must be sparse there
print(issparse(X[0]))
# calculate dispersions
sc.pp.highly_variable_genes(adata, min_mean=0.0125, max_mean=3, min_disp=0.5, inplace=False)
```
True
```


### Error output

```pytb
loop of ufunc does not support argument 0 of type SparseDataset which has no callable expm1 method!
```

goes from https://github.com/scverse/scanpy/blob/bc349b999be62196aa51b59db6e2daa37f428322/scanpy/preprocessing/_highly_variable_genes.py#L206

### Versions

<details>

```
anndata     0.8.0
scanpy      1.9.1
-----
PIL                 9.3.0
appnope             0.1.3
asttokens           NA
backcall            0.2.0
cycler              0.10.0
cython_runtime      NA
dateutil            2.8.2
debugpy             1.6.3
decorator           5.1.1
entrypoints         0.4
executing           1.2.0
google              NA
h5py                3.7.0
igraph              0.10.2
ipykernel           6.17.1
jedi                0.18.2
joblib              1.2.0
kiwisolver          1.4.4
leidenalg           0.9.0
llvmlite            0.39.1
louvain             0.8.0
matplotlib          3.6.2
mpl_toolkits        NA
natsort             8.2.0
numba               0.56.4
numpy               1.23.5
packaging           21.3
pandas              1.2.1
parso               0.8.3
pexpect             4.8.0
pickleshare         0.7.5
pkg_resources       NA
platformdirs        2.5.4
plotly              5.11.0
prompt_toolkit      3.0.33
psutil              5.9.4
ptyprocess          0.7.0
pure_eval           0.2.2
pydev_ipython       NA
pydevconsole        NA
pydevd              2.8.0
pydevd_file_utils   NA
pydevd_plugins      NA
pydevd_tracing      NA
pygments            2.13.0
pyparsing           3.0.9
pytz                2022.6
scipy               1.9.3
session_info        1.0.0
setuptools          62.3.2
sitecustomize       NA
six                 1.16.0
sklearn             1.1.3
stack_data          0.6.1
texttable           1.6.6
threadpoolctl       3.1.0
tornado             6.2
traitlets           5.5.0
typing_extensions   NA
wcwidth             0.2.5
yaml                6.0
zipp                NA
zmq                 24.0.1
-----
IPython             8.6.0
jupyter_client      7.4.7
jupyter_core        5.0.0
-----
Python 3.9.13 (main, May 24 2022, 21:28:31) [Clang 13.1.6 (clang-1316.0.21.2)]
macOS-13.4-x86_64-i386-64bit
-----
```

</details>

flying-sheep:
- [x] Towards #2578 and maybe others? #2764

Note I did not follow:
https://gist.github.com/Intron7/bbf5058794be7b81d3953ae39c17d8b8

This is because this PR is basically very simple.  I just added an `axis_sum` function for dispatching on dask arrays (with sparse chunks) which now handles the needed functionality and then it propagates up to various functions as noted in the release note: `scanpy.pp.scale`, `scanpy.pp.filter_cells`, `scanpy.pp.filter_genes`, `scanpy.pp.scale` and `scanpy.pp.highly_variable_genes`


scanpy/preprocessing/_normalization.py
```diff
@@ -35,7 +35,7 @@ def _normalize_data(X, counts, after=None, copy: bool = False):
     elif isinstance(counts, np.ndarray):
         np.divide(X, counts[:, None], out=X)
     else:
-        X = np.divide(X, counts[:, None])  # dask does not support kwarg "out"
+        X = X / counts[:, None]
```
ilan-gold: This division op causes a `COO` matrix to be spit out.  We could map blocks back to CSR or write a custom division function at a lower level.  However, there's no guarantee on the blocks of X and those of counts really matching up...

ilan-gold: I'm not sure if this is a blocker or not...also just mapping the blocks of `X` to `csr_matrix` or changing `_meta` back also doesn't seem to work amazingly

scanpy/_utils/__init__.py
```diff
@@ -560,6 +560,34 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+@axis_sum.register
+def _(X: DaskArray, axis=None):
+    import dask.array as da
+
+    def sum_drop_keepdims(*args, **kwargs):
+        kwargs.pop("computing_meta", None)
+        if isinstance(X._meta, (sparse.spmatrix, np.matrix)) or isinstance(
+            args[0], (sparse.spmatrix, np.matrix)
+        ):  # forcing the `_meta` to be a sparse array really isn't desirable?
+            kwargs.pop("keepdims", None)
```
ilan-gold, line 742: This is definitely something we don't want to run with the sparse matrices and probably (due to interop) the dense either

scanpy/_utils/__init__.py
```diff
@@ -560,6 +560,34 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+            if isinstance(kwargs["axis"], tuple):
+                kwargs["axis"] = kwargs["axis"][0]
```
ilan-gold: We only ever sum over two axes anyway

scanpy/_utils/__init__.py
```diff
@@ -560,6 +560,34 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+def axis_sum(X: np.ndarray | sparse.spmatrix, axis=None):
+    return np.sum(X, axis=axis)
+
+
+@axis_sum.register
+def _(X: DaskArray, axis=None):
+    import dask.array as da
+
+    def sum_drop_keepdims(*args, **kwargs):
+        kwargs.pop("computing_meta", None)
```
ilan-gold, line 737: This one I am not so sure about.  It doesn't seem to have an impact and also I'm not sure it's used looking at the dask code: https://docs.dask.org/en/stable/search.html?q=computing_meta

scanpy/tests/test_preprocessing_distributed.py
```diff
@@ -140,22 +185,42 @@ def test_write_zarr(adata: AnnData, adata_dist: AnnData):
...
+    npt.assert_allclose(
+        adata_zero_centered.X, adata_dist_zero_centered.X, rtol=1e-6, atol=1e-6
+    )
+
+    adata_non_zero_centered = scale(adata, copy=True, zero_center=False)
+    adata_dist_non_zero_centered = scale(adata_dist, copy=True, zero_center=False)
+    computed_X = adata_dist_non_zero_centered.X.compute()
+    if sp.issparse(computed_X):
+        computed_X = (
+            computed_X.todense()
```
ilan-gold: Same as https://github.com/scverse/scanpy/pull/2856/files#r1505834875

scanpy/_utils/__init__.py
```diff
@@ -560,6 +560,34 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+            args[0], (sparse.spmatrix, np.matrix)
+        ):  # forcing the `_meta` to be a sparse array really isn't desirable?
+            kwargs.pop("keepdims", None)
+            if isinstance(kwargs["axis"], tuple):
+                kwargs["axis"] = kwargs["axis"][0]
+        return da.chunk.sum(*args, **kwargs)
+
+    dtype = getattr(np.zeros(1, dtype=X.dtype).sum(), "dtype", object)
+
+    # operates on `np.matrix` for some reason with sparse chunks in dask so need explicit casting
```
ilan-gold: I imagine this comes from the aggregation of `csr_matrix` instead of `csr_array`

scanpy/testing/_helpers/data.py
```diff
@@ -76,3 +80,49 @@ def _pbmc3k_normalized() -> AnnData:
...
+
+    chunks = []
+    cur_pos = 0
+    for i in range(n_chunks):
+        chunks.append(make_dask_chunk(x, cur_pos, cur_pos + stride))
+        cur_pos += stride
+    if rem:
+        chunks.append(make_dask_chunk(x, cur_pos, x.shape[0]))
+
+    return da.concatenate(chunks, axis=0)
```
ilan-gold: Would be nice to share with anndata/import from anndata

scanpy/tests/test_highly_variable_genes.py
```diff
@@ -85,7 +85,7 @@ def test_no_batch_matches_batch(adata):
 
 
 @pytest.mark.parametrize("batch_key", [None, "batch"], ids=["single", "batched"])
-@pytest.mark.parametrize("array_type", ARRAY_TYPES_SUPPORTED)
+@pytest.mark.parametrize("array_type", ARRAY_TYPES)
```
ilan-gold, line 88: Does the fact this works means we support dask sparse chunks with hvg or is there something else we need to test?  Should update release notes/test that thing if so.

scanpy/_utils/__init__.py
```diff
@@ -560,6 +560,34 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+    def aggregate_sum(*args, **kwargs):
+        return da.chunk.sum(np.array(args[0]), **kwargs)
+
+    return da.reduction(X, sum_drop_keepdims, aggregate_sum, axis=axis, dtype=dtype)
```
ilan-gold: Should we be forcing `meta` to be sparse? I don't think so.  It also seems to break things....

scanpy/preprocessing/_distributed.py
```diff
@@ -35,15 +35,12 @@ def materialize_as_ndarray(
...
     if not isinstance(a, tuple):
         return np.asarray(a)
 
-    if not any(isinstance(arr, DaskArray) for arr in a):
-        return tuple(np.asarray(arr) for arr in a)
-
-    import dask.array as da
-
-    return da.compute(*a, sync=True)
+    return tuple(materialize_as_ndarray(arr) for arr in a)
```
ivirshup: What does changing this do? I would like to keep `da.compute(...)` over multiple arrays so that we schedule them all at once. E.g. allow computing in parallel, while this change seem to make it be serial

scanpy/tests/test_highly_variable_genes.py
```diff
@@ -353,12 +355,13 @@ def test_compare_to_upstream(
     pbmc = pbmc68k_reduced()
     pbmc.X = pbmc.raw.X
     pbmc.var_names_make_unique()
-
     sc.pp.normalize_per_cell(pbmc, counts_per_cell_after=1e4)
     if func == "hvg":
         sc.pp.log1p(pbmc)
+        pbmc.X = array_type(pbmc.X)
```
ilan-gold, line 361: Too many differences propagate from the other functions before this.  In any case, I think this should really only test this individual change anyway.

scanpy/tests/test_highly_variable_genes.py
```diff
@@ -338,23 +338,26 @@ def test_pearson_residuals_batch(pbmc3k_parametrized_small, subset, n_top_genes)
...
         request.node.add_marker(pytest.mark.xfail(reason=msg))
     hvg_info = pd.read_csv(ref_path)
-
     pbmc = pbmc68k_reduced()
     pbmc.X = pbmc.raw.X
-    pbmc.var_names_make_unique()
 
+    pbmc.var_names_make_unique()
     sc.pp.normalize_per_cell(pbmc, counts_per_cell_after=1e4)
+    pbmc.X = array_type(pbmc.X)
```
ilan-gold, line 357: `normalize_per_cell` propagates too many differences and throws off everything downstream of this if this were to happen first

scanpy/tests/test_normalization.py
```diff
@@ -26,21 +26,21 @@
...
-    assert np.allclose(np.ravel(adata.X.sum(axis=1)), [3.0, 3.0, 3.0])
+    assert np.allclose(np.ravel(sc._utils.axis_sum(adata.X, axis=1)), [3.0, 3.0, 3.0])
     sc.pp.normalize_total(adata, target_sum=1, key_added="n_counts2")
-    assert np.allclose(np.ravel(adata.X.sum(axis=1)), [1.0, 1.0, 1.0])
+    assert np.allclose(np.ravel(sc._utils.axis_sum(adata.X, axis=1)), [1.0, 1.0, 1.0])
 
     adata = AnnData(array_type(X_frac).astype(dtype))
     sc.pp.normalize_total(adata, exclude_highly_expressed=True, max_fraction=0.7)
-    assert np.allclose(np.ravel(adata.X[:, 1:3].sum(axis=1)), [1.0, 1.0, 1.0])
+    assert np.allclose(np.ravel(sc._utils.axis_sum(adata.X, axis=1))[:2], [2.0, 4.0])
```
ilan-gold: COO problem mentioned elsewhere forced this change. Can't index into a COO array.

scanpy/_utils/__init__.py
ivirshup: Don't we want `meta` to be dense here?

ilan-gold: Yea, I guess that is the sensible thing to do.  I cleaned up all this stuff now so the type behavior should be clearer

scanpy/preprocessing/_distributed.py
ilan-gold: I think it just seemed cleaner (and handled the case where `a` is a `DaskArray` as the function probably should handle) but you're right, the parallel behavior is nice.

scanpy/preprocessing/_normalization.py
ilan-gold: One option would be to `map_blocks` but use `block_id` and `block_info` to do array access in the `func` passed in (which would use the `scipy.sparse` scaling function).  But I'm not sure how this plays with the parallel model since `counts` would not really be mapped....

ilan-gold: I'm finding this very difficult.  If this coo business is a bug, I think we can just run with "it's a bug, hopefully it's fixed soon"

ivirshup: > However, there's no guarantee on the blocks of X and those of counts really matching up...

Could we just enforce this ourselves?

ilan-gold: @ivirshup I guess this is something I considered but it more likely would mean rechunking `X` (since it might be broken up on the `var` axis) than `counts` which could be expensive?  I guess not though so maybe it's not so bad?

Another option would be to `compute` `counts` and just reference the bug?  I'll try again with rechunking, didn't actually try it before discounting it...

scanpy/preprocessing/_simple.py
```diff
@@ -17,7 +17,7 @@
 from sklearn.utils import check_array, sparsefuncs
 
 from .. import logging as logg
-from .._compat import old_positionals
+from .._compat import old_positionals, sum
```
ivirshup: `sum` is a builtin, we shouldn't be overriding those generally.

It's confusing that `sum(something)` means one thing in any python code, but means something different in this file.

scanpy/preprocessing/_normalization.py
ivirshup: I would assume in most cases `counts` should have a chunk structure like `X` since it was computed from `X`, no?

ivirshup: > X (since it might be broken up on the var axis)

Shouldn't chunking along the `var` axis not mater, since chunking should match along `obs`? I believe `map_blocks` should allow broadcasting when a dimension is missing.

ilan-gold: I agree that it what it says in theory, but in practice this was not what happened....I will need to try again

scanpy/_utils/__init__.py
flying-sheep: then shouldn’t this code make sure that it’s a 1-element tuple?

scanpy/_utils/__init__.py
```diff
@@ -560,6 +560,43 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
     return da.map_blocks(elem_mul, x, y)
 
 
+@singledispatch
+def axis_sum(X: np.ndarray | sparse.spmatrix, axis=None, dtype=None):
```
flying-sheep:
```suggestion
def axis_sum(X: np.ndarray | sparse.spmatrix, *, axis=None, dtype=None):
```

scanpy/preprocessing/_normalization.py
```diff
@@ -187,22 +187,25 @@ def normalize_total(
 
     gene_subset = None
     msg = "normalizing counts per cell"
+
+    counts_per_cell = axis_sum(X, axis=1)
```
flying-sheep, line 213: this should be

```diff
+counts_per_cell = axis_sum(X, axis=1)
 if exclude_highly_expressed:
-    counts_per_cell = X.sum(1)  # original counts per cell
 ...
-else:
-    counts_per_cell = X.sum(1)
```

but you left the identical statements per branch, so it’s now done twice no matter the branch

scanpy/preprocessing/_utils.py
```diff
@@ -1,26 +1,40 @@
...
-from .._utils import AnyRandom, _SupportedArray, elem_mul
+from .._utils import AnyRandom, _MemoryArray, _SupportedArray, axis_sum, elem_mul
 
 if TYPE_CHECKING:
+    import dask.array as da
     from numpy.typing import NDArray
 
 
+@singledispatch
+def mean_func(X: da.Array, axis, dtype):
```
flying-sheep:
```suggestion
def axis_mean(X: da.Array, *, axis: Literal[0, 1], dtype: np.typing.DTypeLike):
```

scanpy/testing/_pytest/params.py
```diff
@@ -52,32 +55,27 @@ def param_with(
     ),
 }
 
-ARRAY_TYPES_MEM = tuple(
+ARRAY_TYPES_MEM = list(
```
flying-sheep: no, constants should have immutable types to avoid accidental mutation

scanpy/tests/test_highly_variable_genes.py
```diff
@@ -338,23 +338,26 @@ def test_pearson_residuals_batch(pbmc3k_parametrized_small, subset, n_top_genes)
...
 )
-def test_compare_to_upstream(
+@pytest.mark.parametrize("array_type", ARRAY_TYPES)
+def test_compare_to_upstream(  # noqa: PLR0917
     request: pytest.FixtureRequest,
     func: Literal["hvg", "fgd"],
     flavor: Literal["seurat", "cell_ranger"],
     params: dict[str, float | int],
     ref_path: Path,
+    array_type: list,
```
flying-sheep: 1. never use `list` as a parameter type (only return type) and
2. always parameterize collection types
3. `array_type` is a callable, not a list

scanpy/preprocessing/_normalization.py
ilan-gold: Fixed (hopefully)

scanpy/_utils/__init__.py
```diff
@@ -560,6 +560,156 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
     return da.map_blocks(elem_mul, x, y)
 
 
+@singledispatch
+def divide(
+    dividend: sparse.spmatrix,
+    divisor,
+    axis: Literal[0, 1],
```
ilan-gold, line 573: Hmmm, should this be inferred from the broadcast-ability of the two...not sure

scanpy/_utils/__init__.py
```diff
@@ -560,6 +560,156 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+    ...
+
+
+@singledispatch
+def axis_sum(
+    X: np.ndarray,
+    *,
+    axis: Optional[Union[tuple[Literal[0, 1], ...], Literal[0, 1]]] = None,  # noqa: UP007
+    dtype: Optional[np.typing.DTypeLike] = None,  # noqa: UP007
+) -> np.ndarray:
```
flying-sheep: Please remove those `noqa` things. You e.g. do `@singledispatch(np.ndarray)`, that’s less ugly.

flying-sheep (commented):
Looks great, except for one detail

scanpy/_utils/__init__.py
ilan-gold: Sorry, I don't understand, @flying-sheep this has to do with an `|` operator.  What does `@singledispatch(np.ndarray)` have to do with that?   Also can you call `singledispatch` directly on something like that?

flying-sheep: I assumed you blocked reformatting these lines because that interferes with singledispatch somehow.

Since singledispatch supports an alternative form of use, we should use that instead of making our type hints uglier.

If that wasn’t the reason, why *did* you add these `noqa` comments?

>  Also can you call singledispatch directly on something like that?

Yup, of course! I’m not trying to mislead you by adding broken code examples :wink:

ilan-gold: I am sorry if I am misunderstanding but

```python
from functools import singledispatch
@singledispatch(int)
def foo(a: int):
     return a
```
raises an error.  I also still don't quite see how this affects the `noqa`. The `noqa` comment out came from just before https://github.com/scverse/scanpy/pull/2856/commits/48a1a1e3c9412efb7d435a1f57a42e4e70aef649.  I was a little confused about why that was happening since the `|` operator works elsewhere with no complaints.  I think it had to do with the multiple generics in the union.

scanpy/_utils/__init__.py
```diff
@@ -560,6 +569,160 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+        )
+
+    import dask.array as da
+
+    scaling_array = broadcast_axis(scaling_array, axis)
+    row_scale = axis == 0
+    column_scale = axis == 1
+
+    if isinstance(scaling_array, DaskArray):
+        warnings.warn("Rechunking scaling_array in user operation", UserWarning)
```
ivirshup: Why does this warn?

scanpy/_utils/__init__.py
ivirshup: I believe this will get rid of the need for the noqa

```suggestion
@singledispatch
def axis_sum(
    X: np.ndarray,
    *,
    axis: tuple[Literal[0, 1], ...] | Literal[0, 1] | None = None,
    dtype: np.typing.DTypeLike | None = None,
) -> np.ndarray:
```

-----


@flying-sheep, `@singledispatch` is what registers the generic. You can't pass a type as argument. I believe it also shouldn't care about the annotations, that's just `.register`. That should also only effect the first argument.

scanpy/_utils/__init__.py
```diff
@@ -560,6 +569,160 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+        ):
+            kwargs.pop("keepdims", None)
+            axis = kwargs["axis"]
+            if isinstance(axis, tuple):
+                if len(axis) != 1:
+                    raise ValueError(
+                        f"`axis_sum` can only sum over one axis when `axis` arg is provided but got {axis} instead"
+                    )
+                kwargs["axis"] = axis[0]
+        # returns a matrix normally, which is undesireable?
```
ivirshup: Do you mean `np.matrix`? Yes that is undesirable.

scanpy/_utils/__init__.py
```diff
@@ -560,6 +569,160 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+    X: np.ndarray,
+    *,
+    axis: Optional[Union[tuple[Literal[0, 1], ...], Literal[0, 1]]] = None,  # noqa: UP007
+    dtype: Optional[np.typing.DTypeLike] = None,  # noqa: UP007
+) -> np.ndarray:
+    return np.sum(X, axis=axis, dtype=dtype)
+
+
+@axis_sum.register
+def _(
```
ivirshup, line 722: When `X._meta` is `np.ndarray`, could this just be `X.sum(axis=axis, dtype=dtype)`?

scanpy/tests/test_highly_variable_genes.py
ivirshup: Do you think this is related to:

* https://github.com/scipy/scipy/issues/20169#issuecomment-1978773475

?

If it is, then we sorta have a fix. We could also wait on upstream.

scanpy/_utils/__init__.py
```diff
@@ -560,6 +569,160 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+    row_scale = axis == 0
+    column_scale = axis == 1
+    # Same as `inplace_{column,row}_scale`, but not in place
+    if X.format == "csr":
+        if row_scale:
+            new_data = X.data * np.repeat(scaling_array, np.diff(X.indptr))
+        elif column_scale:
+            new_data = X.data * scaling_array.take(X.indices, mode="clip")
+        return sparse.csr_matrix((new_data, X.indices, X.indptr), shape=X.shape)
+    return axis_scale(X.T, scaling_array, (axis + 1) % 2).T
```
ivirshup:
```suggestion
    return axis_scale(X.T, scaling_array, axis=1-axis).T
```

scanpy/_utils/__init__.py
```diff
@@ -560,6 +569,160 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+
+    row_scale = axis == 0
+    column_scale = axis == 1
+    # Same as `inplace_{column,row}_scale`, but not in place
+    if X.format == "csr":
+        if row_scale:
+            new_data = X.data * np.repeat(scaling_array, np.diff(X.indptr))
+        elif column_scale:
+            new_data = X.data * scaling_array.take(X.indices, mode="clip")
+        return sparse.csr_matrix((new_data, X.indices, X.indptr), shape=X.shape)
```
ivirshup:
```suggestion
        return sparse.csr_matrix((new_data, X.indices.copy(), X.indptr.copy()), shape=X.shape)
```

We should probably be making copies of these if we're not doing the operation inplace.

scanpy/_utils/__init__.py
```diff
@@ -560,6 +569,160 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+
+@singledispatch
+def axis_scale(
+    X: sparse.spmatrix,
+    scaling_array,
+    axis: Literal[0, 1],
+    *,
+    out: sparse.spmatrix | None = None,
+) -> sparse.spmatrix:
+    if X is out:
```
ivirshup: This should probably error for all other cases where `out` is provided for now.

scanpy/preprocessing/_simple.py
```diff
@@ -859,14 +869,17 @@ def scale_array(
...
+        if issparse(X):
             raise ValueError("Cannot zero-center sparse matrix.")
-        sparsefuncs.inplace_column_scale(X, 1 / std)
-    else:
-        if zero_center:
-            X -= mean
-        X /= std
+        if isinstance(X, DaskArray) and issparse(X._meta):
+            warnings.warn(
+                "zero-center being used with `DaskArray` sparse chunks.  This can be bad if you have large chunks or intend to eventually read the whole data into memory."
```
ivirshup: I think this should either convert to dense or error.

I would go for erroring, since that's what we do for the in-memory case.

scanpy/preprocessing/_simple.py
```diff
@@ -1112,7 +1125,7 @@ def _downsample_per_cell(X, counts_per_cell, random_state, replace):
 
 def _downsample_total_counts(X, total_counts, random_state, replace):
     total_counts = int(total_counts)
-    total = X.sum()
+    total = axis_sum(X)
```
ivirshup:
```suggestion
    total = X.sum()
```

Since this won't work with dask arrays, can we leave it untouched for now?

scanpy/_utils/__init__.py
```diff
@@ -560,6 +569,160 @@ def _elem_mul_dask(x: DaskArray, y: DaskArray) -> DaskArray:
...
+    import dask.array as da
+
+    scaling_array = broadcast_axis(scaling_array, axis)
+    row_scale = axis == 0
+    column_scale = axis == 1
+
+    if isinstance(scaling_array, DaskArray):
+        warnings.warn("Rechunking scaling_array in user operation", UserWarning)
+        if row_scale and not X.chunksize[0] == scaling_array.chunksize[0]:
+            scaling_array = scaling_array.rechunk((X.chunksize[0], 1))
```
ivirshup: Should we be using `.chunks` instead of `chunksize` here? `chunks` can be irregular.

ivirshup (changes requested):
There is some unexpected computing going on:

```python
import scanpy as sc, dask.array as da, numpy as np, pandas as pd, anndata as ad, matplotlib.pyplot as plt

from scipy import sparse

X = da.random.poisson(.5, (100_000, 10_000), chunks = (50_000, 1_000))
X_sp = X.map_blocks(sparse.csr_matrix)

adata = ad.AnnData(X)
adata_sp = ad.AnnData(X_sp)
```

These are all computing something, while I think they should be lazy. When target_sum isn't provided, it's a little more understandable.

```python
sc.pp.normalize_total(adata)
sc.pp.normalize_total(adata_sp)
sc.pp.normalize_total(adata_sp, target_sum=1e4)
```

-----

I'm trying to figure out what the right API is here. Can we put all of these methods that you're making work with sparse dask arrays into their own module?

ilan-gold:
> There is some unexpected computing going on:

There are a number of `np.asarray` calls I did not touch.  I would guess that is what is going on.  I will try to remove them.

scanpy/_utils/__init__.py
ilan-gold: Hmmm I don't know if there is a way to rechunk irregularly though? https://docs.dask.org/en/latest/generated/dask.array.rechunk.html

scanpy/_utils/__init__.py
ilan-gold: @ivirshup I will try again, but I believe that is what I tried previously.  It's possible I erroneously altered the `dtype` one but it was an issue with python 3.9 only and I think the code I have might be necessary.  I will look again though

ilan-gold:
There are also a few conditionals that force some sort of evaluation over the whole frame like https://github.com/scverse/scanpy/blob/14555ba48537995acaa381b8b6ad5fc41e612510/scanpy/preprocessing/_normalization.py#L210

ilan-gold:
The other thing that is taking time is the call to [compute_chunk_sizes](https://github.com/scverse/scanpy/blob/14555ba48537995acaa381b8b6ad5fc41e612510/scanpy/preprocessing/_normalization.py#L26)

ilan-gold:
For the warning, I think we can just skip it for dask.  But there are indeed other places where these sorts of booleans occur, like [filter_genes](https://github.com/scverse/scanpy/blob/14555ba48537995acaa381b8b6ad5fc41e612510/scanpy/preprocessing/_simple.py#L292).   That being said, I think these can both be skipped for dask.

scanpy/preprocessing/_simple.py
ilan-gold: The reason I did not do this is because the matrix could be further reduced downstream.  With out-of-core processing, couldn't it be desirable to let this operation go through, especially if they end up using `X` in a way that reduces its dimensions considerably?

scanpy/_utils/__init__.py
ilan-gold: I think rechunking is a non-trivial operation but this also is not a public API.  So maybe it should be removed then?

scanpy/_utils/__init__.py
ilan-gold: What do you mean by this?  What are "all other cases?"

scanpy/tests/test_highly_variable_genes.py
ilan-gold: Yes!

scanpy/_utils/__init__.py
Intron7: Rechunking can have a big influence on performance

scanpy/_utils/__init__.py
ivirshup: Like this:

```python
import dask.array as da

a = da.ones(100, chunks=5)
a.chunks
# ((5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5, 5),)
a.rechunk(((50, 20, 30),)).chunks
# ((50, 20, 30),)
```

Also: 

* https://github.com/dask/dask/issues/10615

scanpy/preprocessing/_simple.py
ivirshup: I think it's okay to request a `map_blocks` to dense before that. I think it's bad to let people create a completley full sparse array. E.g. the error message could say:

"To enable zero centering, first densify the array like: `adata.X = adata.X.map_blocks(lambda x: x.toarray(), dtype=adata.X.dtype)`"

scanpy/_utils/__init__.py
ivirshup: `axis_scale(x, axis=1, out=y)` where `x is not y`

scanpy/_utils/__init__.py
ivirshup: Sure, but this is just warning whenever a dask array passes through, and isn't specific to the case where rechunking is necessary.

I believe rechunking generally shouldn't be necessary here, since the chunks should be aligned already (though it would be nice to enforce this).

```python
row_chunks = np.random.randint(10, 20, size=10)
X = da.random.poisson(size=(row_chunks.sum(), 100), chunks=(tuple(row_chunks), 50))

assert X.sum(axis=1).chunks[0] == X.chunks[0]
```

ilan-gold: Location of the warning, definitely my bad.  Other than that, it seems we're in agreement then?  I added the conditions so we don't just rechunk for no reason (only if the condition that the chunks are alighned, which in theory is always met anywaym is not met) and that's the enforcement.  Unless you think we should just error out?

scanpy/_utils/__init__.py
ilan-gold: Ah, right ok!

scanpy/preprocessing/_simple.py
ilan-gold: So this should error then?

scanpy/_utils/__init__.py
ilan-gold: A doozy, but hopefully got it right!

scanpy/_utils/__init__.py
ilan-gold: Ok yes https://dev.azure.com/scverse/scanpy/_build/results?buildId=5927&view=logs&j=5ea502cf-d418-510c-3b5f-c4ba606ae534&t=534778bb-2f86-5739-7d3c-59518f7b5a2b&l=31 there is definitely something bizarre going on!  But somehow calling `axis_sum.register(DaskArray)` fixes it?  Ok!

Intron7:
So we dont want to use `futures` and only `map_blocks`

ilan-gold:
@Intron7 I was under the impression @ivirshup had said he wanted this to be done with `map_blocks` and that we would work out the problems that made you use futures later.  But perhaps I misunderstood?  If I remember correctly, the use of futures was more out of not being able to get `map_blocks` to work with you data type, and not about some limitation of `map_blocks` as applied to your problems.

Intron7:
@ilan-gold, I believe `map_blocks` offers several advantages, the most significant being its simplicity. If you don't require `futures`, using `map_blocks` is likely the better choice. I'll delve more into this once I begin working on the multi-GPU version for RSC.

ivirshup:
My main reasoning was that:

* Use dask array/ delayed is more simple
* One can go from dask array/ dask delayed -> `Future` with `client.compute`, but the other way around isn't really possible since `Future`s kick off computation immediately.

So I'd like to use the higher level interface as much as possible, and only use lower level when necessary.

scanpy/preprocessing/_simple.py
ivirshup: I would go for erroring, since that's what we do for the in-memory case.

Intron7 (approved):

scanpy/preprocessing/_simple.py
ilan-gold: Actually, this is a bit of misdirection - we actually densify! https://github.com/scverse/scanpy/blob/14555ba48537995acaa381b8b6ad5fc41e612510/scanpy/preprocessing/_simple.py#L892-L898

ilan-gold: I am actually not sure why that `issparse` check is here even - this is in the dispatch only for `np.ndarray`!

ivirshup (commented):
## Partial review: concerning `normalize_total`

Sorry if this is a bit ramble-y, but I'm trying to figure out how to get normalization to where I'm happy with it.

@ilan-gold, I would consider `normalize_total` feature complete for dask when we can get it to be entirely lazy. I realize there are some barriers there, including:

* Adding an entry to `obs` (we could probably just stop doing this, or keep it and wait for next-gen `obs` dataframes)
* `np.all(cell_subset)`, if we enforce that `X` is chunked along observations then this can be detected within the chunk computation, and we could warn then.
* `exclude_highly_expressed`, especially the warning part. Tbh, I don't think I've seen this warning be thrown. Maybe this is a warning that should come up in `calculate_qc_metrics`. Maybe we can ignore this for now an only target completley delayed computation when `exclude_highly_expressed=None`.
* For calculating median counts: 

```python
counts_greater_than_zero = counts[counts > 0].compute_chunk_sizes()
after = np.median(counts_greater_than_zero, axis=0) if after is None else after
```

First, we should be skipping calculation of `counts_greater_than_zero` if `after` is provided.

Second, I think it should be possible to drop the need for `compute_chunk_sizes`, since we know we need to move the whole `counts` array to one process, then reduce it to a scalar. Maybe this is:

```python
masked = da.ma.masked_array(counts, counts==0)
after = da.median(masked, axis=0)
```

----

So that's maybe a lot, some of which is achievable easily, some of which isn't. Possible I am missing some things. It would be good to figure out scope for this PR, and also maybe go over this in a call or with a  white board.

I do suspect some of these problems may be best solved by splitting to a dask specific branch earlier in the function.

ivirshup:
From slack, @ilan-gold mentioned that the median calculation I suggested may not work (https://dask.discourse.group/t/median-with-masked-array-behaves-unexpectedly/70)

It could be replaced with something like:

```python
da.from_delayed(
    delayed(np.median)(
        delayed(counts[counts > 0])
    ),
    shape=(),
    meta=np.ndarray,
    dtype=counts.dtype,
)
```

Or probably better:

```python
def nonzero_median(x):
    return np.ma.median(np.ma.masked_array(x, x > 0))

da.from_delayed(
    delayed(nonzero_median)(counts),
    shape=(),
    meta=np.ndarray,
    dtype=counts.dtype,
)
```

ilan-gold:
@ivirshup I think this needs some tweaking...

```python
import dask.array as da
import numpy as np
import dask

X = da.random.poisson(2, (1_000, ), chunks = (100, ))
def nonzero_median(x):
    return np.ma.median(np.ma.masked_array(x, x > 0))
dask_median = da.from_delayed(dask.delayed(nonzero_median)(X), shape=(), meta=X._meta, dtype=X.dtype).compute()
np_median = np.median(X.compute()[X.compute() > 0])
np_ma_median = nonzero_median(X.compute())
assert np_median == dask_median
assert np_ma_median == dask_median
```

ilan-gold:
@ivirshup In the last few commits, I added some things to skip some unnecessary computation steps (which were only used for logging/warning).

Here's a quick example:

```python
import fsspec
import zarr
from anndata.experimental import sparse_dataset
from anndata import AnnData
import dask.array as da
from dask import delayed
from scipy import sparse


class AccessTrackingStore(zarr.LRUStoreCache):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)

    def __getitem__(self, key):
        if key not in self._values_cache:
            print(key)
        return super().__getitem__(key)

def csr_callable(shape: tuple[int, int], dtype) -> sparse.csr_matrix:
    if len(shape) == 0:
        shape = (0, 0)
    if len(shape) == 1:
        shape = (shape[0], 0)
    elif len(shape) == 2:
        pass
    else:
        raise ValueError(shape)

    return sparse.csr_matrix(shape, dtype=dtype)

class CSRCallable:
    """Dummy class to bypass dask checks"""
    def __new__(cls, shape, dtype):
        return csr_callable(shape, dtype)

def make_dask_chunk(x, start: int, end: int) -> da.Array:
    def take_slice(x, idx):
        return x[idx]

    return da.from_delayed(
        delayed(take_slice)(x, slice(start, end)),
        dtype=x.dtype,
        shape=(end - start, x.shape[1]),
        meta=CSRCallable,
    )

def sparse_dataset_as_dask(x, stride: int):
    n_chunks, rem = divmod(x.shape[0], stride)

    chunks = []
    cur_pos = 0
    for i in range(n_chunks):
        chunks.append(make_dask_chunk(x, cur_pos, cur_pos + stride))
        cur_pos += stride
    if rem:
        chunks.append(make_dask_chunk(x, cur_pos, x.shape[0]))

    return da.concatenate(chunks, axis=0)

def read_w_sparse_dask(group, obs_chunk: int = 1000) -> AnnData:
    return AnnData(
        X=sparse_dataset_as_dask(sparse_dataset(group["X"]), obs_chunk),
    )
```
After this setup:
```python
mapper = fsspec.get_mapper(
    "https://vitessce-demo-data.storage.googleapis.com/anndata-demos/BALF_VIB-UGent_processed_cleaned.zarr/"
)
store = AccessTrackingStore(mapper, max_size=2**28)
adata = read_w_sparse_dask(zarr.convenience.open_consolidated(store))
```
This takes a moment, but not too long.

Separately, to see that nothing is happening unexpected on these two operations:
```python
import scanpy as sc
sc.pp.normalize_total(adata)
sc.pp.log1p(adata)
```

The missing parts are
1. `filter_{genes,cells}`
2. `scale`
3. `highly_variable_genes`
4. `normalize_total` with `key_added`
5. `normalize_total` with `exclude_highly_expressed` (even minus the warning). 

The first four of these do some storage into `obs`/`var` of values, which is a blocker to being completely lazy because the values need to be computed for that as you pointed out.  We could of course just skip that step for dask for now.  Or keep it and iterate later.

Separately,  both `filter_{genes,cells}`and  `normalize_total` with `exclude_highly_expressed`  also do an subset operation for which we will need to allow `dask` indexing into `AnnData`, which could get hairy.

ilan-gold:
Ultimately I don't think these things should be blockers.  Lazy dataframes are now en-route with the xarray PR accepted in principle, and the subsetting operation is what it is.   I think it's ok to iterate on these things

ilan-gold:
Opened https://github.com/scverse/anndata/issues/1404

scanpy/preprocessing/_simple.py
```diff
@@ -172,7 +175,7 @@ def filter_cells(
     if max_number is not None:
         cell_subset = number_per_cell <= max_number
 
-    s = materialize_as_ndarray(np.sum(~cell_subset))
+    s = axis_sum(~cell_subset)
     if s > 0:
```
ilan-gold, line 180: Why are we allowing this `if` statement to trigger `compute`?  I think this was a mistake unless we landed on some reason to do this and not just skip the logging. @ivirshup

ilan-gold: Is this because there is another `materialize_as_ndarray` call above (L145)? I don't think that should necessarily mean we should also do one here in effect.

ilan-gold: Leaving as-is because these functions need to do `compute` anyway due to subsetting ops

docs/release-notes/1.11.0.md
```diff
@@ -2,6 +2,7 @@
 
 ```{rubric} Features
 ```
+* Support sparse chunks in dask {func}`~scanpy.pp.scale`, {func}`~scanpy.pp.filter_cells`, {func}`~scanpy.pp.filter_genes`, {func}`~scanpy.pp.normalize_total` and {func}`~scanpy.pp.highly_variable_genes` (`seurat` and `cell-ranger` tested) {pr}`2856` {smaller}`ilan-gold`
```
ivirshup:
```suggestion
* Support sparse chunks in dask {func}`~scanpy.pp.scale`, {func}`~scanpy.pp.normalize_total` and {func}`~scanpy.pp.highly_variable_genes` (`seurat` and `cell-ranger` tested) {pr}`2856` {smaller}`ilan-gold`
```

Should we remove those given our experiences today?

ilan-gold: Let me finish doing some investigating....I am not sure yet....

ilan-gold: Also it's possible that the sparse version will not have this problem so that note may be correct

ilan-gold: Let's remove then yes: https://github.com/scverse/anndata/pull/1432 just confirmed this solves the problem...so will depend on `anndata` release maybe then actually? If this goes in as patch release then this problem will solve itself...

ilan-gold:
pre-commit.ci autofix

ivirshup (approved):

PR #2857, opened 2024-02-15
Rename mask parameters

flying-sheep:
- [x] Closes N/A
- [x] Tests included

- [x] Release notes not necessary because: original PR not released

### DIFF
--- a/scanpy/preprocessing/_simple.py
+++ b/scanpy/preprocessing/_simple.py
@@ -746,7 +746,7 @@ def _regress_out_chunk(data):
 
 
 @renamed_arg("X", "data", pos_0=True)
-@old_positionals("zero_center", "max_value", "copy", "layer", "obsm", "mask")
+@old_positionals("zero_center", "max_value", "copy", "layer", "obsm")
 @singledispatch
 def scale(
     data: AnnData | spmatrix | np.ndarray,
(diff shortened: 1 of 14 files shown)


scanpy/preprocessing/_simple.py
```diff
@@ -746,7 +746,7 @@ def _regress_out_chunk(data):
 
 
 @renamed_arg("X", "data", pos_0=True)
-@old_positionals("zero_center", "max_value", "copy", "layer", "obsm", "mask")
```
ilan-gold, line 749: Just to be clear, was `mask` ever needed in `old_positionals` since we never released it as a positional?

scanpy/preprocessing/_simple.py
```diff
@@ -783,6 +783,8 @@ def scale(
         If provided, which element of layers to scale.
     obsm
         If provided, which element of obsm to scale.
+    mask_obs
+        TODO
```
ilan-gold: We should do the "DO" in "TODO!"

ilan-gold (approved):

scanpy/preprocessing/_simple.py
flying-sheep: Ah, forgot to comment on that. You guessed right, it wasn’t!

scanpy/preprocessing/_simple.py
flying-sheep: “DO” is done.

PR #2859, opened 2024-02-15
Convert string input on calculate qc metrics into collection for convenience

### LINKED ISSUE #2858: String conversion of `qc_vars` into a `Collection` when a `str` input is provided in `calculate_qc_metrics`

I often make the mistake of providing a single string input to the `qc_vars` argument in `calculate_qc_metrics` which always fails with a `KeyError` because it beings to iterate over the string. This could be a very simple addition to just convert the argument into a `list` if a `str` is provided as input.

noamteyssier:
- [x] Closes #2858 

- [x] Release notes not necessary because: this is a minor quality of life change, and could be tacked onto any other release as opposed to creating a dedicated release

### DIFF
--- a/scanpy/tests/test_qc_metrics.py
+++ b/scanpy/tests/test_qc_metrics.py
@@ -143,13 +143,10 @@ def test_qc_metrics_format(cls):
         assert np.allclose(adata.var[col], adata_dense.var[col])
 
 
-@pytest.mark.parametrize(
-    "cls", [np.asarray, sparse.csr_matrix, sparse.csc_matrix, sparse.coo_matrix]
-)
-def test_qc_metrics_format_str_qc_vars(cls):
+def test_qc_metrics_format_str_qc_vars():
     adata_dense, init_var = adata_mito()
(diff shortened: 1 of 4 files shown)


scanpy/tests/test_qc_metrics.py
```diff
@@ -143,6 +143,19 @@ def test_qc_metrics_format(cls):
         assert np.allclose(adata.var[col], adata_dense.var[col])
 
 
+@pytest.mark.parametrize(
+    "cls", [np.asarray, sparse.csr_matrix, sparse.csc_matrix, sparse.coo_matrix]
+)
```
ivirshup:
```suggestion
```

I think it's simple enough that we don't need to parametrize this test

ivirshup (commented):
LGTM, thanks for the contribution.

Could you add a release note?

noamteyssier:
Sounds good - added an entry to the release note and updated the tests to not use the parameterization.

ivirshup (approved):

ivirshup:
Thanks! I've moved the release note to 1.10, since this is more of a feature.

PR #2868, opened 2024-02-20
Allow blank issues

ivirshup:
Sometimes the templates just don't fit

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,8 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: Scanpy Community Forum
     url: https://discourse.scverse.org/
     about: If you have questions about “How to do X”, please ask them here.
+  - name: Blank issue
+    url: https://github.com/scverse/scanpy/issues/new
+    about: For things that don't quite fit elsewhere. Please note that other templates should be used in most cases – this is mainly for use by the developers.


flying-sheep:
in which cases?

ivirshup:
Mostly dev process and tracking issues

flying-sheep:
Which are done by us exclusively, so we can just delete the template.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,5 +1,8 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: Scanpy Community Forum
     url: https://discourse.scverse.org/
     about: If you have questions about “How to do X”, please ask them here.
+  - name: Blank issue
+    url: https://github.com/scverse/scanpy/issues/new
+    about: For things that don't quite fit elsewhere. Please note that other templates should be used in most cases – this is mainly for use by the developers.
```
flying-sheep, line 8:
```suggestion
```

flying-sheep (changes requested):
I’d also be OK to just do the first line of the diff, but leave out the ability for users to have a clickable thing.

Then we can just remove the `/choose` from the URL to get a blank form.

I rather press <kbd>Ctrl</kbd>+<kbd>⇐</kbd> → <kbd>⇐</kbd> → <kbd>⏎</kbd> than having to deal with a million blank issues with no reproducible example.

flying-sheep (approved):
OK, this is relatively hidden and not a green button, so it’s fine

PR #2874, opened 2024-02-21
Fix coverage ci

ivirshup:
CI runs that report coverage currently don't fail if the tests fail. This is because the way the coverage job is written swallows the error. I'm updating this use the same approach as anndata, which seems to be working.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -52,14 +52,14 @@ jobs:
 
   - script: |
       python -m pip install --upgrade pip
-      pip install wheel coverage
+      pip install wheel
       pip install .[dev,$(TEST_EXTRA)]
     displayName: 'Install dependencies'
     condition: eq(variables['DEPENDENCIES_VERSION'], 'latest')
 
   - script: |
       python -m pip install --pre --upgrade pip
-      pip install --pre wheel coverage
+      pip install --pre wheel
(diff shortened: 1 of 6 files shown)


flying-sheep:
Using `coverage run` is intentional. Please just configure the shell so it exits when a line fails instead of changing this.

The reason is at we have `addopts = ['-p', 'scanpy.testing.pytest']` (or so). If we use `pytest --cov`, coverage for things run during import breaks.

ivirshup:
> coverage for things run during import breaks.

I don't really understand what you mean here. Do you have an explanation somewhere?

flying-sheep:
This line:

https://github.com/scverse/scanpy/blob/383a61b2db0c45ba622f231f01d0e7546d99566b/pyproject.toml#L162

means that pytest imports that module together with the other plugins, and *then* runs tests.

That means that it will `import scanpy.testing._pytest` (i.e. `scanpy` and everthing imported in there) before pytest-cov is loaded and can do anything.

ivirshup:
So, we could also not early load `scanpy.testing._pytest` or load `pytest-cov` first?

I would like to keep the `xdist` support and use a similar interface.

flying-sheep:
This PR broke the doctest setup.

https://github.com/scverse/scanpy/pull/3057 fixes it again

PR #2875, opened 2024-02-21
Bugfix: Gene score edge case where gene_list gene is chosen as control gene

### LINKED ISSUE #2153: Score genes - control genes index of wrong format
- [ ] I have confirmed this bug exists on the latest version of scanpy.

---
We ran across a case where score genes does not work for one specific gene, but works for others; the gene is in anndata. The same code works on other datasets preprocessed in the same way.

```pytb
genes=['INS']
score_name='ins'
'INS' in adata_rawnorm.var_names
True
Input In [109], in <module>
----> 1 sc.tl.score_genes(adata_rawnorm, gene_list=genes, score_name=score_name+'_score',  use_raw=False)

File ~/miniconda3/envs/block/lib/python3.8/site-packages/scanpy/tools/_score_genes.py:167, in score_genes(adata, gene_list, ctrl_size, gene_pool, n_bins, score_name, random_state, copy, use_raw)
    164 else:
    165     X_list = np.nanmean(X_list, axis=1, dtype='float64')
--> 167 X_control = _adata[:, control_genes].X
    168 if issparse(X_control):
    169     X_control = np.array(_sparse_nanmean(X_control, axis=1)).flatten()

File ~/miniconda3/envs/block/lib/python3.8/site-packages/anndata/_core/anndata.py:625, in AnnData.X(self)
    622         X = _subset(X, (self._oidx, self._vidx))
    623 elif self.is_view:
    624     X = as_view(
--> 625         _subset(self._adata_ref.X, (self._oidx, self._vidx)),
    626         ElementRef(self, "X"),
    627     )
    628 else:
    629     X = self._X

File ~/miniconda3/envs/block/lib/python3.8/functools.py:875, in singledispatch.<locals>.wrapper(*args, **kw)
    871 if not args:
    872     raise TypeError(f'{funcname} requires at least '
    873                     '1 positional argument')
--> 875 return dispatch(args[0].__class__)(*args, **kw)

File ~/miniconda3/envs/block/lib/python3.8/site-packages/anndata/_core/index.py:127, in _subset(a, subset_idx)
    125 if all(isinstance(x, cabc.Iterable) for x in subset_idx):
    126     subset_idx = np.ix_(*subset_idx)
--> 127 return a[subset_idx]

IndexError: arrays used as indices must be of integer (or boolean) type
```

mumichae:
In some edge cases, the control gene selection retrieves the same gene(s) that are also in the gene_list used for scoring.
As a result, when the following line is called, we end up with an empty control gene set, causing the downstream error in #2153
https://github.com/scverse/scanpy/blob/383a61b2db0c45ba622f231f01d0e7546d99566b/scanpy/tools/_score_genes.py#L173

- [x] Closes #2153 
- [x] Tests included


mumichae:
Tests are failing, I believe, because the gene_list genes are removed from control genes before random sampling, not after, resulting in a different control gene set. Not quite sure why the difference in scores is so large though.

ivirshup:
@flying-sheep, I have never been too familiar with the `score_genes` code. Could you take a look at this?

flying-sheep:
There’s two differences:
1. `np.unique(...)` returns the values sorted, `pd.Series(...).unique()` returns them in original order (this already makes the scores not match)

   This probably changes the sampling, but I wonder why the score difference is so large here! With only that change, I get:

   > Arrays are not equal
   >
   > Mismatched elements: 2730 / 2730 (100%)
   > Max absolute difference: 0.22674037
   > Max relative difference: 1581.75673912

2. what you said: The original approach samples from the full list of genes in each bin, then restricts the sample to valid ones. Your approach samples from the valid genes in each bin.

   So if a bin e.g. contains mostly invalid genes, the original code adds only a few genes for that bin, while yours adds the maximum possible number.

   So the questions is: is the sampling bias introduced in the original code wanted? If not, you not only made the code more resilient, but also more objective.

scanpy/tools/_score_genes.py
```diff
@@ -155,12 +155,24 @@ def score_genes(
...
+                f"ctrl_size={ctrl_size} to sample from more control genes."
+            )
         if ctrl_size < len(r_genes):
             r_genes = r_genes.to_series().sample(ctrl_size).index
         control_genes = control_genes.union(r_genes.difference(gene_list))
 
+    assert len(control_genes) > 0, "No control genes found."
+    assert not any(
+        g in control_genes for g in gene_list
+    ), "Genes are in both gene_list and control_genes."
```
flying-sheep: `assert` is only for internal checks. For each of these conditions that can be triggered by user supplied data (which seems to be the case in the test you added), you shouldn’t use `assert`.

Instead, you should use `ValueError` (or `TypeError`) if a faulty parameter is passed in, and `RuntimeError` for when something failed that shouldn’t (i.e. when randomness or so resulted in the error).

Am I correct that the first one is driven by randomness and the second is indeed an internal check? Then you should change the first into a RuntimeError and leave the second an assert.

flying-sheep (changes requested):

scanpy/tools/_score_genes.py
```diff
@@ -155,12 +155,24 @@ def score_genes(
...
+            (obs_cut == cut) & ~obs_cut.index.isin(gene_list)
+        ].index
+        if len(r_genes) == 0:
+            logg.warning(
+                f"No control genes for this cut. You might want to increase "
+                f"ctrl_size={ctrl_size} to sample from more control genes."
+            )
         if ctrl_size < len(r_genes):
             r_genes = r_genes.to_series().sample(ctrl_size).index
         control_genes = control_genes.union(r_genes.difference(gene_list))
```
mumichae: I don't see which part of the [R code](https://github.com/satijalab/seurat/blob/c54e57d3423b3f711ccd463e14965cc8de86c31b/R/utilities.R#L280C3-L303) is equivalent to removing the genes to be scored.

```python
r_genes.difference(gene_list)
```

From what I can tell, if one of the genes to be scored happens to be chosen as the background, it will be included in the calculation

scanpy/tools/_score_genes.py
mumichae: From what I can tell, the first assert will only fail if we don't have enough cuts and we happen to only select the genes from `gene_list` (i.e. the genes we want to score), so I don't think this is something that will happen due to randomness (at least not with the current changes). So I wouldn't say that it's a result of randomness

mumichae:
To answer the following question:

>  2. what you said: The original approach samples from the full list of genes in each bin, then restricts the sample to valid ones. Your approach samples from the valid genes in each bin.
>
>     So if a bin e.g. contains mostly invalid genes, the original code adds only a few genes for that bin, while yours adds the maximum possible number.
>
>     So the questions is: is the sampling bias introduced in the original code wanted? If not, you not only made the code more resilient, but also more objective.

After going through the original [code from Seurat](https://github.com/satijalab/seurat/blob/c54e57d3423b3f711ccd463e14965cc8de86c31b/R/utilities.R#L280C3-L303), it seems to me that there's not equivalent to removing genes to be scored from the control gene set.
From what I can tell, if one of the genes to be scored happens to be chosen as the background, it will be included in the calculation.
But please correct me if that's not the case.

So if the original implementation does not remove score genes from the control gene set, we would simply need to remove the following line: https://github.com/scverse/scanpy/blob/ec4457470618efd85da3c7b29f951cab01a49e3a/scanpy/tools/_score_genes.py#L169

(Note: if we want to keep the current behaviour, we should still remove the line above, since it would be redundant)

scanpy/tools/_score_genes.py
```diff
@@ -155,12 +155,24 @@ def score_genes(
...
-        r_genes: pd.Index[str] = obs_cut[obs_cut == cut].index
+    for cut in obs_cut.loc[gene_list].unique():
+        r_genes: pd.Index[str] = obs_cut[
+            (obs_cut == cut) & ~obs_cut.index.isin(gene_list)
+        ].index
+        if len(r_genes) == 0:
+            logg.warning(
+                f"No control genes for this cut. You might want to increase "
+                f"ctrl_size={ctrl_size} to sample from more control genes."
+            )
```
flying-sheep: What do you mean here? `ctrl_size` only comes into play in the next line, so if the cut is empty, it can’t be fixed by increasing `ctrl_size`, right?

scanpy/tools/_score_genes.py
flying-sheep: > So if the original implementation does not remove score genes from the control gene set, we would simply need to remove the following line:
> https://github.com/scverse/scanpy/blob/ec4457470618efd85da3c7b29f951cab01a49e3a/scanpy/tools/_score_genes.py#L169
> (Note: if we want to keep the current behaviour, we should still remove the line above, since it would be redundant)

Removing that line completely wouldn’t make sense, then `control_genes` would be empty. But with your `& ~obs_cut.index.isin(gene_list)` change, a part of the line currently redundant, so we could do this without changing behavior:

```suggestion
        control_genes = control_genes.union(r_genes)
```

is that what you meant?

flying-sheep (commented):
Sorry for taking so long to look at this.

> From what I can tell, if one of the genes to be scored happens to be chosen as the background, it will be included in the calculation.

So you’re saying that the seurat code neither does `& ~obs_cut.index.isin(gene_list)` nor 
`r_genes.difference(gene_list)`.

Therefore if we’re going with Seurat compat, we’d end up with difference scores than the failing reference test as well, so the reference doesn’t actually reflect any kind of gold standard, only what we did before at some point?

If I’m right, we should probably add an option to get the new better behavior you propose, and in Scanpy 2.0 we’d switch to that by default.

scanpy/tools/_score_genes.py
mumichae: That's correct, but from what I can see, having an empty cut is likely due to an edge case that the user might not want. setting an extremely small control size will leave you with a single gene (or very few genes), which will not really characterise a background distribution. Increasing the control size will make it less likely to get an empty cut

scanpy/tools/_score_genes.py
mumichae: ah yes, you're right, it's just the difference that should be redundant

mumichae:
About question 1: I find also it strange that the order of cuts that you iterate though should have such a large effect on sampling. Should I change it back to `np.unique`?

scanpy/tools/_score_genes.py
flying-sheep: The condition for the warning being shown has nothing to do with `ctrl_size`, yet the warning talks about `ctrl_size`. Therefore following the recommendation in the warning (increasing `ctrl_size`) won’t ever be able to make the warning go away, which is confusing.

Ideally, the warning would only recommend doing a thing if doing the thing has the potential to make the warning go away. Is that possible?

scanpy/tools/_score_genes.py
flying-sheep: Then we should throw a `RuntimeError` here instead of doing an assertion.

As said, `assert`ions mean “this assertion will pass unless we made a false assumption in our code, i.e. scanpy has a bug”

scanpy/tools/_score_genes.py
flying-sheep: OK, I’ll ask that in the data science channel

scanpy/tools/_score_genes.py
mumichae: ah, I see what you mean 😅 You're right, I think the warning isn't necessary actually. The cut can only be empty if the user provides too many genes for scoring and there aren't many control genes left. Do you think it's worth putting a warning for that case or would you prefer no warning at all?

flying-sheep: If we can detect a problem and show a warning that actually has an actionable suggestion, that’s always a good idea.

src/scanpy/tools/_score_genes.py
```diff
@@ -161,14 +165,29 @@ def get_subset(genes: pd.Index[str]):
 
     n_items = int(np.round(len(obs_avg) / (n_bins - 1)))
     obs_cut = obs_avg.rank(method="min") // n_items
-    control_genes = pd.Index([], dtype="string")
+    obs_cut_is_ctrl = False if ctrl_as_ref else obs_cut.index.isin(gene_list)
```
mumichae: Not sure if I fully understand the naming here. I might go with something like `keep_ctrl_in_obs_cut`

mumichae (commented):
LGTM, I like, that we keep the old way of running! Just a question on the naming.

src/scanpy/tools/_score_genes.py
flying-sheep: great name, much better!

PR #2889, opened 2024-02-23
Fix for #2887, update Leiden and Louvain tools to write parameters to user specified key

### LINKED ISSUE #2887: Leiden and louvain params not saved to matching `key_added` key in `uns` dictionary

### What happened?

Leiden and Louvain clustering params are not saved to matching `key_added` key in `uns` dictionary but are ovewritten to hardcoded key instead. One use case is that a user may want to run Leiden/Louvain clustering multiple times with different resolutions / parameters. One may specify different keys to store results under. However, if you do so, the metadata for the parameterization of the clustering algorithms are overwritten because the lines below do not respect the user provided `key_added` parameter.

I think the desired behavior is to store data under `adata.uns[key_added]["params"]`

I think I've found the pertinent lines below. Happy to submit a PR if maintainers agree :D.
- https://github.com/scverse/scanpy/blob/91ea0fbb03392795d1506d297d4b4847c646db04/scanpy/tools/_leiden.py#L206
- https://github.com/scverse/scanpy/blob/91ea0fbb03392795d1506d297d4b4847c646db04/scanpy/tools/_louvain.py#L259

### Minimal code sample

```python
sc.tl.leiden(adata, resolution=0.8, key_added="leiden_0.8")
assert "leiden_0.8" not in adata.uns
params = adata.uns["leiden"] 

sc.tl.leiden(adata, resolution=1.2, key_added="leiden_1.2")
assert "leiden_1.2" not in adata.uns
overwritten_params = adata.uns["leiden"] 
assert params == overwritten_params # should fail
```

### Error output

_No response_

### Versions

<details>
Confirmed that params are overwritten in source in main branch. (see permalinks)
</details>

theJasonFan:
Fix for #2887. Updated and added basic tests to Leiden and Louvain clustering to check that the parameters are written to the user specified key.

- [x] Closes #2887 
- [x] Tests included

- [ ] Release notes not necessary because


scanpy/tests/test_clustering.py
```diff
@@ -186,6 +189,22 @@ def test_louvain_basic(adata_neighbors):
...
+    sc.tl.louvain(
+        adata_neighbors,
+        key_added="louvain_custom",
+        random_state=random_state,
+        resolution=resolution,
+    )
+    assert (
+        adata_neighbors.uns["louvain_custom"]["params"]["random_state"] == random_state
+    )
+    assert adata_neighbors.uns["louvain_custom"]["params"]["resolution"] == resolution
```
theJasonFan, line 205: I'm don't completely understand the details of how codecov works, but I think it fails to pick up this test. Apologies if I'm doing something wrong!

ilan-gold:
@ivirshup Can we stick this in `1.10`?  It seems small enough and IMO basically constitutes a breaking change anyway.  then @theJasonFan can add a release note and we can merge if we don't notice anything else wrong here.

scanpy/tests/test_clustering.py
ilan-gold: No worries, codecov can be flaky.  Can we paramterize this test to do both `leiden` and `louvain` and then also run it twice with two or three (including default) different keys so as to check that we're not doing overriding unintentionally?

theJasonFan: Good idea. I'll add that.

ilan-gold (approved):

PR #2896, opened 2024-03-05
remove self for pynndescent

Intron7:
This removes `self` for `pynndescent` in `scrublet` and makes it reproducible for smaller datasets. I also added a test param that represent this. I also think this might help with larger datasets to remove self from the computation.

### DIFF
--- a/scanpy/preprocessing/_scrublet/core.py
+++ b/scanpy/preprocessing/_scrublet/core.py
@@ -360,7 +360,8 @@ class Scrublet:
             random_state=self._random_state,
         )
         neighbors, _ = _get_indices_distances_from_sparse_matrix(knn.distances, k_adj)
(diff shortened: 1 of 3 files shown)


scanpy/preprocessing/_scrublet/core.py
```diff
@@ -360,7 +360,8 @@ def _nearest_neighbor_classifier(
             random_state=self._random_state,
         )
         neighbors, _ = _get_indices_distances_from_sparse_matrix(knn.distances, k_adj)
-
+        if use_approx_nn:
+            neighbors = neighbors = neighbors[:, 1:]
```
flying-sheep:
```suggestion
            neighbors = neighbors[:, 1:]
```

PR #2900, opened 2024-03-12
adds rsc

Intron7:
Adds Rapids-singlecell to the ecosystem page and formats the decoupler note to be in line with the other notes

### DIFF
--- a/docs/ecosystem.md
+++ b/docs/ecosystem.md
@@ -112,7 +112,15 @@ Interactive manifold viewers.
 
 Analyses using curated prior knowledge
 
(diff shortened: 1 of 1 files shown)


Zethson:
Actually, I think we are deprecating this page in favor of the scverse ecosystem registry https://github.com/scverse/ecosystem-packages. 

It also says so:

> We are no longer accepting new tools on this page.
Instead, please submit your tool to the [scverse ecosystem package listing](https://scverse.org/packages/#ecosystem).

@flying-sheep would you be okay with actually killing the whole content of this page and referring on it to https://scverse.org/packages/#ecosystem?

PR #2901, opened 2024-03-12
Doc submodule

### LINKED ISSUE #2636: Go back to tutorial submodule

Other?

The tutorial submodule was removed as part of a fix-up pass over the docs by @flying-sheep.

The issue at the time was warnings which weren't immediately obvious how to fix.

Since there are a number of reasons to prefer the submodule approach over a separate scanpy-tutorials website, I would like to bring it back.

The plan is: bring back the submodule. In this PR I'll see if the warnings can be easily fixed. If they can't, they'll be silenced and an issue tracking fixing them for the next release will be opened.

ivirshup:
Overhaul of tutorials page

Tutorials can now be versioned with the docs by using a gitsubmodule of the tutorials repo (letting us avoid things like: https://github.com/scverse/scanpy-tutorials/issues/85)

Possible additions, that can be left to later

* It's a lot more bare now, maybe we should use cards? I wasn't sure if there's a great way to do this while keeping overall organization.
* Some tutorials that have been removed may be worth keeping
  * 1.3 million mouse neurons
  * Cell cycle regression


ivirshup:
@flying-sheep, docs are building now if you want to take a look

docs/tutorials/basics/clustering-2017.ipynb
flying-sheep, line 1: What are these? symlinks?

Why don’t we adapt the structure of the tutorials repo to how things are supposed to be?

docs/_tutorials.md
flying-sheep, line 1: is there a redirect fot this in place?

docs/references.bib
```diff
@@ -0,0 +1,100 @@
...
+  volume    = {8},
+  number    = {1},
+  author    = {Grace X. Y. Zheng and Jessica M. Terry and Phillip Belgrader and Paul Ryvkin and Zachary W. Bent and Ryan Wilson and Solongo B. Ziraldo and Tobias D. Wheeler and Geoff P. McDermott and Junjie Zhu and Mark T. Gregory and Joe Shuga and Luz Montesclaros and Jason G. Underwood and Donald A. Masquelier and Stefanie Y. Nishimura and Michael Schnall-Levin and Paul W. Wyatt and Christopher M. Hindson and Rajiv Bharadwaj and Alexander Wong and Kevin D. Ness and Lan W. Beppu and H. Joachim Deeg and Christopher McFarland and Keith R. Loeb and William J. Valente and Nolan G. Ericson and Emily A. Stevens and Jerald P. Radich and Tarjei S. Mikkelsen and Benjamin J. Hindson and Jason H. Bielas},
+  title     = {Massively parallel digital transcriptional profiling of single cells},
+  journal   = {Nature Communications}
+}
+@article{stuart2019comprehensive,
+  title     = {Comprehensive integration of single-cell data},
+  author    = {Stuart, Tim and Butler, Andrew and Hoffman, Paul and Hafemeister, Christoph and Papalexi, Efthymia and Mauck, William M and Hao, Yuhan and Stoeckius, Marlon and Smibert, Peter and Satija, Rahul},
```
flying-sheep, line 83: Seems like the name formats aren’t consistent. The order is neither. Maybe we should run [Bibtool](https://github.com/ge-ne/bibtool) on it to make sure things are formatted consistently and errors can be spotted?

Running it in CI would be better, but it doesn’t seem to have releases unbundled from TeXLive, which might be a bit big to install in CI.

docs/how-to/index.md
```diff
@@ -0,0 +1,7 @@
+# How to
+
+This section contains short examples on how to perform specific tasks with scanpy.
```
flying-sheep, line 3: So there’s where that went!

docs/references.bib
ivirshup: I don't feel strongly about this, and am happy to take whatever cite key is given so long as it doesn't conflict.

I do feel strongly that I don't want to have a manual task beyond "is that info right" for bibtex entries.

docs/tutorials/basics/clustering-2017.ipynb
ivirshup: This gives us a lot more flexibility in how we structure the docs. E.g. this makes it easy to have the "How To" section.

The main reason we have a notebooks repo is that executed notebooks are big. For notebooks that we can execute on the fly, they can just live directly in this repo.

Then, why should the URL for a tutorial be based on how we can render it?

docs/_tutorials.md
ivirshup: No, but this  doesn't work on the stable docs:

https://scanpy.readthedocs.io/en/stable/_tutorials.html

docs/tutorials/basics/clustering-2017.ipynb
flying-sheep: What do you think about this:

1. we move all notebooks into the notebook repo
2. we configure its CI to run the small ones on the fly in PRs (so no broken notebooks get merged in)
3. In the scanpy repo, we could run them in CI again (as tests)

docs/references.bib
flying-sheep: Oh, I’m not talking about cite keys, but `author` names. Bibtex has some bespoke format here and I’d like to make sure we store them in a way that we understand and can verify.

That’s why I proposed bibtool: If all names are in the same format, we can skim new entries and check if they look correct.

docs/tutorials/basics/clustering-2017.ipynb
ivirshup: > we configure its CI to run the small ones on the fly in PRs (so no broken notebooks get merged in)

I'd rather not have to make sure the configs + CI are kept in sync. I think doing most of the CI on this side should be fine.

Plus if a broken PR merges there, it wouldn't effect the doc build here since the submodule should be pointing to a previous commit.

I would like to have "paired" PRs like we do for the template repo.

docs/references.bib
ivirshup: But we can't run `bibtool` in pre-commit due to deps, right?

ivirshup: Can you do the bib tool stuff on another branch?

docs/tutorials/basics/clustering-2017.ipynb
flying-sheep: So for each PR that modifies one of the notebooks that are run in CI here, we need to manually make sure that there’s a paired PR that points to the same commit and makes sure CI runs here? Or we risk “breaking” CI here (read: having to debug things later when maybe multiple breakages have accumulated)?

I’m … not a fan. We have working CI for each commit in the other repo. Sure, notebooks are never run and we need to fix that, but there will never be breakage when pointing scanpy to a new commit there. I like that about our current solution, let’s keep that aspect.

ivirshup: > we need to manually make sure that there’s a paired PR that points to the same commit and makes sure CI runs here?

We automated this for the template repo, couldn't we do the same here?

flying-sheep: We could, sure. It's just much more complicated than just running the CI on one side and then using the artifacts from successful CI runs (which is what we do now by linking them).

flying-sheep (changes requested):
Seems like you accidentally force-pushed over my commits, please re-add: https://github.com/scverse/scanpy/compare/c1d042aaf0a12ecc2da4ee5af290fcfca6bb53b1..b39ce21e57e6ffcfd70a6531a9b105642df0d86f

pyproject.toml
```diff
@@ -122,6 +122,7 @@ doc = [
     "nbsphinx>=0.9",
     "ipython>=7.20",  # for nbsphinx code highlighting
     "matplotlib!=3.6.1",
+    "sphinxcontrib-bibtex",
```
flying-sheep, line 125:
```suggestion
    "sphinxcontrib-bibtex",
    "setuptools",  # https://bitbucket.org/pybtex-devs/pybtex/issues/169
```

docs/news.md
```diff
@@ -7,6 +7,10 @@
 
 ```
 
+### `rapids-singlecell` brings scanpy to the GPU! {small}`2024-03-18`
+
+[`rapids-singlecell`](https://rapids-singlecell.readthedocs.io/en/latest/) by Severin Dicks provides a scanpy-like API with accelerated operations implemented on GPU.
```
flying-sheep: Please don’t use inline links

```suggestion
[`rapids-singlecell`][] by Severin Dicks provides a scanpy-like API with accelerated operations implemented on GPU.

[`rapids-singlecell`]: https://rapids-singlecell.readthedocs.io/en/latest/
```

docs/tutorials/basics/index.md
flying-sheep, line 1: Should this page (and `experimental/index.md`) exist? `tutorials/index.md` already renders full toc trees of these sections, so the sub-pages just contain duplicate information.

flying-sheep (changes requested):
The structure looks good, two main things:

1. The tutorial site looks a bit messy, with toc trees, paragraphs, and images interspersed. If you e.g. remove the sub-indices, put the text together and use a sphinx-design grid for the images, you’d get:

   grafik


3. See https://github.com/scverse/scanpy/pull/2901#pullrequestreview-1949433456

   Bibtool should run either in CI or at least locally.

docs/tutorials/basics/index.md
ivirshup: I believe these are required so we get the full TOC on the higher pages

docs/news.md
ivirshup: Why?

ivirshup:
> Seems like you accidentally force-pushed over my commits, please re-add:

This was intentional as those changes broke the doc build. sphinxcontrib-bibtex couldn't parse the reformatted bibtex file. Over [here](https://github.com/scverse/scanpy/pull/2901#discussion_r1530244562) I requested that if you want to add bib tool working as a precommit, I would like you to do it in a separate branch.

pyproject.toml
ivirshup: I'm not sure I understand this request. I get that there is some upstream bug causing failures on someone else's CI, but we don't seem to be triggering that here?

docs/news.md
flying-sheep: Don’t do inline links, they make text less readable.

ivirshup: I'm not sure I agree. Maybe for really long links. But: 

* there's no effect on readability when rendered
* you have to think about link targets and whether they are already defined
* I don't know if these are handled by sphinx, but we've had issues in the past with sphinx defining reference targets globally and having subtle changes or typos mean we link to the wrong thing. It's harder to get inline links wrong.
* There are 22 links on this page currently, 20 are inline.

docs/tutorials/basics/index.md
ivirshup: If there is a better way, I'm up for it. But I couldn't figure it out.

ivirshup: I would like to have something like what squidpy has here: 
https://squidpy.readthedocs.io/en/latest/notebooks/examples/index.html#

However, I'm not sure if there's a difference in configuration or the pydata theme just doesn't allow this with the way it splits the TOC to either side of the screen (which I am coming to think is actually bad and difficult to use, it just looks nice)

pyproject.toml
flying-sheep: Our CI environment comes with setuptools preinstalled, but that’s not a given. `pip install scanpy[doc]` will only be guaranteed to install all runtime dependencies if we add it here.

docs/news.md
flying-sheep: That’s horrible, we should fix that. Did the MyST conversion do that? I’m getting less and less happy with that as time goes on.

I’m OK with inline links in release notes, but tool/library names like this should definitely be named.

I’d also accept (and in fact probably prefer) an intersphinx link here.

docs/tutorials/basics/index.md
flying-sheep: I don’t understand why the split TOC is supposed to mandate that, could you explain?

Also squidpy has the `notebooks` submodule under the `docs` folder, we should do that too.

docs/news.md
flying-sheep: Intersphinx:

```rst
:doc:`rapids_singlecell:index`
```

docs/tutorials/basics/index.md
ivirshup: * https://github.com/executablebooks/sphinx-book-theme/issues/337

pyproject.toml
```diff
@@ -176,7 +177,6 @@ filterwarnings = [
     "error:The specified parameters:FutureWarning",
     # When calling `.show()` in tests, this is raised
     "ignore:FigureCanvasAgg is non-interactive:UserWarning",
-
```
flying-sheep: I put that blank line there intentionally, because you got confused one time and added a ignore filter in the middle of the error block.

docs/tutorials/basics/index.md
flying-sheep: So we would need to override `sphinx_book_theme/components/sbt-sidebar-nav.html` somehow? They don’t document how to do that.

flying-sheep:
OK, now the only thing missing is to move the notebooks directory into `docs` and to make sure the bibtex isn’t a mess.

You said you don’t want to do that manually, so please add some way to do it automatically.

> sphinxcontrib-bibtex couldn't parse the reformatted bibtex file. 

hmm, so we can’t trust its parser to work properly, that’s a bit concerning. But we control what it sees so that isn’t too bad.

ivirshup:
> OK, now the only thing missing is to move the notebooks directory into docs 

We have already discussed this, and it is staying as is for now. We can think about doing this in the future, but please make your case in an issue.

> bibtex isn’t a mess.

I will be leaving the bibtex as is, I don't see an issue with the bibtex as it renders fine. As I requested last week, you can suggest a formatter for it in a separate PR. Notably this bibtex is copied directly from scverse-tutorials, which doesn't seem to have caused any issues.

FWIW there are some tools for working with bibtex recommended here:

* https://github.com/ge-ne/bibtool/issues/58
  * bibtex-tidy has a pre-commit that doesn't do exactly what you're looking for
  * python-bibtexparser seems like it'd be able to do what you want, and is far more actively maintained than bibtool.

PR #2921, opened 2024-03-15
Add layer argument to score_genes()

### LINKED ISSUE #2910: Add layer argument to `score_genes()`/`score_genes_cell_cycle()`

These functions are some of the few to still not have this argument. Would also help with #2909.

lazappi:
- [x] Closes #2910 


scanpy/tools/_score_genes.py
```diff
@@ -134,10 +137,19 @@ def score_genes(
...
-    _adata = adata.raw if use_raw else adata
+    _adata = adata
+    if layer is not None:
+        if use_raw:
+            raise ValueError("Cannot specify `layer` and have `use_raw=True`.")
+        _adata.X = _adata.layers[layer]
+    else:
+        if use_raw and adata.raw is not None:
+            _adata = adata.raw
+
```
ivirshup: I believe this line:

```
_adata.X = _adata.layers[layer]
```

Will mutate the input, which we'd like to avoid.

Could this use `scanpy.get._get_obs_rep`?

I *think* this should just be:

```suggestion
        X = _get_obs_rep(adata, use_raw=use_raw, layer=layer)
        if use_raw:
            gene_pool_idx = adata.raw.var_names.get_indexer(gene_pool)
        else:
            gene_pool_idx = adata.var_names.get_indexer(gene_pool)
```

Then use gene_pool_idx to do `X[:, gene+pool_idx]` instead of `_adata_subset`?

scanpy/tests/test_score_genes.py
```diff
@@ -228,6 +228,18 @@ def test_use_raw_None():
...
+    adata = _create_adata(100, 1000, p_zero=0, p_nan=0)
+
+    sc.pp.normalize_per_cell(adata, counts_per_cell_after=1e4)
+    sc.pp.log1p(adata)
+    adata.layers["test"] = adata.X.copy()
+
+    gene_set = adata.var_names[:10]
+    sc.tl.score_genes(adata, gene_set, score_name="Test", layer="test")
+    assert adata.obs["Test"].dtype == "float64"
+
```
ivirshup: Can the test be:

Run once with an object that has `X` (`adata_X`), once with an object that has the values in a `layer` (`adata_layer`), then check that the resulting value is the same?

It should also be the case that `not np.array_equal(adata_X.X, adata_layer.X)` so that we can be sure the right element was accessed.

ivirshup (commented):
Thanks for the PR! A few requests

scanpy/tools/_score_genes.py
lazappi: Maybe??? I copied code from another function but maybe that's also out of date. I'm not sure exactly what those internal functions do but I can try something like that. Is it fine to start by accepting this suggestion or was it more of a sketch?

ivirshup: More of a sketch

lazappi:
I had a go a making these changes but I didn't have time to get the tests working locally so there is a fair chance I broke something in the process

flying-sheep:
I’m so confused

```pycon
>>> adata[:, :1].X
ArrayView([[-0.53187567],
           [-0.53187567],
           [-0.01111901],
           ...,
           [-0.53187567],
           [ 1.7015172 ],
           [-0.53187567]], dtype=float32)
>>> adata[:, :2].X
ArrayView([[-0.53187567, -0.38556716],
           [-0.53187567, -0.38556716],
           [-0.01111901, -0.38556716],
           ...,
           [-0.53187567, -0.38556716],
           [ 1.7015172 , -0.38556716],
           [-0.53187567, -0.38556716]], dtype=float32)
>>> adata[:, :1].X.mean(axis=0)
array([2.515185e-08], dtype=float32)
>>> adata[:, :2].X.mean(axis=0)
array([4.3688239e-08, 1.7302814e-07], dtype=float32)
```

why the fuck is the second-to-last line not `array([4.3688239e-08])`?

Similarly:

```pycon
>>> np.nanmean(adata[:, var_names].X, axis=0)
array([ 2.5151850e-08,  4.1919752e-09,  1.7466563e-08, ...,
       -1.8863888e-08,  1.1178600e-08,  1.5370576e-08], dtype=float32)
>>> np.nanmean(adata.X, axis=0)
array([ 4.3688239e-08,  1.7302814e-07,  2.3729962e-08, ...,
        1.4421923e-07, -1.0466837e-07,  4.0828091e-08], dtype=float32)
>>> (adata.var_names == var_names).all()
True
```

Is `mean` bugged when used with certain `ArrayView`s? Or is that change in behavior considered an acceptable precision thing.

flying-sheep:
OK, I made it work, but the above still confuses me. @lazappi you basically did the following, which because of the above precision issue changed the results:

```diff
-_adata_subset = _adata[:, gene_pool] if len(gene_pool) < len(_adata.var_names) else _adata
+_adata_subset = _adata[:, gene_pool]
```

I re-introduced the check here, which restores perfect compatibility.

One thing I noted: There’s a potential (preexisting) bug if `gene_pool` is a shuffled `var_names`, because then `pd.Series(adata.X.means(axis=0), index=gene_pool)` has the values assigned to the wrong genes.

lazappi:
Thanks for checking and the tidying/fixes! I would never have worked out all the optimisations/tricks. I guess I removed that check because it seemed unnecessary but I guess it was there for this precisions thing.

It looks like the remaining failing checks are due to unrelated things (Dask etc.) but I'll let you work that out unless there is something else you would like me to do.

flying-sheep:
Thanks for the PR, that’s it. We know about the two broken tests (e.g. see #3068), they can be ignored for this PR.

Not all changes I made in #3097 were necessary, just re-adding the `if len(gene_pool) < len(var_names)` branch.

I made the refactoring PR since your changes already refactored the function in a good way, I just went a bit further so the repeated code for “get row/col means of a gene subset of `adata`” code could be reused, and your `get_indexer` change would only need to be added in a single location.

PR #2922, opened 2024-03-16
Fixed Coifman 2005 DOI link.

n-hackert:
Very small change: I noticed the Coifman 2005 reference is not linking to the right paper while going through the docs and added the correct link.

(It ain't much but it's honest work.)


flying-sheep:
Thanks! It *is* honest work haha

PR #2923, opened 2024-03-18
(fix): use `delayed` median function in dask normalization

ilan-gold:
- [ ] Closes #


scanpy/preprocessing/_normalization.py
```diff
@@ -207,7 +225,7 @@ def normalize_total(
     counts_per_cell = np.ravel(counts_per_cell)
 
     cell_subset = counts_per_cell > 0
-    if not np.all(cell_subset):
+    if not isinstance(cell_subset, DaskArray) and not np.all(cell_subset):
```
ilan-gold, line 228: I think it's fine to leave this in so that normalization without `exclude_highly_expressed` just works

ilan-gold:
Not sure what's up with the test - is it flaky? If not, I can look into it

maybe https://github.com/scverse/scanpy/pull/2889/commits/4bc1c48bee697bc520720723bf3033dd621544fe?

PR #2942, opened 2024-03-22
updates sparse scale

### LINKED ISSUE #2491: sc.pp.scale() doesn't support adata.X being a dask array

---

In short `sc.pp.scale()` throws an error when `adata.X` is a dask array solely because `sc.pp._simple.scale` doesn't have the `sc.pp._simple.scale_array` function registered for a `dask.array.Array` type. Adding `@scale.register(da.Array)` [here](https://github.com/scverse/scanpy/blob/f03c5b407412de447480733a1a1a0e33e0c871d2/scanpy/preprocessing/_simple.py#LL758C7-L758C7) would fix that, but considering `dask.array` is [only an optional dependency,](https://github.com/scverse/scanpy/blob/f03c5b407412de447480733a1a1a0e33e0c871d2/scanpy/preprocessing/_simple.py#L32) there would have to be a conditional to wrap the function decoration.

This brings me to the larger issue, is scanpy supposed to or working toward supporting dask arrays completely? I'm new to scanpy and I'm not sure if this is a bug report or an enhancement request. I could submit a pull request for this one issue but I'm curious if I'll run into many such issues as I dive in further and trying to figure out if there's existing momentum in this direction or whether I should be following some other parallelization strategy.

I got here by following [AnnData's guide on using dask and zarr](https://anndata.readthedocs.io/en/latest/tutorials/notebooks/%7Bread%2Cwrite%7D_dispatched.html) to try to parallelize processing of a large scRNA-seq file. I come from a microscopy data analysis and ML background where my image data is stored in S3 hosted zarr arrays (in an OME-NGFF schema), handled by using dask arrays wrapped in xarray DataArrays, and parallelized across compute using ray. It would be nice to use a similar stack (dropping in anndata for xarray) for sc-seq analyses.

```python
import zarr
import anndata as ad
import dask.array as da
import scanpy as sc

# write data to zarr file
rel_zarr_path = 'data/pbmc3k_processed.zarr'
adata.write_zarr(f'./{rel_zarr_path}', chunks=[adata.shape[0], 5])
zarr.consolidate_metadata(f'./{rel_zarr_path}')

# read data from zarr file with X as a dask array
def read_dask(store):
    f = zarr.open(store, mode="r")

    def callback(func, elem_name: str, elem, iospec):
        if iospec.encoding_type in (
            "dataframe",
            "csr_matrix",
            "csc_matrix",
            "awkward-array",
        ):
            # Preventing recursing inside of these types
            return ad.experimental.read_elem(elem)
        elif iospec.encoding_type == "array":
            return da.from_zarr(elem)
        else:
            return func(elem)

    adata = ad.experimental.read_dispatched(f, callback=callback)

    return adata
adata_dask = read_dask(f'./{rel_zarr_path}')

# perform preprocessing
sc.pp.log1p(adata)
# sc.pp.highly_variable_genes(adata, min_mean=0.0125, max_mean=3, min_disp=0.5)
sc.pp.scale(adata_dask, max_value=10)
```

```pytb
Cell In[1], line 39
     37 sc.pp.log1p(adata)
     38 # sc.pp.highly_variable_genes(adata, min_mean=0.0125, max_mean=3, min_disp=0.5)
---> 39 sc.pp.scale(adata_dask, max_value=10)

File ~/miniconda3/envs/omics/lib/python3.10/functools.py:889, in singledispatch..wrapper(*args, **kw)
    885 if not args:
    886     raise TypeError(f'{funcname} requires at least '
    887                     '1 positional argument')
--> 889 return dispatch(args[0].__class__)(*args, **kw)

File ~/miniconda3/envs/omics/lib/python3.10/site-packages/scanpy/preprocessing/_simple.py:844, in scale_anndata(adata, zero_center, max_value, copy, layer, obsm)
    842 view_to_actual(adata)
    843 X = _get_obs_rep(adata, layer=layer, obsm=obsm)
--> 844 X, adata.var["mean"], adata.var["std"] = scale(
    845     X,
    846     zero_center=zero_center,
    847     max_value=max_value,
    848     copy=False,  # because a copy has already been made, if it were to be made
    849     return_mean_std=True,
    850 )
    851 _set_obs_rep(adata, X, layer=layer, obsm=obsm)
    852 if copy:

File ~/miniconda3/envs/omics/lib/python3.10/functools.py:889, in singledispatch..wrapper(*args, **kw)
    885 if not args:
    886     raise TypeError(f'{funcname} requires at least '
    887                     '1 positional argument')
--> 889 return dispatch(args[0].__class__)(*args, **kw)

TypeError: scale() got an unexpected keyword argument 'return_mean_std'
```

#### Versions

<details>

-----
anndata     0.9.1
scanpy      1.9.3
-----
PIL                 9.5.0
appnope             0.1.3
asciitree           NA
asttokens           NA
attr                23.1.0
backcall            0.2.0
brotli              NA
certifi             2023.05.07
cffi                1.15.1
charset_normalizer  3.1.0
click               8.1.3
cloudpickle         2.2.1
colorama            0.4.6
colorful            0.5.5
colorful_orig       0.5.5
comm                0.1.3
cycler              0.10.0
cython_runtime      NA
cytoolz             0.12.0
dask                2023.5.0
dateutil            2.8.2
debugpy             1.6.7
decorator           5.1.1
defusedxml          0.7.1
distributed         2023.5.0
entrypoints         0.4
executing           1.2.0
fasteners           0.17.3
filelock            3.12.0
fsspec              2023.5.0
google              NA
grpc                1.43.0
h5py                3.8.0
idna                3.4
igraph              0.10.4
ipykernel           6.23.1
jedi                0.18.2
jinja2              3.1.2
joblib              1.2.0
jsonschema          4.17.3
kiwisolver          1.4.4
leidenalg           0.9.1
llvmlite            0.40.0
locket              NA
lz4                 4.3.2
markupsafe          2.1.2
matplotlib          3.6.3
matplotlib_inline   0.1.6
mpl_toolkits        NA
msgpack             1.0.5
natsort             8.3.1
numba               0.57.0
numcodecs           0.11.0
numpy               1.24.3
packaging           23.1
pandas              2.0.1
parso               0.8.3
patsy               0.5.3
pexpect             4.8.0
pickleshare         0.7.5
pkg_resources       NA
platformdirs        3.5.1
prometheus_client   NA
prompt_toolkit      3.0.38
psutil              5.9.5
ptyprocess          0.7.0
pure_eval           0.2.2
pvectorc            NA
pyarrow             9.0.0
pycparser           2.21
pydev_ipython       NA
pydevconsole        NA
pydevd              2.9.5
pydevd_file_utils   NA
pydevd_plugins      NA
pydevd_tracing      NA
pygments            2.15.1
pynndescent         0.5.10
pynvml              NA
pyparsing           3.0.9
pyrsistent          NA
pytz                2023.3
ray                 2.3.0
rb_analysis         NA
requests            2.29.0
scipy               1.10.1
seaborn             0.12.2
session_info        1.0.0
setproctitle        1.2.2
setuptools          67.7.2
six                 1.16.0
sklearn             1.2.2
socks               1.7.1
sortedcontainers    2.4.0
stack_data          0.6.2
statsmodels         0.14.0
tblib               1.7.0
texttable           1.6.7
threadpoolctl       3.1.0
tlz                 0.12.0
toolz               0.12.0
tornado             6.3.2
tqdm                4.65.0
traitlets           5.9.0
typing_extensions   NA
umap                0.5.3
urllib3             1.26.15
wcwidth             0.2.6
yaml                6.0
zarr                2.14.2
zict                3.0.0
zipp                NA
zmq                 25.0.2
zoneinfo            NA
-----
IPython             8.13.2
jupyter_client      8.2.0
jupyter_core        5.3.0
-----
Python 3.10.11 | packaged by conda-forge | (main, May 10 2023, 19:01:19) [Clang 14.0.6 ]
macOS-13.3.1-arm64-arm-64bit
-----
Session information updated at 2023-05-18 14:00

</details>

### LINKED ISSUE #2941: scale for sparse matrixes and mask

currently `pp.scale` with a `mask_obs` with a sparse matrix and with `zero_center== False` takes a really long time to update the sparse matrix. This also takes up a lot of memory because of the parity calculations. I would suggest a numba kernel that just swaps out the data. This works really well for rapids-singlecell and greatly improves performance and reduces the memory overhead.
I would open a PR with this kernel.

------
Performance for 90k cells and 25k genes:
without mask:
CPU 645 ms | GPU 37 ms | 20x
with mask:
CPU 22 s | GPU 50 ms | 460x

Intron7:
Fixes #2941, fixes #2491

I created some numba.njit() kernels that perform in-place substitutions based on the assumption that we only change existing values and don't add new ones (where all the scipy overhead comes from). 

Benchmarks for 90k cells and 25k genes:
CSR:
old 23 s | new 1 s | 23x
CSC:
old 61 s | 1.6 s | 36x


Intron7:
It seems like I run into segfaults with the csc implementation. I could not reproduce these with on my PC. I run the tests a lot of times. Since the csr-kernel seems to be working and csr is anyway better for row-based subsets I would propose that we transform the `csc` into a `csr` and than return that for `mask_obs`.

docs/release-notes/1.10.1.md
```diff
@@ -0,0 +1,5 @@
+### 1.10.1 {small}`the future`
+
+```{rubric} Features
```
ivirshup: Could this go under "Performance"? We don't do "features" in bugfix releases

ivirshup (commented):
Have you looked at the implementation of `inplace_column_scale`? WDYT about just scaling by one when the value is masked out

Intron7:
That could also work. But this would require a bit of a rewrite. I think the current solution is simpler and also really fast.

ivirshup:
My thinking on this right now is that:

* The code for masking logic (pre this PR) is kind of a mess
* This PR doesn't make the code nicer

But the performance benefit is quite good, and for sure the operation `X[mask_obs, :] = scale_rv` is something we don't want to do with sparse matrices.

I also think we could get even faster, plus a bit cleaner if we instead modified scale array to use something like what I suggest [here](https://github.com/scipy/scipy/issues/20169#issuecomment-1973335172) to accept a `row_mask` argument:

```python
from scipy import sparse
import numpy as np
from operator import mul, truediv

def broadcast_csr_by_vec(X, vec, op, axis):
    if axis == 0:
        new_data = op(X.data, np.repeat(vec, np.diff(X.indptr)))
    elif axis == 1:
        new_data = op(X.data, vec.take(X.indices, mode="clip"))
    return X._with_data(new_data)
```

Which *I think* would be something like:

```python
def broadcast_csr_by_vec(X, vec, op, axis, row_mask: None | np.ndarray):
    if row_mask is not None:
        vec = np.where(row_mask, vec, 1)
    if axis == 0:
        new_data = op(X.data, np.repeat(vec, np.diff(X.indptr)))
    elif axis == 1:
        new_data = op(X.data, vec.take(X.indices, mode="clip"))
    return X._with_data(new_data)
```

Or, since we're doing numba already we could do just write out the operation with a check to see if we're on a masked row (which *should* be even faster since we're not allocating anything extra).

I think either of these solutions would be simpler since we do the masking all in one place, and don't have to have a second update step.

Intron7:
I really like your idea. But I feed like this will complecate the clipping function aswell since we need to subset there aswell than.

ivirshup:
> clipping function

I think it's not so bad. I think you can use similar logic. Numpy version is something like:

```python
    if axis == 0:
        data_mask = np.repeat(row_mask, np.diff(X.indptr))
    elif axis == 1:
        data_mask = obs_mask.take(X.indices, mode="clip")
    X.data[(X.data > max_value) & data_mask] = max_value
```

right?

For numba, I'd just include the clipping in the inner loop so it's still single pass.

Intron7:
I would be open to doing this is you move the whole sparse part out of scale array and than keeping it in scale sparse

ivirshup:
> I would be open to doing this is you move the whole sparse part out of scale array and than keeping it in scale sparse

Absolutely. The sparse definition being in scale_array is a big part of what makes the current code a mess

Intron7:
@ivirshup I don't know where the crash in the build is coming from I change nothing in those parts. However I rewrote the sparse logic and to me it's now a lot better.

ivirshup:
I think it was a bugged `uv` release, but seems to be fixed now.

Intron7:
Ok looks good now

scanpy/preprocessing/_simple.py
```diff
@@ -929,15 +931,65 @@ def scale_sparse(
...
+        )
+    elif isspmatrix_csc(X) and mask_obs is None:
+        return scale_array(
+            X,
+            zero_center=zero_center,
+            copy=copy,
+            max_value=max_value,
+            return_mean_std=return_mean_std,
+            mask_obs=mask_obs,
+        )
```
ivirshup: I'm a little confused by what this branch ends up doing. Does this not hit the numba kernel?

Btw, *I think* you can have:

```python
if mask is None:
```

blocks in numba code where the entire block is just excluded from the compiled code since it's a compile time constant.

Intron7: No that doesn't hit the numba kernel. For CSC matrix I doesnt make sense to make sense to hit the numba kernel. If we use a map obs we need to transform into csr for the mask-obs subset for the `mean-var-calc`. I think what you suggest is fair that we also use the base implementation for none mask csr.

> blocks in numba code where the entire block is just excluded from the compiled code since it's a compile time constant.

I'll look into this and adjust the code based on this

Intron7: So I updated the kernel to use 2 `compile time constant`.

scanpy/preprocessing/_simple.py
```diff
@@ -929,15 +931,75 @@ def scale_sparse(
...
+            has_mask = False
+    mean, var = _get_mean_var(X[mask_obs, :])
+
+    std = np.sqrt(var)
+    std[std == 0] = 1
+
+    @numba.njit()
+    def _scale_sparse_numba(indptr, indices, data, *, std, mask_obs, has_mask, clip):
+        def _loop_scale(cell_ix):
+            for j in numba.prange(indptr[cell_ix], indptr[cell_ix + 1]):
```
ivirshup: Is this multithreaded if you don't pass `parallel=True` to `njit`?

Intron7: its slower, because of the way the memory access happens. I tested it. So no its not multi-threaded

Intron7: Ok after running some more tests. Its not the memory access but the compile time. The speedup only happens for very large matrices in the 2nd run so I dont think its worth it.

Intron7:
Screenshot from 2024-04-04 12-00-02

I have some concerns about the performance of the no numba version for larger datasets. So it might be better to either switch to the numba kernel for larger datasets or take the compile hit for small datasets

ivirshup:
>  So it might be better to either switch to the numba kernel for larger datasets or take the compile hit for small datasets

The compiled versions should get cached, so it's a one time cost per install. No?

ivirshup:
@Intron7, could you open an issue with the follow up things you wanted to investigate here?

I think this is good to merge as it gets ~ a 100x speed up, and we can do comparisons on top of this in follow ups.

ivirshup (approved):

PR #2953, opened 2024-03-25
Fix links

ivirshup:
Some links have been broken in the index page

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,4 +17,4 @@ Contributing code
 -----------------
 
 We love code contributions!
-If you're interested in contributing code, please take a look over the [contributing guide](https://scanpy.readthedocs.io/en/latest/dev/index.html) in the main documentation.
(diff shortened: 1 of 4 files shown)


flying-sheep (changes requested):
That’s the opposite of what I asked for

ivirshup:
These are all one off links that are currently broken in the build and have to be compatible with github markdown since it's in the repo readme, so inter-sphinx is not an option.

flying-sheep:
Because they’re supposed to look like this: https://spec.commonmark.org/0.30/#collapsed-reference-link

Please fix them by using standard markdown instead of making them inline.

flying-sheep:
This should do the trick, without relying on something as fragile as line numbers.

Unless MyST somehow breaks include by converting things to rST before executing the include directive or something silly like that.

/edit: hahaha of course it does:

```
/home/docs/checkouts/readthedocs.org/user_builds/icb-scanpy/checkouts/2953/docs/index.md:1: WARNING: 'include': Invalid option value for 'end-before': None: argument required but none supplied [myst.directive_parse]
```

why? how?

flying-sheep (approved):

PR #2972, opened 2024-04-01
Fix double dot

Zethson:
image

- [x] Release notes not necessary because: Skipping release notes because no one cares about this


ivirshup:
No release notes are needed, but milestones are needed for this to get back-ported.

Also the check for the release notes is actually checking the box in the PR template.

Issue #2969, opened 2024-03-28
leiden alg with igraph flavor causes out of bounds freezing


RubenVanEsch:
### Please make sure these conditions are met

- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the latest version of scanpy.
- [X] (optional) I have confirmed this bug exists on the main branch of scanpy.

### What happened?

was running the standard pipeline on some data and when i run
`sc.tl.leiden(em_adata,flavor='igraph',n_iterations=2,random_state=1653,directed=False) `
it spits out infinite lines of ignored exceptions. it does not actually crash the kernel, but does bog it down and causes everything to to take much more time than necesarry. 
I am working in a conda env on a Win 10 , 64bit, x64 system
the problem also occurs using the pbmc3k dataset

### Minimal code sample

```python
# example with own data, but same happens with pbmc3k data
sc.pp.filter_cells(em_adata, min_genes=200)
sc.pp.filter_genes(em_adata, min_cells=3)
em_adata.shape
# [out] -> (42753, 21636)

sc.pp.calculate_qc_metrics(em_adata, qc_vars=["mt"], percent_top=None, log1p=False, inplace=True)
em_adata.obs["outlier_mt"] = em_adata.obs.pct_counts_mt > 15
em_adata.obs["outlier_total"] = em_adata.obs.total_counts > 30000
em_adata.obs["outlier_ngenes"] = em_adata.obs.n_genes_by_counts > 6000
em_adata = em_adata[~em_adata.obs["outlier_mt"], :]
em_adata = em_adata[~em_adata.obs["outlier_total"], :]
em_adata = em_adata[~em_adata.obs["outlier_ngenes"], :]
sc.pp.filter_genes(em_adata,min_cells=1)

sc.pp.scrublet(em_adata)
em_adata.layers['counts'] = em_adata.X.copy()
sc.pp.normalize_total(em_adata)
sc.pp.log1p(em_adata)
sc.pp.highly_variable_genes(em_adata,flavor='seurat')
sc.pl.highly_variable_genes(em_adata)
em_adata = em_adata[:, em_adata.var["highly_variable"]]
em_adata.shape
# [out] -> (41749, 1425)
sc.pp.pca(em_adata, n_comps=50)
sc.pp.neighbors(em_adata)
sc.tl.umap(em_adata)
sc.tl.leiden(em_adata,flavor='igraph',n_iterations=2,random_state=1653,directed=False)
```

### Error output

[...]

PR #2976, opened 2024-04-03
(chore): run azure on multiple vm images

ilan-gold:
- [ ] Testing for #2969

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -9,10 +9,19 @@ variables:
   DEPENDENCIES_VERSION: "latest"  # |"pre-release" | "minimum-version"
   TEST_TYPE: "standard" # | "coverage"
 
+strategy:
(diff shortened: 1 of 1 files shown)


ivirshup (commented):
For the pearson residual tests, that's an easy fix. `np.dtype("int")` is `np.int32` on windows, and `np.int64` everywhere else. So the tests should just be `is np.int64`.

`test_leiden_random_state[igraph]` does seem to be showing that the random state is at least acting differently here.

ilan-gold:
@ivirshup yup, commented that.  we need to find someone with a windows machine, I think.

ivirshup:
From discussion at triage/ sprint planning:

* We're not sure we want to support windows that much
* We'll leave this untouched for now, and probably won't merge until we also have an official stance on support level for windows

PR #2482, opened 2023-05-05
Benchmarking support with asv

rahulbshrestha:
This PR introduces benchmarking for Scanpy with airspeed-velocity (asv) ([Link](https://asv.readthedocs.io/en/stable/index.html))

PR #2977, opened 2024-04-03
Add benchmarks

flying-sheep:
- [ ] Closes #

Builds on and supersedes #2482


flying-sheep:
pre-commit.ci autofix

benchmarks/asv.conf.json
```diff
@@ -81,7 +81,7 @@
         "zarr": [""],
         "pytest": [""],
         "scanpy": [""],
-        "igraph": [""],
+        "igraph-python": [""],
```
ivirshup: Below, this is written `python-igraph`. Is that important?

flying-sheep: turns out yes, you seem to have commented on an old commit where I haven’t fixed that yet.

`python-igraph` is correct for conda. `python-igraph` also used to be the old PyPI name which is now deprecated.

benchmarks/benchmarks/readwrite.py
```diff
@@ -0,0 +1,101 @@
...
+
+* Read time, write time
+* Peak memory during io
+* File sizes
+
+Parameterized by:
+
+* What method is being used
+* What data is being included
+* Size of data being used
```
ilan-gold: Are any of these parametrized?

ilan-gold: Or is this just a wish-list?

benchmarks/benchmarks/readwrite.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def track_read_full_memratio(self, *_):
+        mem_recording = memory_usage(
+            (sedate(anndata.read_h5ad, seconds=0.005), (self.path,)), interval=0.001
+        )
+
+        base_size = mem_recording[-1] - mem_recording[0]
+        print(np.max(mem_recording) - np.min(mem_recording))
+        print(base_size)
+        return (np.max(mem_recording) - np.min(mem_recording)) / base_size
```
ilan-gold: Why a print here?

benchmarks/benchmarks/tools.py
```diff
@@ -0,0 +1,54 @@
...
+        sc.tl.diffmap(self.adata)
+
+    def peakmem_diffmap(self, *_):
+        sc.tl.diffmap(self.adata)
+
+    # def time_embedding_density(self, *_):
+    #     sc.tl.embedding_density(self.adata, groupby="batch")
+
+    # def peakmem_embedding_density(self, *_):
+    #     sc.tl.embedding_density(self.adata, groupby="batch")
```
ilan-gold: ?

benchmarks/benchmarks/tools.py
```diff
@@ -0,0 +1,54 @@
...
+    #     sc.tl.paga(self.adata, groups='leiden')
+
+    # def peakmem_paga(self, *_):
+    #     sc.tl.paga(self.adata, groups='leiden')
+
+    # def time_draw_graph(self, *_):
+    #     sc.tl.draw_graph(self.adata)
+
+    # def peakmem_draw_graph(self, *_):
+    #     sc.tl.draw_graph(self.adata)
```
ilan-gold: ?

benchmarks/benchmarks/readwrite.py
```diff
@@ -0,0 +1,101 @@
...
+        adata = anndata.read_h5ad(self.path)
+        adata_size = sys.getsizeof(adata)
+
+        return adata_size
+
+    def track_actual_in_memory_size(self, *_):
+        adata = anndata.read_h5ad(self.path)
+        adata_size = get_actualsize(adata)
+
+        return adata_size
```
ilan-gold: Why are we benchmarking anndata i/o here?

benchmarks/benchmarks/utils.py
```diff
@@ -0,0 +1,166 @@
...
+        for obj in objects:
+            if id(obj) not in ids:
+                ids.add(id(obj))
+                memory_size += sys.getsizeof(obj)
+                new.append(obj)
+        objects = gc.get_referents(*new)
+    return memory_size
+
+
+def get_anndata_memsize(adata: AnnData) -> float:
```
ilan-gold: Is this function used anywhere?

benchmarks/benchmarks/utils.py
```diff
@@ -0,0 +1,166 @@
+from __future__ import annotations
```
ilan-gold, line 1: Is this file copy-pasted (mostly) from the anndata repo?  A lot of these functions don't appear to be used at all.

ilan-gold: For example, I don't see anywhere `take_view` is used outside of its declaration in this file.

ilan-gold (dismissed):

benchmarks/benchmarks/readwrite.py
flying-sheep: we just have the dataset “parametrized”, and we only specify one. I feel like that is more like an example of how parametrization looks like too. I could just simplify it to not use a parameter.

benchmarks/benchmarks/readwrite.py
flying-sheep: good point, removed

benchmarks/benchmarks/readwrite.py
flying-sheep: great point. seems like this stuff has just been copied over

benchmarks/benchmarks/utils.py
flying-sheep: no, I’ll remove it.

Issue #2982, opened 2024-04-07
Suggestion to support VisiumHD tissue_position_list files (using parquet files)


Rafael-Silva-Oliveira:
### What kind of feature would you like to request?

Additional function parameters / changed functionality / changed defaults?

### Please describe your wishes

Hey

VisiumHD will now be the main data type for 10X technology on spatial data, and due to the higher number of barcodes used, we can't use a normal .csv file anymore due to row limitations. Instead, they're now using .parquet files, but on scanpy when using the `read_visium` method, there doesn't seem to have support for that.

Lines of code: 

```
    if load_images:
        files = dict(
            tissue_positions_file=path / 'spatial/tissue_positions_list.csv',
            scalefactors_json_file=path / 'spatial/scalefactors_json.json',
            hires_image=path / 'spatial/tissue_hires_image.png',
            lowres_image=path / 'spatial/tissue_lowres_image.png',
        )


```

I would suggest something like this to check if there's a csv file or parquet file:

```
if load_images:
		files = dict(
			tissue_positions_file = next((path / f'spatial/tissue_positions_list{suffix}' for suffix in ['.csv', '.parquet'] if (path / f'spatial/tissue_positions_list{suffix}').exists()), None),
			scalefactors_json_file=path / 'spatial/scalefactors_json.json',
			hires_image=path / 'spatial/tissue_hires_image.png',
			lowres_image=path / 'spatial/tissue_lowres_image.png',
		)

		if files['tissue_positions_file'].suffix == '.csv':
			positions = pd.read_csv(files['tissue_positions_file'], header=None)
		elif files['tissue_positions_file'].suffix == '.parquet':
			positions = pd.read_parquet(files['tissue_positions_file'])

```

This way we can use `read_visium` if the tissue location file is .parquet instead:

```
AnnData object with n_obs × n_vars = 605471 × 18085
    obs: 'in_tissue', 'array_row', 'array_col'
    var: 'gene_ids', 'feature_types', 'genome'
    uns: 'spatial'
    obsm: 'spatial'
```

Using scanpy 1.9.6

111-dep:
Hello, I'd like to ask which version of scanpy you're using? The code you provided doesn't match the latest version, which is causing errors when I try to use it.

Rafael-Silva-Oliveira:
> Hello, I'd like to ask which version of scanpy you're using? The code you provided doesn't match the latest version, which is causing errors when I try to use it.

Hi, I'm using 1.9.6

[...]

PR #2992, opened 2024-04-09
parquet support for visium HD

### LINKED ISSUE #2973: add support for Visium HD Spatial Gene Expression data

10X updated to Space Ranger v3.0, and the output files differ from the v2.0 version. Scanpy is unable to read the output files from v3.0. Can you add support for Space Ranger v3.0?

ktpolanski:
- [x] Closes #2973 

Visium HD stores its coordinates in a `.parquet` file. This loads said file.

### DIFF
--- a/scanpy/readwrite.py
+++ b/scanpy/readwrite.py
@@ -432,6 +432,7 @@ def read_visium(
         tissue_positions_file = (
             path / "spatial/tissue_positions.csv"
             if (path / "spatial/tissue_positions.csv").exists()
+            else path / "spatial/tissue_positions.parquet" if (path / "spatial/tissue_positions.parquet").exists()
             else path / "spatial/tissue_positions_list.csv"
         )
         files = dict(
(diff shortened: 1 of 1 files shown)


Rafael-Silva-Oliveira:
Also refer to https://github.com/scverse/scanpy/issues/2982

ktpolanski:
We ran some data through spaceranger 3.0.1 locally, and in doing so found that 10X have reduced redundancy in the `spatial` folder of the binned outputs by moving the tissue images to a new, central location. This understandably breaks the existing loader.

A hotfix is to copy the images back into the appropriate subdirectory, but that's not a feasible expectation on users. I added an optional argument `spaceranger_image_path` to point to the new folder if need be, which should hopefully be robust with regard to any sort of further restructuring 10X may choose to do in the future.

The code is currently included in [bin2cell](https://github.com/Teichlab/bin2cell) in case anybody needs it or just wants to take it out for a spin, but I think it belongs in a more central location.

giovp:
hi @ktpolanski ,thank you for the heads up

> A hotfix is to copy the images back into the appropriate subdirectory, but that's not a feasible expectation on users. I added an optional argument spaceranger_image_path to point to the new folder if need be, which should hopefully be robust with regard to any sort of further restructuring 10X may choose to do in the future.

> The code is currently included in [bin2cell](https://github.com/Teichlab/bin2cell) in case anybody needs it or just wants to take it out for a spin, but I think it belongs in a more central location.

@LucaMarconato do we have any datasets that test for spaceranger 3.0.1 ?

LucaMarconato:
Last week together with Wouter we identified some good candidate datasets (some of them recommended by 10x). I believe it's 3.0.0 not 3.0.1 though. The list is in the top part of my comment here https://github.com/scverse/spatialdata/issues/700#issuecomment-2326790050

ktpolanski:
I've ran some more spaceranger on my end and noticed that it actually makes symlinks to the new centralised spacial images in all the binned output folders. Symlinks don't tend to hold up well to things like rsync (which is how the original lot vanished, inspiring the new argument), so I think it still makes sense to keep the extra argument in the interest of accessibility.

flying-sheep:
Since spatial APIs in scanpy are now deprecated in favor of spatialdata-io and squidpy, I’m closing this. Thank you for contributing!

PR #2996, opened 2024-04-09
Update coverage job

ivirshup:
CI is bugging us to upgrade this

It's also saying we aren't uploading anything, and we already have another task for uploading...

I'm going to use this pr to

1. try upgrading
2. try removing

and seeing what the effects are.


ivirshup:
This actually seems to fix whatever the issue was so I'll go with it

PR #3015, opened 2024-04-19
`sparse_mean_variance_axis` now uses all cores

### LINKED ISSUE #3011: Update Preprocessing functions with numba

It would be useful to switch from array operations to `numba` kernels in for most of the preprocessing. I would start writing those.

Intron7:
This functions now uses all cores for `mean`  & `var` calculations.


flying-sheep (changes requested):
Really cool!

Please use longer variable names. `nthr`→`num_threads`, `s`→`sums_minor` or so, `s0`→???, `m0`→???

Of course not for things like loop counters, but you know.

Also would be great to have comments, like “calculate sums and sum of squares along the minor axis” (if “minor” is correct here) then “go over minor axis again to calculate means and variances from the sums”.

PS: Benchmarks should be ready in a little bit, so you could add one to check how much this speeds things up! I’ll keep you posted

Intron7:
Some small benchmarks for 32 cores with `CSR.shape=(196943, 20867)`:
| axis |old|new|
|------|-----|------|
|minor|804 ms|96 ms|
|major|520 ms|40 ms|

flying-sheep:
I tentatively added a benchmark that runs just on `_get_mean_var`.

Locally I don’t see any difference though, what’s wrong? Too small data? Numba not set up with correct number of threads?

/edit: also I think the machine is not sufficiently tuned. The original run (before I added the `mean_var` benchmarks) said “No changes in benchmarks.”

flying-sheep (approved):
Looks good! Would be nice to add a sufficiently large dataset to the benchmarks to demonstrate the benefits.

PR #3017, opened 2024-04-22
Hvg seurat v3 numba kernel

### LINKED ISSUE #3011: Update Preprocessing functions with numba

Intron7:
Adds a numba kernel for `seurat_v3` for sparse matrices. This kernel is a lot faster and more memory efficient. I doesn't copy nor promotes the matrix to 64-bit floats.


scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -191,6 +199,37 @@ def _highly_variable_genes_seurat_v3(
...
+    batch_counts_sum = np.zeros(n_cols, dtype=np.float64)
+    for i in numba.prange(n_threads):
+        for row in range(i, len(indptr) - 1, n_threads):
+            for j in range(indptr[row], indptr[row + 1]):
+                idx = indices[j]
+                if idx >= n_cols:
+                    continue
+                element = min(np.float64(data[j]), clip_val[idx])
+                squared_sum_buffer[i, idx] += element**2
+                sum_buffer[i, idx] += element
```
flying-sheep: This part looks almost identical to `sparse_mean_var_minor_axis` (the only difference seems to be the clipping)

Is there a way to deduplicate this?

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -191,6 +199,37 @@ def _highly_variable_genes_seurat_v3(
...
 
+@numba.njit(cache=True, parallel=True)
+def _clip_sparse(
+    indptr: NDArray[np.integer],
+    indices: NDArray[np.integer],
+    data: NDArray[np.floating],
+    *,
+    n_rows: int,
+    n_cols: int,
+    nnz: int,
```
flying-sheep: n_rows and nnz are unused, should they be removed or is that a mistake?

Intron7: This is not 100%final. I'm still unsure if singlethreaded might be better than multithreaded

flying-sheep: OK, why would that be? Memory? Locality? I’m just guessing here haha

Intron7:
~500k Cells.
New Version:
peak memory: 16117.61 MiB, increment: 7709.86 MiB
Wall time: 11 s
Old Version:
peak memory: 40093.42 MiB, increment: 31646.50 MiB
Wall time: 30.9 s

eroell:
Indeed, ~3-7x faster for me & of course quite a bit more memory efficient (quickly checked with scalene).

Tests keep working (for `seurat_v3`/`seurat_v3_paper` somewhat tight numeric comparison with Seurat results), nice.

scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -193,6 +197,26 @@ def _highly_variable_genes_seurat_v3(
         return df
 
 
+@numba.njit(cache=True)
+def _clip_sparse(
```
eroell: Minor point, is the name `_clip_sparse` final..? not sure what else to call it tho, fast clip sum square anything... feel free to ignore this point if its just me, is concise enough to be clear anyways

eroell (approved):

scanpy/preprocessing/_highly_variable_genes.py
ivirshup:
```suggestion
def _sum_and_sum_squares_clipped(
```

?

docs/release-notes/1.10.2.md
```diff
@@ -21,4 +21,5 @@
 ```
 
 * `sparse_mean_variance_axis` now uses all cores for the calculations {pr}`3015` {smaller}`S Dicks`
+* `pp.highly_variable_genes` with `flavor=seurat_v3` now uses all cores for clipping {pr}`3017` {smaller}`S Dicks`
```
ivirshup: This isn't parallel, right?

ivirshup (commented):
Looks good. Kinda suprised how much faster this is.

You could probably get another big speed up by operating on the whole matrix at once, and not creating batch specific matrices.

But I think fine as is.

I added a couple points to address, but that's just naming and docs.

ivirshup:
Would also be nice to have a timing benchmark for this.

docs/release-notes/1.10.2.md
Intron7: No it's not I'll update the release note

Intron7:
It is the whole matrix for each batch. It's just called batch_counts because I needed to make sure the format is csr

ivirshup (approved):
I was thinking more of getting rid of the whole `for b in np.unique(batch_info)` loop and doing the whole thing in two passes over the matrix. 

But LGTM, and I think ready to merge once those tests get fixed

Intron7:
I don't know what makes the tests fail with dask in utils

Issue #3013, opened 2024-04-18
Benchmarks: Add all from Dask tutorial


flying-sheep:
Different data:

1. subsets
2. sparse / dense chunks

Also all from disk.

ilan-gold:
Actually deploying this is probably blocked by correcting CPU affinities on the benchmarking machine, but writing the code for this should be manageable otherwise.

flying-sheep:
We now have the big dataset tests, so this might be easy-ish to do

ilan-gold:
Goal:

Add `dask` use-cases to the scanpy benchmarks so we can understand performance changes.  

Nice links:

1. Example benchmark: https://github.com/scverse/scanpy/blob/main/benchmarks/benchmarks/preprocessing_counts.py
2. Project we use for benchmarking: https://asv.readthedocs.io/projects/asv-runner/en/latest/index.html
3. Dask local cluster: https://distributed.dask.org/en/stable/api.html#cluster
4. Using scanpy and dask: https://scanpy.readthedocs.io/en/stable/tutorials/experimental/dask.html. NOTE: this `read_elem_as_dask` function in the notebook is with anndata 0.11 i.e., `pip install --pre anndata`

ilan-gold:
@mikelkou See https://github.com/scverse/scanpy/issues/2578 for functions that have dask vs. not

PR #3031, opened 2024-04-26
Extend benchmarks from basic tutorial

### LINKED ISSUE #3012: Benchmarks: Add all from basic tutorial

flying-sheep:
- [x] Closes #3012 

I added a sparse dataset to cover more code paths in most benchmarks.

`regress_out` seems pretty slow with sparse data, maybe that should be tackled instead of hidden by disabling the benchmark.


flying-sheep:
Oof, half an hour is definitely too long. I wonder why it takes *that* long.

I can finish this up on friday

ilan-gold:
@flying-sheep Not sure.  Now that you mention it, the rapids benchmark also takes more time than I'd expect as well...The functions themselves are quite fast.

benchmarks/benchmarks/preprocessing.py
```diff
@@ -7,17 +7,22 @@
...
-from .utils import lung93k, pbmc3k, pbmc68k_reduced
+from .utils import bmmc4k, lung93k, pbmc3k, pbmc68k_reduced
 
 if TYPE_CHECKING:
     from anndata import AnnData
 
 
 adata: AnnData
 
+params = ["pbmc68k_reduced", "bmmc4k"]
```
ilan-gold: I am not sure this is what I realized what was happening.  I thought this was a PR for _very_ large datasets

ilan-gold: Ah I see there is a separate issue: https://github.com/scverse/scanpy/issues/3013

benchmarks/benchmarks/preprocessing.py
```diff
@@ -7,116 +7,131 @@
...
-    elif params[0] == "pbmc3k":
-        adata = pbmc3k()
-    elif params[0] == "lung93k":
-        adata = lung93k()
-    else:
-        raise ValueError(f"Unknown dataset {params[0]}")
+# The actual test suite begins here
 
+params: list[Dataset] = ["pbmc68k_reduced", "pbmc3k"]
+param_names = ["dataset"]
```
ilan-gold: Why only the two?

benchmarks/benchmarks/preprocessing.py
```diff
@@ -7,116 +7,131 @@
...
     sc.pp.filter_genes(adata, min_cells=3)
 
 
-def peakmem_filter_genes():
+def peakmem_filter_genes(*_):
     sc.pp.filter_genes(adata, min_cells=3)
 
 
-def time_normalize_total():
+# scublet doesn’t work with these datasets
```
ilan-gold: why?

benchmarks/benchmarks/utils.py
```diff
@@ -1,19 +1,30 @@
...
+
 
 @cache
 def _pbmc68k_reduced() -> AnnData:
-    return sc.datasets.pbmc68k_reduced()
+    adata = sc.datasets.pbmc68k_reduced()
+    mapper = dict(
+        percent_mito="pct_counts_mt",
+        n_counts="total_counts",
+    )
```
ilan-gold: In general, since I saw you did this elsewhere, why do you use `dict`?  I believe it is much slower, but even so, I find it harder to read.

scanpy/preprocessing/_scrublet/__init__.py
```diff
@@ -245,9 +245,11 @@ def _run_scrublet(ad_obs: AnnData, ad_sim: AnnData | None = None):
 
     if batch_key is not None:
         if batch_key not in adata.obs.keys():
-            raise ValueError(
-                "`batch_key` must be a column of .obs in the input annData object."
+            msg = (
```
ilan-gold, line 248: Is there a reason this is failing codecov?  Should we add a test for this?

benchmarks/benchmarks/utils.py
flying-sheep: if the speed of `dict` is noticable, one has bigger problems. I use it whenever there are `snake_case` keys, i.e. when one is creating a kwargs dict. Use kwargs to define kwargs.

scanpy/preprocessing/_scrublet/__init__.py
flying-sheep: There’s a lot of error messages like this that aren’t covered. I think it’s always good to increase coverage, but touching an error message that wasn’t used shouldn’t be a blocker.

benchmarks/benchmarks/preprocessing.py
flying-sheep: Great question, the error is very obscure. I’ll see if I can find out. You suggested that it might have to do with ASV :thinking:

benchmarks/benchmarks/preprocessing.py
flying-sheep: Ideally we want to have useful benchmarks, that cover all code paths we handle separately.

This PR is for the code paths “dense in-memory dataset” and “sparse in-memory dataset” (dask comes later)

I think if we have separate code paths for bigger datasets in some functions, we should benchmark these separately, with a bigger dataset.

benchmarks/benchmarks/preprocessing.py
flying-sheep: Ah, I guess it’s just because that dataset comes preprocessed. scrublet wants raw counts.

ilan-gold: Ok so I didn't comment on this because I thought scanpy was supposed to be idempotent so I guess I'll bring it up now.  It seems like it might be a good idea to somehow mark what was done to which dataset in the functions (for example, I see `log1p` applied sometimes while others already have qc metrics calculated).  Do you think this is feasible?

It might worth at least renaming the datasets based on what has been done to them.

ilan-gold: Sorry, I guess in this case it's not a matter of idempotency.  The point still stands.

flying-sheep: Yeah, it would be great to document what processing has been done and to modernizing the datasets

1. right now: where possible without changing/removing stuff (e.g. add a `counts` layer to `pbmc68k_reduced`, as its `.raw` has the same number of genes and is therefore superflous)
2. for 2.0: remove old stuff, change things (e.g. call `make_obs_names_unique` before returning the data and remove `.raw`

flying-sheep: I added this to the 2.0 board: https://github.com/orgs/scverse/projects/18/views/1?pane=issue&itemId=62702062

flying-sheep: And I added an issue: https://github.com/scverse/scanpy/issues/3051

scanpy/preprocessing/_scrublet/__init__.py
ilan-gold: Fair enough.

benchmarks/benchmarks/preprocessing.py
ilan-gold: Amazing, thanks!  Do you think it's worth doing something here now, like somehow indicating [`pbmc3k` is log-transformed](https://github.com/scverse/scanpy/pull/3031/files#diff-f5f3918ea461b367ebfa4bca0592c40e7fd8bc074e4f8a3c373dabc7a32db0c4R36-R43) by the benchmark code specifically?

flying-sheep: I updated the whole PR to split the preprocessing benchmarks into before-transform and after-transform

ilan-gold (approved):

PR #3035, opened 2024-04-29
Fix benchmark test run

flying-sheep:


ilan-gold (approved):
Oh I didn't realize scanpy was using conda since [AnnData uses mamba](https://github.com/scverse/anndata/blob/main/benchmarks/asv.conf.json#L47) LGTM

PR #3042, opened 2024-05-02
hvg flavors seurat and cellranger with batch: bug in subset

### LINKED ISSUE #3027: Unexplainable sc.pp.highly_variable_genes(subset = True) behavior

### What happened?

When doing HVG selection with batch_key = "something" and subset = True, I noticed unexpected genes to be selected in the subset anndata. Upon further investigation, it seems that somehow the inplace subsetting goes wrong. (Though I checked the source code and could not find any issue that may explain it there.)

### Minimal code sample

```python
import scanpy as sc
import numpy as np
np.random.seed(0)

# Get AnnData
adata = sc.datasets.pbmc3k()
adata.layers['counts'] = adata.X.copy().tocsr()
adata.obs["Age"] = np.random.randint(0, 6, (2700,))
adata.obs["Age"] = adata.obs["Age"].astype('category')

# Filter genes, preprocess
sc.pp.filter_genes(adata, min_counts = 10)
sc.pp.normalize_total(adata)
sc.pp.log1p(adata)

# Subset = False
ad_nosub = adata.copy()
sc.pp.highly_variable_genes(ad_nosub, n_top_genes = 1000, batch_key = "Age", subset = False)

# Subset = False, manual subset afterwards
ad_nosub_subbed = ad_nosub.copy()
ad_nosub_subbed._inplace_subset_var(ad_nosub_subbed.var["highly_variable"].to_numpy())

# Subset = True
ad_sub = adata.copy()
sc.pp.highly_variable_genes(ad_sub, n_top_genes = 1000, batch_key = "Age", subset = True)
```

### Error output

```pytb
>>> # As expected
>>> print(np.sum(ad_nosub.var["highly_variable"]))
1000
>>> 
>>> # As expected
>>> print(np.sum(ad_nosub_subbed.var["highly_variable"]))
1000
>>> 
>>> # Not as expected
>>> print(np.sum(ad_sub.var["highly_variable"]))
101
```

### Versions

<details>

``` bash
→ conda list | grep scanpy
scanpy                    1.10.1             pyhd8ed1ab_0    conda-forge

→ conda list | grep anndata
anndata                   0.10.7             pyhd8ed1ab_0    conda-forge
```

</details>

eroell:
- [x] Closes #3027

- [ ] Release notes not necessary because: they are, added

This PR fixes the bug reported in the linked issue.

A new test which spots the erroneous computations has been added.

I would use this chance to refactor the `_highly_variable_genes.py`, rather than using the 2-lines fix suggested in the first commit:
Doing the multi-batch hvg flagging differently for seurat_v3 and seurat/cell_ranger is what made this bug hard to spot in the first place I think.


flying-sheep:
Looks simple enough! Please deduplicate the tests though, they have too many identical lines.

eroell:
> Please deduplicate the tests though, they have too many identical lines.

To do so did across-setting tests with for loop on top of former test...

Would you prefer one separate test with that for loop for the across-settings check?

flying-sheep:
Oh, sorry for missing this. I’m always for `pytest.parametrize` instead of for loops. That way it’s immediately visible which variants of the test fail.

flying-sheep (changes requested):
Oh, sorry for missing this. I’m always for `pytest.parametrize` instead of for loops. That way it’s immediately visible which variants of the test fail.

eroell:
To test in the future that this bug doesnt happen, the 4 combinations of inplace=True/False and subset=True/False need to be compared for their selected `var_names` - if these arguments were be in `pytest.mark.parametrize`, the different produced anndatas don't live at the same time and can't be compared...

Could
- extract this specific section into a new, dedicated smaller test

```py
    # check that the results are consistent for subset True/False: inplace True
    adata_post_subset = adatas["subset_False_inplace_True"][
        :, adatas["subset_False_inplace_True"].var["highly_variable"]
    ]
    assert adata_post_subset.var_names.equals(
        adatas["subset_True_inplace_True"].var_names
    )

    # check that the results are consistent for subset True/False: inplace False
    df_post_subset = dfs["subset_False_inplace_False"][
        dfs["subset_False_inplace_False"]["highly_variable"]
    ]
    assert df_post_subset.index.equals(dfs["subset_True_inplace_False"].index)

    # check that the results are consistent for inplace True/False: subset True
    assert adatas["subset_True_inplace_True"].var_names.equals(
        dfs["subset_True_inplace_False"].index
    )

```


- hardcode the known-to-be-selected features and check against them

what's your preference here?

flying-sheep:
Ah, silly me, this makes sense.

Since some entries of the dicts are never used, I just removed them and replaced the string with a bool (subset=True/False). This makes it all quite a bit more compact.

Also please remember `itertools`: If we’re not writing numba code, it’s always preferable to use it as opposed to nesting for loops.

flying-sheep (approved):
please add a relnote

docs/release-notes/1.10.3.md
```diff
@@ -6,8 +6,12 @@
 ```{rubric} Docs
 ```
 
-```{rubric} Bug fixes
+```{rubric} Bug
```
flying-sheep: you might want to undo this line change :smile:

eroell: oh no!
yes haha

eroell:
Timeout and Scrublet failing in Python3.12?

eroell:
Ah on main too I see

eroell:
coverage decreased, I think not detected because some pytest.parametrize were removed?
I can also add the new specific test separately do avoid this, else it looks good to me

flying-sheep:
> coverage decreased, I think not detected because some pytest.parametrize were removed?

I think codecov just hadn’t updated its comment yet when you saw that.

What you say can’t be, it doesn’t matter how a line was hit: If a line is run, it’ll be reported as hit, if your changes would have caused it to no longer be it, it would have been reported as a miss.

PR #3044, opened 2024-05-06
Speedup scrublet

Intron7:


Zethson:
scrublet doesn't seem to be part of the benchmarks at the moment

Intron7:
@Zethson I think that this is a bit better. But not as much as I hoped

flying-sheep:
@Zethson I’m adding some in #3031, but it needs more thought: 30 minutes for a benchmark run is too long …

flying-sheep:
With #3056 merged, this now says

> before | after | ratio | benchmark
> --- |  --- |  --- |  ---
> 448±100ms | 381±100ms | ~0.85 | `preprocessing_counts.time_scrublet('pbmc68k_reduced')`

but ASV seems to think that’s not enough to report.

Also unclear why it’s reported as taking 26 minutes by github:

> image

I see in the server logs

> - May 14 10:15:57 scvbench benchmark[1462905]: 2024-05-14T10:15:57.547945Z DEBUG handle_event:HTTP{http.method=PATCH http.url=https://api.github.com/repos/scverse/scanpy/check-runs/24942322156 otel.name="HTTP" otel.kind="client"}: octocrab: requesting
> - […running benchmarks]
> - May 14 10:27:18 scvbench benchmark[1462905]: 2024-05-14T10:27:18.793352Z DEBUG handle_event:HTTP{http.method=PATCH http.url=https://api.github.com/repos/scverse/scanpy/check-runs/24942322156 otel.name="HTTP" otel.kind="client"}: octocrab: requesting

which means that between setting the check run to “running” and to “done”, 11m21s passed.

Maybe GitHub counts the queue time?

PR #3047, opened 2024-05-07
Fixes for sklearn 1.5

flying-sheep:
Scikit-learn 1.5.0 will no longer have `issparse` exported, so let’s import it from where it belongs


flying-sheep:
Hmm, seems like `PCA` with `svd_solver='arpack'` works differently in sklearn 1.5 and flips the coordinates in some tests:

https://github.com/scikit-learn/scikit-learn/issues/28826

E.g. this used to work, now it doesn’t.

```py
from __future__ import annotations

import numpy as np

data = np.asarray([[-1, 2, 0], [3, 4, 0], [1, 2, 0]]).T
expected = np.array(
    [[-1.579575e-15, 1.490712], [-2.44949, -0.745356], [2.44949, -0.745356]],
    dtype=np.float32,
)

pca = PCA(n_components=2, svd_solver="arpack", random_state=np.random.RandomState(0))
transformed = pca.fit_transform(data).astype(np.float32)
np.testing.assert_almost_equal(transformed, expected, decimal=5)
```

PR #3061, opened 2024-05-14
t-SNE optimization using scikit-learn-intelex

ashish615:
This pull request accelerates t-SNE using the scikit-learn-intelex library, resulting in approximately a 10x runtime improvement for the t-SNE implementation in the package for the given example below. The experiment was run on AWS r7i.24xlarge.

```py
import time
import numpy as np

import pandas as pd

import scanpy as sc

import os
import wget

import warnings



# marker genes

# filtering cells

# filtering genes

# PCA

# t-SNE

# k-means

# Gene ranking

# Number of parallel jobs

start=time.time()
tr=time.time()
adata = sc.read(input_file)
adata.var_names_make_unique()
adata.shape



tr=time.time()
USE_FIRST_N_CELLS = 1300000
adata.shape

tr=time.time()
sc.pp.log1p(adata)


for marker in markers:

ts=time.time()


#scale

ts=time.time()
sc.pp.scale(adata, max_value=10)
t0=time.time()

sc.tl.pca(adata, n_comps=n_components)

t0=time.time()
sc.tl.tsne(adata, n_pcs=tsne_n_pcs, use_fast_tsne=True)

'''
from sklearn.manifold import TSNE
from scanpy.tools._utils import _choose_representation
X = _choose_representation(adata, n_pcs=tsne_n_pcs)
X_tsne = TSNE().fit_transform(X.astype(np.float32))
adata.obsm['X_tsne'] = X_tsne
'''
print("Tsne time:", time.time()-t0)
```


sanchit-misra:
For the above code, the time spent in tSNE went down from 2252 secs  to 210 secs due to this PR.

flying-sheep:
pre-commit.ci autofix

flying-sheep:
> For the above code, the time spent in tSNE went down from 2252 secs to 210 secs due to this PR.

What’s the comparison to MulticoreTSNE?

## Defaults

It would probably make sense to use a `flavor: Literal['auto', 'sklearn', 'intelex', 'multicore'] = 'auto'` parameter here, where `auto` would try to import the speedup packages one-by-one and use the preferred one.

`use_fast_tsne` could be deprecated and made to default to `None`, with this logic (too bad we can’t use `match` yet)

```py
if use_fast_tsne is not None:
    warnings.warn("...", FutureWarning)
match (use_fast_tsne, flavor):
    case (None, 'auto'): ...  # try importing 'intelex', fall back to 'sklearn'
    case (None, _): ...  # use specified flavor
    case (True, 'auto'): ...  # use 'multicore'
    case (True, 'sklearn'): ...  # throw error
    case (True, _): ...  # use specified flavor
    case (False, 'auto' | 'sklearn'): ...  # Use 'sklearn'
    case (False, _): ...  # Throw error
    case _: ... raise AssertionError()
```

In the future, we can change `'auto'` to try both intelex and MulticoreTSNE

@ilan-gold what do you think?

ilan-gold:
This seems reasonable to me @flying-sheep .  Does using the patched version change results over the unpatched @ashish615 i.e., for a given random seed, unpatched and patched are the same?  If the two are the same for a given seed/state, then I think what @flying-sheep  is proposing could be done separately (even if we make the dependency optional IMO).  However, if the new version does change results, we will need the handling that @flying-sheep  describes.

ashish615:
> What’s the comparison to MulticoreTSNE?

This is the compare is while using **use_fast_tsne** flag which using MulticoreTSNE.

ashish615:
> This seems reasonable to me @flying-sheep . Does using the patched version change results over the unpatched @ashish615 i.e., for a given random seed, unpatched and patched are the same? If the two are the same for a given seed/state, then I think what @flying-sheep is proposing could be done separately (even if we make the dependency optional IMO). However, if the new version does change results, we will need the handling that @flying-sheep describes.

@ilan-gold , I didn't check that. I will check that and let you guys know.

narendrachaudhary51:
These t-SNE optimizations are mentioned in the following paper. Adding it here for reference.
https://arxiv.org/abs/2212.11506
Accelerating Barnes-Hut t-SNE Algorithm by Efficient Parallelization on Multi-Core CPUs
N Chaudhary, A Pivovar, P Yakovlev, A Gorshkov… - arXiv preprint arXiv:2212.11506, 2022

ilan-gold:
@narendrachaudhary51 You can add the reference here https://github.com/scverse/scanpy/blob/main/docs/references.bib (thank you @flying-sheep for improving this!)

scanpy/preprocessing/__init__.py
```diff
@@ -21,6 +21,7 @@
     subsample,
 )
 
+
```
ilan-gold: Could you remove this and other blank space diffs?  This makes the git blame hard to follow for future users

ashish615:
Hi @flying-sheep @ilan-gold ,
Based on our previous discussion, we observed that applying and then removing a patch while fixing the seed causes the t-SNE output to change. In our experiment, we used 1.3 million data points to run t-SNE and compared the results of the patched and unpatched versions by examining the KL Divergence from both runs. The results are summarized in the table below. 

In the above code use **USE_FIRST_N_CELLS** to set number of records and  use sc.tl.tsne(adata, n_pcs=tsne_n_pcs, **use_fast_tsne=False**) to run optimized run with latest commit. You can get KL divergence numbers by logging  [kl_divergence_](https://scikit-learn.org/stable/modules/generated/sklearn.manifold.TSNE.html)


image

sanchit-misra:
As can be seen with the KL divergence values in the above table, while the output of Intel optimized t-SNE is different, it is equivalent in quality.

ilan-gold:
@sanchit-misra @ashish615 one thing that we thought - why does this need to be in `scanpy`? Can't users turn this on/off outside `scanpy`?

ilan-gold:
Notwithstanding the above question, since your method produces different results than the default, we will need to adopt the conventions outlined: https://github.com/scverse/scanpy/pull/3061#issuecomment-2114783668

This will ensure continued reproducibility with defaults.

ashish615:
Hi @ilan-gold,

Regarding your thought in this [comment](https://github.com/scverse/scanpy/pull/3061#issuecomment-2134651481), we can enable or disable Intel optimization from outside the code. However, users might not be aware of how to use this feature. Instead, if we add it to scanpy directly, all scanpy users will know the same option available.

If we agree with the option discussed in this [comment](https://github.com/scverse/scanpy/pull/3061#issuecomment-2114783668 ), I can proceed with updating the t-SNE file.

PR #3092, opened 2024-06-03
Move to src / tests layout

flying-sheep:
Changes exactly as in

- https://github.com/scverse/anndata/pull/1151 and
- https://github.com/scverse/anndata/pull/1459

but without adding the intermediate cruft


pyproject.toml
```diff
@@ -153,26 +153,23 @@ rapids = ["cudf>=0.9", "cuml>=0.9", "cugraph>=0.9"]  # GPU accelerated calculati
...
 [tool.hatch.build.targets.wheel]
-# TODO: move scanpy to src/
-only-include = ["src/testing", "scanpy"]
-sources = ["src", "."]
+packages = ["src/testing", "src/scanpy"]
 [tool.hatch.version]
 source = "vcs"
 [tool.hatch.build.hooks.vcs]
-version-file = "scanpy/_version.py"
+version-file = "src/scanpy/_version.py"
```
Zethson, line 161: Curiosity of mine: This file does not exist, right? What is it referring to?

Zethson (approved):

pyproject.toml
flying-sheep: the line you’re commenting on causes the file to be created by [`hatch-vcs`](https://pypi.org/project/hatch-vcs/):

https://github.com/scverse/scanpy/blob/ef6af20ab3dfcd784f9edd9ea8dc426835518e18/pyproject.toml#L3

That happens whenever the project is built, e.g. when doing `python -m build` to build a wheel or when doing `pip install -e .` to install the project in editable mode.

flying-sheep: Personally I think `__version__` is silly and shouldn’t exist, and neither should that file. Package metadata is all that should matter. Just call [`importlib.metadata.version('mypkg')`](https://docs.python.org/3/library/importlib.metadata.html#distribution-versions) and you have the canonically installed version.

flying-sheep:
uuh, the auto-backport worked nice!

PR #3099, opened 2024-06-05
scale function(_get_mean_var) updated for dense array, speedup upto ~4.65x

ashish615:
Hi,
We are submitting PR for speed up of the _get_mean_var function. 
|                 |  Time(sec)|
| -----------| ----- |
| Original | 18.49 |
| Updated | 3.97 |
| Speedup | 4.65743073 |

```py
import time
import numpy as np

import pandas as pd

import scanpy as sc

import os
import wget

import warnings



# marker genes

# filtering cells

# filtering genes

# PCA

# t-SNE

# k-means

# Gene ranking

# Number of parallel jobs

start=time.time()
tr=time.time()
adata = sc.read(input_file)
adata.var_names_make_unique()
adata.shape



tr=time.time()
USE_FIRST_N_CELLS = 1300000
adata.shape

tr=time.time()
sc.pp.log1p(adata)


for marker in markers:

ts=time.time()


#scale

ts=time.time()
sc.pp.scale(adata)
```
add timer around _get_mean_var call 

https://github.com/scverse/scanpy/blob/706d4ef65e5d65e04b788831e7fd65dbe6b2a61f/scanpy/preprocessing/_scale.py#L167

we can also create _get_mean_var_std function that return std as well so we don't require to compute it in scale function(L168-L169).


Intron7:
I have some small improvements that I would like to add next week for more precision for larger matrices

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,43 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+@numba.njit(cache=True, parallel=True)
+def _compute_mean_var(
+    X: _SupportedArray, axis: Literal[0, 1] = 0, dtype: DTypeLike | None = None
+) -> tuple[NDArray[np.float64], NDArray[np.float64]]:
+    nthr = numba.get_num_threads()
+    axis_i = 1 if axis == 0 else 0
+    s = np.zeros((nthr, X.shape[axis_i]), dtype=dtype)
+    ss = np.zeros((nthr, X.shape[axis_i]), dtype=dtype)
+    mean = np.zeros(X.shape[axis_i], dtype=dtype)
+    # std=np.zeros(X.shape[axis_i],dtype=dtype)
```
Intron7:
```suggestion

```

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,43 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+        for r in range(i, n, nthr):
+            for c in range(X.shape[axis_i]):
+                v = X[r, c] if axis == 0 else X[c, r]
+                s[i, c] += v
+                ss[i, c] += v * v
+    for c in numba.prange(X.shape[axis_i]):
+        s0 = s[:, c].sum()
+        mean[c] = s0 / n
+        var[c] = (ss[:, c].sum() - s0 * s0 / n) / (n - 1)
+        # std[c]=np.sqrt(var[c])
```
Intron7:
```suggestion

```

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,43 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+def _compute_mean_var(
+    X: _SupportedArray, axis: Literal[0, 1] = 0, dtype: DTypeLike | None = None
+) -> tuple[NDArray[np.float64], NDArray[np.float64]]:
+    nthr = numba.get_num_threads()
+    axis_i = 1 if axis == 0 else 0
+    s = np.zeros((nthr, X.shape[axis_i]), dtype=dtype)
+    ss = np.zeros((nthr, X.shape[axis_i]), dtype=dtype)
+    mean = np.zeros(X.shape[axis_i], dtype=dtype)
+    # std=np.zeros(X.shape[axis_i],dtype=dtype)
+    var = np.zeros(X.shape[axis_i], dtype=dtype)
```
Intron7, line 60:
```suggestion
    s = np.zeros((nthr, X.shape[axis_i]), dtype=np.float64)
    ss = np.zeros((nthr, X.shape[axis_i]), dtype=np.float64))
    mean = np.zeros(X.shape[axis_i], dtype=np.float64))
    # std=np.zeros(X.shape[axis_i],dtype=dtype)
    var = np.zeros(X.shape[axis_i], dtype=np.float64))
```

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,41 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+            mean_sq = axis_mean(elem_mul(X, X), axis=axis, dtype=np.float64)
+            var = mean_sq - mean**2
+        # enforce R convention (unbiased estimator) for variance
+        var *= X.shape[axis] / (X.shape[axis] - 1)
+    return mean, var
+
+
+@numba.njit(cache=True, parallel=True)
+def _compute_mean_var(
+    X: _SupportedArray, axis: Literal[0, 1] = 0, dtype: DTypeLike | None = None
```
Intron7:
```suggestion
    X: _SupportedArray, axis: Literal[0, 1] = 0
```

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,41 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
 def _get_mean_var(
     X: _SupportedArray, *, axis: Literal[0, 1] = 0
 ) -> tuple[NDArray[np.float64], NDArray[np.float64]]:
-    if isinstance(X, sparse.spmatrix):
-        mean, var = sparse_mean_variance_axis(X, axis=axis)
+    if isinstance(X, np.ndarray):
+        mean, var = _compute_mean_var(X, axis=axis, dtype=np.float64)
```
Intron7:
```suggestion
        mean, var = _compute_mean_var(X, axis=axis)
```

Intron7:
@ashish615 after doing some benchmarking myself I found out that your solution for `axis=1` is under performing compared to `axis=0` for larger arrays. I think that is because of the memory access pattern you choose. I rewrote the function with that in mind. I'll again make a PR to you, because for some reason you disallow us from making changes to your PR.

Intron7:
https://github.com/IntelLabs/open-omics-scanpy/pull/2 This would fix the performance. This is still not perfect for F-Array's but its a start

Intron7 (changes requested):
Please merge https://github.com/IntelLabs/open-omics-scanpy/pull/2

docs/release-notes/1.10.2.md
```diff
@@ -31,3 +31,4 @@
 * `pp.highly_variable_genes` with `flavor=seurat_v3` now uses a numba kernel {pr}`3017` {smaller}`S Dicks`
 * Speed up {func}`~scanpy.pp.scrublet` {pr}`3044` {smaller}`S Dicks` and {pr}`3056` {smaller}`P Angerer`
 * Speed up clipping of array in {func}`~scanpy.pp.scale` {pr}`3100` {smaller}`P Ashish & S Dicks`
+* Speed up _get_mean_var used in {func}`~scanpy.pp.scale` {pr}`3099` {smaller}`P Ashish & S Dicks`
```
Intron7: move this to `0.10.3.md`

Intron7 (changes requested):

docs/release-notes/1.10.2.md
ashish615: ok

ashish615: done

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,55 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+        axis_i = 1
+        sums = np.zeros((n_threads, X.shape[axis_i]), dtype=np.float64)
+        sums_squared = np.zeros((n_threads, X.shape[axis_i]), dtype=np.float64)
+        mean = np.zeros(X.shape[axis_i], dtype=np.float64)
+        var = np.zeros(X.shape[axis_i], dtype=np.float64)
+        n = X.shape[axis]
+        for i in numba.prange(n_threads):
+            for r in range(i, n, n_threads):
+                for c in range(X.shape[axis_i]):
+                    value = np.float64(X[r, c])
```
Intron7:
```suggestion
                    value = X[r, c]
```

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,55 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+            sum_ = sums[:, c].sum()
+            mean[c] = sum_ / n
+            var[c] = (sums_squared[:, c].sum() - sum_ * sum_ / n) / (n - 1)
+    else:
+        axis_i = 0
+        mean = np.zeros(X.shape[axis_i], dtype=np.float64)
+        var = np.zeros(X.shape[axis_i], dtype=np.float64)
+        for r in numba.prange(X.shape[0]):
+            for c in range(X.shape[1]):
+                value = np.float64(X[r, c])
```
Intron7:
```suggestion
                value = X[r, c]
```

Intron7 (changes requested):
This fixes the issues

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,55 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+        sums_squared = np.zeros((n_threads, X.shape[axis_i]), dtype=np.float64)
+        mean = np.zeros(X.shape[axis_i], dtype=np.float64)
+        var = np.zeros(X.shape[axis_i], dtype=np.float64)
+        n = X.shape[axis]
+        for i in numba.prange(n_threads):
+            for r in range(i, n, n_threads):
+                for c in range(X.shape[axis_i]):
+                    value = np.float64(X[r, c])
+                    sums[i, c] += value
+                    sums_squared[i, c] += value * value
```
ilan-gold: Remove casting to match previous behavior.  In `1.11.0` we should release a breaking change that changes the behavior of multiplication to preserve the data type better.

Intron7 (approved):
Looks good to me

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,55 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
-    # enforce R convention (unbiased estimator) for variance
-    var *= X.shape[axis] / (X.shape[axis] - 1)
+        if isinstance(X, sparse.spmatrix):
+            mean, var = sparse_mean_variance_axis(X, axis=axis)
+        else:
+            mean = axis_mean(X, axis=axis, dtype=np.float64)
+            mean_sq = axis_mean(elem_mul(X, X), axis=axis, dtype=np.float64)
+            var = mean_sq - mean**2
+        # enforce R convention (unbiased estimator) for variance
+        var *= X.shape[axis] / (X.shape[axis] - 1)
```
flying-sheep, line 47: Before your change, this line ran unconditionally, now it only runs for the `not isinstance(X, np.ndarray)` case. Is that intentional? Then you should mention that in `_compute_mean_var`’s docstring.

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,55 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+    return mean, var
+
+
+@numba.njit(cache=True, parallel=True)
+def _compute_mean_var(
```
flying-sheep, line 52: We already have `_get_mean_var`. Maybe rename this to `_get_mean_var_ndarray` or `_get_mean_var_dense`?

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,55 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+    X: _SupportedArray, axis: Literal[0, 1] = 0, n_threads=1
```
flying-sheep, line 53:
```suggestion
    X: _SupportedArray, axis: Literal[0, 1] = 0, n_threads: int = 1
```

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,14 +33,55 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
+        mean = np.zeros(X.shape[axis_i], dtype=np.float64)
+        var = np.zeros(X.shape[axis_i], dtype=np.float64)
+        for r in numba.prange(X.shape[0]):
+            for c in range(X.shape[1]):
+                value = X[r, c]
+                mean[r] += value
+                var[r] += value * value
+        for c in numba.prange(X.shape[0]):
+            mean[c] = mean[c] / X.shape[1]
+            var[c] = (var[c] - mean[c] ** 2) / (X.shape[1] - 1)
```
flying-sheep, line 83: Please don’t duplicate identical lines.

```suggestion
    axis_i = 1 - axis
    mean = np.zeros(X.shape[axis_i], dtype=np.float64)
    var = np.zeros(X.shape[axis_i], dtype=np.float64)
    if axis == 0:
        sums = np.zeros((n_threads, X.shape[axis_i]), dtype=np.float64)
        sums_squared = np.zeros((n_threads, X.shape[axis_i]), dtype=np.float64)
        n = X.shape[axis]
        for i in numba.prange(n_threads):
            for r in range(i, n, n_threads):
                for c in range(X.shape[axis_i]):
                    value = X[r, c]
                    sums[i, c] += value
                    sums_squared[i, c] += value * value
        for c in numba.prange(X.shape[axis_i]):
            sum_ = sums[:, c].sum()
            mean[c] = sum_ / n
            var[c] = (sums_squared[:, c].sum() - sum_ * sum_ / n) / (n - 1)
    else:
        for r in numba.prange(X.shape[0]):
            for c in range(X.shape[1]):
                value = X[r, c]
                mean[r] += value
                var[r] += value * value
        for c in numba.prange(X.shape[0]):
            mean[c] = mean[c] / X.shape[1]
            var[c] = (var[c] - mean[c] ** 2) / (X.shape[1] - 1)
```

flying-sheep (changes requested):
I don’t see the claimed speedup in the benchmarks, what’s missing?

Also is `numba.get_num_threads()` safe? E.g. I think `_get_mean_var` is also called in each dask chunks. Will `numba.get_num_threads()` return a reasonable number in that case?

Otherwise nice! I’m not a huge fan of how unpythonic numba code looks, but I don’t think anything can be done about that.

src/scanpy/preprocessing/_utils.py
Intron7: I think we can rename the kernel

src/scanpy/preprocessing/_utils.py
Intron7: I think we can slim this down a bit. The two different loops need to be separate though

Intron7:
The function should also work for 1 thread. numba.get_num_threads() is fine it works well with the sparse arrays. But I have no experience with it inside of dask.

src/scanpy/preprocessing/_utils.py
ilan-gold, line 53: I don't think `_SupportedArray` is the right type annotation here.  This doesn't run directly on `dask.Array`, unless I am misunderstanding something.

PR #3110, opened 2024-06-18
Speedup (~20x) of  scanpy.pp.regress_out function using Linear Least Square method.

ashish615:
- [ ] Closes #

Hi,
We are submitting PR for speed up of the regress_out function. Here we finding coefficient using Linear regression (Linear Least Squares) rather then GLM for non categorical data.

|                 |  Time(sec)|
| -----------| ----- |
| Original | 297|
| Updated | 14.91 |
| Speedup | 19.91 |

```python
import time
import numpy as np

import pandas as pd

import scanpy as sc

import os
import wget

import warnings



# marker genes

# filtering cells

# filtering genes

# PCA

# t-SNE

# k-means

# Gene ranking

# Number of parallel jobs

start=time.time()
tr=time.time()
adata = sc.read(input_file)
adata.var_names_make_unique()
adata.shape



tr=time.time()
USE_FIRST_N_CELLS = 1300000
adata.shape

tr=time.time()
sc.pp.log1p(adata)


for marker in markers:

ts=time.time()
```

### DIFF
--- a/src/scanpy/preprocessing/_simple.py
+++ b/src/scanpy/preprocessing/_simple.py
@@ -765,14 +765,12 @@ def regress_out(
     res = None
     if not variable_is_categorical:
         A = regres.to_numpy()
-        # if det(A.T@A) zero, then can not take inverse and regression fails.
-        # if fails then fall back to GLM implemetation of regression.
-        # Numba kernel is used to speedup regression using Linear Least Square method
-        # on regressor of type numpy array.
+        # if det(A.T@A) != 0 we can take the inverse and regress using a fast method.
         if np.linalg.det(A.T @ A) != 0:
             res = numpy_regress_out(X, A)
 
-    # for categorical variable and failed the above code then run original code.
+    # for a categorical variable or if the above checks failed,
+    # we fall back to the GLM implemetation of regression.
     if variable_is_categorical or res is None:
         from joblib import Parallel, delayed
 
--- a/src/scanpy/preprocessing/_simple.py
+++ b/src/scanpy/preprocessing/_simple.py
@@ -620,7 +620,7 @@ def to_dense(
     shape: tuple[int, int],
     indptr: NDArray[np.integer],
     indices: NDArray[np.integer],
-    data: NDArray[np.number],
+    data: NDArray[DT],
 ) -> NDArray[DT]:
     """\
     Numba kernel for np.toarray() function
--- a/src/scanpy/preprocessing/_simple.py
+++ b/src/scanpy/preprocessing/_simple.py
@@ -612,6 +612,47 @@ def normalize_per_cell(  # noqa: PLR0917
     return X if copy else None
 
 
+@numba.njit(cache=True, parallel=True)
+def to_dense(shape, indptr, indices, data):
+    """\
+    Numba kernel for np.toarray() function
+    """
+    X = np.empty(shape, dtype=data.dtype)
+
+    for r in numba.prange(shape[0]):
+        X[r] = 0
+        for i in range(indptr[r], indptr[r + 1]):
+            X[r, indices[i]] = data[i]
+    return X
+
+
+def numpy_regress_out(
+    X: np.ndarray,
+    A: np.ndarray,
+) -> np.ndarray:
+    """\
+    Numba kernel for regress out unwanted sorces of variantion.
+    Finding coefficient using Linear regression (Linear Least Squares).
+    """
+
+    @numba.njit(cache=True, parallel=True)
+    def get_resid(
+        X: np.ndarray,
+        A: np.ndarray,
+        coeff,
+    ) -> np.ndarray:
+        for i in numba.prange(X.shape[0]):
+            X[i] -= A[i] @ coeff
+        return X
+
+    tmp0 = A.T @ A
+    tmp = np.linalg.inv(tmp0)
+    tmp0 = A.T @ X
+    coeff = tmp @ tmp0
+    X = get_resid(X, A, coeff)
+    return X
+
+
 @old_positionals("layer", "n_jobs", "copy")
 def regress_out(
     adata: AnnData,
@@ -664,7 +705,7 @@ def regress_out(
 
     if issparse(X):
         logg.info("    sparse input is densified and may " "lead to high memory use")
-        X = X.toarray()
+        X = to_dense(X.shape, X.indptr, X.indices, X.data)
 
     n_jobs = sett.n_jobs if n_jobs is None else n_jobs
 
@@ -715,14 +756,27 @@ def regress_out(
             regres = regressors
         tasks.append(tuple((data_chunk, regres, variable_is_categorical)))
 
-    from joblib import Parallel, delayed
+    flag = None
+    if not variable_is_categorical:
+        A = regres.to_numpy()
+        # checking if inverse of A.T@A matix is zero or not.
+        if np.linalg.det(A.T @ A) != 0:
+            res = numpy_regress_out(X, A)
+            flag = 1
+    # for categorical variable and failed the above code then run original code.
+    if variable_is_categorical or flag is None:
+        from joblib import Parallel, delayed
 
-    # TODO: figure out how to test that this doesn't oversubscribe resources
-    res = Parallel(n_jobs=n_jobs)(delayed(_regress_out_chunk)(task) for task in tasks)
+        # TODO: figure out how to test that this doesn't oversubscribe resources
+        res = Parallel(n_jobs=n_jobs)(
+            delayed(_regress_out_chunk)(task) for task in tasks
+        )
 
-    # res is a list of vectors (each corresponding to a regressed gene column).
-    # The transpose is needed to get the matrix in the shape needed
-    _set_obs_rep(adata, np.vstack(res).T, layer=layer)
+        # res is a list of vectors (each corresponding to a regressed gene column).
+        # The transpose is needed to get the matrix in the shape needed
+        res = np.vstack(res).T
+
+    _set_obs_rep(adata, res, layer=layer)
     logg.info("    finished", time=start)
     return adata if copy else None
 
--- a/src/scanpy/preprocessing/_simple.py
+++ b/src/scanpy/preprocessing/_simple.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 
 import warnings
 from functools import singledispatch
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, TypeVar
 
 import numba
 import numpy as np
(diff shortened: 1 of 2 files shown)


src/scanpy/preprocessing/_simple.py
```diff
@@ -612,6 +612,47 @@ def normalize_per_cell(  # noqa: PLR0917
     return X if copy else None
 
 
+@numba.njit(cache=True, parallel=True)
+def to_dense(shape, indptr, indices, data):
```
flying-sheep: Is this really faster than `toarray`? In that case something like this should be contributed to scipy!

Also, please do types:

```suggestion
DT = TypeVar("DT")


def to_dense(shape: tuple[int, int], indptr: NDArray[np.integer], indices: NDArray[np.integer], data: NDArray[np.number]) -> NDArray[DT]:
```

src/scanpy/preprocessing/_simple.py
```diff
@@ -612,6 +612,47 @@ def normalize_per_cell(  # noqa: PLR0917
...
+    for r in numba.prange(shape[0]):
+        X[r] = 0
+        for i in range(indptr[r], indptr[r + 1]):
+            X[r, indices[i]] = data[i]
+    return X
+
+
+def numpy_regress_out(
+    X: np.ndarray,
+    A: np.ndarray,
```
flying-sheep: please use speaking parameter names, in `lower_snake_case`.

`X` is reserved for `AnnData.X`, i.e. the main matrix of an AnnData object.

src/scanpy/preprocessing/_simple.py
```diff
@@ -612,6 +612,47 @@ def normalize_per_cell(  # noqa: PLR0917
...
+        coeff,
+    ) -> np.ndarray:
+        for i in numba.prange(X.shape[0]):
+            X[i] -= A[i] @ coeff
+        return X
+
+    tmp0 = A.T @ A
+    tmp = np.linalg.inv(tmp0)
+    tmp0 = A.T @ X
+    coeff = tmp @ tmp0
```
flying-sheep: ditto, `tmp0` and `tmp` aren’t good names, and re-declaring them makes this more confusing.

E.g. this would be a start:

```suggestion
    todo_rename_me = np.linalg.inv(A.T @ A)
    coeff = todo_rename_me @ (A.T @ X)
```

src/scanpy/preprocessing/_simple.py
```diff
@@ -612,6 +612,47 @@ def normalize_per_cell(  # noqa: PLR0917
...
+    """\
+    Numba kernel for regress out unwanted sorces of variantion.
+    Finding coefficient using Linear regression (Linear Least Squares).
+    """
+
+    @numba.njit(cache=True, parallel=True)
+    def get_resid(
+        X: np.ndarray,
+        A: np.ndarray,
+        coeff,
```
flying-sheep: ditto, also please add a type for `coeff`

src/scanpy/preprocessing/_simple.py
```diff
@@ -715,14 +756,27 @@ def regress_out(
             regres = regressors
         tasks.append(tuple((data_chunk, regres, variable_is_categorical)))
 
-    from joblib import Parallel, delayed
+    flag = None
+    if not variable_is_categorical:
+        A = regres.to_numpy()
+        # checking if inverse of A.T@A matix is zero or not.
```
flying-sheep: That comment doesn’t add anything that the code on the next line doesn’t say. Can you replace it with a comment explaining *why* you do that?

src/scanpy/preprocessing/_simple.py
```diff
@@ -715,14 +756,27 @@ def regress_out(
...
+        if np.linalg.det(A.T @ A) != 0:
+            res = numpy_regress_out(X, A)
+            flag = 1
```
flying-sheep: if `flag` was necessary, it should have a speaking name and be a `bool`.

But it isn’t, you can just do `res = None` and set it in the branch, then check `if variable_is_categorical or res is None` later

flying-sheep (changes requested):

flying-sheep:
Thanks for keeping them coming in! It would be helpful if you documented a bit why your newly introduced branch is faster with comments.

src/scanpy/preprocessing/_simple.py
ashish615: @flying-sheep , Yes it is faster then the toarray. for above example toarray was taking 8 sec vs numba kernel is taking 2 sec.

flying-sheep:
ooh, this time the benchmark shows really nicely how much faster it is!

> **Ratio**: 0.01 **Benchmark**: preprocessing_log.time_regress_out('pbmc68k_reduced')

src/scanpy/preprocessing/_simple.py
flying-sheep: wild. That should really go back into scipy as C++ or Cython code or whatever they use.

ashish615:
> ooh, this time the benchmark shows really nicely how much faster it is!

Looks like preprocessing_log.time_regress_out('pbmc68k_reduced') , regress out those variables that is not inside it. It should regress_out  ['n_counts', 'percent_mito'] instead of ["total_counts", "pct_counts_mt"]. For the both commit it fails so report the same time.

src/scanpy/preprocessing/_simple.py
```diff
@@ -612,8 +612,16 @@ def normalize_per_cell(  # noqa: PLR0917
...
+DT = TypeVar("DT")
+
+
 @numba.njit(cache=True, parallel=True)
-def to_dense(shape, indptr, indices, data):
+def to_dense(
+    shape: tuple[int, int],
+    indptr: NDArray[np.integer],
+    indices: NDArray[np.integer],
+    data: NDArray[np.number],
```
flying-sheep, line 623:
```suggestion
data: NDArray[DT],
```

sorry, that’s what I meant! “the datatype stays the same” is the point here.

src/scanpy/preprocessing/_simple.py
```diff
@@ -612,6 +612,53 @@ def normalize_per_cell(  # noqa: PLR0917
...
 
+DT = TypeVar("DT")
+
+
+@numba.njit(cache=True, parallel=True)
+def to_dense(
+    shape: tuple[int, int],
+    indptr: NDArray[np.integer],
+    indices: NDArray[np.integer],
+    data: NDArray[np.number],
```
flying-sheep:
```suggestion
    data: NDArray[DT],
```

sorry, that’s what I meant! “the datatype stays the same” is the point here.

src/scanpy/preprocessing/_simple.py
```diff
@@ -715,14 +762,30 @@ def regress_out(
...
+    if not variable_is_categorical:
+        A = regres.to_numpy()
+        # if det(A.T@A) zero, then can not take inverse and regression fails.
+        # if fails then fall back to GLM implemetation of regression.
+        # Numba kernel is used to speedup regression using Linear Least Square method
+        # on regressor of type numpy array.
+        if np.linalg.det(A.T @ A) != 0:
+            res = numpy_regress_out(X, A)
+
+    # for categorical variable and failed the above code then run original code.
```
flying-sheep: I think this cleans up the comments a bit

```suggestion
        # if det(A.T@A) != 0 we can take the inverse and regress using a fast method.
        if np.linalg.det(A.T @ A) != 0:
            res = numpy_regress_out(X, A)

    # for a categorical variable or if the above checks failed,
    # we fall back to the GLM implemetation of regression.
```

flying-sheep (commented):
looks good from my side, do you have opinions, @Intron7?

Intron7:
It looks really good. I would love to have a closer look next week

ashish615:
> > ooh, this time the benchmark shows really nicely how much faster it is!
> 
> Looks like preprocessing_log.time_regress_out('pbmc68k_reduced') , regress out those variables that is not inside it. It should regress_out ['n_counts', 'percent_mito'] instead of ["total_counts", "pct_counts_mt"]. For the both commit it fails so report the same time.

@flying-sheep , can you look at the this benchmark test?

flying-sheep:
Hi, everything is OK with the benchmarks, `regress_out` would fail if called with variables that doesn’t exist.

The reason these are named differently is here: https://github.com/scverse/scanpy/blob/ad657edfb52e9957b9a93b3a16fc8a87852f3f09/benchmarks/benchmarks/_utils.py#L27-L31

I did that to be able to run benchmarks benchmarks on multiple data sets with the same code.

Intron7:
The `to_dense` function only works for csr matrices. I think we need another kernel that handles `csc` or just `.T` the resulting array if csc. I also get performance warnings about the matmul in the numba_kernel. I'll investigate this further.

ashish615:
This will work for csc matrix
```bash 
@numba.njit(cache=True, parallel=True)
def to_dense_csc(
    shape: tuple[int, int],
    indptr: NDArray[np.integer],
    indices: NDArray[np.integer],
    data: NDArray[DT],
) -> NDArray[DT]:
    """\
    Numba kernel for np.toarray() function
    """
    X = np.empty(shape, dtype=data.dtype)

    for c in numba.prange(shape[1]):
        X[:,c] = 0
        for i in range(indptr[c], indptr[c + 1]):
            X[indices[i],c] = data[i]
    return X
```

Intron7:
That kernel technically works but is way slower than the default scipy operation due to inefficient memory layout.

src/scanpy/preprocessing/_simple.py
```diff
@@ -612,6 +612,53 @@ def normalize_per_cell(  # noqa: PLR0917
...
+    Numba kernel for regress out unwanted sorces of variantion.
+    Finding coefficient using Linear regression (Linear Least Squares).
+    """
+
+    @numba.njit(cache=True, parallel=True)
+    def get_resid(
+        data: np.ndarray,
+        regressor: np.ndarray,
+        coeff: np.ndarray,
+    ) -> np.ndarray:
```
ilan-gold, line 651: Is the `DT` not important here as above?

src/scanpy/preprocessing/_simple.py
```diff
@@ -715,14 +762,28 @@ def regress_out(
             regres = regressors
         tasks.append(tuple((data_chunk, regres, variable_is_categorical)))
 
-    from joblib import Parallel, delayed
+    res = None
+    if not variable_is_categorical:
+        A = regres.to_numpy()
```
ilan-gold, line 767: I think the code before this should be refactored as well to include these new methods in the condition (i.e. at https://github.com/scverse/scanpy/pull/3110/files#diff-f9e0bdcdfc04622c421f0a5788bbba6ee0303750580d2915caba3239d799322fR759).  We can just check `regressors.to_numpy()` for non-zero determinant before iterating through the chunk list.  We don't need to make the list at all, from what I can tell, if `not variable_is_categorical` and the determinant is non-zero.  This will also make things cleaner because we will need fewer checks.  We can then use a boolean to check whether we should move into the next clause (which will include chunk creation) instead of checking `res=None` or not.  

Does this make sense?

ilan-gold (changes requested):

PR #3115, opened 2024-06-20
Unpin numpy 2

### LINKED ISSUE #3114: Numpy 2 support (unpin)

Other?

https://github.com/lmcinnes/pynndescent/pull/242 is released in [0.5.13](https://github.com/lmcinnes/pynndescent/releases/tag/release-0.5.13), so we can do that now

flying-sheep:
- [x] Closes #3114

Other things done:

- Rename some variables and reorder functions so the diff between the metrics is minimal for a future unification
- Skip seurat v3 tests with numpy 2 because skmisc isn’t ready: https://github.com/has2k1/scikit-misc/issues/34

  This will skip the seurat v3 tests for all but the minimum versions test job for now, but I think that’s OK?

### DIFF
--- a/src/scanpy/_utils/compute/is_constant.py
+++ b/src/scanpy/_utils/compute/is_constant.py
@@ -81,7 +81,7 @@ def is_constant(
 def _(a: NDArray, axis: Literal[0, 1] | None = None) -> bool | NDArray[np.bool_]:
     # Should eventually support nd, not now.
     if axis is None:
-        return (a == a.flat[0]).all()
+        return bool((a == a.flat[0]).all())
     if axis == 0:
         return _is_constant_rows(a.T)
     elif axis == 1:
@@ -116,9 +116,9 @@ def _is_constant_csr_rows(
     indptr: NDArray[np.integer],
     shape: tuple[int, int],
 ):
-    N = len(indptr) - 1
-    result = np.ones(N, dtype=np.bool_)
-    for i in range(N):
+    n = len(indptr) - 1
+    result = np.ones(n, dtype=np.bool_)
+    for i in range(n):
         start = indptr[i]
         stop = indptr[i + 1]
         if stop - start == shape[1]:
--- a/src/scanpy/metrics/_gearys_c.py
+++ b/src/scanpy/metrics/_gearys_c.py
@@ -1,3 +1,5 @@
+"""Geary's C autocorrelation."""
+
 from __future__ import annotations
 
 from functools import singledispatch
@@ -87,7 +89,7 @@ def gearys_c(
     Examples
     --------
 
-    Calculate Gearys C for each components of a dimensionality reduction:
+    Calculate Geary’s C for each components of a dimensionality reduction:
 
     .. code:: python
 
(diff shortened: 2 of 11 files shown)


src/scanpy/_utils/compute/is_constant.py
```diff
@@ -81,7 +81,7 @@ def is_constant(
 def _(a: NDArray, axis: Literal[0, 1] | None = None) -> bool | NDArray[np.bool_]:
     # Should eventually support nd, not now.
     if axis is None:
-        return (a == a.flat[0]).all()
+        return bool((a == a.flat[0]).all())
```
flying-sheep, line 84: This makes the function return a `bool` instead of a `np.bool_` scalar as the type hints say.

The two types behave almost identically but have a different repr with numpy 2, so this change passes the doctests.

src/scanpy/metrics/_gearys_c.py
```diff
@@ -1,3 +1,5 @@
+"""Geary's C autocorrelation."""
```
flying-sheep, line 1: I didn’t change anything in the two metrics modules, just reordered and reformatted (and replaced the removed `np.float_` with `np.float64`, which in numpy 1.x are identical)

src/scanpy/preprocessing/_simple.py
```diff
@@ -121,19 +121,19 @@ def filter_cells(
     >>> sc.pp.filter_cells(adata, min_genes=0)
     >>> adata.n_obs
     640
-    >>> adata.obs['n_genes'].min()
+    >>> int(adata.obs['n_genes'].min())
     1
```
flying-sheep, line 125: more repr changes: this is shown as `np.int64(1)` without the cast to int

tests/test_datasets.py
```diff
@@ -173,7 +173,7 @@ def test_download_failure():
                 *DS_MARKS[ds],
             ],
         )
-        for ds in set(sc.datasets.__all__) - DS_DYNAMIC
+        for ds in sorted(set(sc.datasets.__all__) - DS_DYNAMIC)
```
flying-sheep, line 176: This fixes `pytest -n auto`. Pytest complained locally that different jobs collected “different tests” which were just differently sorted.

ilan-gold:
> Skip seurat v3 tests with numpy 2 

Sorry does this mean it doesn't work with numpy 2?

flying-sheep:
It doesn’t. We could also

1. wait to merge this until `skmisc` has a new release or
2. throw a special error when people try to use seurat v3 with numpy 2

Sadly(?) Python doesn’t allow packages to add constraint to other packages’ dependencies, else we could tell the resolver that all currently release skmisc versions are incompatible with numpy 2

ilan-gold:
> wait to merge this until skmisc has a new release

Are we in a rush?  We are upper bounding right now anyway on the current release, no?  I would just wait.

flying-sheep:
We’re not upper-bounding in 1.10.1.

ilan-gold:
We discussed outside @flying-sheep - we will release with the bound then and just wait here?

flying-sheep:
release done, we can wait here …

ilan-gold:
Looks like there is a release out for the `scikit-misc`: https://github.com/has2k1/scikit-misc/compare/v0.4.0...main

Hopefully this fixes it

ilan-gold (approved):

PR #3138, opened 2024-07-02
feat: add layer argument for sc.tl.score_genes_cell_cycle

shinthor:
Simple addition of the `layer` argument which is already included in `sc.tl.score_genes` so that you can use your own layer instead of being restricted to what is stored in `adata.X` 

- [ ] Closes #


flying-sheep (changes requested):
Looks good! Could you add a release note? This is a feature, so it would be in `1.11.0.md`.

I wonder if it would be a good idea to do something like

```py
s_name, g2m_name, phase_name = (
    f"{layer}_{name}" if layer else name for name in ("S_score", "G2M_score", "phase") 
)
```

and then use that, then multiple layers could be scored?

flying-sheep:
pre-commit.ci autofix

flying-sheep:
don’t forget to mark this as ready for review if you think it’s ready!

PR #3147, opened 2024-07-08
Run benchmarks for off axis

flying-sheep:
- [ ] Closes #

@Intron7 

This adds benchmarks for the off axis for all parameters.

The off axis peak memory is lower since we `.pop` that layer. If you think that’s confusing I could change it.

Regarding big datasets for `_get_mean_var`, I already added that benchmark. You could maybe check if there’s anything else that should go into the `FastSuite`

### DIFF
--- a/benchmarks/benchmarks/_utils.py
+++ b/benchmarks/benchmarks/_utils.py
@@ -166,7 +166,22 @@ def get_count_dataset(
 def param_skipper(
     param_names: Sequence[str], params: tuple[Sequence[object], ...]
 ) -> ParamSkipper:
-    """Creates a decorator that will skip all combinations that contain any of the given parameters."""
+    """Creates a decorator that will skip all combinations that contain any of the given parameters.
+
+    Examples
+    --------
+
+    >>> param_names = ["letters", "numbers"]
+    >>> params = [["a", "b"], [3, 4, 5]]
+    >>> skip_when = param_skipper(param_names, params)
+
+    >>> @skip_when(letters={"a"}, numbers={3})
+    ... def func(a, b):
+    ...     print(a, b)
+    >>> run_as_asv_benchmark(func)
+    b 4
+    b 5
+    """
 
     def skip(**skipped: Set) -> Callable[[C], C]:
         skipped_combs = [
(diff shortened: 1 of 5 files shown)


benchmarks/benchmarks/_utils.py
```diff
@@ -1,29 +1,49 @@
...
     Dataset = Literal["pbmc68k_reduced", "pbmc3k", "bmmc", "lung93k"]
+    KeyX = Literal[None, "off-axis"]
+    KeyCount = Literal["counts", "counts-off-axis"]
 
 
 @cache
 def _pbmc68k_reduced() -> AnnData:
+    """A small datasets with a dense `.X`"""
     adata = sc.datasets.pbmc68k_reduced()
+    adata.var["mt"] = adata.var_names.str.startswith("MT-")
```
flying-sheep: this seems to be the same everywhere, maybe extract it into `_get_dataset_raw`

benchmarks/benchmarks/_utils.py
```diff
@@ -43,8 +63,11 @@ def _pbmc3k() -> AnnData:
     sc.pp.calculate_qc_metrics(
         adata, qc_vars=["mt"], percent_top=None, log1p=False, inplace=True
     )
+    assert isinstance(adata.X, sparse.csr_matrix)
     adata.layers["counts"] = adata.X.astype(np.int32, copy=True)
+    adata.layers["counts-off-axis"] = adata.layers["counts"].tocsc()
     sc.pp.log1p(adata)
+    adata.layers["off-axis"] = adata.X.tocsc()
```
flying-sheep: same here, this just differs based on `X` being dense or sparse

Intron7:
I think once we add more preprocessing functions with numba we can add more. I would maybe think about `hvg` with `seurat_v3`. Other than that this looks good to me. I would maybe add the calculate gene qc and normalize_total to Fast already

benchmarks/benchmarks/_utils.py
Intron7: I don't this this is a bad idea.

benchmarks/benchmarks/_utils.py
Intron7: I mean we could make a helper function for that.

Intron7 (changes requested):

PR #3155, opened 2024-07-12
Fix `layers` parameter in `score_genes` with `.raw`

Intron7:
This would fix this interaction
```py
sc.tl.score_genes(adata, layer="counts", gene_list=adata.var.index[:5])
```

This requires #3154


src/scanpy/_utils/__init__.py
```diff
@@ -247,14 +247,18 @@ def _check_array_function_arguments(**kwargs):
         )
 
 
-def _check_use_raw(adata: AnnData, use_raw: None | bool) -> bool:
+def _check_use_raw(
+    adata: AnnData, use_raw: None | bool, layer: str | None = None
```
flying-sheep:
```suggestion
    adata: AnnData, *, use_raw: None | bool, layer: str | None = None
```

since it’s no longer only one parameter aside from adata, better specify them by name

src/scanpy/tools/_score_genes.py
```diff
@@ -122,7 +122,7 @@ def score_genes(
     """
     start = logg.info(f"computing score {score_name!r}")
     adata = adata.copy() if copy else adata
-    use_raw = _check_use_raw(adata, use_raw)
+    use_raw = _check_use_raw(adata, use_raw, layer)
```
flying-sheep:
```suggestion
    use_raw = _check_use_raw(adata, use_raw=use_raw, layer=layer)
```

see above

tests/test_score_genes.py
```diff
@@ -257,6 +257,24 @@ def test_layer():
...
+    # score X
+    gene_set = adata.var_names[:10]
+    sc.tl.score_genes(adata, gene_set, score_name="X_score")
+    # score layer (`del` makes sure it actually uses the layer)
+    adata.layers["test"] = adata.X.copy()
+    adata.raw = adata
+    del adata.X
+    sc.tl.score_genes(adata, gene_set, score_name="test_score", layer="test")
+
+    np.testing.assert_array_equal(adata.obs["X_score"], adata.obs["test_score"])
```
flying-sheep: this is almost completely copied, please don’t do that, use `parametrize` instead

flying-sheep (changes requested):
Looks good, aside from the duplicated test code.

~~#3154 still needs to be backported before we can merge this, please do that or ask me for help doing it!~~

src/scanpy/_utils/__init__.py
Intron7: This would break a lot of functions if i would change this

flying-sheep: Then `layer=layer` should probably be added in most of these other places as well, so this PR should touch these anyway, no?

flying-sheep (changes requested):
- [x] see second paragraph in https://github.com/scverse/scanpy/pull/3155#pullrequestreview-2198787535
- [x] see https://github.com/scverse/scanpy/pull/3155#discussion_r1691269183

src/scanpy/_utils/__init__.py
Intron7: I think we should leave the plotting alone for the time beeing

flying-sheep (approved):

src/scanpy/_utils/__init__.py
```diff
@@ -247,14 +247,18 @@ def _check_array_function_arguments(**kwargs):
...
 
-def _check_use_raw(adata: AnnData, use_raw: None | bool) -> bool:
+def _check_use_raw(
+    adata: AnnData, use_raw: None | bool, *, layer: str | None = None
+) -> bool:
     """
     Normalize checking `use_raw`.
 
     My intentention here is to also provide a single place to throw a deprecation warning from in future.
     """
```
flying-sheep, line 257: Shouldn’t it be like this? Otherwise one overrides the other.

```suggestion
    """
    if use_raw and layer is not None:
        msg = "Cannot use both `use_raw` and `layer`."
        raise ValueError(msg)
```

flying-sheep: OK, that’s checked later

PR #3162, opened 2024-07-23
Fix tests for dask PCA

flying-sheep:
Fixes the current test failures.

I could also do a simpler version where I just rechunk in one way instead of adding parameters here.

### DIFF
--- a/tests/test_pca.py
+++ b/tests/test_pca.py
@@ -405,13 +405,10 @@ def test_mask(request: pytest.FixtureRequest, array_type):
     )
 
 
-def test_mask_order_warning(request):
+def test_mask_order_warning(request: pytest.FixtureRequest):
     if Version(ad.__version__) >= Version("0.9"):
-        request.node.add_marker(
-            pytest.mark.xfail(
-                reason="Not expected to warn in later versions of anndata"
-            )
-        )
+        reason = "Not expected to warn in later versions of anndata"
+        request.applymarker(pytest.mark.xfail(reason=reason))
 
     adata = ad.AnnData(X=np.random.randn(50, 5))
     mask = np.array([True, False, True, False, True])
--- a/tests/test_pca.py
+++ b/tests/test_pca.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import warnings
+from functools import wraps
 from typing import TYPE_CHECKING
 
 import anndata as ad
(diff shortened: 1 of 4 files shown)


ilan-gold:
Sorry what is the provenance of this fix? Why isn't the fix something in AnnData?  This seems somewhat complicated

tests/test_pca.py
```diff
@@ -62,13 +64,31 @@
 )
 
 
+def dask_array_dir_chunks(
+    a: object, *, conv: Callable[[object], DaskArray], dir: Literal["horiz", "vert"]
+):
```
ilan-gold: Return type?

tests/test_pca.py
```diff
@@ -62,13 +64,31 @@
...
+    arr = conv(a)
+    chunks = (arr.chunksize[0], -1) if dir == "horiz" else (-1, arr.chunksize[1])
+    return arr.rechunk(chunks)
+
+
+def svdify(array_types: Iterable[ParameterSet]) -> Generator[ParameterSet, None, None]:
```
ilan-gold: I think the name could be better - this is preparing for svd, not actually doing it?

flying-sheep:
Yeah, I was concerned about the complexity.

But AnnData doesn’t do anything wrong, it just uses chunks in both directions, which that one specific algorithm doesn’t support.

But I think in general we should test for 2D chunks, that’s why I think AnnData’s test helpers work as intended.

Any idea how to do this more simply?

ilan-gold:
@flying-sheep I mean why this came up all of a sudden in terms of provenance.  And furthermore, instead of `rechunk`, why not make a `chunks` argument or the like for the anndata function?

flying-sheep:
I’m not sure why it came up now, I saw changes to the chunking in https://github.com/scverse/anndata/pull/1550 and assumed that introduced it, but maybe not.

regarding complexity and parameters, we could simplify things by doing

```py
def maybe_rechunk_1d(a: NDArray | DaskArray | ...) -> NDArray | DaskArray | ...:
    if isinstance(a, DaskArray):
        return a.rechunk((a.chunksize[0], -1))
    return a
```

and using that in all the functions that fail (after running things through `array_type`). Would you prefer that?

flying-sheep:
I simplified this quite a bit. I decided against blocking this on an upstream `anndata` helper for multiple reasons:

1. we test on older anndata versions that won’t have that parameter immediately
2. needs some designing, e.g. the API should be able to do “use `ceil(shape[0] / 2)` as chunk size for dim 0, and `-1` (=full size) for dim 1”
3. it would make no sense for the non-dask array types, should we still add it for them?

PR #3167, opened 2024-07-25
Bug fix for scanpy.tl.score_genes

### LINKED ISSUE #3168: score_genes doesn’t produce the expected number of bins of equal or approximately equal size

### What happened?

(extracted from #3167)

Utilizing the murine hematopoietic progenitors from [Nestorowa et al., 2016](https://doi.org/10.1182/blood-2016-05-716480), as well as the regev_lab_cell_cycle_genes.txt, one issue is apparent.

Currently the code doesn’t produce the expected number of bins of equal or approximately equal size. Bin 24 is empty when n_bins = 25.
current_hist

### The current ranking system code within score_genes()
```py
n_items = int(np.round(len(obs_avg) / (n_bins - 1)))
obs_cut = obs_avg.rank(method="min") // n_items
```

### The modified code in #3167
```py
obs_avg.sort_values(ascending=True, inplace=True)
n_items = int(np.ceil(len(obs_avg) / (n_bins)))
rank = np.repeat(np.arange(n_bins), n_items)[:len(obs_avg)]
obs_cut = pd.Series(rank, index=obs_avg.index)
```

The modified code performs as expected producing 25 bins containing approximately equal number of genes. The last bin can have up to 24 less than expected because the total number of genes is not perfectly divisible by 25.
modified_hist

### Minimal code sample

```python
TODO
```

### Error output

_No response_

### Versions

<details>

```
Python 3.10.12 
scanpy==1.10.2 anndata==0.10.8 umap==0.5.6 numpy==1.26.4 scipy==1.14.0 pandas==2.2.2 scikit-learn==1.5.0 statsmodels==0.14.2 igraph==0.11.6 louvain==0.8.2 pynndescent==0.5.13
```

</details>

### LINKED ISSUE #3169: score_genes fails completely when the gene set has zero expression in some cells

### What happened?

(extracted from https://github.com/scverse/scanpy/pull/3167)

The code fails completely when the gene set has zero expression in some cells. The data used to illustrate this problem is located here: [test_data.h5ad](https://github.com/gmvaccaro/scanpy.tl.score_genes_fix/blob/main/test_data.h5ad) 

Since relatively few genes are typically selected for measurement, this will be a common occurrence.
hist

If the expression data is **scaled** the problem is more intrusive since it occurs in the middle of the expression range. The bins with no genes give a Nan result.
hist2

The proposed code modification in  #3167 has no problem with this data set, either scaled or unscaled.
image

### Minimal code sample

```python
TODO
```

### Error output

_No response_

### Versions

<details>

```
Python 3.10.12 
scanpy==1.10.2 anndata==0.10.8 umap==0.5.6 numpy==1.26.4 scipy==1.14.0 pandas==2.2.2 scikit-learn==1.5.0 statsmodels==0.14.2 igraph==0.11.6 louvain==0.8.2 pynndescent==0.5.13
```

</details>

gmvaccaro:
The score_genes procedure currently uses a ranking system to split genes into bins of similar expression levels. The current approach fails with some datasets.

- Currently the code doesn’t produce the expected number of bins of equal or approximately equal size, see #3168
- The code fails completely when the gene set has zero expression in some cells, see #3169

- [x] Closes #3168, closes #3169


flying-sheep:
Hi, that certainly seems like an improvement! I took the liberty to split out two bugs: #3168 and #3169

We had a similar fix #2875, which is hidden behind a `ctrl_as_ref` flag.

I think since that fix is not yet released, we should rename the flag and unify both fixes behind it.

Could you please
1. check if issue #3169 is already fixed by installing scanpy’s git version and setting `ctrl_as_ref=True`
2. Our backwards compatibility means that changes to the scoring need to be optional. This is why `tests/test_score_genes.py::test_score_with_reference` is failing here, and that’s what `ctrl_as_ref` is for.

	So since that option is not yet released and in order to fix the test, we should probably change that option to incorporate both improvements. We can rename it to reflect the two things it does.

flying-sheep:
pre-commit.ci autofix

gmvaccaro:
Hi, 

Thank you!

1. I've installed scanpy's git version and set ctrl_as_ref=True with the same results, it does not fix https://github.com/scverse/scanpy/issues/3169

2. I am not entirely sure what you mean here, would you mind elaborating? This bug fix is specific to the way score_genes() ranks the genes into bins & when there's a lot of zero expression in cells, I'm a bit unsure how this connects to ctrl_as_ref and what you would like me to do. 

Thanks again

flying-sheep:
Sure, I’ll happily elaborate!

As you can see, our test suite is failing. That’s because we have a test `tests/test_score_genes.py::test_score_with_reference` which checks if the scores emitted by the functions are exactly equal to older versions of the function. We have that test because we’d like people to be able to rely on consistent calculations.

Now of course we’d also like to fix things eventually, so we’d implement your fix behind an option (so people have to opt-in to the changes). Eventually, we’d switch to the new behavior (likely scanpy 2.0).

That’s why I propose to rename the `ctrl_as_ref` option and use it to gate both the change it already affects as well as your change.

----

Another thing: We need to test that this works as intended. Can you create a reproducer with built-in datasets (or synthetic data that you create using `numpy`) that would show the degraded binning behavior with the old behavior? We could then add a test like this:

```py
def test_score_genes():
    adata = TODO  # create test data here
    gene_list = TODO
    gene_pool = TODO
    gene_list, gene_pool, get_subset = _check_score_genes_args(
        adata, gene_list, gene_pool, use_raw=use_raw, layer=layer
    )

    bins = list(_score_genes_bins(
        gene_list,
        gene_pool,
        ctrl_as_ref=False,  # needs to be renamed
        ctrl_size=50,
        n_bins=25,
        get_subset=get_subset,
    ))

    assert 0 not in map(len, bins)
```

flying-sheep:
Hi @gmvaccaro, it would be awesome if you can give us some code to reproduce the problem!

Originally, you said

> Utilizing the murine hematopoietic progenitors from [Nestorowa et al., 2016](https://doi.org/10.1182/blood-2016-05-716480), as well as the regev_lab_cell_cycle_genes.txt, one issue is apparent.

but what code would reproduce the problem with the zero-sized bin from these data?

flying-sheep:
From the data you used, I assume you tried this, but I don’t get an empty bin:

https://gist.github.com/flying-sheep/b2ae449ab70a9358e07a82f284de5dca

gmvaccaro:
Hi! Here is the source code with data [Nestorowa et al., 2016](https://doi.org/10.1182/blood-2016-05-716480) with added diagnostic tests to produce these results.



https://github.com/gmvaccaro/scanpy.tl.score_genes_fix/blob/main/score_genes_diagnostics_tests2.ipynb

flying-sheep:
Wonderful, thank you!

flying-sheep:
Hm, I adapted your reproducer to use scanpy 1.10.3’s code and it doesn’t seem to be an issue: https://gist.github.com/flying-sheep/b2ae449ab70a9358e07a82f284de5dca#file-score_genes_diagnostics_tests2-ipynb

I’m going to assume that this is fixed in 1.10.3. If you can reproduce it with 1.10.3, we can reopen it!

PR #3181, opened 2024-07-31
Fix vector rasterizing bug (#1405)

### LINKED ISSUE #1405: sc.plt.umap save as pdf bug

---
I tried to save an umap result to a pdf, however, the dots in Umap are blurred and it just looks like png. Only the legend is in vector mode. I was wondering if this is a bug?

soymintc:
Vector images would be rasterized when `settings._vector_friendly` is True. This should be fixed so that when `settings._vector_friendly` is True, image is NOT rasterized, and when False, image should be rasterized.

- [x] Closes #1405
- [x] Tests included or not required because: simple reversion of booleans

- [x] Release notes not necessary because: did not create a release on this

### DIFF
--- a/src/scanpy/plotting/_tools/scatterplots.py
+++ b/src/scanpy/plotting/_tools/scatterplots.py
@@ -349,7 +349,7 @@ def embedding(
                 coords[:, 1],
                 coords[:, 2],
                 c=color_vector,
-                rasterized=settings._vector_friendly,
+                rasterized=(True - settings._vector_friendly),
                 marker=marker[count],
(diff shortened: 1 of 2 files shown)


flying-sheep:
pre-commit.ci autofix

flying-sheep:
Hi! Thank you for your contribution!

> Vector images would be rasterized when settings._vector_friendly is True.

Yes, that’s intentional, and that’s how `vector_friendly` is documented.

We won’t invert its behavior, but if you have a recommendation on how to improve the docs, please share!

PR #3206, opened 2024-08-12
Fix *Plot.style() methods

### LINKED ISSUE #1633: DotPlot `.style()` resets previous value assignments
Multiple calls to `sc.pl.Dotplot.style()` cause resetting of parameters. This issue can be observed as a change in the default colormap shown by #1632  Other methods that set default parameters are also affected like `.add_totals()`.

The following example should show the dots using the `Reds` colormap, but instead it uses the `winter` colormap because the second call sets the color map to `winter` if not given. This double call happens because when `sc.pl.dotplot()` is used (instead of `sc.pl.DotPlot`), internally a call to `.style()` is made and a subsequent explicit calls to `.style()` is required to tune the parameters as suggested in the documentation.

```python
markers = ['C1QA', 'PSAP', 'CD79A', 'CD79B', 'CST3', 'LYZ']
sc.pl.DotPlot(adata, markers, groupby='bulk_labels').style(cmap='Reds').style(dot_edge_color='black').show()
```
image

The problem is caused by the current implementation of `sc.pl.Dotplot.style()` that set the default parameters as:

```
  def style(
        self,
        cmap: str = DEFAULT_COLORMAP,
        color_on: Optional[Literal['dot', 'square']] = DEFAULT_COLOR_ON,
        dot_max: Optional[float] = DEFAULT_DOT_MAX,
        dot_min: Optional[float] = DEFAULT_DOT_MIN,
        .....
```

Where DEFAULT_* are the default values defined at the beginning of the file (see https://github.com/theislab/scanpy/blob/master/scanpy/plotting/_dotplot.py#L84) 

What is nice about this is that the documentation clearly shows the default values.

The downside is that optional values are assigned a default value that rewrites previous calls to style. Ideally, all optional values should be `None`, then is easy to know if a new value is passed or a previous call has already set a value. But, doing so will remove the defaults from the documentation.

@flying-sheep suggested to use a code he wrote to add default annotations to the documentation

https://github.com/theislab/scanpydoc/blob/875b441212830678cf9fc81c52f5af29bbb8715f/scanpydoc/elegant_typehints/formatting.py#L101-L107

flying-sheep:
- [x] Closes #1633

This adds relies on `_empty` to check which parameters have been specified, and only changes those


src/scanpy/plotting/_dotplot.py
```diff
@@ -303,18 +301,18 @@ def __init__(
...
-        dot_min: float | None = DEFAULT_DOT_MIN,
-        smallest_dot: float | None = DEFAULT_SMALLEST_DOT,
-        largest_dot: float | None = DEFAULT_LARGEST_DOT,
-        dot_edge_color: ColorLike | None = DEFAULT_DOT_EDGECOLOR,
-        dot_edge_lw: float | None = DEFAULT_DOT_EDGELW,
-        size_exponent: float | None = DEFAULT_SIZE_EXPONENT,
-        grid: float | None = False,
-        x_padding: float | None = DEFAULT_PLOT_X_PADDING,
-        y_padding: float | None = DEFAULT_PLOT_Y_PADDING,
+        cmap: str | None | Empty = _empty,
```
ilan-gold: Why can this be `None` now?

ilan-gold: and others

ilan-gold (commented):
Can we add a test for this?

src/scanpy/plotting/_dotplot.py
flying-sheep: `self.cmap` was always allowed to be None, the type annotations for `style` were just wrong.

flying-sheep:
@ilan-gold I cleaned up all the style parameters around where `None` is valid or not

ilan-gold (approved):

src/scanpy/plotting/_stacked_violin.py
```diff
@@ -814,6 +815,7 @@ def stacked_violin(
         var_group_labels=var_group_labels,
         var_group_rotation=var_group_rotation,
         layer=layer,
+        categories_order=order,
```
ilan-gold: What prompted this?

src/scanpy/plotting/_stacked_violin.py
```diff
@@ -823,7 +825,9 @@ def stacked_violin(
     )
 
     if dendrogram:
-        vp.add_dendrogram(dendrogram_key=dendrogram)
+        vp.add_dendrogram(
+            dendrogram_key=dendrogram if isinstance(dendrogram, str) else None
```
ilan-gold: The typing then of `add_dendrogram` with `dendrogram_key: str | None = None,` seems off, no? 

https://github.com/scverse/scanpy/blob/578e57297e8cfa48e5492e58b45567217a6ae43b/src/scanpy/plotting/_baseplot_class.py#L239C9-L245

ilan-gold (changes requested):

src/scanpy/plotting/_stacked_violin.py
ilan-gold: Oh apologies, I misread this.

flying-sheep: I think I got to the bottom of this. commit incoming

src/scanpy/plotting/_stacked_violin.py
flying-sheep: just a bug I found, I saw that `order` was unused.

ilan-gold: Ok, do you think it needs a test? Some way to check this? Perhaps overkill, though....up to you

flying-sheep:
OK, I undid the `dendrogram` change and instead added the effective typing for all the dendrogram-related functions.

I was confused by the unclear naming convention and thought that `dendrogram` was bool|str|None whereas `dendrogram_key` was str|None, but the names are just used interchangeably.

src/scanpy/plotting/_stacked_violin.py
flying-sheep: I changed my mind. `categories_order` is the standard name everywhere else so I just removed the unused `order` instead.

src/scanpy/plotting/_stacked_violin.py
```diff
@@ -727,10 +727,6 @@ def stacked_violin(
         See :func:`~seaborn.stripplot`.
     size
         Size of the jitter points.
-    order
-        Order in which to show the categories. Note: if `dendrogram=True`
-        the categories order will be given by the dendrogram and `order`
-        will be ignored.
```
ilan-gold: Also,isn't this a breaking change?  This is a public API, no?

src/scanpy/plotting/_stacked_violin.py
```diff
@@ -696,8 +696,10 @@ def stacked_violin(
     jitter: float | bool = StackedViolin.DEFAULT_JITTER,
     size: int | float = StackedViolin.DEFAULT_JITTER_SIZE,
     row_palette: str | None = StackedViolin.DEFAULT_ROW_PALETTE,
-    scale: Literal["area", "count", "width"] = StackedViolin.DEFAULT_DENSITY_NORM,
+    density_norm: Literal["area", "count", "width"] | Empty = _empty,
     yticklabels: bool = StackedViolin.DEFAULT_PLOT_YTICKLABELS,
+    # deprecated
+    scale: Literal["area", "count", "width"] | Empty = _empty,
```
ilan-gold: This is from an old PR? #2844? And you noticed it here? Should it be tested?

src/scanpy/plotting/_stacked_violin.py
ilan-gold: https://scanpy.readthedocs.io/en/stable/generated/scanpy.pl.stacked_violin.html So I see that we have the `categories_order` docstring (not sure how - common args?) but `order` still seems to be a public API?  And now it's gone: https://icb-scanpy--3206.com.readthedocs.build/en/3206/generated/scanpy.pl.stacked_violin.html?

ilan-gold (changes requested):
Might be off on the public API thing, but seems like it would require a minor version bump then....not sure if it's actually relevant anymore to the goal of this PR so might be worth separating out into its own PR.

src/scanpy/plotting/_stacked_violin.py
flying-sheep: It’s a public API … that used to be completely ignored. I can add a warning like

> `order` is ignored. use `categories_order` instead

src/scanpy/plotting/_stacked_violin.py
flying-sheep: I’ll start pulling unrelated things like this out!

src/scanpy/plotting/_stacked_violin.py
ilan-gold: Interesting ok! In that case, yea a warning would be better than removal

flying-sheep: OK, moved to #3252

ilan-gold (approved):

PR #3209, opened 2024-08-13
Less hacky plotting classes

flying-sheep:
An attempt to make the plotting classes less weird

contains a descriptor to allow the DEFAULT_BLAH class attributes to still work.

@ilan-gold please take a preliminary look


src/scanpy/plotting/_baseplot_class.py
```diff
@@ -74,141 +81,126 @@ class BasePlot:
...
+
+    @property
+    def height(self) -> float | None:
+        return self.figsize[1] if self.figsize is not None else None
+
+    @property
+    def vboundnorm(self) -> VBoundNorm:
+        return VBoundNorm(self.vmin, self.vmax, self.vcenter, self.norm)
+
+    # deprecated class vars
```
ilan-gold, line 156: Why deprecated? Because they are already set from elsewhere (the function-based API)?

ilan-gold: If so, why re-introduce?

ilan-gold: So these are part of the public API, the idea being to get rid of these eventually? https://scanpy.readthedocs.io/en/stable/api/generated/classes/scanpy.pl.StackedViolin.html

src/scanpy/plotting/_baseplot_class.py
```diff
@@ -221,17 +213,16 @@ def swap_axes(self, *, swap_axes: bool | None = True) -> Self:
...
 
         """
-        self.DEFAULT_CATEGORY_HEIGHT, self.DEFAULT_CATEGORY_WIDTH = (
-            self.DEFAULT_CATEGORY_WIDTH,
-            self.DEFAULT_CATEGORY_HEIGHT,
+        # TODO: this doesn’t make much sense
+        self.category_height, self.category_width = (
+            self.category_width,
+            self.category_height,
         )
```
ilan-gold, line 238: Agreed needs some investigating (what's the counter-factual when we remove it?)

src/scanpy/plotting/_dotplot.py
```diff
@@ -93,143 +96,108 @@ class DotPlot(BasePlot):
...
-            var_group_rotation=var_group_rotation,
-            layer=layer,
-            ax=ax,
-            vmin=vmin,
-            vmax=vmax,
-            vcenter=vcenter,
-            norm=norm,
-            **kwds,
-        )
+    DEFAULT_DOT_MAX: ClassVar[DefaultProxy[float | None]] = DefaultProxy("dot_max")
```
ilan-gold, line 136: Unfortunate that the types need to be duplicated i.e., `dot_max: float | None` and then here `DEFAULT_DOT_MAX: ClassVar[DefaultProxy[float | None]]`

ilan-gold: Maybe worth creating type variables?  Seems like that is more lines of code, though...

ilan-gold (commented):
Instead of explicit old all-caps defaults, could be possible it to make a lookup of properties -> old all-caps defaults?  And then have some sort of `__get_attr__` or `__get_attribute__` do the lookup and use the default?  This might also make typing a bit easier maybe? Not sure though, just a quick first pass

src/scanpy/plotting/_baseplot_class.py
flying-sheep: Yeah, the default would be accessible like `default = Baseplot.cmap`, and overridable in subclasses:

```py
class DotPlot(BasePlot):
    cmap = "Blues"
```

or just by setting it on the instance.

src/scanpy/plotting/_baseplot_class.py
flying-sheep: This just needs to be implemented in a way that actually respects the boolean parameter.

PR #3263, opened 2024-09-26
Implement sparse `covariance_eigh` PCA using Dask

### LINKED ISSUE #3226: Dask Sparse PCA

@Intron7 and I set up a nice implementation using GPU here: https://github.com/scverse/rapids_singlecell/pull/179/files#diff-483d6f872ddf4abd63e32af66a8566e0dcb40ba853a8672771dcbffb0235b7f9

It should be straightforward (and easier) without GPU.  Basic outline of computational-heavy steps:

1. Chunked covariance matrix calculation
2. Use already existing mean-var
3. Eigenvalue decomp

It works well on GPU so hopefully we can get nice performance on GPU.

I suspect using numba-in-dask for the covariance matrix calculation (similar to @Intron7 's kernel) would be super helpful.

flying-sheep:
- [x] Closes #3226

An alternative would be to subclass `PCA`, but that would involve erroring out or reimplementing all of its options.

Ideally #3267 would be merged first and this one integrated into its improved decision tree.


src/scanpy/preprocessing/_pca_dask_sparse.py
```diff
@@ -0,0 +1,195 @@
...
+        else:
+            self.n_components_ = self.n_components
+
+        self.n_samples_ = x.shape[0]
+        self.n_features_in_ = x.shape[1] if x.ndim == 2 else 1
+        self.dtype = x.dtype
+        covariance, self.mean_, _ = _cov_sparse_dask(x=x, return_mean=True)
+        self.explained_variance_, self.components_ = cp.linalg.eigh(
+            covariance, UPLO="U"
+        )
```
flying-sheep: this part doesn’t need to be in dask, `_cov_sparse_dask` and `transform` have to

src/scanpy/preprocessing/_pca_dask_sparse.py
```diff
@@ -0,0 +1,195 @@
...
+            (
+                x_part.indptr,
+                x_part.indices,
+                x_part.data,
+                x_part.shape[0],
+                n_cols,
+                gram_matrix,
+            ),
+        )
+        return gram_matrix[None, ...]  # need new axis for summing
```
flying-sheep: this can probably just be `(x_part.T @ x_part).to_dense()`

tests/test_pca.py
```diff
@@ -224,9 +235,12 @@ def test_pca_transform_randomized(array_type):
     assert np.linalg.norm(A_pca_abs - np.abs(adata.obsm["X_pca"])) < 2e-05
 
 
-def test_pca_transform_no_zero_center(array_type):
+def test_pca_transform_no_zero_center(request: pytest.FixtureRequest, array_type):
     adata = AnnData(array_type(A_list).astype("float32"))
     A_svd_abs = np.abs(A_svd)
+    if isinstance(adata.X, DaskArray) and issparse(adata.X._meta):
+        reason = "TruncatedSVD is not supported for sparse Dask yet"
+        request.applymarker(pytest.mark.xfail(reason=reason))
```
flying-sheep, line 284: @Intron7 you said it would be possible to adapt the code to do TruncatedSVD as well?

I think this PR is pretty clean, so maybe we can extend it in another PR?

ilan-gold: Agree, PR into this PR if we start pre-merge

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -0,0 +1,183 @@
...
+        import dask.array as da
+
+    if dtype is None:
+        dtype = np.float64 if np.issubdtype(x.dtype, np.integer) else x.dtype
+    else:
+        dtype = np.dtype(dtype)
+
+    def gram_block(x_part: CSMatrix):
+        gram_matrix: CSMatrix = x_part.T @ x_part
+        return gram_matrix.toarray()[None, ...]  # need new axis for summing
```
ilan-gold, line 183: Did we not consider speedups @Intron7 ?

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -0,0 +1,183 @@
...
+        self.components_ = self.components_.T[: self.n_components_, :]
+
+        self.explained_variance_ratio_ = self.explained_variance_ / np.sum(
+            self.explained_variance_
+        )
+        if self.n_components_ < min(self.n_samples_, self.n_features_in_):
+            self.noise_variance_ = self.explained_variance_[self.n_components_ :].mean()
+        else:
+            self.noise_variance_ = np.array([0.0])
+        self.explained_variance_ = self.explained_variance_[: self.n_components_]
```
ilan-gold, line 77:
```suggestion
        self.explained_variance_ = self.explained_variance_[:self.n_components_]
```
what's up here?

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -0,0 +1,183 @@
...
+
+        self.explained_variance_ratio_ = self.explained_variance_ratio_[
+            : self.n_components_
+        ]
+        return self
+
+    def transform(self: PCASparseFit, x: DaskArray) -> DaskArray:
+        if TYPE_CHECKING:
```
ilan-gold, line 103: Why do we need this internally?

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def fit_transform(self, x: DaskArray, y: DaskArray | None = None) -> DaskArray:
+        if y is None:
+            y = x
+        return self.fit(x).transform(y)
+
+
+@overload
+def _cov_sparse_dask(
+    x: DaskArray, *, return_gram: Literal[False] = False, dtype: DTypeLike | None = None
```
ilan-gold, line 131: We never use `return_gram` so let's remove

ilan-gold (changes requested):
Let's update the notebook as well.  Would be great to understand performance difference before merging + get rid of the horrible densifying operation in `dask.ipynb`

Also CI

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: this is how ruff formats this

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: because it makes my editor happy.

ideally someone would publish dask type hints or dask provides them instead of these hacks …

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: we do in the test I added.

src/scanpy/preprocessing/_pca/_dask_sparse.py
Intron7: I think this is quite well optimzed atm. I have to do tests on how multicore numbakernels work within dask. If it turns out to be faster we can still update.

src/scanpy/preprocessing/_pca/_dask_sparse.py
ilan-gold: Yup, but why are we testing it if we are not using it? Will we ever?

src/scanpy/preprocessing/_pca/_dask_sparse.py
ilan-gold: Ok, I guess I was asking because we don't use type hints in this function from `da.`

flying-sheep: It doesn’t have any result at runtime, so are you OK with keeping it that way until the dask typing situation improves?

flying-sheep:
> Let's update the notebook as well. Would be great to understand performance difference before merging + get rid of the horrible densifying operation in `dask.ipynb`

On it: https://github.com/scverse/scanpy-tutorials/pull/137

src/scanpy/preprocessing/_pca/_dask_sparse.py
ilan-gold: Sure, no worries, I just wanted to understand.  If it's just "my computer wants it" and we don't know why, that's ok.  I don't even care to investigate.  Just wanted to learn :)

ilan-gold: Maybe best to comment in this case

src/scanpy/preprocessing/_pca/_dask_sparse.py
ilan-gold: We do we need multicore kernels?  Dask workers are roughly mappable to cores.  Do we think the algorithm would be so different from optimized densification procedures on the GPU?

tests/test_pca.py
```diff
@@ -84,10 +81,20 @@ def wrapper(a: np.ndarray) -> DaskArray:
...
 
 
+def dask_compute(*adatas: AnnData, X: bool = False, pca: bool = True) -> AnnData:
+    """Helper to compute dask arrays in AnnData objects, e.g. before comparison."""
+    for adata in adatas:
+        if X and isinstance(adata.X, DaskArray):
+            adata.X = adata.X.compute()
+        if pca and isinstance(adata.obsm["X_pca"], DaskArray):
+            adata.obsm["X_pca"] = adata.obsm["X_pca"].compute()
+    return adatas
```
ilan-gold: I think we have `to_memory()` on the `AnnData` object for this purpose? Why do we need to check these elements?

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -0,0 +1,183 @@
...
+@dataclass
+class PCASparseDask:
+    n_components: int | None = None
+
+    def fit(self, x: DaskArray) -> PCASparseFit:
+        # this method makes `self` into the fitted version
+        self = cast(PCASparseFit, self)
+        assert isinstance(x.shape, tuple)
+        self.n_components_ = (
+            min(x.shape[:2]) if self.n_components is None else self.n_components
```
ilan-gold:
```suggestion
            min(x.shape) if self.n_components is None else self.n_components
```
why slice?

hatch.toml
```diff
@@ -15,7 +15,7 @@ scripts.clean = "git restore --source=HEAD --staged --worktree -- docs/release-n
 
 [envs.hatch-test]
 default-args = []
-features = ["test"]
+features = ["test", "dask-ml"]
```
ilan-gold: Why add this here?  I thought this PR specifically doesn't use an implementation?  Is this related to the other PR?

ilan-gold (changes requested):
Seems like still some issues with CI

src/scanpy/preprocessing/_pca/_dask_sparse.py
ilan-gold: What was the resolution here?

hatch.toml
flying-sheep: I could also add the dask extra. It’s a bit unclear which features should be in the env specification.

I think if we’d have added the hatch config to `pyproject.toml`, we could maybe use local overrides in `hatch.toml` to specify extras we’re currently working on, but right now, there’s no good way to locally modify the set of extras used here I think.

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: I’d say `if TYPE_CHECKING` is the comment here, it means “this is the version I want the type checker to see instead of the real code”, but OK, I’ll be more specific!

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: I think testing intermediate results is useful because it makes it easier to debug when something breaks after the “checkpoint”.

If we want to change the internal structure, it’s easy to remove or adapt that part of the test code.

hatch.toml
ilan-gold: Hmmm, ok.  I think have a fully featured env is good, but I guess I am wondering what this has to do with the PR.  Am I misunderstanding this PR i.e., do we need something with `dask-ml` to make this PR work?

ilan-gold: Ah, ok I see this is indeed duplicated from #3267.

tests/test_pca.py
flying-sheep: `np.array_equal` throws a warning when run on dask arrays.

I’ll try `to_memory`, great point!

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -0,0 +1,185 @@
...
+@dataclass
+class PCASparseDask:
+    n_components: int | None = None
+
+    def fit(self, x: DaskArray) -> PCASparseFit:
+        # this method makes `self` into the fitted version
+        self.__class__ = PCASparseFit
+        self = cast(PCASparseFit, self)
+
+        assert isinstance(x.shape, tuple)
```
ilan-gold: remove? no asserts?

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -0,0 +1,185 @@
...
+        self.n_components_ = (
+            min(x.shape) if self.n_components is None else self.n_components
+        )
+        self.n_samples_ = x.shape[0]
+        self.n_features_in_ = x.shape[1] if x.ndim == 2 else 1
```
ilan-gold: Maybe make a little cleaner now that we are planning on allowing `X` to have more than one dimension.  Should we raise a `ValueError` here?

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -0,0 +1,185 @@
...
+    CSMatrix = sparse.csr_matrix | sparse.csc_matrix
+
+
+@dataclass
+class PCASparseDask:
+    n_components: int | None = None
+
+    def fit(self, x: DaskArray) -> PCASparseFit:
+        # this method makes `self` into the fitted version
+        self.__class__ = PCASparseFit
```
ilan-gold: But why are these two different classes?

ilan-gold: Maybe add a comment?

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -0,0 +1,185 @@
...
+    )
+    gram_matrix = gram_matrix.astype(dtype)
+    gram_matrix /= x.shape[0]
+
+    cov_result = gram_matrix.copy() if return_gram else gram_matrix
+    cov_result -= mean_x[:, None] @ mean_x[None, :]
+
+    if return_gram:
+        return cov_result, gram_matrix, mean_x
+    return cov_result, mean_x
```
ilan-gold, line 206: I still don't get why this is here if the only place we actually use this is in the tests.  We just don't need to test it?

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: I think this one is just for typing, because somehow `shape` is inferred to be `None` here, but maybe I can add a comment for that. Should probably also fix that upstream.

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: good point!

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: This is the only way I could found to have sklearn’s fit pattern work together with typing.

The idea is that before `fit`, all `trailing_dunder_` attributes don’t exist. Making them all `| None` and initializing them with `None` would be extremely ugly and isn’t accurate: after `fit`, they’re never `None`.

My pattern makes many ways to use the class work accurately: `PCASparseDask().fit()` returns a `PCASparseFit` instance which has all the `trailing_dunder_` attributes.

The only thing that doesn’t work is

```py
pca = PCASparseDask()
pca.fit(...)  # `pca` is actually a `PCASparseFit` now, but the type system doesn’t know
pca = pca.fit(...)  # this works
```

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: https://github.com/scverse/scanpy/pull/3263#discussion_r1783023544

I don’t feel super strongly about this, but I feel like having a checkpoint is better than having none.

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: I’ll happily add more comments, but I don’t know what you want explained in them, please elaborate :D

ilan-gold: 1. I guess some condensed version of your entire previous comment would not be a bad comment
2. Do you think it's possible this might affect code readability/debugging?  I'm having trouble reasoning what you mean by "type system doesn't know" - why else would we do this?

For example:

```python
class Foo:
   bar = "zar"

class Baz:
     def __init__(self):
         self.__class__ = Foo

assert not isinstance(Baz(), Baz)
```

This sort of thing seems a bit confusing to people who might be debugging

src/scanpy/preprocessing/_pca/_dask_sparse.py
ilan-gold: Ok, We can leave it.

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: `PCASparseDaskFit` is a subclass of `PCASparseDask` though, so all’s good 😉 

I’ll add that comment!

ilan-gold: True, but I think the point of this was just for readability i.e.,:

```python
type(Baz())
# Foo
```

If you feel like it's ok, I don't mind.

flying-sheep:
Isaac says that lobpcg seems much less precise and is probably not worth it, so we should probably just adapt the warning to mention the imprecision and call it a day

src/scanpy/preprocessing/_pca/_dask_sparse.py
flying-sheep: Try `type(pathlib.Path())` :wink:

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -265,6 +265,17 @@ def pca(
         for chunk, start, end in adata_comp.chunked_X(chunk_size):
             chunk = chunk.toarray() if issparse(chunk) else chunk
             X_pca[start:end] = pca_.transform(chunk)
+    elif is_dask and issparse(X._meta):
+        from ._dask_sparse import PCASparseDask
+
+        if random_state != 0:
+            msg = f"Ignoring {random_state=} when using a sparse dask array"
+            warnings.warn(msg)
+        if svd_solver not in {None, "arpack", "auto"}:
```
ilan-gold: Can we make `svd_solver` to be `None` as the "our covariance implementation when dask+sparse" and the rest can be ignored for now, and then error later?

tests/test_pca.py
```diff
@@ -481,3 +519,27 @@ def test_pca_layer():
     )
     np.testing.assert_equal(X_adata.obsm["X_pca"], layer_adata.obsm["X_pca"])
     np.testing.assert_equal(X_adata.varm["PCs"], layer_adata.varm["PCs"])
+
+
+@needs.dask
+@pytest.mark.skipif(
+    pkg_version("anndata") < Version("0.10"),
+    reason="Old AnnData doesn’t have dask test helpers",
+)
```
ilan-gold:
```suggestion
# Skipping this test during min-deps testing shouldn't be an issue because the sparse-in-dask feature is not available on anndata<0.10 anyway
@pytest.mark.skipif(
    pkg_version("anndata") < Version("0.10"),
    reason="Old AnnData doesn’t have dask test helpers",
)
```

ilan-gold: It took me a bit of time to come up with that explanation, so would be good for the next person

ilan-gold (approved):

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: I don’t quite get that, can you elaborate please?

flying-sheep: OK, added `'covariance_eigh'` option.

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -44,12 +44,14 @@
 SvdSolvTruncatedSVDDaskML = Literal["tsqr", "randomized"]
 SvdSolvDaskML = SvdSolvPCADaskML | SvdSolvTruncatedSVDDaskML
 
-SvdSolvPCASklearn = Literal["auto", "full", "arpack", "randomized"]
+SvdSolvPCASklearn = Literal["auto", "full", "arpack", "covariance_eigh", "randomized"]
```
ilan-gold: Are we saying then that we support this in this PR? Maybe a different PR for this?  Not sure what the protocol for this sort of thing is

ilan-gold: i.e., scipy sparse without dask + covariance_eigh?

flying-sheep: The [sklearn PCA docs](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html) say:

> svd_solver{‘auto’, ‘full’, ‘covariance_eigh’, ‘arpack’, ‘randomized’}, default=’auto’
> <dl><dt>auto
> <dd>The solver is selected by a default ‘auto’ policy is based on X.shape and n_components: if the input data has fewer than 1000 features and more than 10 times as many samples, then the “covariance_eigh” solver is used. Otherwise, if the input data is larger than 500x500 and the number of components to extract is lower than 80% of the smallest dimension of the data, then the more efficient “randomized” method is selected. Otherwise the exact “full” SVD is computed and optionally truncated afterwards.

so I think now supporting explicit use of this is an oversight, since we have been supporting implicit use (through `auto`), and explicit use of `randomized` (the other implicit option picked by `auto`)

I can extract it to another PR so it gets its own release snippet if you want, but we can also avoid changing the docs for this twice in a row.

flying-sheep: OK, extracted to #3296

ilan-gold (commented):
Can we add this to the test suite? Also maybe the test suite should just depend on the literals in this file?

ilan-gold:
Can we add a test here to find out (or to at least know) how far our implementation here is from https://github.com/scverse/scanpy/pull/3296 for sparse?

flying-sheep:
Done!

ilan-gold (approved):
If everything passes CI, good to go from me

tests/test_pca.py
```diff
@@ -499,3 +549,45 @@ def test_pca_layer():
...
+        np.linalg.norm(
+            np.abs(adata_mem.obsm["X_pca"]) - np.abs(adata_dask.obsm["X_pca"].compute())
+        )
+        < 2e-05
+    )
+
+
+@needs.dask
+# Skipping this test during min-deps testing shouldn't be an issue because the sparse-in-dask feature is not available on anndata<0.10 anyway
+@pytest.mark.skipif(
```
ilan-gold: Might need to add this mark to the other test for CI actually!

flying-sheep: ah yes!

tests/test_pca.py
```diff
@@ -551,6 +551,23 @@ def test_pca_layer():
...
+
+    sc.pp.pca(adata_mem, svd_solver="covariance_eigh")
+    sc.pp.pca(adata_dask, svd_solver="covariance_eigh")
+
+    assert (
+        np.linalg.norm(
+            np.abs(adata_mem.obsm["X_pca"]) - np.abs(adata_dask.obsm["X_pca"].compute())
+        )
+        < 2e-05
+    )
```
ilan-gold: Why do we do this `np.abs` + `np.linalg.norm` business? Why not just check their closeness?

Intron7 (approved):
LGTM

tests/test_pca.py
flying-sheep: Math.

- `abs`: I think it’s mathematically equivalent to flip the components because they’re linearly independent.
- `norm`: no clue, but maybe some other way the PCA components can be mathematically equivalent while not being numerically identical

ilan-gold (approved):

tests/test_pca.py
flying-sheep: hm, maybe the norm isn’t necessary, I’ll try without

PR #3267, opened 2024-09-27
Use upstream sklearn PCA if possible

### LINKED ISSUE #2744: Broken test: PCA solver “LOBPCG” can’t handle small test data

### What happened?

The `pca_params` fixture randomly picks one of the possible solvers, but since a little while, the `"lobpcg"` solver fails for our small test data, [as explained by its docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.lobpcg.html):

> The LOBPCG code internally solves eigenproblems of the size 3k on every iteration by calling the dense eigensolver eigh, so if k is not small enough compared to n, it makes no sense to call the LOBPCG code. Moreover, if one calls the LOBPCG algorithm for 5k > n, it would likely break internally, so the code calls the standard function eigh instead. It is not that n should be large for the LOBPCG to work, but rather the ratio n / k should be large. It you call LOBPCG with k=1 and n=10, it works though n is small. The method is intended for extremely large n / k.

The workaround was to rerun failed tests until the fixture randomly returned another solver, which isn’t great.

I therefore  simply `xfail` that test for now to not block PRs in https://github.com/scverse/scanpy/pull/2745

### Minimal code sample

```bash
# try one of the following to reproduce (if necessary multiple times)
pytest --runxfail 'scanpy/tests/test_pca.py::test_pca_warnings[scipy_csr-zero_center-valid]'
pytest --runxfail 'scanpy/tests/test_pca.py::test_pca_warnings[scipy_csc-zero_center-valid]'
```

### Error output

(note that since the code is run with `warnings.simplefilter('error')`, the below is an error that fails the test)

```pytb
UserWarning: The problem size 5 minus the constraints size 0 is too small relative to the block size 4. Using a dense eigensolver instead of LOBPCG iterations.No output of the history of the iterations.
```

### Versions

<details>

```
-----
anndata     0.10.2
scanpy      1.10.0.dev156+gd1a2c8f8.d20231110
-----
PIL                 10.0.1
asciitree           NA
cloudpickle         3.0.0
cycler              0.12.1
cython_runtime      NA
dask                2023.10.0
dateutil            2.8.2
fasteners           0.19
h5py                3.10.0
igraph              0.10.8
iniconfig           NA
jinja2              3.1.2
joblib              1.3.2
kiwisolver          1.4.5
leidenalg           0.10.1
llvmlite            0.41.1
markupsafe          2.1.3
matplotlib          3.8.0
mpl_toolkits        NA
msgpack             1.0.7
natsort             8.4.0
numba               0.58.1
numcodecs           0.12.1
numpy               1.26.1
packaging           23.2
pandas              2.1.1
pluggy              1.3.0
psutil              5.9.6
py                  NA
pyparsing           3.1.1
pytest              7.4.3
pytz                2023.3.post1
scipy               1.11.3
session_info        1.0.0
setuptools          68.2.2
setuptools_scm      NA
six                 1.16.0
sklearn             1.3.2
sparse              0.14.0
tblib               3.0.0
texttable           1.7.0
threadpoolctl       3.2.0
tlz                 0.12.0
toolz               0.12.0
typing_extensions   NA
yaml                6.0.1
zarr                2.16.1
zipp                NA
-----
Python 3.11.5 (main, Sep  2 2023, 14:16:33) [GCC 13.2.1 20230801]
Linux-6.6.1-zen1-1-zen-x86_64-with-glibc2.38
-----
Session information updated at 2023-11-10 09:29
```

</details>

flying-sheep:
this PR fixes #2744 and

- moves `_pca_with_sparse` behind a check for scipy <1.4, which has @ivirshup’s port of that code https://github.com/scikit-learn/scikit-learn/pull/18689
- simplifies our logic around which parameters lead to which dispatch

  this makes it useful to get this in before #3263

- throws a warning when people use the `lobpcg` solver, since the closure of https://github.com/scikit-learn/scikit-learn/issues/12794#issuecomment-2118064158 makes it unlikely that we can count on that getting in any time soon

  I filed https://github.com/scikit-learn/scikit-learn/pull/30075. Depending on how that PR is received, we can update the warning here: either they like it, then we can remove the warning (we remove our legacy code once we depend on a scipy version that has lobpcg upstream), or they don’t, then we leave the warning for now and remove lobpcg support in the future.


tests/test_pca.py
```diff
@@ -180,48 +181,41 @@ def test_pca_warnings(array_type, zero_center, pca_params):
         raise
 
 
-# This warning test is out of the fixture because it is a special case in the logic of the function
-def test_pca_warnings_sparse():
-    for array_type in (sparse.csr_matrix, sparse.csc_matrix):
-        A = array_type(A_list).astype("float32")
-        adata = AnnData(A)
-        with pytest.warns(UserWarning, match="not work with sparse input"):
-            sc.pp.pca(adata, svd_solver="randomized", zero_center=True)
```
flying-sheep, line 190: this is intentional, it’s an old backwards compat hack, time to remove it.

tests/test_pca.py
```diff
@@ -180,48 +181,41 @@ def test_pca_warnings(array_type, zero_center, pca_params):
...
+
+    with (
+        pytest.warns(UserWarning, match="Ignoring.*'randomized'")
+        if sparse.issparse(adata.X)
+        else nullcontext()
+    ):
         sc.pp.pca(
             adata,
-            n_comps=5,
+            n_comps=4,
```
flying-sheep, line 195: this is iffy, the `_pca_with_sparse` was able to do 5 earlier. probably need to fix this.

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -1,33 +1,52 @@
...
+    T = TypeVar("T", bound=LiteralString)
+
+
+SvdSolvPCADaskML = Literal["auto", "full", "tsqr", "randomized"]
+SvdSolvTruncatedSVDDaskML = Literal["tsqr", "randomized"]
+SvdSolvDaskML = Union[SvdSolvPCADaskML, SvdSolvTruncatedSVDDaskML]
+
+SvdSolvPCASklearn = Literal["auto", "full", "arpack", "randomized"]
+SvdSolvTruncatedSVDSklearn = Literal["arpack", "randomized"]
+SvdSolvPCASparseSklearn = Literal["arpack"]
```
flying-sheep, line 49: also iffy, `_pca_with_sparse` was able to handle `lobpcg`, upstream only supports `arpack`

tests/test_pca.py
flying-sheep: actually that was the backwards compat code I eliminated, so it’s fine.

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: fixed

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -265,35 +280,42 @@ def pca(
...
             chunk = chunk.toarray() if issparse(chunk) else chunk
             X_pca[start:end] = pca_.transform(chunk)
-    elif (not issparse(X) or svd_solver == "randomized") and zero_center:
-        if is_dask:
-            from dask_ml.decomposition import PCA
-
-            svd_solver = _handle_dask_ml_args(svd_solver, "PCA")
+    elif zero_center:
+        if issparse(X) and (
+            pkg_version("scikit-learn") < Version("1.4") or svd_solver == "lobpcg"
```
ilan-gold, line 291: Shouldn't `lobpcg` be included in the above `Literal`s somewhere?

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -265,35 +280,42 @@ def pca(
...
+            if svd_solver not in {"lobpcg", "arpack"}:
+                if svd_solver is not None:
+                    msg = (
+                        f"Ignoring {svd_solver=} and using 'arpack', "
+                        "sparse PCA with sklearn < 1.4 only supports 'lobpcg' and 'arpack'."
+                    )
+                    warnings.warn(msg)
+                svd_solver = "arpack"
+            elif svd_solver == "lobpcg":
+                msg = f"{svd_solver=} for sparse relies on legacy code and will not be supported in the future."
```
ilan-gold: So shouldn't this be amended to something like "may not" since it could if we upstream this?

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -486,13 +444,25 @@ def _handle_dask_ml_args(svd_solver: str, method: str) -> str:
...
+def _handle_sklearn_args(
+    svd_solver: str | None, method: Literal["PCA"]
+) -> SvdSolvPCASklearn: ...
+@overload
+def _handle_sklearn_args(
+    svd_solver: str | None, method: Literal["TruncatedSVD"]
+) -> SvdSolvTruncatedSVDSklearn: ...
+@overload
+def _handle_sklearn_args(
+    svd_solver: str | None, method: Literal["PCA (with sparse input)"]
```
ilan-gold: Why make this a string? Why not use the backing array type as an argument?

ilan-gold: (just not in general a fan of strings as arguments when concrete types are available, or rather like to minimize them)

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -501,11 +471,19 @@ def _handle_sklearn_args(svd_solver: str | None, method: str) -> str:
     return _handle_x_args("sklearn", svd_solver, method, method2args, method2default)
 
 
-def _handle_x_args(lib, svd_solver: str | None, method, method2args, method2default):
+def _handle_x_args(
+    lib: str,
+    svd_solver: str | None,
```
ilan-gold, line 496: is `str` right here?

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -501,11 +471,19 @@ def _handle_sklearn_args(svd_solver: str | None, method: str) -> str:
     return _handle_x_args("sklearn", svd_solver, method, method2args, method2default)
 
 
-def _handle_x_args(lib, svd_solver: str | None, method, method2args, method2default):
+def _handle_x_args(
+    lib: str,
```
ilan-gold: `Literal`?

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -501,11 +471,19 @@ def _handle_sklearn_args(svd_solver: str | None, method: str) -> str:
...
+    svd_solver: str | None,
+    method: str,
```
ilan-gold: Worth making `Literal` and re-using above and below?

src/scanpy/preprocessing/_pca/_compat.py
```diff
@@ -0,0 +1,79 @@
...
+    from numpy.typing import NDArray
+    from scipy import sparse
+    from sklearn.decomposition import PCA
+
+    from .._utils import AnyRandom
+
+    CSMatrix = sparse.csr_matrix | sparse.csc_matrix
+
+
+def _pca_with_sparse(
```
ilan-gold: maybe worth renaming?

src/scanpy/preprocessing/_pca/_compat.py
```diff
@@ -0,0 +1,79 @@
...
+    x: CSMatrix,
+    n_pcs: int,
+    *,
+    solver: Literal["arpack", "lobpcg"],
```
ilan-gold, line 30: I think this function no longer is used in the `arpack` case?

tests/test_pca.py
```diff
@@ -127,9 +125,7 @@ def pca_params(
                 else {"tsqr", "randomized"}
             )
         elif array_type in {sparse.csr_matrix, sparse.csc_matrix}:
-            svd_solver = (
-                {"lobpcg", "arpack"} if zero_center else {"arpack", "randomized"}
-            )
+            svd_solver = {"arpack"} if zero_center else {"arpack", "randomized"}
```
ilan-gold: Are we losing `lobpcg` here?

ilan-gold: Huh, looking at it, we never had a `lobpcg` test ok

ilan-gold:
Some comments and questions, some changes needed.   Overall, much cleaner :)

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: Not if we want to deprecate it. This is the legacy branch with its own manual argument handling.

As said above: if they like my scikit-learn PR, we can handle the warnings earlier, and yes, then it would be added.

ilan-gold: Ok fair, that makes sense.

ilan-gold: Sorry wait, but then we probably shouldn't make a decision about deprecating it yet, no? maybe give your PR some time?

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: I’ll wait for the reaction to that PR. If we upstream it, there will be no warning at all. Users don’t need to care then, they will just be silently upgraded from our implementation to the upstream one some time.

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: `svd_solver`? If you call this function with `"hello i do not exist"` then you’ll get the default, so `str` is right.

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: See above.

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: That’s a free form string that gets interpolated into the warning message, so anything is OK here.

src/scanpy/preprocessing/_pca/__init__.py
ilan-gold: Sorry, no `method` having the type within it

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: done

src/scanpy/preprocessing/_pca/__init__.py
ilan-gold: I guess the fact that `library` probably has finite possibilities was the motivation but it is just a message so free-form

src/scanpy/preprocessing/_pca/_compat.py
flying-sheep: It is, after this line:
https://github.com/scverse/scanpy/blob/f8986683b243fa32857a26565460d29620766041/src/scanpy/preprocessing/_pca/__init__.py#L291
…it’s used here:
https://github.com/scverse/scanpy/blob/f8986683b243fa32857a26565460d29620766041/src/scanpy/preprocessing/_pca/__init__.py#L302-L303

src/scanpy/preprocessing/_pca/_compat.py
flying-sheep: done

src/scanpy/preprocessing/_pca/_compat.py
ilan-gold: Ah apologies, must have misread the line as being within the conditional block for block for `lobpcg`

tests/test_pca.py
flying-sheep: This test is (sometimes) the lobpcg test, so we had one. And we still have one, this will run the code (sometimes) with the lobpcg arg.

This change just means that lobpcg went from “we expect no warning” to “we expect a warning”

If we remove the warning, I’ll move it back in here.

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: Update from standup yesterday: Isaac said that lobpcg is not accurate, so I proposed that we get rid of it and everyone seemed in favor.

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: updated

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: I see! Can do!

ilan-gold (approved):

PR #3275, opened 2024-10-08
Catch PerfectSeparationWarning during regress_out

### LINKED ISSUE #3260: `regress_out` does not capture PerfectSeparation with statsmodels >= 0.14

### What happened?

Hi,
since statsmodels 0.14, perfect separation no longer raises an error but a warning (see [function doc here](https://www.statsmodels.org/stable/generated/statsmodels.genmod.generalized_linear_model.GLM.html#statsmodels-genmod-generalized-linear-model-glm)). Because scanpy currently only catches the now-outdated error (instead of catch the warning), users may see many warnings from `regress_out` when no perfect separation exists (see usage in scanpy [here](https://github.com/scverse/scanpy/blob/main/src/scanpy/preprocessing/_simple.py#L759-L761)). It seems to follow on the heels of [this issue](https://github.com/statsmodels/statsmodels/issues/2680) in statsmodels. I propose to implement that the warning is caught just as the errors were being caught.
Cheers,
Jesko

### Minimal code sample

```python
import anndata as ad
import scanpy as sc
import numpy as np
import pandas as pd
adata = ad.AnnData(np.array([[0,0,1,1]]).T, obs=pd.DataFrame({"a":[0,0,1,1]}))
sc.pp.regress_out(adata, "a")
```

### Error output

```pytb
.../statsmodels/genmod/generalized_linear_model.py:1257: PerfectSeparationWarning: Perfect separation or prediction detected, parameter may not be identified
  warnings.warn(msg, category=PerfectSeparationWarning)
```

### Versions

<details>

```
anndata                     0.10.4
scanpy                      1.9.6
statsmodels                 0.14.0
```

</details>

jeskowagner:
- [x] Closes #3260
- [x] Tests included or not required because: minor change to maintain compat with `statsmodels`>=14.0

Inspired by @mwaskom's [fix for seaborn](https://github.com/mwaskom/seaborn/pull/3356), which promotes the warning to an error and catches it (in this case with the same logic `scanpy` was using for prior versions of `statsmodels`).


Zethson:
Thanks! I'll check it out soon.

Zethson (approved):
Great, thank you!

src/scanpy/preprocessing/_simple.py
```diff
@@ -754,11 +754,17 @@ def _regress_out_chunk(data):
         else:
             regres = regressors
         try:
-            result = sm.GLM(
-                data_chunk[:, col_index], regres, family=sm.families.Gaussian()
-            ).fit()
-            new_column = result.resid_response
-        except PerfectSeparationError:  # this emulates R's behavior
+            err_classes = (sme.PerfectSeparationError,)
```
flying-sheep, line 757: this needs to be outside of the try-catch bock

src/scanpy/preprocessing/_simple.py
```diff
@@ -754,11 +754,17 @@ def _regress_out_chunk(data):
...
+            with warnings.catch_warnings():
+                if hasattr(sme, "PerfectSeparationWarning"):
```
flying-sheep, line 759: when is that not the case?

src/scanpy/preprocessing/_simple.py
Zethson: Huh, I'll have another look.

src/scanpy/preprocessing/_simple.py
Zethson: older statsmodels versions I assume.

PR #3280, opened 2024-10-11
scale function(_get_mean_var) updated for dense array, speedup upto ~4.65x

kaushalprasadhial:
We had created this PR before https://github.com/scverse/scanpy/pull/3099. This one is the same PR with editing enabled for maintainers.
Hi,
We are submitting PR for speed up of the _get_mean_var function.
| | Time(sec) |
| -- | -- |
| Original | 18.49 |
| Updated | 3.97 |
| Speedup | 4.65743073 |

```python
import time
import numpy as np

import pandas as pd

import scanpy as sc

import os
import wget

import warnings



# marker genes

# filtering cells

# filtering genes

# PCA

# t-SNE

# k-means

# Gene ranking

# Number of parallel jobs

start=time.time()
tr=time.time()
adata = sc.read(input_file)
adata.var_names_make_unique()
adata.shape



tr=time.time()
USE_FIRST_N_CELLS = 1300000
adata.shape

tr=time.time()
sc.pp.log1p(adata)


for marker in markers:

ts=time.time()


#scale

ts=time.time()
sc.pp.scale(adata)
```
add timer around _get_mean_var call
https://github.com/scverse/scanpy/blob/706d4ef65e5d65e04b788831e7fd65dbe6b2a61f/scanpy/preprocessing/_scale.py#L167
we can also create _get_mean_var_std function that return std as well so we don't require to compute it in scale function(L168-L169).


src/scanpy/preprocessing/_utils.py
```diff
@@ -33,17 +33,53 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
 def _get_mean_var(
     X: _SupportedArray, *, axis: Literal[0, 1] = 0
 ) -> tuple[NDArray[np.float64], NDArray[np.float64]]:
-    if isinstance(X, sparse.spmatrix):
-        mean, var = sparse_mean_variance_axis(X, axis=axis)
+    if isinstance(X, np.ndarray):
+        n_threads = numba.get_num_threads()
+        mean, var = _compute_mean_var(X, axis=axis, n_threads=n_threads)
```
ilan-gold: What happened to sparse?

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,17 +33,53 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
     else:
         mean = axis_mean(X, axis=axis, dtype=np.float64)
         mean_sq = axis_mean(elem_mul(X, X), axis=axis, dtype=np.float64)
         var = mean_sq - mean**2
     # enforce R convention (unbiased estimator) for variance
     if X.shape[axis] != 1:
         var *= X.shape[axis] / (X.shape[axis] - 1)
+
```
ilan-gold: Please remove erroneous diffs

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,17 +33,53 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
...
     return mean, var
 
+@numba.njit(cache=True, parallel=True)
+def _compute_mean_var(
```
ilan-gold:
```suggestion
def _compute_mean_var_dense(
```
and elsewhere

ilan-gold (changes requested):

src/scanpy/preprocessing/_utils.py
Intron7: I still try to integrate all the stuff it's a mess though.

src/scanpy/preprocessing/_utils.py
```diff
@@ -33,7 +33,10 @@ def _(X: np.ndarray, *, axis: Literal[0, 1], dtype: DTypeLike) -> np.ndarray:
 def _get_mean_var(
     X: _SupportedArray, *, axis: Literal[0, 1] = 0
 ) -> tuple[NDArray[np.float64], NDArray[np.float64]]:
-    if isinstance(X, sparse.spmatrix):
+    if isinstance(X, np.ndarray):
```
flying-sheep, line 37: instead of adding a second code path that handles np.ndarray, you should replace the existing one above:

```py
@axis_mean.register(np.ndarray)
def _(X: np.ndarray, ...): ...
```

src/scanpy/preprocessing/_utils.py
```diff
@@ -45,6 +48,41 @@ def _get_mean_var(
...
+                    sums[i, c] += value
+                    sums_squared[i, c] += value * value
+        for c in numba.prange(X.shape[axis_i]):
+            sum_ = sums[:, c].sum()
+            mean[c] = sum_ / n
+            var[c] = (sums_squared[:, c].sum() - sum_ * sum_ / n) / (n - 1)
+    else:
+        axis_i = 0
+        mean = np.zeros(X.shape[axis_i], dtype=np.float64)
+        var = np.zeros(X.shape[axis_i], dtype=np.float64)
```
flying-sheep, line 76: pull this out of the if branch

src/scanpy/preprocessing/_utils.py
```diff
@@ -45,6 +48,41 @@ def _get_mean_var(
...
+        for r in numba.prange(X.shape[0]):
+            for c in range(X.shape[1]):
+                value = X[r, c]
+                mean[r] += value
+                var[r] += value * value
+        for c in numba.prange(X.shape[0]):
+            mean[c] = mean[c] / X.shape[1]
+            var[c] = (var[c] - mean[c] ** 2) / (X.shape[1] - 1)
```
flying-sheep, line 84: there is no return statement

flying-sheep (changes requested):
Hi! This seems a bit unfinished. If you need our help with a local test environment or so, we’re happy to help.

src/scanpy/preprocessing/_utils.py
kaushalprasadhial: i have updated the code please check

flying-sheep: Well, the tests aren’t passing, so it still doesn’t seem to be working

src/scanpy/preprocessing/_utils.py
flying-sheep: I mean these assignments. When you have two branches, and both start with the same 3 lines, just do those before the if statement instead.

Intron7:
This also doesn't work for F-Arrays

PR #3283, opened 2024-10-14
Remove 3.9 support

### LINKED ISSUE #3282: Update minimum Python version to 3.10
See https://scientific-python.org/specs/spec-0000/

flying-sheep:
- [x] Closes #3282


src/scanpy/preprocessing/_scrublet/core.py
```diff
@@ -28,13 +27,7 @@
 __all__ = ["Scrublet"]
 
 
-if sys.version_info > (3, 10):
```
flying-sheep, line 31: this check was wrong, should have been `>=`: all the `kw_only` features were added in 3.10: https://docs.python.org/3/library/dataclasses.html#dataclasses.KW_ONLY

Zethson (approved):
Interesting that we weren't using `from __future__ import annotations` everywhere.

ilan-gold (approved):

flying-sheep:
@Zethson we did though! Where do you see `from __future__ import annotations` being added?

The changes here aren’t in a type position (like `param: int`), so they have a runtime effect that isn’t influenced by `from __future__ import annotations`:

- `isinstance(obj, A | B)` results in the second argument having the type `types.UnionType` instead of `tuple`, which I guess 3.9 doesn’t support
- `SomeAlias = A | B` is also evaluated when the module is imported (when not in a `if TYPE_CHECKING` block), or at least has to be parsed at module import time

The same would be true for `cast(A | B, thing)` but I think we don’t do that anywhere because we don’t run MyPy here yet.

PR #3284, opened 2024-10-15
Speedup (~20x) of scanpy.pp.regress_out function using Linear Least Square method.

kaushalprasadhial:
This pull request is same as https://github.com/scverse/scanpy/pull/3110 with allowed edits to maintainers

Hi,
We are submitting PR for speed up of the regress_out function. Here we finding coefficient using Linear regression (Linear Least Squares) rather then GLM for non categorical data.

| | Time(sec) |
| -- | -- |
| Original | 297 |
| Updated | 14.91 |
| Speedup | 19.91 |

```python
import time
import numpy as np

import pandas as pd

import scanpy as sc

import os
import wget

import warnings



# marker genes

# filtering cells

# filtering genes

# PCA

# t-SNE

# k-means

# Gene ranking

# Number of parallel jobs

start=time.time()
tr=time.time()
adata = sc.read(input_file)
adata.var_names_make_unique()
adata.shape



tr=time.time()
USE_FIRST_N_CELLS = 1300000
adata.shape

tr=time.time()
sc.pp.log1p(adata)


for marker in markers:

ts=time.time()
```

- [ ] Closes #


src/scanpy/preprocessing/_simple.py
```diff
@@ -623,6 +623,77 @@ def normalize_per_cell(
...
+
+
+@numba.njit(cache=True, parallel=True)
+def _to_dense_csc_numba(
+    indptr: NDArray,
+    indices: NDArray,
+    data: NDArray,
+    X: NDArray,
+    shape: tuple[int, int],
+):
```
ilan-gold:
```suggestion
) -> None:
```
and elsewhere

src/scanpy/preprocessing/_simple.py
```diff
@@ -623,6 +623,77 @@ def normalize_per_cell(
...
+
+    @numba.njit(cache=True, parallel=True)
+    def get_resid(
+        data: np.ndarray,
+        regressor: np.ndarray,
+        coeff: np.ndarray,
+    ) -> np.ndarray:
+        for i in numba.prange(data.shape[0]):
+            data[i] -= regressor[i] @ coeff
+        return data
```
ilan-gold: Why is this inlined?  I think we need either a comment for this or not do it.

src/scanpy/preprocessing/_simple.py
```diff
@@ -720,14 +791,29 @@ def regress_out(
         regres = regressors_chunk[idx] if variable_is_categorical else regressors
         tasks.append(tuple((data_chunk, regres, variable_is_categorical)))
 
-    from joblib import Parallel, delayed
+    res = None
+    if not variable_is_categorical:
```
ilan-gold: So we should be able to test whether the old implementation and the new one are equal by parametrizing a test that has `variable_is_categorical`.  In any case, this fact (whether or not the old and new are equal) needs to be checked.  If they are not the same, then we need to put this behind a flag.

src/scanpy/preprocessing/_simple.py
```diff
@@ -720,14 +791,29 @@ def regress_out(
...
+        A = regres.to_numpy()
```
ilan-gold: It seems like this `regres` variable is [coming from the above loop](https://github.com/scverse/scanpy/pull/3284/files#diff-f9e0bdcdfc04622c421f0a5788bbba6ee0303750580d2915caba3239d799322fR788-R792) which could be bad.  I think the `A` variable should be clearly created from something that isn't declared in a loop, unless this is intentional (I assume not).  I would also move that loop into the `variable_is_categorical` block below.  That will make reasoning what `A` is easier anyway because it will need to be clearly declared as `regressor`

ilan-gold: More generally, I would split this and the above code into two blocks `variable_is_categorical` and `not variable_is_categorical`

ilan-gold (changes requested):

src/scanpy/preprocessing/_simple.py
Intron7: We also still have to check the logic and figure out a better task dispatch tree for the numpy shortcut. I also have to investigate How we format the array (C/F) for both versions

Intron7:
Reminder: Check Benchmarks

src/scanpy/preprocessing/_simple.py
Intron7: I think that the implementation as both working. Because we clear the tests that are out there. Because in the end we test `variable_is_categorical` and the algorithm is the same.

src/scanpy/preprocessing/_simple.py
Intron7: I think this is out of scope for this update. To resolve the ideas you have would result in a complete rewrite of the regress_out function. There are more things that need to be updated for this to happen. So I would leave the logic as is for now.

Intron7 (approved):

src/scanpy/preprocessing/_simple.py
flying-sheep: So we do have that test? Which test function is that?

flying-sheep (commented):
- @Intron7 what’s up with the benchmarks?
- There’s a TODO left

If it’s a speedup across the board, and the test @ilan-gold asks for already exists, it looks clean enough to me.

Intron7:
Yes the todo will be a seperate PR once this one is merged. I have the kernel ready

flying-sheep:
what about the benchmarks?

src/scanpy/preprocessing/_simple.py
ilan-gold: @Intron7 I think all that is required here is to use `regressors` instead of `regres`.  And then move the loop after this if-block into an `else` block.  It's unnecessary computation in the case that `not variable_is_categorical and np.linalg.det(A.T @ A) != 0`.

Previously, there were no `if-then` blocks around `variable_is_categorical` so seems pretty in-scope since we're adding those here.

ilan-gold (changes requested):

Intron7:
@flying-sheep I don't know why the benchmarks are failing. But I think we should already be seeing improvement. If not we have to adjust the big dataset to have a hvg column and work from there.

src/scanpy/preprocessing/_utils.py
```diff
@@ -159,3 +159,46 @@ def sample_comb(
...
+    Numba kernel for np.toarray() function
+    """
+    out = np.zeros(X.shape, dtype=X.dtype, order=order)
+    if X.format == "csr":
+        _to_dense_csr_numba(X.indptr, X.indices, X.data, out, X.shape)
+    elif X.format == "csc":
+        _to_dense_csc_numba(X.indptr, X.indices, X.data, out, X.shape)
+    else:
+        out = X.toarray(order=order)
+    return out
```
ilan-gold, line 179: We should cover these things

ilan-gold (changes requested):

ilan-gold:
It looks like the benchmark doesn't like how little time the new implementation takes:

1.59±0.02s>>>>>>>>6.89±0.9ms

src/scanpy/preprocessing/_utils.py
Intron7: Do you want a test?

src/scanpy/preprocessing/_simple.py
```diff
@@ -678,7 +707,6 @@ def regress_out(
 
     if issparse(X):
         logg.info("    sparse input is densified and may lead to high memory use")
-        X = X.toarray()
```
ilan-gold, line 681: As a follow up probably worth replacing any instance of this in scanpy with the approporiate numba'ed version

ilan-gold (approved):

ilan-gold (changes requested):
https://github.com/scverse/scanpy/pull/3284#discussion_r1827868863 has this ben addressed?

We need to ensure stability

src/scanpy/preprocessing/_simple.py
Intron7: Yes we will do that. However that will be a separate PR.

tests/test_preprocessing.py
```diff
@@ -321,6 +325,14 @@ def test_regress_out_constants():
     assert_equal(adata, adata_copy)
 
 
+def test_regress_out_reproducible():
+    adata = pbmc68k_reduced()
+    adata = adata.raw.to_adata()[:200, :200].copy()
+    sc.pp.regress_out(adata, keys=["n_counts", "percent_mito"])
+    tester = np.load(DATA_PATH / "regress_test_small.npy")
```
ilan-gold, line 334: Please add a comment mentioning how this file was generated i.e., the commit of `main` it was made from or something.

ilan-gold (approved):

PR #3285, opened 2024-10-15
Update `test_rank_genes_groups.py` reference

emmanuel-ferdman:
# PR Summary
Small PR. Commit f03d4f40594a1f1b3216212dc60273c809882a4c moved the location of `test_rank_genes_groups.py`. This PR adjusts sources to changes.


flying-sheep (approved):
Thanks! Since this is such a small change, I assume you’re OK with not writing a patch note. I’ll merge this under that assumption, but if I’m wrong, just make another PR with the patch note. You can create one by running `hatch run towncrier:create`. If you do, type `3285` for the number it asks you.

PR #3294, opened 2024-10-18
Test all PCA param combinations

flying-sheep:
This replaces the random choice with iterating over all combinations. That way, if you want to debug a certain combination, you can just do so instead of rerunning the test and hoping it gets picked.

Sadly AFAIK it’s not possible to have a fixture that depends on other fixture values and generates a variable amount of values depending on their arguments: either you have `fixture(params=some_list)` which creates `len(some_list)` values or not, then it creates one. Therefore I had to get rid of the fixtures and use a static list instead.

It’s not that much slower to run them all:

before: 24 passed, 6 xfailed in 1.81s 
after: 56 passed, 14 xfailed in 3.41s


tests/test_pca.py
```diff
@@ -83,75 +81,105 @@ def wrapper(a: np.ndarray) -> DaskArray:
...
 
 
-@pytest.fixture(
-    params=[
-        param_with(at, marks=[needs.dask_ml]) if "dask" in at.id else at
-        for at in ARRAY_TYPES_SPARSE_DASK_UNSUPPORTED
-    ]
-)
-def array_type(request: pytest.FixtureRequest):
+def conv_at(array_type):
```
ilan-gold: type please but also name of the function to `maybe_convert_array_to_dask` or something?

tests/test_pca.py
```diff
@@ -83,75 +81,105 @@ def wrapper(a: np.ndarray) -> DaskArray:
...
-        and zero_center
-        and svd_solver == "lobpcg"
-    ):
-        warn_pat_expected = r"legacy code"
+SVDSolver = Literal["auto", "full", "arpack", "randomized", "tsqr", "lobpcg"]
+
 
-    return (svd_solver, warn_pat_expected)
+def gen_pca_params(
+    *, array_type, svd_solver_type: Literal[None, "valid", "invalid"], zero_center: bool
```
ilan-gold: Ok, so there's an issue with typing `array_type`?

tests/test_pca.py
```diff
@@ -324,7 +352,7 @@ def test_pca_n_pcs():
 
 # We use all ARRAY_TYPES here since this error should be raised before
```
ilan-gold:
```suggestion
# We use all ARRAY_TYPES_ALL_SUPPORTED here since this error should be raised before
```

ilan-gold (approved):

tests/test_pca.py
flying-sheep: no, I can do it!

PR #3295, opened 2024-10-18
lets umap run in parallel

### LINKED ISSUE #3211: Umap parallel

Umap support parallel compute with numba. However we are not using it. I would suggest either setting this to true by default or adding a flag to set this to true for people to use.

Intron7:
Closes #3211

Lets you run all parts of UMAP in Parallel. Default will still be False for reproducibility. 
Benchmarks (95k Cells AMD5950X)
`parallel = False` 33 s
`parallel = True` 18 s

I copied the doc string from UMAP to explain the impact of parallel execution.

### DIFF
--- a/src/scanpy/tools/_umap.py
+++ b/src/scanpy/tools/_umap.py
@@ -56,6 +56,7 @@ def umap(
     key_added: str | None = None,
     neighbors_key: str = "neighbors",
     copy: bool = False,
+    parallel: bool = False,
 ) -> AnnData | None:
     """\
     Embed the neighborhood graph using UMAP :cite:p:`McInnes2018`.
@@ -146,6 +147,8 @@ def umap(
         :attr:`~anndata.AnnData.obsp`\\ ``[.uns[neighbors_key]['connectivities_key']]`` for connectivities.
     copy
         Return a copy instead of writing to adata.
+    parallel
+        Whether to run the computation using numba parallel. Running in parallel is non-deterministic, and is not used if a random seed has been set, to ensure reproducibility.
 
     Returns
     -------
@@ -232,6 +235,7 @@ def umap(
             densmap_kwds={},
             output_dens=False,
             verbose=settings.verbosity > 3,
+            parallel=parallel,
         )
     elif method == "rapids":
         msg = (
--- a/src/scanpy/tools/_umap.py
+++ b/src/scanpy/tools/_umap.py
@@ -148,8 +148,7 @@ def umap(
     copy
         Return a copy instead of writing to adata.
     parallel
-        Whether to run the computation using numba parallel. Running in parallel is non-deterministic, and is not used if a random seed has been set, to ensure reproducibility.
-
+        Whether to run the computation using numba parallel. Running in parallel is non-deterministic.
     Returns
     -------
     Returns `None` if `copy=False`, else returns an `AnnData` object. Sets the following fields:
--- a/src/scanpy/tools/_umap.py
+++ b/src/scanpy/tools/_umap.py
@@ -216,6 +216,12 @@ def umap(
         # for the init condition in the UMAP embedding
         default_epochs = 500 if neighbors["connectivities"].shape[0] <= 10000 else 200
         n_epochs = default_epochs if maxiter is None else maxiter
+        if parallel and random_state is not None:
+            warnings.warn(
+                "Parallel execution was expected to be disabled when both `parallel=True` and `random_state` are set, "
+                "to ensure reproducibility. However, parallel execution still seems to occur, which may lead to "
+                "non-deterministic results."
+            )
         X_umap, _ = simplicial_set_embedding(
             data=X,
             graph=neighbors["connectivities"].tocoo(),
--- a/tests/test_embedding.py
+++ b/tests/test_embedding.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import warnings
+
 import numpy as np
 import pytest
 from numpy.testing import assert_array_almost_equal, assert_array_equal, assert_raises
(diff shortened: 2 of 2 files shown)


Intron7:
I didn't add release note yet because I'm not sure if this should already go into 1.10.4

src/scanpy/tools/_umap.py
```diff
@@ -146,6 +147,8 @@ def umap(
         :attr:`~anndata.AnnData.obsp`\\ ``[.uns[neighbors_key]['connectivities_key']]`` for connectivities.
     copy
         Return a copy instead of writing to adata.
+    parallel
```
ilan-gold, line 150: ~~Has this been tested? It seems like if the underlying implementation would error out every time a random seed is set, and we set a random seed, then setting this argument as true won't work?~~

I don't know where I got that this should error, but then we should definitely warn users about this sort of thing.   Reproducibility is very important

ilan-gold: On the other hand:

```python
sc.tl.umap(adata, parallel=True, random_state=42)
```

works so I think this needs a test + updated comment to reflect whatever is supposed to be going on here.

ilan-gold: As does

```python
sc.tl.umap(adata, parallel=True, random_state=np.random.RandomState(42))
```

ilan-gold (changes requested):

src/scanpy/tools/_umap.py
Intron7: I added a test to see if it errors (it doesn't)

ilan-gold: Yes, but we should also warn users about this random state business.  And check that the warning is raised every time `parallel` is True because this is very important!

Intron7: I also thing the docs in umap is wrong. The speedup stays if you submit a randomseed

ilan-gold: Ok! But then we should _definitely_ warn in that case.

Intron7: I added a warning

src/scanpy/tools/_umap.py
```diff
@@ -56,6 +56,7 @@ def umap(
     key_added: str | None = None,
     neighbors_key: str = "neighbors",
     copy: bool = False,
+    parallel: bool = False,
```
flying-sheep, line 59: I think we have the `n_jobs: int | None` convention for this (with `None` meaning `sc.settings.N_JOBS`), but `simplicial_set_embedding` just passes `parallel` on to numba.

We should think about how the two integrate before we add this.

src/scanpy/tools/_umap.py
```diff
@@ -216,6 +216,12 @@ def umap(
         # for the init condition in the UMAP embedding
         default_epochs = 500 if neighbors["connectivities"].shape[0] <= 10000 else 200
         n_epochs = default_epochs if maxiter is None else maxiter
+        if parallel and random_state is not None:
+            warnings.warn(
+                "Parallel execution was expected to be disabled when both `parallel=True` and `random_state` are set, "
+                "to ensure reproducibility. However, parallel execution still seems to occur, which may lead to "
+                "non-deterministic results."
+            )
```
ilan-gold, line 224: Link to the umap repo to give user more context.  Something like "UMAP reports that parallel execution should error with a random seed to ensure reproducibility.  Parallel execution may occur nonetheless, which can lead to non-deterministic results." with a link to their docs

ilan-gold: And then make sure this warning is raised in tests

src/scanpy/tools/_umap.py
Intron7: ok there is in issue with `check_random_state` as it transforms `None` into a seed.

tests/test_embedding.py
```diff
@@ -88,3 +88,9 @@ def test_diffmap():
     sc.tl.diffmap(pbmc, random_state=1234)
     d3 = pbmc.obsm["X_diffmap"].copy()
     assert_raises(AssertionError, assert_array_equal, d1, d3)
+
+
+def test_umap_parallel_randomstate():
+    pbmc = pbmc68k_reduced()[:100, :].copy()
+    sc.tl.umap(pbmc, parallel=True, random_state=42)
+    sc.tl.umap(pbmc, parallel=True, random_state=np.random.RandomState(42))
```
ilan-gold: Only one action per test.  This should be parametrized.  Also, we don't need both of these given what I said above.  We should just check for the warning with `random_state` as `None` or not-`None`

src/scanpy/tools/_umap.py
Intron7: ok the function is bugged on the umap side. If you force `None` as the random seed, it errors. I'll make an issue with umap for clarification. I'll but the PR on ice while i wait for an update.

Intron7: https://github.com/lmcinnes/umap/issues/1155

src/scanpy/tools/_umap.py
Intron7: Since this is numba parallel, we can only set it to use everything you got or nothing

tests/test_embedding.py
Intron7: Yes I updated the test already to catch the warning

src/scanpy/tools/_umap.py
Intron7: I'll update this once i have clarification from https://github.com/lmcinnes/umap/issues/1155

src/scanpy/tools/_umap.py
flying-sheep: Yes, that’s why we should talk about the parameter name. IIRC that would be the first parallelization parameter not called `n_jobs`.

flying-sheep:
Now with `rng` in, this is very simple, I recreated it as #4036.

Sorry for forgetting about this one @Intron7!

PR #3296, opened 2024-10-18
Add explicit support to PCA for `'covariance_eigh'` svd_solver

flying-sheep:
Extracted from #3263

Also deprecates `lobpcg` in docs which we forgot in a previous PR

### DIFF
--- a/tests/test_pca.py
+++ b/tests/test_pca.py
@@ -16,6 +16,7 @@ from scipy import sparse
 from scipy.sparse import issparse
 
 import scanpy as sc
+from scanpy.preprocessing._pca import SvdSolver as SvdSolverSupported
 from testing.scanpy import _helpers
 from testing.scanpy._helpers.data import pbmc3k_normalized
 from testing.scanpy._pytest.marks import needs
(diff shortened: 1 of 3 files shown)


tests/test_pca.py
```diff
@@ -110,7 +111,8 @@ def array_type(request: pytest.FixtureRequest) -> ArrayType:
     return request.param
 
 
-SVDSolver = Literal["auto", "full", "arpack", "randomized", "tsqr", "lobpcg"]
```
ilan-gold, line 113: What happened to `tsqr` in the actual implementation type?

ilan-gold: Never there?

flying-sheep: Doesn’t it work like advertised?

It’s in `SvdSolvPCADaskML` so if you have a dense dask array, you can use it

ilan-gold: Apologies, was looking locally with a filter on for tests

ilan-gold (approved):

PR #3306, opened 2024-10-22
(fix): clarify sparse pca usage

ilan-gold:
This PR clarifies the docs/handling to make it clear we _only_ support this for correctly chunked CSR-dask.  I think that not handling the other case is fine for a few reasons:

1. CSC chunking would basically require multiple passes at the data.  Every chunk (of size `(adata.shape[0], N)`) would need to be `X.T @ Y`-ed over the entire matrix where `X` is the chunk and `Y` is any given column-chunk from the matrix.  I can't think of a way around this
2. Given the above, I don't think there is any reason why we should implement this algorithm.  People should just re-write their data to disk as CSR.  I can't imagine its worse than this modulo the fact that you need to load the whole matrix into memory.  This might be a good reason to change our on-disk format but at the moment, this is about all I think we should do.

- [x] Just remembered that this fact needs to be stated clearly.  @Intron7 please update the RSC PR as well if you haven't already!

- [x] Release notes not necessary because: edited

### DIFF
--- a/src/scanpy/preprocessing/_pca/_dask_sparse.py
+++ b/src/scanpy/preprocessing/_pca/_dask_sparse.py
@@ -50,14 +50,14 @@ class PCASparseDask:
         """
         if x._meta.format != "csr":
             msg = (
-                "Only dask arrays with CSR-meta format are supported."
-                f" Got {x._meta.format} as meta."
+                "Only dask arrays with CSR-meta format are supported. "
+                f"Got {x._meta.format} as meta."
             )
             raise ValueError(msg)
         if x.chunksize[1] != x.shape[1]:
             msg = (
-                "Only dask arrays with chunking along the first axis are supported."
-                f" Got chunksize {x.chunksize} with shape {x.shape}."
+                "Only dask arrays with chunking along the first axis are supported. "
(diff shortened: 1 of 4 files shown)


flying-sheep (approved):

src/scanpy/preprocessing/_pca/_dask_sparse.py
```diff
@@ -48,6 +48,19 @@ def fit(self, x: DaskArray) -> PCASparseDaskFit:
...
+            msg = (
+                "Only dask arrays with CSR-meta format are supported. "
+                f"Got {x._meta.format} as meta."
+            )
+            raise ValueError(msg)
+        if x.chunksize[1] != x.shape[1]:
+            msg = (
+                "Only dask arrays with chunking along the first axis are supported. "
+                f"Got chunksize {x.chunksize} with shape {x.shape}. "
+                "Rechunking should be simple and cost nothing from AnnData's on-disk format the on-disk layout has this chunking."
```
flying-sheep:
```suggestion
                "Rechunking should be simple and cost nothing from AnnData's on-disk format when the on-disk layout has this chunking."
```

flying-sheep (changes requested):

flying-sheep (approved):

PR #3307, opened 2024-10-22
(feat): `calculate_qc_metrics` with `dask`

### LINKED ISSUE #2937: `calculate_qc_metrics` for `dask`

It would appear that the main blocker for `calculate_qc_metrics` being dask-ified is the fact that the calculation of top proportions is complicated and not easily dask-able.  So at the minimum, we can skip that and then do the rest of the computations which should be trivial.

ilan-gold:
- [x] Closes #2937


src/scanpy/preprocessing/_qc.py
```diff
@@ -432,7 +456,7 @@ def top_segment_proportions_dense(
     return values / sums[:, None]
 
 
-@numba.njit(cache=True, parallel=True)
+# @numba.njit(cache=True, parallel=True)
```
ilan-gold: This is the current blocker.   Without this commented, I just get a "quiet" failure.  Asked on zulip: https://scverse.zulipchat.com/#narrow/channel/456188-Dask/topic/Numba.20in.20Dask/near/478248965

ilan-gold:
Apparently the problem I saw here with `numba` within dask is a bit more widespread: #3317

src/scanpy/preprocessing/_qc.py
```diff
@@ -106,13 +109,20 @@ def describe_obs(
     obs_metrics = pd.DataFrame(index=adata.obs_names)
     if issparse(X):
         obs_metrics[f"n_{var_type}_by_{expr_type}"] = X.getnnz(axis=1)
+    elif isinstance(X, DaskArray) and issparse(X._meta):
+        obs_metrics[f"n_{var_type}_by_{expr_type}"] = X.map_blocks(
+            lambda x: x.getnnz(axis=1),
+            dtype=np.int64,
+            meta=np.array([], dtype=np.int64),
+            drop_axis=1,
+        ).compute()
```
flying-sheep: maybe we should make and use a `getnnz` helper like `.._utils.axis_sum`?

src/scanpy/preprocessing/_qc.py
```diff
@@ -191,7 +202,15 @@ def describe_var(
...
             X.eliminate_zeros()
     var_metrics = pd.DataFrame(index=adata.var_names)
-    if issparse(X):
+    if isinstance(X, DaskArray) and issparse(X._meta):
+        var_metrics["n_cells_by_{expr_type}"] = X.map_blocks(
+            lambda x: x.getnnz(axis=0),
+            dtype=np.int64,
+            meta=np.array([], dtype=np.int64),
+            drop_axis=0,
+        ).compute()
```
flying-sheep: ditto

src/scanpy/preprocessing/_qc.py
```diff
@@ -219,6 +238,7 @@ def describe_var(
         adata.var[var_metrics.columns] = var_metrics
     else:
         return var_metrics
+    return None
```
flying-sheep:
```suggestion
        adata.var[var_metrics.columns] = var_metrics
        return None
    return var_metrics
```

src/scanpy/preprocessing/_qc.py
```diff
@@ -393,6 +413,34 @@ def top_segment_proportions(
...
+            lambda x: top_segment_proportions_in_memory(mtx=x, ns=ns), meta=np.array([])
+        ).compute()
+    return top_segment_proportions_in_memory(mtx, ns)
+
+
+def top_segment_proportions_in_memory(
+    mtx: np.ndarray | spmatrix, ns: Collection[int]
+) -> np.ndarray:
+    """
+    Calculates total percentage of counts in top ns genes.
```
flying-sheep: please delete the docstring here, this is just a helper function to dispatch to, people should go through `top_segment_proportions`, right?

src/testing/scanpy/_helpers/__init__.py
```diff
@@ -158,3 +159,22 @@ def __enter__(self):
...
+    if find_spec("dask"):
+        import dask
+
+        prev_scheduler = dask.config.get("scheduler", "threads")
+        dask.config.set(scheduler="single-threaded")
+        yield
+        dask.config.set(scheduler=prev_scheduler)
+    else:
+        dask = None
+        yield
```
flying-sheep:
```suggestion
@contextmanager
def maybe_dask_process_context():
    """
    Running numba with dask's threaded scheduler causes crashes,
    so we need to switch to single-threaded (or processes, which is slower)
    scheduler for tests that use numba.
    """
    if not find_spec("dask"):
        yield
        return

    import dask.config

    prev_scheduler = dask.config.get("scheduler", "threads")
    dask.config.set(scheduler="single-threaded")
    try:
        yield
    finally:
        dask.config.set(scheduler=prev_scheduler)
```

flying-sheep (commented):
looks great!

The diff for `test_qc_metrics` is a bit hard to follow. Am I right that you simply
1. pulled out `test_qc_metrics_no_log1p` and `test_qc_metrics_idempotent`
2. parametrized it with all array types
3. added `with maybe_dask_process_context():`

Thanks for pulling the tests apart!

ilan-gold:
> Am I right that you simply

You are right!

src/scanpy/preprocessing/_qc.py
ilan-gold: Ah great idea!

flying-sheep:
great! then apart from my suggestions and [the crash problem](https://scverse.zulipchat.com/#narrow/channel/456188-Dask/topic/Numba.20in.20Dask/near/478897406), this looks ready.

src/scanpy/preprocessing/_qc.py
ilan-gold: why should we delete the docstring, seems better than not having one?

src/scanpy/preprocessing/_qc.py
```diff
@@ -110,12 +110,7 @@ def describe_obs(
...
     elif isinstance(X, DaskArray) and issparse(X._meta):
-        obs_metrics[f"n_{var_type}_by_{expr_type}"] = X.map_blocks(
-            lambda x: x.getnnz(axis=1),
-            dtype=np.int64,
-            meta=np.array([], dtype=np.int64),
-            drop_axis=1,
-        ).compute()
+        obs_metrics[f"n_{var_type}_by_{expr_type}"] = axis_nnz(X, axis=1).compute()
     else:
         obs_metrics[f"n_{var_type}_by_{expr_type}"] = np.count_nonzero(X, axis=1)
```
flying-sheep: This should probably be collapsed into something like:

```py
obs_metrics[f"n_{var_type}_by_{expr_type}"] = materialize_as_ndarray(axis_nnz(X, axis=1))
```

src/scanpy/preprocessing/_qc.py
```diff
@@ -203,12 +198,7 @@ def describe_var(
...
-        var_metrics["n_cells_by_{expr_type}"] = X.map_blocks(
-            lambda x: x.getnnz(axis=0),
-            dtype=np.int64,
-            meta=np.array([], dtype=np.int64),
-            drop_axis=0,
-        ).compute()
+        var_metrics["n_cells_by_{expr_type}"] = axis_nnz(X, axis=0).compute()
         var_metrics["mean_{expr_type}"] = _get_mean_var(X, axis=0)[0].compute()
     elif issparse(X):
         # Current memory bottleneck for csr matrices:
```
flying-sheep: same here, the whole elif tree can probably become something like

```py
var_metrics["n_cells_by_{expr_type}"], var_metrics["mean_{expr_type}"] = (
    materialize_as_ndarray(axis_nnz(X, axis=0), _get_mean_var(X, axis=0)[0])
)
```

src/scanpy/_utils/__init__.py
```diff
@@ -713,6 +713,27 @@ def _(
     )
 
 
+@singledispatch
+def axis_nnz(X: np.ndarray, axis: Literal[0, 1]) -> np.ndarray:
```
flying-sheep, line 717: maybe `X: ArrayLike`? With the idea that this is the fallback branch that’s able to handle everything that numpy can convert into an array?

flying-sheep: maybe `X: ArrayLike` (and still `-> np.ndarray`)? With the idea being that this is the fallback branch that’s able to handle everything that numpy can convert into an array?

src/scanpy/_utils/__init__.py
```diff
@@ -713,6 +713,27 @@ def _(
...
+
+
+@axis_nnz.register(sparse.spmatrix)
+def _(X: sparse.spmatrix, axis: Literal[0, 1]) -> np.ndarray:
+    return X.getnnz(axis=axis)
+
+
+@axis_nnz.register(DaskArray)
+def _(X: DaskArray, axis: Literal[0, 1]) -> DaskArray:
+    return X.map_blocks(
```
flying-sheep, line 728: strange, codecov says this line isn’t covered.

I assume it’s just lying, but please check if that message has vanished before you merge.

src/scanpy/preprocessing/_qc.py
```diff
@@ -191,21 +194,17 @@ def describe_var(
...
             X.eliminate_zeros()
     var_metrics = pd.DataFrame(index=adata.var_names)
-    if issparse(X):
-        # Current memory bottleneck for csr matrices:
-        var_metrics["n_cells_by_{expr_type}"] = X.getnnz(axis=0)
-        var_metrics["mean_{expr_type}"] = mean_variance_axis(X, axis=0)[0]
-    else:
-        var_metrics["n_cells_by_{expr_type}"] = np.count_nonzero(X, axis=0)
-        var_metrics["mean_{expr_type}"] = X.mean(axis=0)
+    var_metrics["n_cells_by_{expr_type}"], var_metrics["mean_{expr_type}"] = (
```
flying-sheep, line 197: I’m just realizing that none of these are `f`-literals. `expr_type` won’t be interpolated into this!

mayyybe we should fix that.

src/scanpy/preprocessing/_qc.py
flying-sheep: it’s duplicated, so when someone modifies it, there’s a >0 chance that they only modify one version and not both, which is worse than having to search for the docstring for a few seconds.

if you want, you can rename these internal helper functions to `_leading_underscore` names, then it’s even more likely that future us will find the entrypoint `top_proportions` and its docstring.

flying-sheep (approved):
beautiful. we should decide how to deal with the missing `f`-literals, otherwise LGTM

src/scanpy/preprocessing/_qc.py
ilan-gold: i think they are! `col.format` is called!

flying-sheep: ah! bizarre! but sure, let’s do that in another PR then

src/scanpy/_utils/__init__.py
ilan-gold: Screenshot 2024-10-30 at 09 49 45
gone!

src/scanpy/preprocessing/_qc.py
ilan-gold: great point, refactored for singledispatch

flying-sheep:
OK, https://github.com/scverse/scanpy/pull/3335 is ready.

We should manually check if it can get rid of `maybe_dask_process_context`

/edit: it’s not ready

PR #3311, opened 2024-10-23
Fix some `Returns` docstrs re: `inplace` semantics

ryan-williams:
- [x] Tests included or not required because: comment/docstring nits
- [x] Release notes not necessary because: comment/docstring nits


Zethson (approved):
Thank you very much!

PR #3316, opened 2024-10-24
Refactor regress_out

flying-sheep:
- [ ] Closes #


Zethson (approved):
Can't vouch for a speedup but I guess the benchmark CI could? :) Currently it doesn't seem to though. Maybe joblib adds an overhead here? You know these things better than me.

flying-sheep:
it shouldn’t affect performance at all, just readability

PR #3317, opened 2024-10-24
Fix zappy compatibility for clip_array

flying-sheep:
- [ ] Closes #


src/scanpy/preprocessing/_scale.py
```diff
@@ -203,25 +217,13 @@ def scale_array(
...
         else:
-            if isinstance(X, DaskArray):
-                X = X.map_blocks(
-                    clip_array, max_value=max_value, zero_center=zero_center
-                )
-            elif issparse(X):
-                X.data = clip_array(X.data, max_value=max_value, zero_center=False)
-            else:
-                X = clip_array(X, max_value=max_value, zero_center=zero_center)
+            X = clip_array(X, max_value=max_value, zero_center=zero_center)
```
ilan-gold, line 226: This works with zappy? is that the fix?

ilan-gold (approved):
Don't know much about zappy but it's certainly cleaner

src/scanpy/preprocessing/_scale.py
flying-sheep: Ah sorry; context.

No, the logic is untouched, only the `maybe_dask_process_context` is new.

This PR is part of me figuring out the whole “numba abort” situation, and this logic refactor is a side effect. I should probably retitle this PR.

ilan-gold: "This" being this line that was the `else` with `clip_array` which appears to only be for `numpy`

flying-sheep: I think that was another casualty of the flaky numba-on-M1 thing. Looking at it now, the logic seems identical.

PR #3333, opened 2024-10-31
(fix): sort pca test args

ilan-gold:
Since `gen_pca_params` is a generator, not sure how we can control the order besides rolling back the use of sets.  So I landed on just sorting the args based on `id` of the pytest.  

I was running in to https://github.com/pytest-dev/pytest-xdist/issues/432

- [ ] Closes #

- [x] Release notes not necessary because: just fixing some tests

### DIFF
--- a/src/scanpy/_utils/__init__.py
+++ b/src/scanpy/_utils/__init__.py
@@ -15,11 +15,11 @@ import warnings
 from collections import namedtuple
 from contextlib import contextmanager, suppress
 from enum import Enum
-from functools import partial, singledispatch, wraps
-from operator import mul, truediv
+from functools import partial, reduce, singledispatch, wraps
+from operator import mul, or_, truediv
 from textwrap import dedent
-from types import MethodType, ModuleType
-from typing import TYPE_CHECKING, overload
+from types import MethodType, ModuleType, UnionType
+from typing import TYPE_CHECKING, Literal, Union, get_args, get_origin, overload
 from weakref import WeakSet
 
 import h5py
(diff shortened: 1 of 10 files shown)


flying-sheep:
Would probably be more elegant to just do this in the generator instead:

```diff
-for svd_solver in svd_solvers:
+for svd_solver in sorted(svd_solvers):
```

That being said, there are ordered sets in Python, they’re just hidden:

```py
def ordered_union(a: KeysView[T1], b: KeysView[T2]) -> KeysView[T1 | T2]:
    """Ordered set union
    
    Examples
    --------
    >>> ordered_union(dict.fromkeys("ab").keys(), dict.fromkeys("cb").keys())
    dict_keys(['a', 'b', 'c'])
    """    
    return (dict.fromkeys(a) | dict.fromkeys(b)).keys()
```

flying-sheep (changes requested):
Would probably be more elegant to just do this in the generator instead:

```diff
-for svd_solver in svd_solvers:
+for svd_solver in sorted(svd_solvers):
```

That being said, there are ordered sets in Python, they’re just hidden:

```py
def ordered_union(a: KeysView[T1], b: KeysView[T2]) -> KeysView[T1 | T2]:
    """Ordered set union
    
    Examples
    --------
    >>> ordered_union(dict.fromkeys("cb").keys(), dict.fromkeys("ca").keys())
    dict_keys(['c', 'b', 'a'])
    """    
    return dict.fromkeys(a) | dict.fromkeys(b)
```

ilan-gold:
@flying-sheep `SVDSolver` can be a union of literals so any sort of sorting on it is not out-of-the-box from what I can tell

flying-sheep:
`svd_solvers` is just a `set[str]`, no?

/edit: oof it isn’t:

```py
>>> from typing import Literal, get_args
>>> get_args(Literal["a"] | Literal["b"])
(typing.Literal['a'], typing.Literal['b'])
```

I was sure these get collapsed into one `Literal` … that’s a bug!

flying-sheep (approved):

PR #3336, opened 2024-10-31
Fix PCA tests

flying-sheep:


src/scanpy/_utils/__init__.py
```diff
@@ -532,6 +533,19 @@ def update_params(
     return updated_params
 
 
+# `get_args` returns `tuple[Any]` so I don’t think it’s possible to get the correct type here
+def get_literal_vals(typ: UnionType | Any) -> KeysView[Any]:
```
ilan-gold, line 537: Why make this return `KeysView`? I would think a `List` would be a more natural return a list

ilan-gold: Or even a set.  But forcing keysview is a strong choice

ilan-gold (changes requested):
something up with the CI, looks relevant

src/scanpy/_utils/__init__.py
flying-sheep: `KeysView` is an “ordered set” type.

This could also return a `collections.abc.Set` (aka `AbstractSet`), but I think it’s fine to be specific regarding return types.

flying-sheep:
we need to finally fix that CI failure in this job, I only look into the `minimum_versions` job anymore before I merge PRs

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -471,7 +472,7 @@ def _handle_dask_ml_args(
 def _handle_dask_ml_args(svd_solver: str | None, method: MethodDaskML) -> SvdSolvDaskML:
     import dask_ml.decomposition as dmld
 
-    args: tuple[SvdSolvDaskML, ...]
+    args: AbstractSet[SvdSolvDaskML]
```
ilan-gold, line 474: I would say we should probably use a different name since `AbstractSet` is deprecated (I understand that this is a renaming, but I don't really get why.  I'm fine with the reason for renaming, whatever it is, but just would want something that isn't the same as something that is deprecated).

flying-sheep: As discussed in private: https://docs.astral.sh/ruff/rules/unaliased-collections-abc-set-import/

Added `PYI` to https://github.com/scverse/anndata/issues/1241

PR #3337, opened 2024-11-04
Actually working min-deps job

flying-sheep:
Hia everyone, this should actually make`hatch test -i deps=min` work even on Macbooks.

Maybe there’s a way to put the constraint file inside of the venv, then it’ll survive reboots on Linux.

I tagged you all because I thought you might be interested in this.


pyproject.toml
```diff
@@ -52,7 +52,7 @@ dependencies = [
     "pandas >=1.5",
     "scipy>=1.8",
     "seaborn>=0.13",
-    "h5py>=3.6",
+    "h5py>=3.7",
```
flying-sheep, line 55: h5py 3.6.0 doesn’t build on macOS M1, 3.7.0 has wheels.

Idk if that’s a good reason to bump the min deps …

hatch.toml
```diff
@@ -19,7 +19,11 @@ features = ["test", "dask-ml"]
 extra-dependencies = ["ipykernel"]
 overrides.matrix.deps.env-vars = [
     { if = ["pre"], key = "UV_PRERELEASE", value = "allow" },
-    { if = ["min"], key = "UV_RESOLUTION", value = "lowest-direct" },
+    #{ if = ["min"], key = "UV_RESOLUTION", value = "lowest-direct" },
+    { if = ["min"], key = "UV_CONSTRAINT", value = "/tmp/scanpy-min-deps.txt" },
```
ilan-gold: 1. Delete commented out?
2. Why not put this in the `ci` folder or something and then `.gitignore` it?  I don't know enough about filesystems, but I seem to remember Helmholtz's having no accessible `tmp`

flying-sheep: can do!

hatch.toml
```diff
@@ -19,14 +19,17 @@ features = ["test", "dask-ml"]
 extra-dependencies = ["ipykernel"]
 overrides.matrix.deps.env-vars = [
     { if = ["pre"], key = "UV_PRERELEASE", value = "allow" },
-    { if = ["min"], key = "UV_RESOLUTION", value = "lowest-direct" },
+    { if = ["min"], key = "UV_CONSTRAINT", value = "ci/scanpy-min-deps.txt" },
+]
+overrides.matrix.deps.pre-install-commands = [
+    { if = ["min"], value = "uv pip install -c '' tomli packaging && python ci/scripts/min-deps.py pyproject.toml -o ci/scanpy-min-deps.txt" },
```
grst: Does installing `tomli` and `packaging` here potentially affect the CI env? 
Maybe this would be a good fit for using `uv run ci/scripts/min-deps.py` and declare the dependencies [within the script](https://docs.astral.sh/uv/guides/scripts/#running-a-script-with-dependencies)?

flying-sheep: that’s a great point! I’ll do that as it’s the most elegant way.

What I always preach about reusing environments is that when you install into them multiple times, you can’t rely on things from earlier install calls to still be there. But what I did here was

1. install script deps
2. use script deps in script
3. install package deps
4. use package deps in tests

so the only thing that changes is that the script deps used to maybe still still be present at step 4, whereas now step 3 works on a fresh environment.

PR #3347, opened 2024-11-10
Change default matplotlib.rcParams["axes.grid"] from True to False

senseibence:
# Description

When creating heatmaps, matrices, and various graphs using large datasets with seaborn, `scanpy.set_figure_params()` provides helpful default configurations for the plot. However, one default, caused by `matplotlib.rcParams["axes.grid"] = True`, provides no benefit to the vast majority of plots, making them unreadable and confusing (especially with very large plots). Below is a simple example; evidently, the distortion increases with graph complexity.

# Images
### Graph with default grid:
graph_with_grid<br>
### Graph without default grid:
graph_without_grid

# Conclusion
This change simply modifies `matplotlib.rcParams["axes.grid"]` from `True` to `False`, not affecting any other components of the module. We believe this adjusted default would benefit the majority of graphs compared to the original. If one would like to use the grid lines for any reason (which we believe is a niche use case), they may run `matplotlib.pyplot.rcParams["axes.grid"] = True`.

### DIFF
--- a/src/scanpy/plotting/_rcmod.py
+++ b/src/scanpy/plotting/_rcmod.py
@@ -61,7 +61,7 @@ def set_rcParams_scanpy(fontsize=14, color_map=None):
     rcParams["ytick.labelsize"] = fontsize
 
     # axes grid
-    rcParams["axes.grid"] = True
+    rcParams["axes.grid"] = False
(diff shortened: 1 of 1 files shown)


senseibence:
Could someone help with the checks for this PR?

PR #3351, opened 2024-11-11
Simplify scale

flying-sheep:
`scale_sparse` has a lot of code that’s not needed.

This PR also fixes zappy compatibility for `scale`, not that anyone is using it as far as I know.


flying-sheep:
@Intron7 any idea what I did wrong? Why do the benchmarks say that it uses more memory?

flying-sheep:
@Intron7 ping

docs/release-notes/3351.bugfix.md
```diff
@@ -0,0 +1 @@
+Fix zappy compatibility for clip_array {smaller}`P Angerer`
```
ilan-gold, line 1: What does this have to do with the PR?

src/scanpy/preprocessing/_scale.py
```diff
@@ -29,20 +29,32 @@
...
-    for i in numba.prange(len(indptr) - 1):
-        if mask_obs[i]:
-            for j in range(indptr[i], indptr[i + 1]):
-                if clip:
-                    data[j] = min(clip, data[j] / std[indices[j]])
-                else:
-                    data[j] /= std[indices[j]]
+
+@singledispatch
+def clip(x: ArrayLike | _A, *, max_value: float, zero_center: bool = True) -> _A:
```
ilan-gold, line 40: Why do we need `ArrayLike`?

src/scanpy/preprocessing/_scale.py
```diff
@@ -144,155 +148,141 @@ def scale(
...
     if zero_center:
-        if isinstance(X, DaskArray) and issparse(X._meta):
+        if isinstance(x, csr_matrix | csc_matrix) or (
+            isinstance(x, DaskArray) and issparse(x._meta)
+        ):
             warnings.warn(
                 "zero-center being used with `DaskArray` sparse chunks. "
                 "This can be bad if you have large chunks or intend to eventually read the whole data into memory.",
                 UserWarning,
             )
```
ilan-gold: Your intention was to warn people against doing this with sparse data? Maybe update the message then?

docs/release-notes/3351.bugfix.md
ilan-gold: Ah I see your mention of it in the PR description, ok

src/scanpy/preprocessing/_scale.py
ilan-gold: For `zappy`?

src/scanpy/preprocessing/_scale.py
```diff
@@ -144,155 +148,141 @@ def scale(
...
         )
 
-    if np.issubdtype(X.dtype, np.integer):
+    if np.issubdtype(x.dtype, np.integer):
         logg.info(
             "... as scaling leads to float results, integer "
             "input is cast to float, returning copy."
         )
-        X = X.astype(float)
+        x = x.astype(float)
```
Intron7:
```suggestion
    # Create default mask to ones for CSR matrices if mask_obs is None
    # This speeds up the default code path for CSR matrices
    if isinstance(x, csr_matrix) and mask_obs is None:
        mask_obs = np.ones(x.shape[0], dtype=bool)

```

Intron7 (commented):
This would make the default code path  for csr even faster. otherwise LGTM

src/scanpy/preprocessing/_scale.py
flying-sheep: Yeah, this makes `scale` work for things that behave sufficiently like a numpy array.

src/scanpy/preprocessing/_scale.py
flying-sheep: What would you like to see? The message already mentions sparse chunks.

flying-sheep: I see, done!

PR #3353, opened 2024-11-11
Speed up categorical regressor with numba

Intron7:
Use numba to create the regressor for categorical regression


tests/test_preprocessing.py
```diff
@@ -337,6 +339,39 @@ def test_regress_out_reproducible():
...
 
 
+def test_regress_out_reproducible_category():
+    adata = sc.datasets.pbmc68k_reduced()
+    adata = adata.raw.to_adata()[:200, :200].copy()
+    sc.pp.regress_out(adata, keys=["bulk_labels"])
+    # This file was generated from the original implementation in version 1.10.3
+    # Now we compare new implementation with the old one
+    tester = np.load(DATA_PATH / "regress_test_small_cat.npy")
+    np.testing.assert_array_almost_equal(adata.X, tester)
```
ilan-gold: Parametrize this so `test_regress_out_reproducible_category` and `test_regress_out_reproducible` are one test

src/scanpy/preprocessing/_simple.py
```diff
@@ -628,6 +628,19 @@ def normalize_per_cell(
 DT = TypeVar("DT")
 
 
+@njit
+def _create_regressor_categorical(X, cats, filters):
```
ilan-gold: types?

ilan-gold: Also `cats`->`num_categories`? It's an `int`?

tests/test_preprocessing.py
```diff
@@ -337,6 +339,39 @@ def test_regress_out_reproducible():
...
+
+
+def test_regressor_categorical():
```
ilan-gold: I would

1. explain why this test exists (to test against a previous implementation? I am impartial whether it's necessary TBH since we are already testing for reproducibility, could see getting rid of this)
2. refactor the "Create org regressors" into a helper function like `create_original`

tests/test_preprocessing.py
```diff
@@ -337,6 +339,39 @@ def test_regress_out_reproducible():
...
+    for category in adata.obs[keys[0]].cat.categories:
+        mask = (category == adata.obs[keys[0]]).values
+        for ix, x in enumerate(X.T):
+            regressors[mask, ix] = x[mask].mean()
+
+    # Create new regressors
+    cats = np.int64(len(adata.obs[keys[0]].cat.categories))
+    filters = adata.obs[keys[0]].cat.codes.to_numpy()
+    cats = cats.astype(filters.dtype)
+    X = _to_dense(X_org, order="F") if issparse(X_org) else X_org
```
ilan-gold: This test is not parameterized so `X_org` is either sparse or not, no? So this line could be simplified or the test should be parameterized to handle both cases

ilan-gold (changes requested):

tests/test_preprocessing.py
Intron7: I removed the is sparse part

tests/test_preprocessing.py
Intron7: I can see your point here

ilan-gold (changes requested):
I think this is missing: https://github.com/scverse/scanpy/pull/3353#discussion_r1836825935 and the first part of https://github.com/scverse/scanpy/pull/3353/files#r1836830351

tests/test_preprocessing.py
```diff
@@ -329,23 +329,20 @@ def test_regress_out_constants():
...
-    sc.pp.regress_out(adata, keys=["n_counts", "percent_mito"])
-    # This file was generated from the original implementation in version 1.10.3
-    # Now we compare new implementation with the old one
-    tester = np.load(DATA_PATH / "regress_test_small.npy")
-    np.testing.assert_allclose(adata.X, tester)
-
-
-def test_regress_out_reproducible_category():
+@pytest.mark.parametrize(
+    ("keys", "tester_file"),
```
ilan-gold:
```suggestion
    ("keys", "expected_result_file_path"),
```

tests/test_preprocessing.py
```diff
@@ -329,23 +329,20 @@ def test_regress_out_constants():
...
+    [
+        (["n_counts", "percent_mito"], "regress_test_small.npy"),
+        (["bulk_labels"], "regress_test_small_cat.npy"),
+    ],
+)
+def test_regress_out_reproducible(keys, tester_file):
```
ilan-gold:
```suggestion
def test_regress_out_reproducible(keys, expected_result_file_path):
```

tests/test_preprocessing.py
```diff
@@ -329,23 +329,20 @@ def test_regress_out_constants():
...
     adata = sc.datasets.pbmc68k_reduced()
     adata = adata.raw.to_adata()[:200, :200].copy()
-    sc.pp.regress_out(adata, keys=["bulk_labels"])
+    sc.pp.regress_out(adata, keys=keys)
     # This file was generated from the original implementation in version 1.10.3
-    # Now we compare new implementation with the old one
-    tester = np.load(DATA_PATH / "regress_test_small_cat.npy")
+    # Now we compare the new implementation with the old one
+    tester = np.load(DATA_PATH / tester_file)
```
ilan-gold:
```suggestion
    tester = np.load(DATA_PATH / expected_result_file_path)
```

src/scanpy/preprocessing/_simple.py
```diff
@@ -628,6 +628,21 @@ def normalize_per_cell(
 DT = TypeVar("DT")
 
 
+@njit
+def _create_regressor_categorical(
+    X: np.ndarray, cats: int, filters: np.ndarray
```
ilan-gold: https://github.com/scverse/scanpy/pull/3353#discussion_r1836827147

ilan-gold: I asked the name be changed because `cats` implies categories which on a pandas object is the values, not size.

src/scanpy/preprocessing/_simple.py
```diff
@@ -722,13 +737,13 @@ def regress_out(
                 "we regress on the mean for each category."
             )
         logg.debug("... regressing on per-gene means within categories")
-        regressors = np.zeros(X.shape, dtype="float32")
+        # Create numpy array's from categorical variable
+        cats = np.int64(len(adata.obs[keys[0]].cat.categories))
```
ilan-gold: Ditto

ilan-gold: Also comment why `np.int64`

ilan-gold (changes requested):

src/scanpy/preprocessing/_simple.py
Intron7: because it has be done because of weird typing from pandas. So this ensures that it works within the kernel

tests/test_preprocessing.py
ilan-gold: looks like you missed the above

src/scanpy/preprocessing/_simple.py
```diff
@@ -628,6 +628,21 @@ def normalize_per_cell(
...
+    X: np.ndarray, number_categories: int, filters: np.ndarray
+) -> np.ndarray:
+    # create regressor matrix faster for categorical variables
+    regressors = np.zeros(X.shape, dtype=X.dtype)
+    XT = X.T
+    for category in range(number_categories):
+        mask = category == filters
+        for ix in numba.prange(XT.shape[0]):
+            x = XT[ix]
+            regressors[mask, ix] = x[mask].mean()
```
ilan-gold:
```suggestion
            regressors[mask, ix] = XT[ix][mask].mean()
```
why not?

tests/test_preprocessing.py
ilan-gold: Do you have an an opinion on the first point? Is this test necessary? If so, perhaps a comment then?

ilan-gold (changes requested):

src/scanpy/preprocessing/_simple.py
```diff
@@ -628,6 +628,20 @@ def normalize_per_cell(
...
+def _create_regressor_categorical(
+    X: np.ndarray, number_categories: int, filters: np.ndarray
+) -> np.ndarray:
+    # create regressor matrix faster for categorical variables
+    regressors = np.zeros(X.shape, dtype=X.dtype)
+    XT = X.T
+    for category in range(number_categories):
+        mask = category == filters
+        for ix in numba.prange(XT.shape[0]):
+            regressors[mask, ix] = XT[ix][mask].mean()
```
flying-sheep:
```suggestion
    for category in range(number_categories):
        mask = category == filters
        for ix in numba.prange(X.T.shape[0]):
            regressors[mask, ix] = X.T[ix][mask].mean()
```

src/scanpy/preprocessing/_simple.py
```diff
@@ -722,13 +736,13 @@ def regress_out(
                 "we regress on the mean for each category."
             )
         logg.debug("... regressing on per-gene means within categories")
-        regressors = np.zeros(X.shape, dtype="float32")
+        # Create numpy array's from categorical variable
+        number_categories = np.int64(len(adata.obs[keys[0]].cat.categories))
+        filters = adata.obs[keys[0]].cat.codes.to_numpy()
+        number_categories = number_categories.astype(filters.dtype)
```
flying-sheep: Either this or add a comment (to the code) explaining why it needs to be the other way.
Also if I do this, the test still passes, so …

```suggestion
        number_categories = len(adata.obs[keys[0]].cat.categories)
        filters = adata.obs[keys[0]].cat.codes.to_numpy()
```

tests/test_preprocessing.py
flying-sheep: yes, please parametrize instead of adding another test that is 85% identical with another

flying-sheep (changes requested):

src/scanpy/preprocessing/_simple.py
Intron7: done

src/scanpy/preprocessing/_simple.py
Intron7: I added a comment. Other wise you have a dtype missmatch and crash of the kernel

ilan-gold: > Other wise you have a dtype missmatch and crash of the kernel

I would say that this is the important part for the comment!

src/scanpy/preprocessing/_simple.py
```diff
@@ -628,6 +628,19 @@ def normalize_per_cell(
 DT = TypeVar("DT")
 
 
+@njit
+def _create_regressor_categorical(
+    X: np.ndarray, number_categories: int, filters: np.ndarray
+) -> np.ndarray:
+    # create regressor matrix faster for categorical variables
```
ilan-gold: What does this comment mean?

src/scanpy/preprocessing/_simple.py
flying-sheep: 100%!

- refactor your code until the “what” is obvious.
- if the “why” isn’t obvious from understanding the “what”, add the missing parts as a comment

I see that you’re

1. convert the cat codes into a numpy array
2. creating a numpy scalar with the same dtype as `filters`, holding the number of categories

So  you don’t need to comment that you do any of that.

I asked because I’m confused why a Python integer is converted to a numpy scalar: Usually APIs accept either and do the converting themselves. So I’d like to see a comment removing that confusion by explaining *why* you convert to a numpy scalar. (a crash is a great reason)

----

but I also see that `_create_regressor_categorical` has `number_categories: int` and then does `range(number_categories)`, so I’m still very confused why numba crashes unless the dtypes match.

I can’t reproduce the crash. leaving the thing as a Python int just works for me.

src/scanpy/preprocessing/_simple.py
flying-sheep: so `len` doesn’t return a Python `int`? That’s a pandas bug.

src/scanpy/preprocessing/_simple.py
flying-sheep: Also the way to do this in one step is

```suggestion
        filters = adata.obs[keys[0]].cat.codes.to_numpy()
        number_categories = filters.dtype.type(len(adata.obs[keys[0]].cat.categories))
```

Intron7: That worked.

Intron7:
I renamed one variable to make is clearer what it is. Added some comments that should add more context what the code is doing.

src/scanpy/preprocessing/_simple.py
```diff
@@ -637,6 +637,22 @@ def normalize_per_cell(
...
+) -> np.ndarray:
+    # create regressor matrix for categorical variables
+    regressors = np.zeros(X.shape, dtype=X.dtype)
+    # iterate over categories
+    for category in range(number_categories):
+        # iterate over genes and calculate mean expression
+        # for each gene per category
+        mask = category == cat_array
+        for ix in numba.prange(X.T.shape[0]):
+            regressors[mask, ix] = X.T[ix][mask].mean()
```
ilan-gold:
```suggestion
            regressors[mask, ix] = X.T[ix, mask].mean()
```

tests/test_preprocessing.py
```diff
@@ -464,14 +464,21 @@ def test_regress_out_constants():
...
+)
+def test_regress_out_reproducible(keys, test_file):
+    adata = sc.datasets.pbmc68k_reduced()
     adata = adata.raw.to_adata()[:200, :200].copy()
-    sc.pp.regress_out(adata, keys=["n_counts", "percent_mito"])
+    sc.pp.regress_out(adata, keys=keys)
     # This file was generated from the original implementation in version 1.10.3
     # Now we compare new implementation with the old one
-    tester = np.load(DATA_PATH / "regress_test_small.npy")
-    np.testing.assert_allclose(adata.X, tester)
```
ilan-gold, line 427: Why the change in testing function? There's a pretty big note (see https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_almost_equal.html) about why we should use `assert_allclose` in a floating point situation, which I believe is what we're in right now.

src/scanpy/preprocessing/_simple.py
```diff
@@ -732,13 +748,13 @@ def regress_out(
             )
             raise ValueError(msg)
         logg.debug("... regressing on per-gene means within categories")
-        regressors = np.zeros(X.shape, dtype="float32")
+
+        # set number of categories to the same dtype as the categories
+        cat_array = adata.obs[keys[0]].cat.codes.to_numpy()
+        number_categories = cat_array.dtype.type(len(adata.obs[keys[0]].cat.categories))
+
         X = _to_dense(X, order="F") if issparse(X) else X
```
ilan-gold: Out of curiosity, why not write a kernel that can handle sparse arrays, at least CSC?

Intron7: good question. It's not that hard tbh. But since we need dense anyway for the later step I just didn't think it was important.

ilan-gold: Ah ok that's a good enough reason then for me not to do it.

src/scanpy/preprocessing/_simple.py
```diff
@@ -637,6 +637,22 @@ def normalize_per_cell(
 DT = TypeVar("DT")
 
 
+@njit
+def _create_regressor_categorical(
+    X: np.ndarray, number_categories: int, cat_array: np.ndarray
+) -> np.ndarray:
+    # create regressor matrix for categorical variables
+    regressors = np.zeros(X.shape, dtype=X.dtype)
```
ilan-gold: check dtype for behavior with integer dtype i.e., need to ensure this is a floating point matrix

ilan-gold (changes requested):

tests/test_preprocessing.py
```diff
@@ -464,14 +464,21 @@ def test_regress_out_constants():
...
-def test_regress_out_reproducible():
-    adata = pbmc68k_reduced()
+@pytest.mark.parametrize(
+    ("keys", "test_file", "atol"),
+    [
+        (["n_counts", "percent_mito"], "regress_test_small.npy", 0.0),
+        (["bulk_labels"], "regress_test_small_cat.npy", 1e-6),
+    ],
+)
+def test_regress_out_reproducible(keys, test_file, atol):
```
ilan-gold, line 446: Shouldn't we add a test for integer + float as a param to this test?

src/scanpy/preprocessing/_simple.py
```diff
@@ -732,13 +748,16 @@ def regress_out(
...
+        number_categories = cat_array.dtype.type(len(adata.obs[keys[0]].cat.categories))
+
         X = _to_dense(X, order="F") if issparse(X) else X
-        # TODO figure out if we should use a numba kernel for this
-        for category in adata.obs[keys[0]].cat.categories:
-            mask = (category == adata.obs[keys[0]]).values
-            for ix, x in enumerate(X.T):
-                regressors[mask, ix] = x[mask].mean()
+        if np.issubdtype(X.dtype, np.integer):
+            target_dtype = np.float32 if X.dtype.itemsize == 4 else np.float64
```
ilan-gold: What's the point of `float32` here? Why not just always use `float64`? Wouldn't it be more accurate?

ilan-gold: Also, if there is a reason, just to be more clean:

```suggestion
            target_dtype = np.float32 if X.dtype.itemsize < 4 else np.float64
```

ilan-gold: Lastly, maybe I'm misremembering, but did we not discuss putting this in `to_dense` as well (so some sort of `target_dtype` created before the `X = _to_dense(X, order="F") if issparse(X) else X` line)? That way we don't have to make two copies in the sparse case.

ilan-gold: Should also apply to #3461

ilan-gold (changes requested):
Why no test for the dtype if we're also fixing that bug here? or in #3461?

docs/release-notes/3353.performance.md
```diff
@@ -0,0 +1 @@
+* Speed up for a categorical regressor in {func}`~scanpy.pp.regress_out` {smaller}`S Dicks`
```
flying-sheep:
```suggestion
Speed up for a categorical regressor in {func}`~scanpy.pp.regress_out` {smaller}`S Dicks`
```

src/scanpy/preprocessing/_simple.py
ilan-gold: Yes! discoverered a bug! `regressors` needs to be `float32` to match old behavior, not sure why it was hardcoded

tests/test_preprocessing.py
```diff
@@ -15,6 +15,7 @@
 
 import scanpy as sc
 from scanpy._compat import CSBase
+from scanpy.datasets._datasets import pbmc3k
```
flying-sheep: wrong import, please be careful that you import these from the testing helpers as intended

src/scanpy/preprocessing/_simple.py
```diff
@@ -632,7 +632,8 @@ def _create_regressor_categorical(
     X: np.ndarray, number_categories: int, cat_array: np.ndarray
 ) -> np.ndarray:
     # create regressor matrix for categorical variables
-    regressors = np.zeros(X.shape, dtype=X.dtype)
+    # would be best to use X dtype but this matches old behavior
+    regressors = np.zeros(X.shape, dtype=np.float32)
```
flying-sheep, line 636: That does seem suboptimal. I wonder how an upgrade path would look here

docs/release-notes/3353.performance.md
```diff
@@ -0,0 +1 @@
+Speed up for a categorical regressor in {func}`~scanpy.pp.regress_out` {smaller}`S Dicks`
```
flying-sheep: Ilan, you also contributed, maybe add yourself?

src/scanpy/preprocessing/_simple.py
ilan-gold: We could easily add an argument, `upcast_regressor` or similar, and tack it onto scanpy 2.0 or something

flying-sheep (approved):

PR #3357, opened 2024-11-12
(chore): generate 1.10.4 release notes

flying-sheep:
Making a release because of the `anndata.io` compat fix in #3289

Also
- pins `patsy!=1.0.0` because of https://github.com/pydata/patsy/issues/215
- fixes `ci/scripts/min-deps.py` to handle deps with just an `!=` dependency spec
- removes the `get-annotations; python_version < '3.10'` dependency since we missed that in #3283

@meeseeksmachine backport to main

### DIFF
--- a/ci/scripts/min-deps.py
+++ b/ci/scripts/min-deps.py
@@ -1,4 +1,11 @@
 #!/usr/bin/env python3
+# /// script
+# dependencies = [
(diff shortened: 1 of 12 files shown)


flying-sheep:
@meeseeksmachine backport to main

flying-sheep:
@meeseeksdev backport to main

PR #3362, opened 2024-11-12
Updated Harmony Integrate Docs to better match interface to Harmonypy package

DaminK:
Fixed harmony documentation where a crucial feature of Harmony, namely the integration w.r.t. multiple covariates was not mentioned despite being supported by the interface. Harmony can take strings and lists of strings as its "vars_use" parameter which is named "key" in the Scanpy interface.

- [ ] Closes #
- [X] Tests included or not required because:

Small non-breaking change to docs. However, pre-commit throws: 
```
ruff.....................................................................Failed
- hook id: ruff
- files were modified by this hook
```
Will wait if some maintainer know an easy fix for this before trying to debug this myself. Guess its more related to incorrect flags somewhere than the code.
- [X] Release notes not necessary because:
Change docs to better reflect current code behaviour/features

### DIFF
--- a/docs/release-notes/3362.doc.md
+++ b/docs/release-notes/3362.doc.md
@@ -1,0 +1,1 @@
+Improve {func}`~scanpy.external.pp.harmony_integrate` docs {smaller}`D Kühl`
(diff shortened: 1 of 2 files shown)


flying-sheep:
Thanks!

PR #3379, opened 2024-11-21
(perf): remove out of core median in normalize_total

Intron7:
Computing counts per cell for `normalize_total` leads to better performance.


flying-sheep:
Fair enough. Can this blow up? I.e. is `counts` ever big enough that this is a problem?

Intron7:
So I did some testing with 1Billion cells. 1Billion cells at 64bit are roughly 8GB. The median costs another 8GB. Thats roughly 16GB. I dont think 16GB of Ram is an issue for someone that analyses 1Billion cells.

ilan-gold:
Why is this for the 1.12 milestone?

flying-sheep:
probably because no 1.11.1 milestone existed for a bit (my fault, I’ll be cleaner about this in the future)

ilan-gold (commented):
Please add a test to get the coverage to pass, but the code itself looks fine.

flying-sheep:
Looks like the tests don’t pass anymore. Any idea what’s wrong?

selmanozleyen:
They passed the same tests on the PR right? The results I checked were due to small differences I am not sure why

flying-sheep:
Ah, it’s unrelated. The test that failed doesn’t call `normalize_total`.

PR #3384, opened 2024-11-22
Switch to session-info2

### LINKED ISSUE #3202: Depend on session-info2
https://github.com/flying-sheep/session-info2

flying-sheep:
- [x] Closes #3202

TODO: update notebooks


Intron7:
I like it a lot. However we might need a fallback mode if `ipywidgets` are not installed.

flying-sheep:
Great point! It doesn’t fail if the import isn’t possible, but the current fallback is just printing the stack trace. That could be nicer. Adding this in https://github.com/flying-sheep/session-info2/pull/81

Intron7 (approved):

PR #3389, opened 2024-11-28
A memory efficient implementation of the .mtx reading function

gjeuken:
- [ ] Closes #
- [x] Tests included or not required because: test_datasets.py already implemented

- [x] Release notes not necessary because: This is a backend change

Pandas `read_csv` function is very memory intensive, and this makes loading data (especially large datasets from EBI Single Cell Expression Atlas) impossible on computers with 16gb of ram or less. The subsequent analysis of such datasets with scanpy, however, works well on such computers.

Loading the data into chunks, using the same pandas function, solves this problem.


src/scanpy/datasets/_ebi_expression_atlas.py
```diff
@@ -67,13 +67,19 @@ def read_mtx_from_stream(stream: BinaryIO) -> sparse.csr_matrix:
...
-    data = pd.read_csv(
+    chunks = pd.read_csv(
         stream,
         sep=r"\s+",
         header=None,
         dtype={0: coord_dtype, 1: coord_dtype, 2: np.float32},
+        chunksize=1e7,
     )
-    mtx = sparse.csr_matrix((data[2], (data[1] - 1, data[0] - 1)), shape=(m, n))
+    mtx = sparse.csr_matrix(([0], ([0], [0])), shape=(m, n))
```
flying-sheep:
```suggestion
    mtx = sparse.csr_matrix((m, n), dtype=np.float64)
```

src/scanpy/datasets/_ebi_expression_atlas.py
```diff
@@ -67,13 +67,19 @@ def read_mtx_from_stream(stream: BinaryIO) -> sparse.csr_matrix:
...
+    for data in chunks:
+        mtx_chunk = sparse.csr_matrix(
+            (data[2], (data[1] - 1, data[0] - 1)), shape=(m, n)
+        )
+        mtx = mtx + mtx_chunk
```
flying-sheep: This is probably slightly slower than necessary, since we know the chunks don‘t overlap and `=` needs to deal with actually summing up things. But I imagine it could also be pretty well optimized, so if the following is not faster, please just add a comment instead explaining that `+` is well-enough optimized.

The way `csr_matrix((data, (i, j)), [shape])` works is that it first creates a `coo_matrix`, then converts it to `csr`.
I think the best way is probably:

1. build up `data, i, j` arrays in a loop
2. create a `csr_matrix` from the final arrays as the last step

flying-sheep (commented):
Good idea! some small notes:

src/scanpy/datasets/_ebi_expression_atlas.py
gjeuken: The improvement in speed is not really significant, but you are correct, and this solution has now been implemented.

gjeuken:
Please note that the selection of the chunk size of ``1e7`` is rather arbitrary.
This value solves the loading issue for computers with 8 or 16gb of RAM without looping through too many chunks.

src/scanpy/datasets/_ebi_expression_atlas.py
flying-sheep: great!

flying-sheep:
Thank you! Since you don’t seem interested in appearing in a release note:

grafik

… I’ll merge this as-is. If you want a release note after all, please comment, and I’ll add one!

gjeuken:
If you think that it would be informative to include this in the release notes, then I have no problem with it :)

flying-sheep:
OK! I’m working on it a bit more anyway in #3484

gjeuken:
Ah, I see now that I missed the already established dtypes, a few lines above.

PR #3392, opened 2024-12-05
(feat): `elem_mul` 64 bit upcast option

### LINKED ISSUE #3127: `elem_mul` in variance calculation should use `float64` casting

### What happened?

When we calculate `X*X` for variance, we preserve the data type of the incoming `X`, but this actually can cause downstream inaccuracies from overflow differences.   This has been the [case for many years](https://github.com/scverse/scanpy/blame/27e5f1fa62ec05ee9db0dfa7a9decbf8f25caa31/scanpy/preprocessing/_utils.py#L10)

Really we should do something like `np.multiply(X, X, dtype="float64)`.  This would be more accurate/sensible.  This came up in the context of https://github.com/scverse/scanpy/pull/3099/files#diff-afb2fb35cbde7ff5e7d9b79874ede22605918cdba923250dd554f23353702e45R65-R67 where @Intron7 was casting first, and then multiplying (because it should be more accurate), but this revealed that we are _not_ doing this at the moment, despite the fact that it is more accurate.  And then downstream analyses can change quite a bit.

Thus we should remedy this for the next minor release as it is a breaking change to e.g., https://dev.azure.com/scverse/scanpy/_build/results?buildId=7094&view=logs&j=5ea502cf-d418-510c-3b5f-c4ba606ae534&t=534778bb-2f86-5739-7d3c-59518f7b5a2b&l=2171
Screenshot 2024-06-26 at 16 21 27

### Minimal code sample

```python
import numpy as np
arr = np.random.random((10, 10_000)).astype("float32")

print(np.multiply(arr, arr, dtype="float64"))
print(np.multiply(arr, arr))
```

### Error output

```pytb
N/A
```

### Versions

<details>

```
-----
anndata     0.10.7
scanpy      1.10.0rc2.dev74+g1c98fd19
-----
IPython             8.24.0
PIL                 10.3.0
asciitree           NA
asttokens           NA
cloudpickle         3.0.0
cycler              0.12.1
cython_runtime      NA
dask                2024.5.1
dateutil            2.9.0.post0
decorator           5.1.1
defusedxml          0.7.1
distutils           3.12.3
executing           2.0.1
h5py                3.11.0
igraph              0.11.5
jedi                0.19.1
jinja2              3.1.4
joblib              1.4.2
kiwisolver          1.4.5
legacy_api_wrap     NA
leidenalg           0.10.2
llvmlite            0.42.0
louvain             0.8.2
markupsafe          2.1.5
matplotlib          3.9.0
mpl_toolkits        NA
msgpack             1.0.8
natsort             8.4.0
numba               0.59.1
numcodecs           0.12.1
numpy               1.26.4
packaging           24.0
pandas              2.2.2
parso               0.8.4
pkg_resources       NA
prompt_toolkit      3.0.45
psutil              5.9.8
pure_eval           0.2.2
pyarrow             16.1.0
pygments            2.18.0
pyparsing           3.1.2
pytz                2024.1
scipy               1.13.1
session_info        1.0.0
setuptools          70.0.0
setuptools_scm      NA
sitecustomize       NA
six                 1.16.0
sklearn             1.5.0
sparse              0.15.4
sphinxcontrib       NA
stack_data          0.6.3
tblib               3.0.0
texttable           1.7.0
threadpoolctl       3.5.0
tlz                 0.12.1
toolz               0.12.1
traitlets           5.14.3
wcwidth             0.2.13
yaml                6.0.1
zarr                2.18.2
-----
Python 3.12.3 (main, Apr  9 2024, 08:09:14) [Clang 15.0.0 (clang-1500.1.0.2.5)]
macOS-13.6.1-arm64-arm-64bit
-----
Session information updated at 2024-06-26 16:19

```

</details>

ilan-gold:
- [x] Closes #3127 


ilan-gold:
I'm putting this in scanpy2.0 because this setting should be turned on then

flying-sheep (commented):
So this will just use 64 bits for computation and downcast again after?

ilan-gold:
> So this will just use 64 bits for computation and downcast again after?

I hadn't considered that.  There's no downcasting at the moment, it just uses 64 bits for the multiplication.  Maybe it should downcast to the original?

I could see that being beneficial, but then the question is what do we downcast to? The data type of the right or left matrix?

We would have to make this clear in the documentation.

flying-sheep:
@ivirshup [said on Zulip](https://scverse.zulipchat.com/#narrow/channel/328272-scanpy/topic/upcasting.20to.2064.20bits.20for.20elementwise.20multiplication/near/492783006):

> I think the intermediary upcast is an implementation detail (in theory we could rewrite the calculation to not upcast, and still be accurate) and should not change the result dtype. For what dtype the result should be, we should use numpy type promotion rules, which should be: `f32 * f32 -> f32`, `f32 * f64 -> f64`, `f64 * f64 -> f64`.

ilan-gold:
Ok I actually believe `numpy` does this by default:

```python
import numpy as np
arr = np.random.random((10, 10_000)).astype("float32")

np.testing.assert_array_equal(np.multiply(arr, arr), np.multiply(arr, arr, dtype="float64").astype(np.float32), strict=True)
```
i.e., multiplying two 32 bit arrays together allocates 64 bits under the hood, and then lops off the bits.  This can be seen since without the `astype(np.float32)` in the assert, the dtypes don't match but with it, the assertion passes

My mistake was not checking if the result of the array multiplication without `float64` is the same, but with the bits lopped off it seems

ilan-gold:
See https://github.com/numpy/numpy/issues/28171 for an answer

flying-sheep:
Is this PR necessary then?

ilan-gold:
I should have clarified, _future answer_ - no one has replied to the issue yet

flying-sheep:
I’m pretty sure numpy is stable enough that we can rely on this behavior sticking around, but confirmation is better than my guess!

ilan-gold:
I also agree that the behavior appears to be such that this PR is unnecessary, but I'd rather have verbal confirmation that I'm not missing something.

ilan-gold:
Ok, I think I was just extremely confused.  Looking back the original issue, a few things arise:

1. We actually do do an upcast to float64 in mean-var calculation, as Severin does.  We just don't do it everywhere `elem_mul` is used (which is fine).
2. Because of this, there should be no issue.  Furthermore, according to the linked issue (and also this makes sense), numpy does indeed do this internally.  So I will close this.

PR #3413, opened 2024-12-20
Metadata updates

flying-sheep:
@ilan-gold @Intron7 @Zethson do you think this is a fair assessment that we 4 are the only active-ish scanpy devs, and the 3 of us that meet weekly can be considered maintainers?

@ilan-gold @Intron7 Can I add your emails to the maintainer list? (I never get emails through that I think)


ilan-gold (approved):

Zethson:
Yeah that seems fair. I'd say I borderline qualify given that I mostly just triage issues but I'd happily step it up at tiny bit in the future to keep my position on this list.

But yes.

PR #3414, opened 2024-12-20
Formatting

flying-sheep:
I don’t think we want pyproject-fmt, as it sorts our extras. We have so many of those that organizing becomes important. We probably just want to use taplo directly.

I configured Taplo to use pyproject-fmt style (120 columns, spaces inside of arrays and dicts), otherwise left the structure alone.


pyproject.toml
```diff
@@ -1,30 +1,30 @@
...
+    { name = "Tobias Callies" },
+    { name = "Marius Lange" },
+    { name = "Andrés R. Muñoz-Rojas" },
 ]
 maintainers = [
-    {name = "Philipp Angerer", email = "phil.angerer@gmail.com"},
-    {name = "Ilan Gold"},
-    {name = "Severin Dicks"},
+    { name = "Philipp Angerer", email = "phil.angerer@gmail.com" },
+    { name = "Ilan Gold" },
```
ilan-gold:
```suggestion
    { name = "Ilan Gold", email = "ilan.gold@helmholtz-munich.de" },
```

ilan-gold (approved):

pyproject.toml
Intron7: Do we want our scverse emails for all of the maintainers

ilan-gold:
```suggestion
    { name = "Ilan Gold", email = "ilan.gold@scverse.org" },
```

PR #3416, opened 2024-12-24
(chore): generate 1.10.5 release notes

dinakazemi:
@meeseeksdev backport to main


Zethson:
This PR doesn't make a lot of sense to me. What was it's intended purpose? The title? We can reopen it if I get more justification.

flying-sheep:
I think @dinakazemi found our towncrier automation script and ran it, not knowing that it would create that PR.

PR #3428, opened 2025-01-09
Implement missing plot functionality for `pl.rank_genes_groups`

### LINKED ISSUE #3205: Rank_genes_groups -> does not return Axes

### What happened?

Function Rank_genes_groups() does not return a figure -> returns None type
Cannot get Figure via plt.gcf(), plt.gca().

Potential Fix:
https://github.com/scverse/scanpy/blob/main/src/scanpy/plotting/_tools/__init__.py
Line 485:  cann be extended to the following as in other functions below:
```
    savefig_or_show(f"{key}_", show=show, save=save)
    show = settings.autoshow if show is None else show
    if show:
        return None
    return ax
```

### Minimal code sample

```python
fig = sc.pl.rank_genes_groups(adata, show=False)
type(fig)
#NoneType
plt.gca() -> empty axes
plt.gcf() -> empty figure
```

### Error output

_No response_

### Versions

<details>

```
-----
anndata     0.10.8
scanpy      1.10.2
-----
```

</details>

flying-sheep:
- [x] Closes #3205


src/scanpy/plotting/_tools/__init__.py
```diff
@@ -474,15 +489,21 @@ def rank_genes_groups(
         if count % n_panels_x == 0:
             ax.set_ylabel("score")
 
-    if sharey is True:
+    if sharey is True and ax is not None:
```
ilan-gold: What case is `ax` `None` here? It looks like `ax` is set inside the above loop so that would mean `group_names` is a length-0 object?

ilan-gold (commented):
Why no test?

src/scanpy/plotting/_tools/__init__.py
flying-sheep: Yeah!

flying-sheep:
I guess because plotting is broken enough and wants a rework enough that I didn’t bother but sure, why not, I’ll add one.

ilan-gold (commented):
actual
expected
Do you expected/actual look like this? Should the actual not have grid lines? I've never noticed the grid lines before on the exepcted

flying-sheep:
Oof our plotting test setup sucks.

OK, so the `plt.subplots()` creates a big `Axes` (plot area) with its own frame, grid lines, … then `rank_genes_groups(..., ax=ax)` plots its multiple small plots over that area. In order to not have to make a new test image, I just call `set_axis_off`. Then I noticed that the test failed because of tiny distortions in aspect ratio, and raised the tolerance for that.

Unfortunately the problem with comparing images is that “line is shifted by 1px” has a *larger* diff than “line is added/removed”. So raising the tolerance until almost imperceptible distortions are accepted made it accept basically anything.

I think to take plotting tests seriously while sticking with image comparisons, we’d have to
1. set *all* tolerances super low (our most common value of 15 is honestly far too large and will miss some added/removed lines)
2. aggressively bump the matplotlib min version whenever its behavior changed (updating all test reference image in the process), so the min-deps tests and stable/pre-deps tests can use the same images.

for this specific test, I could research 20 minutes until I find the matplotlib knob to adjust the aspect ratio so the images match, make a new test image that’s imperceptibly different from an existing one, or call it a day.

tests/test_plotting.py
```diff
@@ -853,17 +853,17 @@ def test_rank_genes_groups(image_comparer, name, fn):
...
     sc.tl.rank_genes_groups(pbmc, "louvain", n_genes=pbmc.raw.shape[1])
 
     pbmc.var["symbol"] = pbmc.var.index + "__"
 
-    with plt.rc_context({"axes.grid": True, "figure.figsize": (12, 16)}):
-        fig, ax = plt.subplots()
+    fig, ax = plt.subplots(figsize=(12, 16))
     ax.set_axis_off()
-    axes: list[Axes] = fn(pbmc, ax=ax, show=False)
+    with plt.rc_context({"axes.grid": True}):
```
ilan-gold, line 865: Maybe just worth adding a comment why this is here

ilan-gold (approved):

tests/test_plotting.py
flying-sheep: it’s there because Fidel added it here, I can’t really say: https://github.com/scverse/scanpy/pull/369

PR #3431, opened 2025-01-10
Replace spmatrix with _CSMatrix as appropriate

flying-sheep:
- [ ] Closes #

- [x] Release notes not necessary because: The only user-facing change is some subtly typing fixes

Preparation for sparray support: make sure `spmatrix` is only used where it makes sense.

I checked every single instance of `spmatrix` and either tested that `coo_matrix` also works or replaced it with `_CSMatrix = csr_matrix | csc_matrix`.

TODO after this PR:
1. replace spmatrix with `_SpBase = sparray | spmatrix` and `_CSMatrix` with `_CSBase = cs{cr}_matrix | cs{cr}_array` and test
2. figure out `bsr_array`

### DIFF
--- a/src/scanpy/metrics/_common.py
+++ b/src/scanpy/metrics/_common.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import warnings
 from functools import singledispatch
-from typing import TYPE_CHECKING, TypeVar
+from typing import TYPE_CHECKING, overload
 
 import numpy as np
 import pandas as pd
@@ -11,13 +11,24 @@ from scipy import sparse
 from .._compat import DaskArray
 
 if TYPE_CHECKING:
+    from typing import NoReturn, TypeVar
+
     from numpy.typing import NDArray
+
+    T_NonSparse = TypeVar("T_NonSparse", bound=NDArray | DaskArray)
+    V = TypeVar("V", bound=NDArray | sparse.csr_matrix)
+
+
+@overload
+def _resolve_vals(val: T_NonSparse) -> T_NonSparse: ...
+@overload
+def _resolve_vals(val: sparse.spmatrix) -> sparse.csr_matrix: ...
+@overload
+def _resolve_vals(val: pd.DataFrame | pd.Series) -> NDArray: ...
 
 
 @singledispatch
-def _resolve_vals(
-    val: NDArray | sparse.spmatrix | DaskArray,
-) -> NDArray | sparse.csr_matrix | DaskArray:
+def _resolve_vals(val: object) -> NoReturn:
     msg = f"Unsupported type {type(val)}"
     raise TypeError(msg)
 
@@ -38,16 +49,11 @@ def _(val: sparse.spmatrix) -> sparse.csr_matrix:
 
 @_resolve_vals.register(pd.DataFrame)
 @_resolve_vals.register(pd.Series)
-def _(val):
+def _(val: pd.DataFrame | pd.Series) -> NDArray:
     return val.to_numpy()
 
 
-V = TypeVar("V", np.ndarray, sparse.csr_matrix)
-
-
-def _check_vals(
-    vals: V,
-) -> tuple[V, NDArray[np.bool_] | slice, NDArray[np.float64]]:
+def _check_vals(vals: V) -> tuple[V, NDArray[np.bool_] | slice, NDArray[np.float64]]:
     """\
     Checks that values wont cause issues in computation.
 
--- a/src/scanpy/metrics/_common.py
+++ b/src/scanpy/metrics/_common.py
@@ -15,19 +15,24 @@ if TYPE_CHECKING:
 
 
 @singledispatch
-def _resolve_vals(val: NDArray | sparse.spmatrix) -> NDArray | sparse.csr_matrix:
-    return np.asarray(val)
+def _resolve_vals(
+    val: NDArray | sparse.spmatrix | DaskArray,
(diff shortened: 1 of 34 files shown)


src/scanpy/metrics/_common.py
```diff
@@ -15,19 +15,24 @@
 
 
 @singledispatch
-def _resolve_vals(val: NDArray | sparse.spmatrix) -> NDArray | sparse.csr_matrix:
-    return np.asarray(val)
+def _resolve_vals(
+    val: NDArray | sparse.spmatrix | DaskArray,
+) -> NDArray | sparse.csr_matrix | DaskArray:
+    msg = f"Unsupported type {type(val)}"
+    raise TypeError(msg)
```
ilan-gold: Does this type hint make sense? Why not `Any`?

src/scanpy/preprocessing/_qc.py
```diff
@@ -298,9 +298,9 @@ def calculate_qc_metrics(
         )
     # Pass X so I only have to do it once
     X = _choose_mtx_rep(adata, use_raw=use_raw, layer=layer)
-    if isspmatrix_coo(X):
+    if isinstance(X, spmatrix) and not isinstance(X, _CSMatrix):
```
ilan-gold: This check seems different than the one before, and I'm not sure we want to be _more_ permissive

src/scanpy/tools/_utils.py
```diff
@@ -105,28 +105,33 @@ def preprocess_with_pca(adata, n_pcs: int | None = None, random_state=0):
 
 
 def get_init_pos_from_paga(
-    adata, adjacency=None, random_state=0, neighbors_key=None, obsp=None
+    adata: AnnData,
+    adjacency: spmatrix | None = None,
```
ilan-gold, line 110: Was the `spmatrix` here intentional?  I could see an argument for it because this function doesn't use it (so this type is permissive) but also, it fetches it out of `obsp` if it's not present suggesting that it has an "expected" type

ilan-gold (approved):
Looks good, about the plan:

> replace spmatrix with _SpBase = sparray | spmatrix and _CSMatrix with _CSBase = cs{cr}_matrix | cs{cr}_array and test

I am not sure this will go so smoothly because of differing APIs and version compatibility.  It was something of a nightmare in `AnnData`...

src/scanpy/metrics/_common.py
flying-sheep: Two points here:
- the `@singledispatch`-decorated function is the one actually seen by typing. So either we also add `@overload` or give it the catch-all typing for all variants. (See https://github.com/python/mypy/issues/11727)
- typing a function that *might* not return is by pretending the exception case doesn’t exist (i.e. every function also has type `(Any) -> NoReturn`). Only functions that indeed never return should have type `(Any) -> NoReturn` or so.

So together, *if* typing would work as expected together with `@singledispatch` you’d be right: this branch should have type `(Any) -> NoReturn`.

So I’m making you right by adding `@overload`s

src/scanpy/preprocessing/_qc.py
flying-sheep: Yeah, do we actually support `coo_matrix` values in AnnData? If not, we can probably just remove this code.

I made the change because if the code is “just coerce to `csr_matrix`”, there’s no reason to single out `coo_matrix`, we can just coerce everything that we know can be coerced.

src/scanpy/tools/_utils.py
flying-sheep: I went by what’s actually done with the type, and this one just gets passed to `networkx`, which says it can handle all sparse matrix classes.

src/scanpy/metrics/_common.py
flying-sheep: Also I’d generally ban `Any` as it’s infectious (`any_val.attr` -> `Any`, `any_val["x"]` -> `Any`, `func(any_val)` -> `Any`). A parameter should either have a type or be `object` in case we don’t know anything. (`object`s can still be made into other types in `if isinstance(val, SomeType)` blocks)

src/scanpy/preprocessing/_qc.py
ilan-gold: > Yeah, do we actually support coo_matrix values in AnnData? If not, we can probably just remove this code.

Very soon we will not actually:

```python
import numpy as np
import scipy.sparse as sp
import anndata as ad

arr = sp.random(10, 100, format="coo", random_state=np.random.default_rng())
ad.AnnData(X=arr)
```
yields
```
/Users/ilangold/Projects/Theis/anndata/src/anndata/_core/storage.py:48: FutureWarning: AnnData previously had undefined behavior around matrices of type <class 'scipy.sparse._coo.coo_matrix'>.In 0.12, passing in this type will throw an error. Please convert to a supported type.Continue using for this minor version at your own risk.
  warnings.warn(msg, FutureWarning)
```

src/scanpy/metrics/_common.py
ilan-gold: Thanks so much for the explanation.  Super interesting, had never thought of it in terms of "all functions do that" before, and had forgotten about `singledispatch` typing weirdness.  Thanks!

src/scanpy/preprocessing/_qc.py
flying-sheep: Then I’ll just go back to `coo`, letting anndata handle the warning.

PR #3438, opened 2025-01-14
Grammar fixes in `sc.tl` docstrings

zm711:
Been working with scanpy for a bit now and really appreciate all the work you're all doing. Noticed a few grammar mistakes in docstrings. If it's helpful I figured I could submit a couple PRs to fix up docstrings :) . As a demonstration this is just a quick rundown of the `tools` submodule. If this isn't desired just let me know.

- [x] Closes # NA
- [x] Tests included or not required because: I was just updating some docstrings to be slightly more idiomatic English so no api or meaning changes.

- [x] Release notes not necessary because: Same as above. Meaning not changed for anything. Just grammar fixes. Although if you want something here let me know.

### DIFF
--- a/src/scanpy/tools/_dendrogram.py
+++ b/src/scanpy/tools/_dendrogram.py
@@ -61,9 +61,9 @@ def dendrogram(
 
     The hierarchical clustering can be visualized using
     :func:`scanpy.pl.dendrogram` or multiple other visualizations
-    including a dendrogram: :func:`~scanpy.pl.matrixplot`,
+    that can include a dendrogram: :func:`~scanpy.pl.matrixplot`,
     :func:`~scanpy.pl.heatmap`, :func:`~scanpy.pl.dotplot`,
-    or :func:`~scanpy.pl.stacked_violin`.
+    and :func:`~scanpy.pl.stacked_violin`.
 
     .. note::
         The computation of the hierarchical clustering is based on predefined
--- a/src/scanpy/tools/_dendrogram.py
+++ b/src/scanpy/tools/_dendrogram.py
@@ -60,10 +60,10 @@ def dendrogram(
     to compute a correlation matrix.
 
     The hierarchical clustering can be visualized using
-    :func:`scanpy.pl.dendrogram` or multiple other visualizations that can
-    include a dendrogram: :func:`~scanpy.pl.matrixplot`,
+    :func:`scanpy.pl.dendrogram` or multiple other visualizations
+    including a dendrogram: :func:`~scanpy.pl.matrixplot`,
     :func:`~scanpy.pl.heatmap`, :func:`~scanpy.pl.dotplot`,
-    and :func:`~scanpy.pl.stacked_violin`.
+    or :func:`~scanpy.pl.stacked_violin`.
 
     .. note::
         The computation of the hierarchical clustering is based on predefined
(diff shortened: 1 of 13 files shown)


src/scanpy/tools/_dendrogram.py
```diff
@@ -60,10 +60,10 @@ def dendrogram(
     to compute a correlation matrix.
 
     The hierarchical clustering can be visualized using
-    :func:`scanpy.pl.dendrogram` or multiple other visualizations that can
-    include a dendrogram: :func:`~scanpy.pl.matrixplot`,
+    :func:`scanpy.pl.dendrogram` or multiple other visualizations
+    including a dendrogram: :func:`~scanpy.pl.matrixplot`,
```
Zethson: Not sure about this one. I'd keep the current wording.

src/scanpy/tools/_dendrogram.py
```diff
@@ -60,10 +60,10 @@ def dendrogram(
...
     :func:`~scanpy.pl.heatmap`, :func:`~scanpy.pl.dotplot`,
-    and :func:`~scanpy.pl.stacked_violin`.
+    or :func:`~scanpy.pl.stacked_violin`.
```
Zethson: Not sure about this one. I'd keep the current wording.

src/scanpy/tools/_draw_graph.py
```diff
@@ -56,14 +56,14 @@ def draw_graph(
     Force-directed graph drawing :cite:p:`Islam2011,Jacomy2014,Chippada2018`.
 
     An alternative to tSNE that often preserves the topology of the data
-    better. This requires to run :func:`~scanpy.pp.neighbors`, first.
+    better. This requires running :func:`~scanpy.pp.neighbors`, first.
```
Zethson, line 59: Maybe even `having run`?

Zethson (commented):
Great! Please keep the docstring improvements and fixes coming. A few minor points. I am not a native speaker and am happy to be convinced that my comments are not valid.

src/scanpy/tools/_dendrogram.py
zm711: That's fine to keep the current wording if your goal is to include the ability to include them all together/in different combos. Happy to change this back.

src/scanpy/tools/_dendrogram.py
zm711: Yep if they can be included together can change back. The current wording meant you can choose between either a dendrogram or any of the other three in combination.

My wording states you can choose dendrogram or any of the other three. 

To be honest I think you mean that you can do a dendrogram AND you could add a combination of any of the other three (or maybe any one of the other three?)

src/scanpy/tools/_draw_graph.py
zm711: In this case either having run or running is fine especially because we have the word first to emphasize the timing. \
I've noticed in the docstring there is a preference for using the perfect tense in the past when in (American) English we would tend to use simple past or gerunds (maybe British English emphasizes the perfect more than we do??). In this case the infinitive sounds wrong (although I can't give a grammatical reason off the top of my head), but either the present perfect or a gerund would be fine.

src/scanpy/tools/_dendrogram.py
Zethson: Yeah, I'd say we revert this one.

src/scanpy/tools/_dendrogram.py
Zethson: Yeah that's what I meant. I don't feel too strongly about this one so whatever you prefer.

Zethson:
Feel free to add a release note if you want this PR to be mentioned in the next release. Docstring improvements are important.

src/scanpy/tools/_dendrogram.py
zm711: Done !

zm711:
> Feel free to add a release note if you want this PR to be mentioned in the next release. Docstring improvements are important.

Since I have a few modules to go maybe I could add a release note after I finish a few others (unless the release is super soon ). Then it would be something like `docstring fixes (pr  x pr x pr x)`. That seems cleaner to me.

Zethson:
Sure, fine with me. We can group them later.

Zethson (approved):

PR #3461, opened 2025-02-11
fix int for shortcut

Intron7:
Fixes a bug where the regress_out short cut would error for integer dyte


src/scanpy/preprocessing/_simple.py
```diff
@@ -752,6 +752,9 @@ def regress_out(
     # if the regressors are not categorical and the matrix is not singular
     # use the shortcut numpy_regress_out
     if not variable_is_categorical and np.linalg.det(regressors.T @ regressors) != 0:
+        if np.issubdtype(X.dtype, np.integer):
+            target_dtype = np.float32 if X.dtype.itemsize <= 4 else np.float64
+            X = X.astype(target_dtype)
```
ilan-gold: For dense you can also do the ordering here, since you're copying anyway

Intron7: Done

src/scanpy/preprocessing/_simple.py
```diff
@@ -754,7 +754,7 @@ def regress_out(
     if not variable_is_categorical and np.linalg.det(regressors.T @ regressors) != 0:
         if np.issubdtype(X.dtype, np.integer):
             target_dtype = np.float32 if X.dtype.itemsize <= 4 else np.float64
-            X = X.astype(target_dtype)
+            X = X.astype(target_dtype, order="C")
```
ilan-gold: But this could also apply to sparse? And we don't want the argument then, no? https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.astype.html nothing here about `order`

ilan-gold: indeed it was a bug, tested for and resolved

PR #3484, opened 2025-02-25
Further .mtx reading improvements

flying-sheep:
After merging #3389, I noticed that it changes the dtypes. Then I realized that while the .mtx format has no dtypes (beyond real|complex|pattern), it does store the number of entries, so we can pre-allocate.


src/scanpy/datasets/_ebi_expression_atlas.py
```diff
@@ -58,31 +61,35 @@ def download_experiment(accession: str):
...
 
-def read_mtx_from_stream(stream: BinaryIO) -> sparse.csr_matrix:
+def read_mtx_from_stream(stream: ReadCsvBuffer[bytes]) -> sparse.csr_matrix:
     curline = stream.readline()
     while curline.startswith(b"%"):
         curline = stream.readline()
-    n, m, _ = (int(x) for x in curline[:-1].split(b" "))
+    n, m, e = map(int, curline[:-1].split(b" "))
 
+    dtype_data = np.float32
```
ilan-gold, line 70: I'm a little lost - so before the referenced PR, the default was float32 and now this just goes back to that?

flying-sheep: yup!

ilan-gold (approved):

PR #3485, opened 2025-02-25
Adopt the Scientific Python deprecation schedule:

flying-sheep:
- [x] Closes #3493
- [x] Tests included or not required because:

- [ ] Release notes not necessary because:

Change supported Python versions to 3.11–3.13

See https://github.com/scverse/anndata/pull/1768

The rest flowed from there:
- h5py 3.7 failed to compile, so 3.8 it is 
- only `numba>=0.58` support the new numpy min version
- the lowest `scikit-image` version that has wheels for Python 3.11 depends on `networkx>=0.28`
- https://github.com/scverse/anndata/pull/790 is only available in anndata 0.9 (I wonder how the minimum versions job passed before with anndata 0.8!)

PR #3490, opened 2025-02-27
Use pydocstyle instead of test

flying-sheep:
Wow this is a huge one.

This is necessary for #3485, as Python 3.13 dedents docstrings and breaks that test.

Ruff does a better job than us at testing that.

Findings for people wanting to do the same:

- Do this first, else some auto“fixes” break things

  ```py
  [tool.ruff.lint.pydocstyle]
  convention = "numpy"
  ```

- Don’t rely on unsafe fixes for [D301](https://docs.astral.sh/ruff/rules/escape-sequence-in-docstring/), as it doesn’t replace `\\` with `\` when replacing `"""` with `r"""`

### DIFF
--- a/src/scanpy/_utils/__init__.py
+++ b/src/scanpy/_utils/__init__.py
@@ -17,6 +17,7 @@ from contextlib import contextmanager, suppress
 from enum import Enum
 from functools import partial, reduce, singledispatch, wraps
 from operator import mul, or_, truediv
+from textwrap import indent
 from types import MethodType, ModuleType, UnionType
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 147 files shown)


src/scanpy/_utils/__init__.py
```diff
@@ -243,21 +246,35 @@ def annotate_doc_types(mod: ModuleType, root: str):
...
+            len(m.group(1))
+            for m in _leading_whitespace_re.finditer(obj.__doc__[start_line_2:])
+        )
+
+        # The placeholder is already indented, so only indent subsequent lines
+        indented_replacements = {
+            k: indent(v, " " * n_spaces)[n_spaces:] for k, v in replacements.items()
+        }
+        obj.__doc__ = obj.__doc__.format_map(indented_replacements)
         return obj
```
flying-sheep, line 270: this is probably the only interesting part of this PR

PR #3501, opened 2025-03-11
Deduplicate _reorder_categories_after_dendrogram and _plot_var_groups_brackets

### LINKED ISSUE #3489: Remove duplicate _reorder_categories_after_dendrogram

### What happened?

There are two `_reorder_categories_after_dendrogram`

### Minimal code sample

```python
from scanpy.plotting._anndata import _reorder_categories_after_dendrogram
from scanpy.plotting._baseplot_class import BasePlot

assert BasePlot._reorder_categories_after_dendrogram is not _reorder_categories_after_dendrogram
```

### Error output

```pytb

```

### Versions

master

flying-sheep:
- [x] Closes #3489

- [x] Release notes not necessary because: dev change

Only change should be that `BasePlot._reorder_categories_after_dendrogram` sets `self.var_groups` which can be reordered

### DIFF
--- a/src/scanpy/plotting/_anndata.py
+++ b/src/scanpy/plotting/_anndata.py
@@ -6,7 +6,7 @@ from collections import OrderedDict
 from collections.abc import Collection, Mapping, Sequence
 from itertools import product
 from types import NoneType
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, NamedTuple, TypedDict, cast
 
 import matplotlib as mpl
 import numpy as np
@@ -69,6 +69,11 @@ if TYPE_CHECKING:
     # TODO: is that all?
     _Basis = Literal["pca", "tsne", "umap", "diffmap", "draw_graph_fr"]
     _VarNames = str | Sequence[str]
+
+
+class VarGroups(NamedTuple):
+    labels: Sequence[str]
+    positions: Sequence[tuple[int, int]]
 
 
 @old_positionals(
@@ -1149,9 +1154,10 @@ def heatmap(
     tl.rank_genes_groups
 
     """
-    var_names, var_group_labels, var_group_positions = _check_var_names_type(
+    var_names, var_groups = _check_var_names_type(
         var_names, var_group_labels, var_group_positions
     )
+    del var_group_labels, var_group_positions
 
     categories, obs_tidy = _prepare_dataframe(
         adata,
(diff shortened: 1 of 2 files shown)


src/scanpy/plotting/_anndata.py
```diff
@@ -1451,11 +1471,11 @@ def heatmap(
             )
 
         # plot group legends next to the heatmap_ax (if given)
-        if var_group_positions is not None and len(var_group_positions) > 0:
+        if var_groups is not None:
             gene_groups_ax = fig.add_subplot(axs[1, 1])
             arr = []
             for idx, (label, pos) in enumerate(
-                zip(var_group_labels, var_group_positions)
+                zip(var_groups.labels, var_groups.positions)
```
ilan-gold:
```suggestion
                zip(*var_groups)
```
or no? `NamedTuple` doesn't?

src/scanpy/plotting/_anndata.py
```diff
@@ -2173,17 +2188,16 @@ def _plot_gene_groups_brackets(
...
-    right = [x[1] + right_adjustment for x in group_positions]
+    left = [x[0] + left_adjustment for x in var_groups.positions]
+    right = [x[1] + right_adjustment for x in var_groups.positions]
 
     # verts and codes are used by PathPatch to make the brackets
     verts = []
     codes = []
     if orientation == "top":
         # rotate labels if any of them is longer than 4 characters
-        if rotation is None and group_labels:
```
ilan-gold, line 2185: I guesss there was no way for `group_labels` to be `None` in practice (at least judging by the typing)?

ilan-gold: asking re: why you dropped the check

ilan-gold (approved):
Small questions but otherwise looks good.  A little scared of any plotting change given how brittle the testing code but this one seems fairly good, especially with the validate-on-creation business :)

src/scanpy/plotting/_anndata.py
flying-sheep: that should totally work, yes!

src/scanpy/plotting/_anndata.py
flying-sheep: Before, there were many checks in the code like `if var_groups_?? is None or len(var_groups_??) == 0`, but the validation now takes care that neither can be the case. Either `var_groups is None` or it contains nonempty sequences

flying-sheep: Ah and fun fact: the `wide` parameter I added was because there was some code drift between the two versions of bracket drawing code, and our tests *did* catch that!

So actually maybe we should decide on one version and update the test images.

ilan-gold: Got it

PR #3509, opened 2025-03-14
Switch to GH actions

flying-sheep:
See https://github.com/scverse/anndata/pull/1906


flying-sheep:
@ilan-gold I think you figured out how to make the HDF5 tests more robust?

> Can't synchronously determine if attribute exists by name (invalid identifier type to function)

ilan-gold:
> Can't synchronously determine if attribute exists by name (invalid identifier type to function)

What is this referring to?

flying-sheep:
This is an error that various tests throw when they try to read some HDF5 file, e.g: https://github.com/scverse/scanpy/actions/runs/13859873447/job/38785429107?pr=3509

E.g. the doctest for a dataset or tests using the backed_adata fixture.

I assumed that that’s what your “reopen trick” is there to solve but maybe I guessed wrong.

flying-sheep:
OK, so comparing the old and new job config:

| Azure | GitHub |
|--------|--------|
| py=3.11 | py=3.11 |
| py=3.13 | py=3.13 |
| py=3.13, test-extra=test-min | py=3.13, deps=min-optional |
| py=3.13, deps=pre, tests=cov | py=3.13, deps=pre, tests=cov |
| py=3.11, deps=minimum, (test-extra=test), tests=cov | py=3.11, deps=minimum, tests=cov |

so the only difference is that `TEST_EXTRA` (default: test-full) is gone and replaced by `dependencies-version=min-optional`

before, the job logic respected `TEST_EXTRA` in the `deps=latest` (default) and the `deps=pre` cases (in the `deps=minimum` case the `test` extra was hardcoded).

now, we have a new `dependencies-version=min-optional`, which is the same as the default `dependencies-version` except that the `test-min` extra is used.

Behavior is therefore the same.

PR #3523, opened 2025-03-19
Add in bws() method

### LINKED ISSUE #3503: Add  Baumgartner-Weiss-Schindler test to sc.tl.rank_genes_groups()

Hello,

Long-time user first-time complainer. Love the package. It quite literally has changed the way I do science.

So I have a request for an enhancement for the sc.tl.rank_genes_groups() function. I'm curious why there isn't an option to select a Baumgartner-Weiss-Schindler test when research groups are interested in ranking genes that are more highly variable and could be subject to drop-out in a dataset and thus would have heavy tails in their distributions. I recently encountered this issue while working on a family of genes that I'm interested in but which are also expressed at lower values and so many of the cells have a read count of 0. I got some interesting results and was thinking about the data but I was feeling cautious about how to interpret the results from the Wilcoxon comparison when I compare my groups. 

I was thinking about using some autoencoder deep learning to impute the dropout values in the genes with scanpy.external.pp.dca() and then seeing how much my sc.get.rank_genes_groups_df() changes but I would also like to compare those results to the outputs from a SciPy bws_test() on just the raw data and nothing super processed.

I'm sure other people have hit this issue and because those statistical functions exist they have made a weird hacky way to compute them on their own sanity check but I figured It might be interesting to put this up here and see if this is something the community would like to see incorporated just to streamline this kind of analysis.

Thanks,
Jake Lehle

JakeLehle:
- [X] Closes #3503 

### DIFF
--- a/src/scanpy/tools/_rank_genes_groups.py
+++ b/src/scanpy/tools/_rank_genes_groups.py
@@ -7,6 +7,9 @@ from typing import TYPE_CHECKING, Literal
 import numpy as np
 import pandas as pd
 from scipy.sparse import issparse, vstack
+from scipy.stats import bws_test 
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from tqdm import tqdm
 
 from .. import _utils
 from .. import logging as logg
@@ -31,7 +34,7 @@ if TYPE_CHECKING:
 
 
 # Used with get_literal_vals
-_Method = Literal["logreg", "t-test", "wilcoxon", "t-test_overestim_var"]
+_Method = Literal["logreg", "t-test", "wilcoxon", "bws", "t-test_overestim_var"]
 
 _CONST_MAX_SIZE = 10000000
 
@@ -109,6 +112,7 @@ class _RankGenes:
         layer: str | None = None,
         comp_pts: bool = False,
     ) -> None:
+        self.n_cpu = 1
         self.mask_var = mask_var
         if (base := adata.uns.get("log1p", {}).get("base")) is not None:
             self.expm1_func = lambda x: np.expm1(x * np.log(base))
@@ -150,7 +154,6 @@ class _RankGenes:
         if self.mask_var is not None:
             self.X = X[:, self.mask_var]
             self.var_names = adata_comp.var_names[self.mask_var]
-
         else:
             self.X = X
             self.var_names = adata_comp.var_names
@@ -161,9 +164,11 @@ class _RankGenes:
 
         self.means = None
         self.vars = None
+        self.medians = None
 
         self.means_rest = None
         self.vars_rest = None
+        self.medians_rest = None
 
         self.comp_pts = comp_pts
         self.pts = None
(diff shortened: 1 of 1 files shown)


JakeLehle:
Okay Let's kick this pull request off cause I want some more eyeballs on what I have been doing so I made some changes to the rank_genes() function to add the bws() method which I set up similar to the wilcoxon method in format. I solved a computational bottleneck of trying to compute an integral on a ~25K long list of numbers by dropping zeros for the bws_test() portion and then multithreaded the computation adding that to the function. Finally, I modified how logfold change was being determined for both the wilcoxon and bws test to switch it from a comparison of means to that of medians which is what the wilcoxon test should be using. 

My main issue with this is that I think the setup might actually need to be modified to fit closer to the way you have things set up for the t-test rather than the wilcoxon where you split out your genes and then instead of calculating means and var for each group I can just pipe each of these into the bws_test() 

```        group_data = self.X[mask_obs, i].toarray().flatten() if issparse(self.X) else self.X[mask_obs, i]
        reference_data = self.X[mask_obs_rest, i].toarray().flatten() if issparse(self.X) else self.X[mask_obs_rest, i]

        # Drop zero counts to get over the dataset size bottleneck
        group_data = group_data[group_data != 0]
        reference_data = reference_data[reference_data != 0]
```
        
 And then play around with setting everything up like you do with the t-test so something like this
 
 ```
         for group_index, (mask_obs) in enumerate(
            zip(self.groups_masks_obs)
        ):
        ....
        with np.errstate(invalid="ignore"):
                result = bws_test(group_data, reference_data)
                )
```
        
Take a look and let me know your thoughts and I can play around more with polishing this.

JakeLehle:
I'm making some more changes to the script based on my previous comment which seem like it is working better. Let me play around with it more and read over the contribution documentation so I can make sure I'm doing this right. 

Please ignore this pull request for the moment. I'll fix my issues and try again in a day or so.

flying-sheep:
Hi! Sorry that I didn’t reply, but based on your last comment you wanted to make some changes.

But now I realized that there was no response to you at all! Are you still interested in moving this forward?

JakeLehle:
Hi @flying-sheep yes, thank you for circling back to this. I'm still interested in pushing this forward. My thought that this was going to be a simple patch got derailed the more I tried fitting things in with the previous functions on the rank_genes_groups.py. My interest to run everything off of the comparison of medians for calculating log fold change while still preserving the behavior of the other functions like the t-test method that needs to run on means and thus the calculation of log fold change  using means for that is giving me some errors with how I'm setting up my conditional statements in the existing function. I'm probably overcomplicating the script to make it fancy and trying to fit everything together rather than add in new functions specifically that calculate the quick summary stats differently for the bws and Wilcoxon method. 

Then life happened and PIs make demands and I lose a few weeks setting up scripts for some mutation signature calling and working on some other collaborative projects.

I probably need like a full 2-3 days to finish my changes and figure out the pre-commit checks that you guys have set up. I'll see if I can't block off some time this week to make some solid progress on this and if not detail where I'm stuck and ask for some help.

Thank you for your patience.

flying-sheep:
Don’t worry too much about the pre-commit checks, we can take care of that for you!

JakeLehle:
Hey just an update. I'm gonna pull back for the time being and close out this request and issue. I have the bws() method working on my independent code where I compute "DEGs" separately from the log-normalized anndata count matrix but I don't know if I am in love with the process to get there and the output. 

I think doing a two sides distribution comparison between gene pairs does have some benefits for the identification of DEGs that suffer from high variability and large dropout, rather than the traditional approach of simply comparing mean or median values which would be highly influenced by genes with large numbers of 0 counts. However, the bws approach comes with a high computational cost to find the integral under a curve for each gene to do the comparison and again the genes I'm interested in are those that have low expression and are an edge cases for most users.

While interesting and possible, I think it would be a function that would take a long time to run and be more difficult to interpret for users than I would like. I'm still debating if getting a "DEG" with this approach is biologically relevant or a form of p-value hacking.

I appreciate the support from the maintainers and will circle back and open this up down the road if I feel the results are reliable and will add value to scanpy.

PR #3528, opened 2025-03-26
(fix): allow dense dask pca using `covariance_eigh`

ilan-gold:
- [x] An alternative to https://github.com/dask/dask-ml/issues/985 as someone in the lab complained about this issue (separately tried to convince them not to densify their data for no reason).  Not clear whether this is a feature or not though...basically an oversight on our part that this is super straightforward and it does kind of fix a bug but it is a feature in the API sense.


ilan-gold:
looking into the docs issue btw

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -123,7 +123,7 @@ def pca(
             Not available with *dask* arrays.
         `'covariance_eigh'`
             Classic eigendecomposition of the covariance matrix, suited for tall-and-skinny matrices.
-            With dask, array must be CSR and chunked as (N, adata.shape[1]).
+            With dask, array must be CSR or dense and chunked as (N, adata.shape[1]).
```
flying-sheep, line 126: for `None` above it says

> For sparse *dask* arrays, will use `'covariance_eigh'`.

With this PR, what does `None` do for dense dask arrays?

src/scanpy/preprocessing/_pca/_dask.py
```diff
@@ -187,9 +186,11 @@ def _cov_sparse_dask(
...
         dtype = np.dtype(dtype)
 
-    def gram_block(x_part: _CSMatrix):
-        gram_matrix: _CSMatrix = x_part.T @ x_part
-        return gram_matrix.toarray()[None, ...]  # need new axis for summing
+    def gram_block(x_part: _CSMatrix | NDArray):
+        gram_matrix = x_part.T @ x_part
+        if isinstance(gram_matrix, _CSMatrix):
+            gram_matrix = gram_matrix.toarray()
+        return gram_matrix[None, ...]  # need new axis for summing
```
flying-sheep, line 193: codecov says these lines aren’t covered, is that a lie?

flying-sheep (approved):
Looks indeed very straightforward! Some minor comments

src/scanpy/preprocessing/_pca/__init__.py
ilan-gold: What it did before i.e., 

```
if *dask-ml* :class:`~dask_ml.decomposition.PCA` or :class:`~dask_ml.decomposition.IncrementalPCA` is used, will give `'auto'`
````

src/scanpy/preprocessing/_pca/_dask.py
ilan-gold: It says they weren't covered before either, just now there is more of them.  I think both are lying.

ilan-gold:
@flying-sheep What do you think about `feat` vs `fix`?  Is this a feature or a fix?

flying-sheep:
borderline, I wouldn’t complain about `fix` though!

PR #3553, opened 2025-04-01
Deduplicate graph metrics

flying-sheep:
One step towards scverse/fast-array-utils#82

One outstanding question: https://github.com/scverse/scanpy/pull/915#issuecomment-2769441638

I wouldn’t think about it if these metrics weren’t super precisely tested

### DIFF
--- a/src/scanpy/metrics/_common.py
+++ b/src/scanpy/metrics/_common.py
@@ -45,11 +45,11 @@ class _SparseMetric(ABC):
 
     @abstractmethod
     def mtx(self, vals_het: NDArray | sparse.csr_matrix, /) -> NDArray:
-        """Calculate metric when ``.vals`` is a 2D matrix (on an easier to handle version of ``.vals``)."""
+        """Calculate metric when ``._vals`` is a 2D matrix (on an easier to handle version of ``._vals``)."""
 
     @abstractmethod
     def vec(self) -> np.float64:
-        """Calculate metric when ``.vals`` is a 1D vector."""
+        """Calculate metric when ``._vals`` is a 1D vector."""
 
     def __call__(self) -> np.ndarray:
         match self._vals, self._vals.ndim:
--- a/src/scanpy/metrics/_common.py
+++ b/src/scanpy/metrics/_common.py
@@ -1,22 +1,84 @@
 from __future__ import annotations
 
 import warnings
+from abc import ABC, abstractmethod
+from dataclasses import InitVar, dataclass, field
 from functools import singledispatch
-from typing import TYPE_CHECKING, overload
+from typing import TYPE_CHECKING, ClassVar, overload
 
 import numpy as np
 import pandas as pd
 from scipy import sparse
 
-from .._compat import DaskArray
+from .._compat import DaskArray, fullname
 
 if TYPE_CHECKING:
     from typing import NoReturn, TypeVar
 
+    from anndata import AnnData
     from numpy.typing import NDArray
 
     T_NonSparse = TypeVar("T_NonSparse", bound=NDArray | DaskArray)
     V = TypeVar("V", bound=NDArray | sparse.csr_matrix)
+
+    _Vals = NDArray | sparse.spmatrix | DaskArray | pd.DataFrame | pd.Series
+
+
+__all__ = ["_get_graph", "_SparseMetric"]
+
+
+@dataclass
+class _SparseMetric(ABC):
+    g: sparse.csr_matrix
+    vals: InitVar[_Vals]
+    _vals: NDArray | sparse.csr_matrix | DaskArray = field(init=False)
+
+    name: ClassVar[str]
+
+    def __post_init__(self, vals: _Vals) -> None:
+        assert isinstance(type(self).name, str)
+        assert self.g.shape[0] == self.g.shape[1], (
+            "`g` should be a square adjacency matrix"
+        )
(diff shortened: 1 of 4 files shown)


src/scanpy/metrics/_common.py
```diff
@@ -1,23 +1,85 @@
...
+            "`g` should be a square adjacency matrix"
+        )
+        self.g = self.g.astype(np.float64, copy=False)
+        self._vals = _resolve_vals(vals)
+
+    @abstractmethod
+    def mtx(self, new_vals: NDArray | sparse.csr_matrix) -> NDArray: ...
+
+    @abstractmethod
+    def vec(self) -> np.float64: ...
```
ilan-gold: doc strings maybe? just for us develoeprs

src/scanpy/metrics/_common.py
```diff
@@ -1,23 +1,85 @@
...
 
+    _Vals = NDArray | sparse.spmatrix | DaskArray | pd.DataFrame | pd.Series
+
+
+__all__ = ["_get_graph", "_SparseMetric"]
+
+
+@dataclass
+class _SparseMetric(ABC):
+    g: sparse.csr_matrix
```
ilan-gold: I think we should use a name like `graph` or `weights` but I can be verbose sometimes

src/scanpy/metrics/_common.py
```diff
@@ -1,23 +1,85 @@
...
+
+    name: ClassVar[str]
+
+    def __post_init__(self, vals: _Vals) -> None:
+        assert isinstance(type(self).name, str)
+        assert self.g.shape[0] == self.g.shape[1], (
+            "`g` should be a square adjacency matrix"
+        )
+        self.g = self.g.astype(np.float64, copy=False)
+        self._vals = _resolve_vals(vals)
```
ilan-gold, line 44: I think maybe something like `data` is better than `vals` which us "`_SparseMetric` operates on its `graph` and `data`" instead of "`_SparseMetric` operates on its `g` and `_vals`"

Separately, this `_resolve_val` should be called by the caller IMO and we can use the `__post_init__` to validate and raise `TypeError`.  Code duplication in exchange for not having `vals` (which is never used) and `_vals` is worth it IMO

src/scanpy/metrics/_gearys_c.py
```diff
@@ -109,22 +110,33 @@ def gearys_c(
...
+
+class _GearysC(_SparseMetric):
+    name = "Geary’s C"
+
+    def mtx(self, new_vals: NDArray | sparse.csr_matrix) -> NDArray:
+        g_parts = (self.g.data, self.g.indices, self.g.indptr)
+        if isinstance(new_vals, np.ndarray):
+            return _gearys_c_mtx(*g_parts, new_vals)
+        v_parts = (new_vals.data, new_vals.indices, new_vals.indptr)
+        return _gearys_c_mtx_csr(*g_parts, *v_parts, new_vals.shape)
```
ilan-gold: This and `vec` are basically identical between Moran's and Geary's unless I am missing something with the exception of the actual function i.e., they do this `g_parts` thing, have an instance check etc.  Maybe deduplicate?

src/scanpy/metrics/_common.py
flying-sheep: `vals` is used in the 1D case, but good point!

src/scanpy/metrics/_gearys_c.py
flying-sheep: I’ll see what I can do. The issue is that numba functions necessarily need to have ugly signatures I think. I’ll see if I can improve that and if not, at least wrap them somehow.

flying-sheep: Hmm numba-scipy exists, but it seems pretty dead and has no support for csr_array/csc_array: https://github.com/numba/numba-scipy/issues/109

Maybe we should vendor its code into fast-array-utils and maintain it there?

src/scanpy/metrics/_common.py
ilan-gold: If it is then I don't see it, maybe I'm missing something.

https://github.com/scverse/scanpy/pull/3553/files/c4e037f9a36f9618a1fe3fb3c2d24f17f261b226..9f0f18589fe81c389c8b3072b2331bd0f79f446f#diff-fb08f0d4e96aca4f02fb2014b003a0e560fcddff14bbc532428f407ef5d1aaabR125

https://github.com/scverse/scanpy/pull/3553/files/c4e037f9a36f9618a1fe3fb3c2d24f17f261b226..9f0f18589fe81c389c8b3072b2331bd0f79f446f#diff-e6b16594f2d52a49a1e95379f8387e9c8951396a2f34eb51f12d81326eb6a67bR123

src/scanpy/metrics/_gearys_c.py
flying-sheep: OK, @ilan-gold agreed, so I’ll leave that for now

src/scanpy/metrics/_common.py
flying-sheep: https://github.com/scverse/scanpy/blob/f5ef8f76f4e31d2f5eceb3f007f594811d5099ab/src/scanpy/metrics/_gearys_c.py#L125

ilan-gold (approved):

PR #3554, opened 2025-04-02
(fix): correct `dtype` for `transform` in `PCAEighDaskFit.transform`

### LINKED ISSUE #3487: `dask` `sparse` PCA `dtype`

### What happened?

This isn't so much a bug as it doesn't have an impact on runtime behavior as a "we should be more careful because it could change" (dask ignores this when returning data and just returns the actually computed data, which is floating point)

https://github.com/scverse/scanpy/blob/f6a665b9502d93db67d1a7b3ce6dfa21d62911ca/src/scanpy/preprocessing/_pca/_dask_sparse.py#L129-L137

The output `dtype` of the PCA should be float, not that of the underlying data so we should likely add an argument or repurpose the `dtype_` attribute to reflect this

This same issue is present in the covariance calculation

### Minimal code sample

```python
N/A
```

### Error output

```pytb
N/A
```

### Versions

N/A

ilan-gold:
- [x] Closes #3487

- [x] Release notes not necessary because: Internal change


src/scanpy/preprocessing/_pca/_dask.py
```diff
@@ -33,16 +33,16 @@ def fit(self, x: DaskArray) -> PCAEighDaskFit:
         >>> import dask.array as da
         >>> import scipy.sparse as sp
         >>> x = (
-        ...     da.array(sp.random(100, 200, density=0.3, dtype="float32").toarray())
+        ...     da.array(sp.random(100, 200, density=0.3, dtype="int64").toarray())
```
ilan-gold, line 37: Confirmed that this test now fails on `main` without the change

src/scanpy/preprocessing/_pca/_dask.py
```diff
@@ -131,8 +131,8 @@ def transform_block(
             mean_=self.mean_,
             components_=self.components_,
             chunks=(x.chunks[0], self.n_components_),
-            meta=np.zeros([0], dtype=x.dtype),
-            dtype=x.dtype,
+            meta=np.zeros([]),
+            dtype=np.result_type(x.dtype, self.components_.dtype, self.mean_.dtype),
```
flying-sheep: [`dask.array.map_blocks`](https://docs.dask.org/en/stable/generated/dask.array.map_blocks.html)

> **dtype**: The `dtype` of the output array. […]
> **meta**: The `meta` of the output array, when specified is expected to be an array of the same type and dtype of that returned when calling `.compute()` on the array returned by this function. […]

so shouldn’t this be the following?

I don’t get why dask wants both to avoid running the function on test data. Shouldn’t `meta` be enough, and allow dask to infer `dtype` from `meta`?

```suggestion
            dtype=dtype_out := np.result_type(x.dtype, self.components_.dtype, self.mean_.dtype),
            meta=np.zeros((), dtype=dtype_out),
```

flying-sheep (commented):
I understand your fix, but the fact that you don’t specify the dtype twice confuses me.


PS: I don’t get why dask wants both to avoid running the function on test data. Shouldn’t `meta` be enough, and allow dask to infer `dtype` from `meta`?

src/scanpy/preprocessing/_pca/_dask.py
ilan-gold: Yes! we can just use `meta` all good

ilan-gold:
> I understand your fix, but the fact that you don’t specify the dtype twice confuses me.

My impression was always that dask uses the meta for classes and dtype for dtype, but if you give `meta` a dtype without specifying one for map_blocks, it will use that one.  But if you specify a `dtype`, it will override meta

ilan-gold:
> I don’t get why dask wants both to avoid running the function on test data.

Right so not withstanding that other reason, I also don't like letting dask do this just from past experience but I should get over myself.  So https://github.com/scverse/scanpy/pull/3554/commits/0411969d08b5011362c2d3d91401657a9fe7a480

flying-sheep:
> My impression was always that dask uses the meta for classes and dtype for dtype, but if you give meta a dtype without specifying one for map_blocks, it will use that one. But if you specify a dtype, it will override meta

that sounds like it makes more sense than what the docs imply. Good.

> I also don't like letting dask do this just from past experience

I’m totally fine with specifying everything to avoid that, I just wanted to be clear how to best do it!

The way it is now, we have a `meta` (of course with a dtype, as every numpy array has one), so if Dask gets it from `meta` as you say, it’ll now have the wrong one, right?

ilan-gold:
> The way it is now, we have a meta (of course with a dtype, as every numpy array has one), so if Dask gets it from meta as you say, it’ll now have the wrong one, right?

Right it would be wrong except that `float64` is always the right `dtype` from what I can tell.  If the underlying data is `float32`, the operation promotes to `float64` anyway and all the integer cases end up at `float64` and we don't support `complex`.  So I just dropped the type promotion.  If you can come up with a case where we need it, all ears!

I also did a little similar change as well below the one we were talking about if you want to look

src/scanpy/preprocessing/_pca/_dask.py
```diff
@@ -132,8 +132,7 @@ def transform_block(
             mean_=self.mean_,
             components_=self.components_,
             chunks=(x.chunks[0], self.n_components_),
-            meta=np.zeros([0], dtype=x.dtype),
-            dtype=x.dtype,
+            meta=np.zeros([]),
```
flying-sheep: OK, great, then maybe we should be specific? Otherwise I think I finally understand everything, took me long enough.

```suggestion
            meta=np.array([], dtype=np.float64),
```

ilan-gold:
@flying-sheep any idea what the bot is talking about? The backport seems to have worked

flying-sheep:
no clue! I also don’t know where to find the logs!

Issue #3423, opened 2025-01-01
Support for `scipy.csr_array`/`scipy.csc_array`


alam-shahul:
### Please make sure these conditions are met

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of scanpy.
- [ ] (optional) I have confirmed this bug exists on the main branch of scanpy.

### What happened?

It seems like `sc.pp.pca` doesn't work when `.X` is a `csr_array` (but it works fine when it's a `csr_matrix`); seems like some low level matrix multiplication subroutine is not implemented for `csr_array`s.

### Minimal code sample

```python
import scanpy as sc
import anndata as ad
import numpy as np
from scipy.sparse import csr_array, csr_matrix

data = np.random.random((2, 5))
dataset = ad.AnnData(
    X=csr_array(data),
    # X=csr_matrix(data),
)
sc.pp.pca(dataset)
```

### Error output

```pytb
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/env/lib/python3.10/site-packages/scanpy/preprocessing/_pca.py", line 285, in pca
    X_pca, pca_ = _pca_with_sparse(
  File "/env/lib/python3.10/site-packages/scanpy/preprocessing/_pca.py", line 453, in _pca_with_sparse
    total_var = _get_mean_var(X)[1].sum()
  File "/env/lib/python3.10/site-packages/scanpy/preprocessing/_utils.py", line 40, in _get_mean_var
    mean_sq = axis_mean(elem_mul(X, X), axis=axis, dtype=np.float64)
  File "/env/lib/python3.10/functools.py", line 889, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/env/lib/python3.10/site-packages/scanpy/_utils/__init__.py", line 546, in elem_mul
    raise NotImplementedError
NotImplementedError
```

### Versions

<details>

```
1.10.4
```

</details>

flying-sheep:
you forgot to fill out the version information. Please do that. It’s likely that this is true, but I still need that info.

alam-shahul:
Hmm... well, I just tried it again on my current installation, and I got a different error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/work/magroup/shahula/bin/miniconda3/envs/popari_distribution/lib/python3.10/site-packages/scanpy/preprocessing/_pca.py", line 200, in pca
[...]

PR #3556, opened 2025-04-03
Switch to CSBase

flying-sheep:
Another step towards #3423, should have done it this way to begin with.

The idea is that there will be fewer merge conflicts if I switch everything to `CSBase` now and then change what CSBase is in the real PR.

TODO:

- [x] replace instances of csr_matrix with CSRBase after merging #3553

after this PR, continue with:

```py
if TYPE_CHECKING:
    from scipy.sparse import csc_array, csr_array
else:
    try:
        from scipy.sparse import csc_array, csr_array
    except ImportError:
        csr_array = type("csr_array", (), {})
        csc_array = type("csc_array", (), {})
        csr_array.__module__ = csc_array.__module__ = "scipy.sparse"
CSArray = csr_array | csc_array
CSMatrix = sparse.csr_matrix | sparse.csc_matrix
CSBase = CSArray | CSMatrix
```


src/scanpy/_utils/__init__.py
```diff
@@ -781,7 +778,7 @@ def _(X: DaskArray, axis: Literal[0, 1]) -> DaskArray:
 
 @overload
 def axis_sum(
-    X: sparse.spmatrix,
+    X: _CSMatrix,
```
flying-sheep, line 781: one of the few cases where really only matrices are meant, will be gone when switching to fast-array-utils (which will never emit np.matrix instances)

src/scanpy/_utils/__init__.py
```diff
@@ -825,8 +822,8 @@ def _(
     def sum_drop_keepdims(*args, **kwargs):
         kwargs.pop("computing_meta", None)
         # masked operations on sparse produce which numpy matrices gives the same API issues handled here
-        if isinstance(X._meta, sparse.spmatrix | np.matrix) or isinstance(
-            args[0], sparse.spmatrix | np.matrix
+        if isinstance(X._meta, _CSMatrix | np.matrix) or isinstance(
+            args[0], _CSMatrix | np.matrix
```
flying-sheep, line 826: this is the other case, will also be gone with the switch to fast-array-utils

src/scanpy/metrics/_common.py
```diff
@@ -19,19 +19,19 @@
     from numpy.typing import NDArray
 
     T_NonSparse = TypeVar("T_NonSparse", bound=NDArray | DaskArray)
-    V = TypeVar("V", bound=NDArray | sparse.csr_matrix)
+    V = TypeVar("V", bound=NDArray | CSRBase)
 
-    _Vals = NDArray | sparse.spmatrix | DaskArray | pd.DataFrame | pd.Series
+    _Vals = NDArray | SpBase | DaskArray | pd.DataFrame | pd.Series
```
flying-sheep, line 24: this and `_resolve_vals` are only places SpBase appears in this file, because we convert to CSR in `_resolve_vals`.

flying-sheep (commented):
This change is big, but mechanical. Most importantly, I added test coverage for everything in the last PR that touched all this.

1. all instantiations stay the same (e.g. `sparse.csr_matrix(...)`)
2. all type annotations or `isinstance` checks for `cs{rc}_matrix` become `CSBase` to accomodate `cs{rc}_array` in the future
3. same for checking/annotating only `csr_matrix`, which becomes `CSRBase`

Since AnnData only supports CSBase, I made all the checks explicitly for CSBase. The exceptions are obvious (paga adjacency stuff, comment in code) or commented below

PR #3570, opened 2025-04-08
perf: more efficient _sparse_nanmean

### LINKED ISSUE #1894: _sparse_nanmean is inefficient

<details>

```python
import numpy as np

@njit(parallel=True)
    N, M = shape
    for i in prange(N):
        start = indptr[i]
        stop = indptr[i+1]
        window = data[start:stop]
        n_nan = np.int64(0)
        i_sum = np.float64(0.)
        for j_val in window:
            if np.isnan(j_val):
                n_nan += 1
            else:
                i_sum += j_val
        sums[i] = i_sum
        nans[i] = n_nan
    sums /= (M - nans)
    return sums
```

</details>

Reovirus:
Fixes [#1894](https://github.com/scverse/scanpy/issues/1894). Reduced redundant data copying; the original matrix is now copied once instead of twice. One copy remains necessary to replace NaNs with zeros without modifying the original matrix.​

- [x] Closes [#1894](https://github.com/scverse/scanpy/issues/1894)
- [x] Tests included 
- [x] Release notes not necessary because


Reovirus:
@flying-sheep Can you assign the reviewer please? Fall test looks like testing bug

Reovirus:
And I can rewrite logic without any coping using numba, but it can be slowly that implemented methods

Reovirus:
I know that I have "No milestone failure" but I haven't permissions to set milestone, probably I need help of maintainers to set it

Zethson:
@Reovirus I restarted the CI as there should hopefully be no flaky tests at the moment.

Don't worry about the milestone, please.

Reovirus:
failure is scipy.sparse.csr_matrix.count_nonzero which has axis argument since scipy 1.15.0. I'll rewrite with numba

Reovirus:
@Zethson cund you please restart CI? I catch very strange bug with http, it's not mine code

Reovirus:
CI pased, can somebody review in some time pls

flying-sheep:
I made the benchmarks run, let’s see how well this works!

Could you please add a release note fragment?

```
hatch run towncrier:create 3570.performance.md
```

Reovirus:
Yes, l'll make a note, thanks!)

flying-sheep:
according to the benchmarks, this is actually slower than before.

the benchmarks are a bit flaky, so it could be wrong, but usually a factor of 2.5 like here is real.

Reovirus:
Understand. Try to change and speed up

Reovirus:
I've replace np.add.at by np.bincount (it was slowest place). 

Now results is on pictures (old is old functions, new is new implementation, new_parallel is new with some njint in one case. Matrix was (5000, 5000) and I use 100 random matrix (~20% nans, ~20% zeros in one matrix) for testing). X axis like {input_format: csc|csr}_{axis: 0|1}_{version: old|new|new_with_numba}, y-axis is _sparse_nanmean spent time in seconds on my ubuntu VM (low-resources)

test_2000_2000_100_boxplot
test_2000_2000_100_violin

UPD. I've seen benchmark is added automatically

Reovirus:
@flying-sheep 
 We have a small win (23.4±2ms | 21.8±0.8ms | 0.93 | tools.time_score_genes, too high p-value to display cause stderr is high, implementation is lessly effective on smal-amount-of-zeros data) in score_genes time and small lose in memory. Should I optimize memory usage? (it can be caused by mask=np.isnan(data), instead of thic I can backgroundly recalc mask in np.nan_to_num and reduce memory on 1 array)))

Reovirus:
What should we do with optimization?)
@flying-sheep

Intron7:
I think these should be parallel numba kernels with mask arguments for major and minor axis. I have a [working implementation](https://github.com/scverse/rapids_singlecell/blob/main/src/rapids_singlecell/tools/_kernels/_nan_mean_kernels.py) in rapids-singlecell that doesnt need to copy at all. Starting from _get_mean_var is I think the best way forward

Intron7 (changes requested):
Refactor kernels to be 0 copy

Reovirus:
@Intron7  I rewrite logics. But my local benchmarking gives another result (I just measure peak memory by ), I'm trying to fix it. And I have no idea about some metrics like preprocessing_counts.FastSuite.time_log1p('pbmc3k', 'counts-off-axis'). It change sign randomly, is it normal?

Intron7:
The kernels already look good. However my main point I was trying to make is that you can create the means without subsetting. That means you wouldnt need to use the get subset function but just use masks within your kernel.

src/scanpy/tools/_score_genes.py
```diff
@@ -28,28 +29,74 @@
     _GetSubset = Callable[[_StrIdx], np.ndarray | CSBase]
 
 
+@njit
+def _get_sparce_nanmean_indptr(
```
Zethson:
```suggestion
def _get_sparse_nanmean_indptr(
```

src/scanpy/tools/_score_genes.py
```diff
@@ -28,28 +29,74 @@
...
+                count -= 1
+        if count == 0:
+            result[i] = np.nan
+        else:
+            result[i] = total / count
+    return result
+
+
+@njit
+def _get_sparce_nanmean_indices(
```
Zethson:
```suggestion
def _get_sparse_nanmean_indices(
```

Zethson (commented):
Just typos I think. These should be renamed.

flying-sheep:
I merged upstream changes. Do you still plan to work in the changes @Intron7 requested?

Zethson:
Kind check in @Reovirus - would you be up for finalizing this PR, please? It'd be very much appreciated.

Reovirus:
Hello)
I'll up testa abd optimisations for bench in 2 days

PR #3572, opened 2025-04-09
Add tests for more coverage

selmanozleyen:
Since I don't have the permission to commit I will do it this way. 

After the discussion here https://github.com/scverse/scanpy/pull/3379#pullrequestreview-2623027800 I seperated the median function so that I can tests the uncovered/untested lines directly.

### DIFF
--- a/src/scanpy/preprocessing/_normalization.py
+++ b/src/scanpy/preprocessing/_normalization.py
@@ -26,13 +26,13 @@ if TYPE_CHECKING:
     from anndata import AnnData
 
 
-def _compute_nnz_median(counts: np.ndarray | DaskArray, axis: int = 0):
-    """Compute median of counts, ignoring zeros."""
+def _compute_nnz_median(counts: np.ndarray | DaskArray) -> np.floating:
+    """Given a 1D array of counts, compute the median of the non-zero counts."""
     if isinstance(counts, DaskArray):
         counts = counts.compute()
     counts_greater_than_zero = counts[counts > 0]
-    after = np.median(counts_greater_than_zero, axis=0)
-    return after
+    median = np.median(counts_greater_than_zero)
+    return median
 
 
 def _normalize_data(X, counts, after=None, *, copy: bool = False):
--- a/src/scanpy/preprocessing/_normalization.py
+++ b/src/scanpy/preprocessing/_normalization.py
@@ -26,15 +26,21 @@ if TYPE_CHECKING:
     from anndata import AnnData
 
 
+def _compute_nnz_median(counts: np.ndarray | DaskArray, axis: int = 0):
+    """Compute median of counts, ignoring zeros."""
+    if isinstance(counts, DaskArray):
+        counts = counts.compute()
+    counts_greater_than_zero = counts[counts > 0]
+    after = np.median(counts_greater_than_zero, axis=0)
+    return after
+
+
 def _normalize_data(X, counts, after=None, *, copy: bool = False):
     X = X.copy() if copy else X
     if issubclass(X.dtype.type, int | np.integer):
         X = X.astype(np.float32)  # TODO: Check if float64 should be used
     if after is None:
-        if isinstance(counts, DaskArray):
-            counts = counts.compute()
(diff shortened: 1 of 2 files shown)


src/scanpy/preprocessing/_normalization.py
```diff
@@ -26,15 +26,21 @@
     from anndata import AnnData
 
 
+def _compute_nnz_median(counts: np.ndarray | DaskArray, axis: int = 0):
```
flying-sheep: The axis parameter should be removed and a return type added

src/scanpy/preprocessing/_normalization.py
```diff
@@ -26,15 +26,21 @@
...
+    """Compute median of counts, ignoring zeros."""
+    if isinstance(counts, DaskArray):
+        counts = counts.compute()
+    counts_greater_than_zero = counts[counts > 0]
+    after = np.median(counts_greater_than_zero, axis=0)
```
flying-sheep: you should ~~actually pass~~ remove the `axis` parameter here

src/scanpy/preprocessing/_normalization.py
```diff
@@ -26,15 +26,21 @@
...
 def _normalize_data(X, counts, after=None, *, copy: bool = False):
     X = X.copy() if copy else X
     if issubclass(X.dtype.type, int | np.integer):
         X = X.astype(np.float32)  # TODO: Check if float64 should be used
     if after is None:
-        if isinstance(counts, DaskArray):
-            counts = counts.compute()
-        counts_greater_than_zero = counts[counts > 0]
-        after = np.median(counts_greater_than_zero, axis=0)
+        after = _compute_nnz_median(counts)
```
flying-sheep, line 43: you should pass `axis=0` here

tests/test_normalization.py
```diff
@@ -323,3 +324,13 @@ def test_normalize_pearson_residuals_recipe(pbmc3k_parametrized_small, n_hvgs, n
     assert adata.varm["PCs"].shape == (n_genes, n_comps)
     # number of all-zero-colums should be number of non-hvgs
     assert sum(np.sum(np.abs(adata.varm["PCs"]), axis=1) == 0) == n_genes - n_hvgs
+
+
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_DENSE)
+@pytest.mark.parametrize("dtype", ["float32", "int64"])
+def test_compute_nnz_median(array_type, dtype):
+    data = np.array([0, 0, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=dtype)
```
flying-sheep, line 332: please give it a shape, and parametrize this test with `axis`

flying-sheep (changes requested):
Please remove the axis parameter if the input is 1D / we always compute across all axes.

tests/test_normalization.py
selmanozleyen: but doesn't this always take 1d array? like the counts per cell. Also if it is 2d should there be at least one non zero for each axis we are computing the median on?

flying-sheep: then the `axis` parameter should be completely dropped (also from `np.median`) and it should always return a float.

selmanozleyen: true, I forgot about removing them

flying-sheep: OK, then please do that!

tests/test_normalization.py
```diff
@@ -323,3 +324,13 @@ def test_normalize_pearson_residuals_recipe(pbmc3k_parametrized_small, n_hvgs, n
...
+    data = array_type(data)
+    computed = _compute_nnz_median(data)
+    expected = 5
+    assert np.allclose(computed, expected), f"Expected {expected}, got {computed}"
```
flying-sheep:
```suggestion
    np.testing.assert_allclose(_compute_nnz_median(data), 5)
```

flying-sheep (approved):

flying-sheep:
thanks!

PR #3577, opened 2025-04-11
Add Pylint lints

flying-sheep:
See https://github.com/scverse/anndata/issues/1241


src/scanpy/plotting/_anndata.py
```diff
@@ -458,10 +456,10 @@ def add_centroid(centroids, name, Y, mask):
         centroids[name] = Y_mask[i]
 
     # loop over all categorical annotation and plot it
-    for ikey, palette in zip(categoricals, palettes, strict=False):
+    for ikey, pal in zip(categoricals, palettes, strict=False):
```
flying-sheep, line 459: PLW2901 ([redefined-loop-name](https://docs.astral.sh/ruff/rules/redefined-loop-name/)) is great to disentangle hairy code!

PR #3590, opened 2025-04-14
Centralize random utils

### LINKED ISSUE #3580: _create_adata in test_score_genes doesen't work properly

_create_random_gene_names have no checks to name unique, as a result we have gene name duplicates and errors like `pandas.errors.InvalidIndexError: Reindexing only valid with uniquely valued Index objects when testing`

1) Fixed random seed is needed to test reproducibility
2) Rewrite function

Maybe we should use set.add and check set length

flying-sheep:
- [x] Closes #3580

- [ ] Release notes not necessary because: dev changes

TODO:

- [x] check if we can get rid of the `subsample` helpers that are only used in one place


src/scanpy/_utils/random.py
```diff
@@ -0,0 +1,202 @@
...
+
+                return wrapper
+
+            setattr(cls, name, mk_wrapper(name, meth))
+
+
+_FakeRandomGen._delegate()
+
+
+# Random k-tuples
```
ilan-gold: A little weird to just have this here with a single `#` stylistically

ilan-gold (approved):

ilan-gold:
slick algorithm!

flying-sheep:
haha thanks!

PR #3597, opened 2025-04-17
Fix `pp.scale` returning a Dask Array with `np.matrix` chunks

### LINKED ISSUE #3594: sc.pp.scale and sc.pp.pca does not work together in dask

### What happened?

I got this error when I tried to use pca after scale when loading the data with anndata dask. I think the pca is calling dask-ml pca but my scaled matrix is sparse, therefore resulting in this error/ 

### Minimal code sample

```python
%%time
with h5py.File(cell_atlas_path, "r") as f:
    adata = ad.AnnData(
        obs=ad.io.read_elem(f["obs"]),
        var=ad.io.read_elem(f["var"]),
    )
    adata.X = ad.experimental.read_elem_as_dask(
        f["X"], chunks=(SPARSE_CHUNK_SIZE, adata.shape[1])
    )

adata.layers["counts"] = adata.X.copy()  # Making sure we keep access to the raw counts
sc.pp.log1p(adata)
sc.pp.highly_variable_genes(adata)
sc.pp.scale(adata)
sc.pp.pca(adata)
```

### Error output

```pytb
/usr/local/lib/python3.11/dist-packages/scanpy/preprocessing/_scale.py:199: UserWarning: zero-center being used with `DaskArray` sparse chunks. This can be bad if you have large chunks or intend to eventually read the whole data into memory.
  warnings.warn(
/usr/local/lib/python3.11/dist-packages/dask/array/_reductions_generic.py in partial_reduce(func, x, split_every, keepdims, dtype, name, reduced_meta)
    321         try:
--> 322             meta = func(reduced_meta, computing_meta=True)
    323         # no meta keyword argument exists for func, and it isn't required

16 frames
TypeError: matrix.sum() got an unexpected keyword argument 'keepdims'

During handling of the above exception, another exception occurred:

/usr/local/lib/python3.11/dist-packages/numpy/matrixlib/defmatrix.py in __getitem__(self, index)
    194 
    195         try:
--> 196             out = N.ndarray.__getitem__(self, index)
    197         finally:
    198             self._getitem = False

IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

### Versions

<details>

pathlib	1.0.1
dask	2025.3.0
scanpy	1.11.1
anndata	0.11.4
h5py	3.13.0

</details>

flying-sheep:
- [x] Closes #3594
- [x] Tests included

### DIFF
--- a/tests/test_preprocessing.py
+++ b/tests/test_preprocessing.py
@@ -298,8 +298,10 @@ def test_sample_copy_backed_error(tmp_path):
 
 
 @pytest.mark.parametrize("array_type", ARRAY_TYPES)
-@pytest.mark.parametrize("zero_center", [True, False])
-@pytest.mark.parametrize("max_value", [None, 1.0])
+@pytest.mark.parametrize(
+    "zero_center", [True, False], ids=["zero_center", "no_zero_center"]
+)
+@pytest.mark.parametrize("max_value", [None, 1.0], ids=["no_clip", "clip"])
 def test_scale_matrix_types(array_type, zero_center, max_value):
     adata = pbmc68k_reduced()
     adata.X = adata.raw.X
@@ -310,6 +312,7 @@ def test_scale_matrix_types(array_type, zero_center, max_value):
         sc.pp.scale(adata_casted, zero_center=zero_center, max_value=max_value)
     X = adata_casted.X
     if "dask" in array_type.__name__:
+        assert not isinstance(X._meta, np.matrix)
         X = X.compute()
     if isinstance(X, CSBase):
         X = X.todense()
--- a/tests/test_scaling.py
+++ b/tests/test_scaling.py
@@ -73,46 +73,36 @@ X_scaled_for_mask_clipped = [
     [np.array, sparse.csr_matrix, sparse.csc_matrix],  # noqa: TID251
     ids=lambda x: x.__name__,
 )
-@pytest.mark.parametrize("dtype", ["float32", "int64"])
+@pytest.mark.parametrize("container", ["anndata", "array"])
+@pytest.mark.parametrize("dtype", [np.float32, np.int64])
+@pytest.mark.parametrize("zero_center", [True, False], ids=["center", "no_center"])
 @pytest.mark.parametrize(
-    ("mask_obs", "X", "X_centered", "X_scaled"),
+    ("mask_obs", "x", "x_centered", "x_scaled"),
     [
-        (None, X_original, X_centered_original, X_scaled_original),
-        (
+        pytest.param(
+            None, X_original, X_centered_original, X_scaled_original, id="no_mask"
+        ),
+        pytest.param(
             np.array((0, 0, 1, 1, 1, 0, 0), dtype=bool),
             X_for_mask,
             X_centered_for_mask,
             X_scaled_for_mask,
+            id="mask",
         ),
     ],
 )
-def test_scale(*, typ, dtype, mask_obs, X, X_centered, X_scaled):
-    # test AnnData arguments
-    # test scaling with default zero_center == True
-    adata0 = AnnData(typ(X).astype(dtype))
-    sc.pp.scale(adata0, mask_obs=mask_obs)
-    assert np.allclose(sparse.csr_matrix(adata0.X).toarray(), X_centered)  # noqa: TID251
-    # test scaling with explicit zero_center == True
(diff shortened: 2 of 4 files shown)


tests/test_scaling.py
flying-sheep, line 1: Changes to this file are purely cosmetic. More parametrization is better to pinpoint issues, but no actual changes to what’s being tested happen in here.

tests/test_preprocessing.py
```diff
@@ -310,6 +312,7 @@ def test_scale_matrix_types(array_type, zero_center, max_value):
         sc.pp.scale(adata_casted, zero_center=zero_center, max_value=max_value)
     X = adata_casted.X
     if "dask" in array_type.__name__:
+        assert not isinstance(X._meta, np.matrix)
```
flying-sheep, line 315: I verified that this catches the issue without my fix

selmanozleyen: so this means it's ok to ignore the coverage warning right?

tests/test_scaling.py
```diff
@@ -135,22 +125,23 @@ def test_clip(zero_center):
...
-def test_scale_sparse(*, mask_obs, X, X_scaled, X_clipped):
-    adata0 = AnnData(sparse.csr_matrix(X).astype(np.float32))  # noqa: TID251
-    sc.pp.scale(adata0, mask_obs=mask_obs, zero_center=False)
-    assert np.allclose(sparse.csr_matrix(adata0.X).toarray(), X_scaled)  # noqa: TID251
-    # test scaling with explicit zero_center == True
-    adata1 = AnnData(sparse.csr_matrix(X).astype(np.float32))  # noqa: TID251
-    sc.pp.scale(adata1, zero_center=False, mask_obs=mask_obs, max_value=1)
-    assert np.allclose(sparse.csr_matrix(adata1.X).toarray(), X_clipped)  # noqa: TID251
+@pytest.mark.parametrize("clip", [False, True], ids=["no_clip", "clip"])
+def test_scale_sparse(*, mask_obs, x, x_scaled, x_clipped, clip):
```
selmanozleyen, line 143: should there be tests for zero_center=True case as well?

selmanozleyen (commented):
The fix itself looks good but I have some questions about `sc.pp.scale` in general. `sc.pp.scale` calls the following function but `CSBase` is not annotated in `scale_array`, it needs to be updated right? Or am I missing something?
https://github.com/scverse/scanpy/blob/184ecc92bd2e6e98e575d92cd4899228d83d6133/src/scanpy/preprocessing/_scale.py#L145

flying-sheep:
> sc.pp.scale calls the following function but CSBase is not annotated in scale_array, it needs to be updated right?

No, `scale_array` will run on numpy ndarrays and dask arrays with a `isinstance(x._meta, np.ndarray | CSBase)`, but `scale_sparse` will run on sparse matrices/arrays

tests/test_preprocessing.py
flying-sheep: At the time of writing, there is no coverage warning, must have been that you reviewed when a test job hadn’t run yet.

What I meant is that I first modified the test, confirmed that it indeed works (fails), then made my change and ran the test again to see the test now passing.

tests/test_scaling.py
flying-sheep: good point! I think `test_scale_matrix_types` already tests that case enough though.

check the references that are compared against here and in `test_scale`

selmanozleyen:
Ah ok I just saw the `scale_anndata` function. I thought
``
@singledispatch
def scale(
    data: AnnData | CSBase | np.ndarray | DaskArray,
``
would take the AnnData because of the annotation but I guess later registers override it.

flying-sheep:
@selmanozleyen yeah, singledispatch is confusing:

the fallback function (the one decorated with `@singledispatch`) actually only ever receives the types for which no registered functions exist, but type checkers for some reason still want it annotated with all the types. So:

```py
@singledispatch
def foo(thing: int | str) -> int:
    assert isinstance(thing, int), "`foo_str` will be called for a string argument"
    return thing

@foo.register(str)
def foo_str(thing: str) -> int:
    return len(thing)
```

IMHO this would make more sense, but mypy will complain:

```py
@singledispatch
def foo(thing: int) -> int:  # Mypy error here
    return thing

@foo.register(str)
def foo_str(thing: str) -> int:
    return len(thing)
```

Issue #2717, opened 2023-10-26
Remove `sc.external`


flying-sheep:
External has outlived its usefulness with the emergence of the scverse ecosystem, and should be removed.

^ This text should be expanded

# TODO

- [ ] Finalize methods to keep/ merge into main namespace
  - Currently: `scanorama`, `hashsolo`, `scrublet` or equivalent, maybe `bbknn`
- [ ] Make removal plan (probably deprecation message + maybe email to authors) for
  - [ ]  `harmony`
  - [ ] `dca`
  - [ ] `magic`
  - [ ] `phate`
  - [ ] `palantir`
  - [ ] `trimap`
  - [ ] `sam`
  - [ ] `phenograph`
  - [ ] `wishbone`
  - [ ] `sandbag`
  - [ ] `cyclone`
  - [ ] Export functionality for
    - [ ] `spring_project`
    - [ ] `cellbrowser`

flying-sheep:
@Intron7 you said you have a sense which ones are still used.

Could you please weigh in here?

flying-sheep:
TODO: create sub-issues for all tools and invite feedback by maintainers and users
- what’s still maintained (and therefore people should use the upstream one)
- what’s unmaintained but still used (and therefore we should take over)

PR #3607, opened 2025-04-22
(chore): derive CI matrix from hatch env

flying-sheep:
See https://github.com/scverse/cookiecutter-scverse/blob/d4bdfc5ec4c5029aebf7c0cba65609e20358144d/%7B%7Bcookiecutter.project_name%7D%7D/.github/workflows/test.yaml

This also updates the hatch env matrix to have the following jobs

- `stable` (formerly `full`): run all pretty much tests except external tools
- `pre` (unchanged): run tests with pre-release versions
- `min-vers` (formerly `min`): run tests with the minimum possible versions
- `min-deps` (new): like in anndata, tests if everything imports OK with no optional dependencies installed

The only regression is that we no longer test `scanorama`, because it depends on `annoy`, which hasn’t been updated in years and fails to build on my system.

We should probably think of a better strategy for it, but I think running the maximum set of tests that we test online also locally is a good idea.

We also need to check if codecov still works, I’m typing this while tests still run.


.github/workflows/ci.yml
```diff
@@ -17,43 +17,49 @@ defaults:
...
-            dependencies-version: minimum
-            test-type: coverage
-
-    env:  # for use codecov’s env_vars tagging
-      PYTHON: ${{ matrix.python-version }}
-      DEPS: ${{ matrix.dependencies-version || 'default' }}
-      TESTS: ${{ matrix.test-type || 'default' }}
-
+        env: ${{ fromJSON(needs.get-environments.outputs.envs) }}
+    env:  # environment variable for use codecov’s env_vars tagging
```
ilan-gold:
```suggestion
    env:  # environment variable for use in codecov’s env_vars tagging
```

hatch.toml
```diff
@@ -15,25 +15,27 @@ scripts.clean = "git restore --source=HEAD --staged --worktree -- docs/release-n
...
 overrides.matrix.deps.extra-dependencies = [
     { if = [ "pre" ], value = "anndata[dev,test] @ git+https://github.com/scverse/anndata.git" },
 ]
+overrides.matrix.deps.features = [
+    { if = [ "stable", "pre", "min-vers" ], value = "test" },
+]
 
 [[envs.hatch-test.matrix]]
-deps = [ "stable", "full", "pre", "min" ]
+deps = [ "stable", "pre", "min-vers", "min-extras" ]
```
ilan-gold: I think this should be changed to `minimal-dependencies` because if I am not mistaken, that's what it's testing for.  Right now it reads like "minimum version of the extras" but that's actually what `min-vers` is for.

pyproject.toml
```diff
@@ -140,14 +135,14 @@ dev = [
...
 paga = [ "igraph" ]
-louvain = [ "igraph", "louvain>=0.6.0,!=0.6.2" ] # Louvain community detection
-leiden = [ "igraph>=0.10", "leidenalg>=0.9.0" ]  # Leiden community detection
-bbknn = [ "bbknn" ]                              # Batch balanced KNN (batch correction)
-magic = [ "magic-impute>=2.0" ]                  # MAGIC imputation method
-skmisc = [ "scikit-misc>=0.1.3" ]                # highly_variable_genes method 'seurat_v3'
-harmony = [ "harmonypy" ]                        # Harmony dataset integration
-scanorama = [ "scanorama" ]                      # Scanorama dataset integration
-scrublet = [ "scikit-image>=0.20" ]              # Doublet detection with automatic thresholds
+louvain = [ "igraph", "louvain>=0.8.2" ]          # Louvain community detection
```
ilan-gold, line 136: Are we allowed to bump minimum versions like this on the minor release?

pyproject.toml
```diff
@@ -95,22 +95,17 @@ test-min = [
...
     "zappy",
+    "zarr<3",
     # additional tested algorithms
+    "scanpy[scrublet]",
     "scanpy[louvain]",
+    "scanpy[leiden]",
     "scanpy[magic]",
     "scanpy[skmisc]",
     "scanpy[harmony]",
-    "scanpy[scanorama]",
```
ilan-gold, line 113: I read the description of this PR:

```
The only regression is that we no longer test scanorama, because it depends on annoy, which hasn’t been updated in years and fails to build on my system.
```

Presumably this is coming from https://github.com/spotify/annoy/pull/680.  Can we put a condition on this for 3.13? Separately, I would be curious how this has been working the past few months at all.  I semi-recently downloaded the package no issue.  On first glance, for example, https://github.com/scverse/scanpy/actions/runs/15063089888/job/42341857082 has annoy.

ilan-gold: I'm not saying I actually know what is going on on your system but it seems like perhaps we could use https://packaging.python.org/en/latest/specifications/dependency-specifiers/#dependency-specifiers as a middle ground between completely removing and leaving something in that breaks your system.

ilan-gold (commented):
Few comments/questions

pyproject.toml
flying-sheep: We’re not, this is the minimum possible version that actually works. Anyone who tried running scanpy with e.g. louvain 0.8.0 would have run into an error. Also bumping patch versions in a patch version is fine, no? Just fixes.

This PR changes the min-deps script so it doesn’t try to do fuzzy matching of the “patch” version. (see below on why I’m using scare quotes). We added a `.*` to the end of the version, i.e. we had an actually functional version of what Ilia Kats tried in the anndata PR. His version didn’t work, because it used `~=` together with adding a `.0` but it didn’t guard against doing that multiple times, therefore e.g. making `thing>=1.5` into `thing~=1.5.0.0.0.0` or so, which just means `thing==1.5` (`~=` in Python is *not* a semver-aware operator like in JS, it just transforms `=~stuff.x` into `>=stuff.(x+1), <(stuff+1)` no matter how many dots `stuff` has)

I was always against this fuzzy matching: it meant
1. that we *can’t* specify the *actual* minimum version for which we wanted (as said, `louvain>=0.8` is a lie, only `>=0.8.2` is functional)
2. we’re adding special magical semantics based on the number of dots to our version boundaries which one would need to know to meaningfully change the boundaries

The only way fuzzy matching *could* work is if we maintained a list of dependencies that we know actually use semver and use it. But if we’d do that in `min-vers.py`, we wouldn’t actually test if our minimum version bounds are actually meaningful, so I don’t like that either.

The perfect solution would be:
1. what I do in this PR
2. we add a dependabot or so config that auto-bumps just the patch versions of known semver dependencies regularly

That’d mean our minimum version bounds are actually meaningful, but we’d still make sure to use the latest patches

ilan-gold: > Anyone who tried running scanpy with e.g. louvain 0.8.0 would have run into an error. Also bumping patch versions in a patch version is fine, no? Just fixes.

In this case, it's a minor version but if this is a no-change due to the non-functioning status of louvain, then we're fine.  So since it was a minor version change (previously `0.6.0`), why was that not working? Was not properly tested because of fuzzy matching? I don't think the two are related, so would be good to undersand here.

> therefore e.g. making thing>=1.5 into thing~=1.5.0.0.0.0 or so, which just means thing==1.5

I noticed this as well once you brought up the potential pitfalls of what he did, thanks for that.

> we add a dependabot

Neutral on this, but would support if we add some sort of auto-merge strategy and/or automatically merge if tests pass.  With JS projects I'm always hesitant to merge given the complexity of literal user-facing graphical applications (i.e., button size changes or something).

pyproject.toml
flying-sheep: Yes, we should fix this before this PR gets merged.

`scanorama→annoy` is the only dependency that requires building a native package on my system, I think, and annoy doesn’t seem to be very actively maintained, that’s why I thought we should evaluate this a little.

hatch.toml
flying-sheep: Hm, I struggled with the naming. I settled on `min-vers` and `min-extras`, to disambiguate what we’re minimizing here: the versions and the active extra features.

I don’t think that `min-extras` next to `min-vers` reads as “minimum *versions* of the extras”, otherwise `vers` wouldn’t be omitted from this one and present in the other one.

pyproject.toml
flying-sheep: > So since it was a minor version change (previously 0.6.0), why was that not working?

because we had no job that both tested minimum versions and also enabled the `louvain` extra. The minimum versions job on `main` uses the `test` extra which (on `main`) doesn’t contain the following extras:

https://github.com/scverse/scanpy/blob/eb8974a6e5161f5e2b6d4b1521176494cffc8a2a/pyproject.toml#L106-L114

pyproject.toml
flying-sheep: Ah, let’s discuss in https://github.com/scverse/scanpy/issues/2717 actually

hatch.toml
flying-sheep: maybe `low-vers` and `few-extras`?

ilan-gold (approved):

PR #3613, opened 2025-04-25
feat: add modularity to scanpy.metrics

### LINKED ISSUE #2908: Get modularity score after community detection in leiden/louvain

This was already discussed in https://github.com/scverse/scanpy/pull/819, is there any reason why it is not possible now?

amalia-k510:
This adds two functions to compute modularity scores from a given graph and a clustering like Leiden or Louvain. The goal is to make it easier to compare different community detection methods using an external metric. To my knowledge, there is no built-in way to compare clustering results nor ways to calculate modularity score. 

Fixes #2908

Functions:

- modularity(): accepts a connectivity matrix and label array, generates an igraph, and returns the modularity score
- modularity_adata(): AnnData wrapper that pulls graph and clustering labels


src/scanpy/metrics/_metrics.py
```diff
@@ -89,3 +95,61 @@ def confusion_matrix(
     df = df.loc[np.array(orig_idx), np.array(new_idx)]
 
     return df
+
+
+def modularity(connectivities, labels, mode="UNDIRECTED") -> float:
```
flying-sheep: This comment also applies to the other function.

please use type annotations and remove the `: free-text type` from the parameters in the docstring.

E.g. instead of `connectivities : array-like or sparse matrix` in the docstring, it should be `connectivities: ArrayLike | CSBase` in the function definition. (search the code for examples if you’re unsure where to import them from).

Also since there are limited options, use `Literal["UNDIRECTED", "DIRECTED"]` instead of `str`. Text for defaults is added automatically, so don‘t write them into the docstring either. But here a `is_directed: bool` parameter would be better anyway, there will never be more than two options.

src/scanpy/metrics/_metrics.py
```diff
@@ -4,11 +4,15 @@
 
 from typing import TYPE_CHECKING
 
+import igraph as ig
```
flying-sheep: This is an optional dependency and therefore can’t be imported on the top-level.

Optimally we’d have multiple implementations so this works without igraph, but it’s OK to just import igraph in the function and have it fail when igraph can’t be imported.

src/scanpy/metrics/_metrics.py
```diff
@@ -4,11 +4,15 @@
...
 import numpy as np
 import pandas as pd
+from anndata import AnnData
 from natsort import natsorted
 from pandas.api.types import CategoricalDtype
 
+from scanpy._compat import CSRBase
```
flying-sheep: please use relative imports (`from .._compat`)

src/scanpy/metrics/_metrics.py
```diff
@@ -89,3 +95,61 @@ def confusion_matrix(
     df = df.loc[np.array(orig_idx), np.array(new_idx)]
 
     return df
+
+
+def modularity(connectivities, labels, mode="UNDIRECTED") -> float:
+    # default mode is undirected?? can be specified as directed or undirected
```
flying-sheep: why is this a question? :smiley:

src/scanpy/metrics/_metrics.py
```diff
@@ -89,3 +95,61 @@ def confusion_matrix(
...
+
+    Returns
+    -------
+    float
+        The modularity of the graph based on the provided clustering.
+    """
+    if isinstance(connectivities, CSRBase):
+        # Convert sparse matrix to dense format so that igraph can handle it
+        # Weighted_Adjacency expects with nested lists or numpy arrays and not sparse matrices
+        dense_connectivities = connectivities.toarray()
```
flying-sheep: The graphs we use are inherently sparse, so we should use APIs that operate on sparse graphs and never densify if possible.

Are you sure there’s no support for initializing igraph from a sparse matrix?

src/scanpy/metrics/_metrics.py
```diff
@@ -89,3 +95,61 @@ def confusion_matrix(
...
+    else:
+        dense_connectivities = connectivities
+    # creating igraph graph from the dense connectivity matrix
+    graph = ig.Graph.Weighted_Adjacency(dense_connectivities.tolist(), mode=mode)
+    if isinstance(labels, pd.Series):
+        labels = labels.values
+    # making sure labels are in the right format, i.e., a list of integers
+    labels = pd.Categorical(np.asarray(labels)).codes
```
flying-sheep: that’s a lot of converting around! Are you sure you can’t just pass `labels` into `pd.Categorical` directly?

src/scanpy/metrics/_metrics.py
```diff
@@ -89,3 +95,61 @@ def confusion_matrix(
...
+    -------
+    float
+        The modularity of the graph based on the provided clustering.
+    """
+    # if user passes leiden or louvain as a string, this will get the actual labels
+    # from adata.obs
+    label_array = adata.obs[labels] if isinstance(labels, str) else labels
+    # extracting the connectivities from adata.obsp["connectivities"]
+    connectivities = adata.obsp[obsp]
+    return modularity(connectivities, label_array)
```
flying-sheep: Hmm, we don’t currently store if the connectivity calculation assumes a directed or undirected graph …

Just a thought.

flying-sheep (changes requested):
Hi! Apart from the issue with `igraph` being an optional dependency, this looks good!

I think we have `get_igraph_from_adjacency` which might be useful, but maybe not.

Could you please add tests? We have many examples on how, best would probably be
1. for the direct variant, manually create very small graphs to run this on so you can be sure the results are correct
2. for the anndata version, use `neighbors` to create the connectivity matrix.

Please add `@needs.igraph` so the test only runs when igraph is installed

if you’re unsure about anything, please search the code for examples or ask me!


If you end up implementing a non-igraph flavor for this, please test using [parametrization](https://docs.pytest.org/en/stable/how-to/parametrize.html), e.g.: `@pytest.mark.parametrize("directed", [True, False], ids=["directed", "undirected"])`

src/scanpy/metrics/_metrics.py
amalia-k510: I went ahead and moved it to where it's being used, and added a try-except block to handle any potential import issues. Please let me know if you have any suggestions for a better approach.

flying-sheep: Great, that works! The try-except for a slightly better message is optional. We do that everywhere but I always thought it’s of dubious value.

tests/test_metrics.py
```diff
@@ -196,3 +199,123 @@ def test_confusion_matrix_api():
     pd.testing.assert_frame_equal(
         expected, sc.metrics.confusion_matrix(data["a"], "b", data)
     )
+
+
+# importing igraph, louvain, leiden if available
+HAS_IGRAPH = importlib.util.find_spec("igraph") is not None
+HAS_LOUVAIN = importlib.util.find_spec("louvain") is not None
+HAS_LEIDEN = importlib.util.find_spec("leidenalg") is not None
```
flying-sheep: please use `@need.igraph` and so on instead of `if CONDITION: pytest.skip(...)`

You’ll find examples about where to import the `needs` object and how to use it all over the tests!

src/scanpy/neighbors/__init__.py
```diff
@@ -248,6 +258,65 @@ def neighbors(  # noqa: PLR0913
     return adata if copy else None
 
 
+# def neighbors_from_distance(
```
flying-sheep: did these changes sneak in from a different PR?

src/scanpy/metrics/_metrics.py
```diff
@@ -89,3 +98,95 @@ def confusion_matrix(
...
+        The modularity of the graph based on the provided clustering.
+    """
+    try:
+        # try to import igraph in case the user wants to calculate modularity
+        # not in the main module to avoid import errors
+        import igraph as ig
+    except ImportError as e:
+        msg = "igraph is require for computing modularity"
+        raise ImportError(msg) from e
+    if isinstance(connectivities, CSRBase | SpBase):
```
flying-sheep:
```suggestion
    if isinstance(connectivities, SpBase):
```

`SpBase` is a supertype of `CSRBase`:

```py
# scipy.sparse:
class spmatrix: ...
class coo_matrix(spmatrix): ...
class csr_matrix(spmatrix): ...
class csc_matrix(spmatrix): ...
class sparray: ...
class coo_array(sparray): ...
class csr_array(sparray): ...
class csc_array(sparray): ...

# our types:
SpBase = spmatrix | sparray  # so any csr_*, csc_*, coo_*, … class
CSRBase = csr_matrix | csr_array  # use this when `.format == "csr"` is important
CSCBase = csc_matrix | csc_array  # use this when `.format == "csc"` is important
CSBase = CSRBase | CSCBase  # almost always use this for type checks or type annotations
```

flying-sheep (changes requested):
Hi, this is shaping up! The tests are looking particularly great!

Please address these previous comments:
1. Please follow the doc style as described here: https://github.com/scverse/scanpy/pull/3613#discussion_r2060063701

  Also I mentioned in there “But here a `is_directed`: bool parameter would be better anyway, there will never be more than two options.”
  
  What do you think?

2. Don’t densify (in `modularity_adata` or anywhere): https://github.com/scverse/scanpy/pull/3613#discussion_r2060073331
3. Why the `np.asarray`? https://github.com/scverse/scanpy/pull/3613#discussion_r2060080099

flying-sheep:
Oh, I just saw that we have `get_igraph_from_adjacency`, can you use that?

src/scanpy/neighbors/__init__.py
amalia-k510: Yep, got rid off it.

src/scanpy/metrics/_metrics.py
amalia-k510: Resolved it

src/scanpy/metrics/_metrics.py
amalia-k510: Updated the `modularity()` function to avoid densifying sparse matrices.
If the input is sparse, it now uses the COO format to extract edges and weights directly, and builds the igraph from those. This keeps things memory-efficient and avoids unnecessary `.toarray()` calls. Dense inputs are still supported as before.

src/scanpy/metrics/_metrics.py
amalia-k510: Thanks for pointing this out. I’ve left the mode argument in for now, defaulting to "UNDIRECTED." It’s not strictly necessary at this point, but I figured it might be worth keeping in case support for directed graphs is added down the line. Let me know if you’d prefer to remove it to keep things simpler.

flying-sheep (changes requested):
Looks good! A few points:

- please add a release note (see the other PR, please adapt the PR number in the command I mention there when running it for this PR)
- please move the correct change to `unique_labels` (`to_numpy`) from the other PR into this one: https://github.com/scverse/scanpy/blob/040b8b715a871df85b6af3fefb0edd3a3fd72e68/src/scanpy/metrics/_metrics.py#L63
- please change `mode` to `is_directed: bool` (no default value) for `modularity`

Once your other PR (the `neighbors` one) is merged, let’s finish this one up:
- we should make `neighbors` store an `is_directed` param in `.uns[key_added or "neighbors"]["params"]`
- then we change the `modularity_adata` parameter `obsp: str` parameter to `key: str = "neighbors"`
- then we use `adata.uns[key]["connectivities_key"]` and `adata.uns[key]["params"]["is_directed"]` in `modularity_adata` to call `modularity`.

flying-sheep:
there should be
- 1 release note file
- named `3613.feature.md` (this PR’s number is 3613)
- without “Add your info here” in the file contents
- with the file contents ending with `` {smaller}`A. Karesh` ``

src/scanpy/metrics/_metrics.py
```diff
@@ -89,3 +100,120 @@ def confusion_matrix(
...
+) -> float:
+    try:
+        import igraph as ig
+    except ImportError as e:  # pragma: no cover
+        msg = "igraph is require for computing modularity"
+        raise ImportError(msg) from e
+    igraph_mode = ig.ADJ_DIRECTED if is_directed else ig.ADJ_UNDIRECTED
+    graph = ig.Graph.Weighted_Adjacency(connectivities, mode=igraph_mode)
+    # cluster labels to integer codes required by igraph
+    labels = pd.Categorical(np.asarray(labels)).codes
```
ilan-gold: Can we exploit the fastpath when `labels` is a `pd.Series` with categorical dtype (i.e., call `.codes` directly on a `pandas` object)?

tests/test_metrics.py
```diff
@@ -241,3 +247,151 @@ def test_confusion_matrix_api() -> None:
...
+    """Test domain and API of modularity score."""
+    adata = pbmc3k()
+    sc.pp.pca(adata)
+    sc.pp.neighbors(adata)
+    sc.tl.leiden(adata, flavor="igraph")
+
+    scores = {}
+    with monkeypatch.context() as m:
+        # make sure we don’t calculate it
+        m.delattr(sc.metrics._metrics, "modularity_array")
```
ilan-gold, line 326: Just to be clear, the next operation would not calculate anyway? This is just a double-triple extra bonus check?

ilan-gold (approved):
Nice!

tests/test_metrics.py
flying-sheep: Yup! Just checking that it actually doesn’t

flying-sheep (approved):

PR #3615, opened 2025-04-29
Catch all exceptions getting version from VCS, fallback to `importlib`

rgoya:
`scanpy` has the same issue reported for `anndata` here https://github.com/scverse/anndata/issues/1960 and fixed here https://github.com/scverse/anndata/pull/1972

Mainly, the `_get_version_from_vcs()` function in `src/scanpy/_version.py` breaks when the user has `hatchling` installed and a `pyproject.toml` file is found, but cannot be parsed by `hatch`/`hatchling`.

This fix is a replicate of https://github.com/scverse/anndata/pull/1972, with `_PROJECT_NAME = "scanpy"`

- [X] "Closes" `scanpy` replicate of https://github.com/scverse/anndata/issues/1960

- [X] Release notes not necessary because:

### DIFF
--- a/src/scanpy/_version.py
+++ b/src/scanpy/_version.py
@@ -5,12 +5,19 @@ See <https://github.com/maresb/hatch-vcs-footgun-example>.
 
 from __future__ import annotations
 
+import warnings
 from pathlib import Path
 
 __all__ = ["__version__"]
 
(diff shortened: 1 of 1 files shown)


flying-sheep:
Thanks!

PR #3627, opened 2025-05-08
feat: add `distances` parameter to `sc.pp.neighbors`

amalia-k510:
Adding a function, `neighbors_from_distance`, that allows users to compute connectivities and distances directly from precomputed distance matrix. Might be useful where distances are generated externally, and we want to plug them into the standard scanpy graph-based pipeline without recomputing neighbors. 

This function should support the following:
- supports both dense and sparse matrices
- zeroes the diagonal to avoid self-loops
- builds connectivities using either the UMAP or Gauss method
- stores results in .obsp and .uns using the same structure as sc.pp.neighbors,
- optionally allows key namespacing via key_added


src/scanpy/neighbors/__init__.py
```diff
@@ -74,6 +76,7 @@ def neighbors(  # noqa: PLR0913
     n_neighbors: int = 15,
     n_pcs: int | None = None,
     *,
+    distance_matrix: np.ndarray | None = None,
```
flying-sheep: Should probably be called `distances` and have the correct type: if people could only pass a dense ndarray here, the option would be mostly useless for real-life data sizes.

src/scanpy/metrics/_metrics.py
```diff
@@ -89,3 +98,95 @@ def confusion_matrix(
     df = df.loc[np.array(orig_idx), np.array(new_idx)]
 
     return df
+
+
+def modularity(
```
flying-sheep: and this ended up here from the wrong branch again :D

src/scanpy/neighbors/__init__.py
```diff
@@ -248,6 +260,102 @@ def neighbors(  # noqa: PLR0913
     return adata if copy else None
 
 
+def neighbors_from_distance(
```
flying-sheep: please add a `*` at the appropriate place, probably after the parameters without a default.

src/scanpy/neighbors/__init__.py
```diff
@@ -248,6 +260,102 @@ def neighbors(  # noqa: PLR0913
...
+    adata: AnnData,
+    distance_matrix: np.ndarray | SpBase,
+    n_neighbors: int = 15,
+    method: Literal["umap", "gauss"] = "umap",  # default to umap
```
flying-sheep: we already have a variable for this exact type imported in this file, please use that instead of `Literal[...]`

src/scanpy/neighbors/__init__.py
```diff
@@ -248,6 +260,102 @@ def neighbors(  # noqa: PLR0913
...
+    Returns
+    -------
+    adata
+        Annotated data with computed distances and connectivities.
+    """
+    if isinstance(distance_matrix, SpBase):
+        # spare matrices can save memory for large datasets
+        # csr_matrix is the most efficient format for sparse matrices
+        # setting the diagonal to 0 is important = distance to self must not affect umap or gauss
+        # elimimate zeros is important to save memory, avoids storing explicit zeros
```
flying-sheep: I don’t think this is true, I’m pretty sure `sparse.csr_matrix(...)` will already call `eliminate_zeros` implicitly. Unless I’m wrong, please remove that and this comment line.

tests/test_metrics.py
flying-sheep, line 1: this also belongs into the other PR, please revert changes in this PR that aren’t `neighbors`-related

src/scanpy/neighbors/__init__.py
```diff
@@ -248,6 +260,102 @@ def neighbors(  # noqa: PLR0913
...
+        )
+    else:
+        msg = f"Method {method} not implemented."
+        raise NotImplementedError(msg)
+    # defining where to store graph info
+    key = "neighbors" if key_added is None else key_added
+    dists_key = "distances" if key_added is None else key_added + "_distances"
+    conns_key = "connectivities" if key_added is None else key_added + "_connectivities"
+    # storing the actual distance and connectivitiy matrices as obsp
+    adata.uns[dists_key] = sparse.csr_matrix(distance_matrix)  # noqa: TID251
```
flying-sheep: `uns`? that doesn’t seem right

flying-sheep (changes requested):
looks good except for the metrics changes being here instead of remaining in their own PR :wink: 

It also looks like some code is copied. Please check if you can introduce a helper function for everything after `if method == "umap":`, looks like most of that is identical to what’s in `def neighbors`. But if it’s too complicated or different, don’t do it!

Please also (after removing the `metrics` tests from this PR) add tests to check that passing `distances` results in the same stuff being added to `adata` as not doing it. Something like

```py
adata_n = ...
adata_d = adata_n.copy()

sc.pp.neighbors(adata_n)
sc.pp.neighbors(adata_d, distances=adata_n.obsp["distances"])

# for all the parts that should be there:
np.testing.assert_almost_equal(...)
…

# for all the parts that can’t be derived from distances
assert "somekey" in adata_n.…
assert "somekey" not in adata_d.…
```

src/scanpy/neighbors/__init__.py
amalia-k510: I’ve added SpBase as a possible input for distance_matrix. Please let me know if that’s correct.

flying-sheep: sounds good!

src/scanpy/metrics/_metrics.py
```diff
@@ -60,7 +60,9 @@ def confusion_matrix(
     orig, new = pd.Series(orig), pd.Series(new)
     assert len(orig) == len(new)
 
-    unique_labels = pd.unique(np.concatenate((orig.values, new.values)))
+    unique_labels = pd.unique(
+        np.concatenate((np.asarray(orig.values), np.asarray(new.values)))
+    )
```
flying-sheep: `np.asarray(thing.values)` is better written as `thing.to_numpy()`

src/scanpy/neighbors/__init__.py
amalia-k510: I looked through the SciPy code base right now, specifically `__init__`, `_compressed`, `_csr`, and `csr.py`, and from what I can tell, eliminate_zeros() doesn’t seem to be called implicitly. Is there a case I might be missing where it’s triggered automatically?

src/scanpy/neighbors/__init__.py
amalia-k510: switched to `.obsp`

src/scanpy/neighbors/__init__.py
flying-sheep: yeah, you’re right:

```pycon
>>> sp.coo_matrix(([0, 1], ([0, 0], [1, 0])), (2, 2)).data
array([0, 1])
>>> sp.coo_matrix(([0, 1], ([0, 0], [1, 0])), (2, 2)).tocsr().data
array([1, 0])
```

I wouldn’t say it’s important to call `elimimate_zeros` to save memory though, usually there won’t be many (I think the maximum I’ve seen is like 1 per row or so)

flying-sheep (approved):
looks good! please add a release note by running `hatch run towncrier:create 3627.feature.md` and editing that (don’t forget to add your name as in the other release note fragments)

src/scanpy/neighbors/__init__.py
```diff
@@ -203,51 +229,124 @@ def neighbors(  # noqa: PLR0913
...
 
 
+def neighbors_from_distance(
+    adata: AnnData,
+    distances: np.ndarray | SpBase,
+    *,
+    n_neighbors: int = 15,
+    metric: _Metric | None = None,
+    method: _Method = "umap",  # default to umap
+    key_added: str | None = None,
```
flying-sheep: please remove the defaults here, and fix the bug that gets uncovered by this action.

tests/test_neighbors.py
```diff
@@ -241,3 +242,26 @@ def test_restore_n_neighbors(neigh, conv):
         ad.uns["neighbors"] = dict(connectivities=conv(neigh.connectivities))
     neigh_restored = Neighbors(ad)
     assert neigh_restored.n_neighbors == 1
+
+
+def test_neighbors_distance_equivalence():
```
flying-sheep: please parametrize this test over `n_neighbors`, `method`, and `key_added` (2 options each) to make sure they all get passed down correctly

flying-sheep (changes requested):

flying-sheep (approved):

PR #3647, opened 2025-05-27
Backport PR #3607 on branch 1.11.x ((chore): derive CI matrix from hatch env)

flying-sheep:
Backports #3607

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,43 +17,49 @@ defaults:
     shell: bash -e {0} # -e to fail on error
 
 jobs:
-  pytest:
+  get-environments:
     runs-on: ubuntu-latest
+    outputs:
+      envs: ${{ steps.get-envs.outputs.envs }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          filter: blob:none
+          fetch-depth: 0
+      - uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: false
+      - id: get-envs
+        run: |
+          ENVS_JSON=$(NO_COLOR=1 uvx hatch env show --json | jq -c 'to_entries
+            | map(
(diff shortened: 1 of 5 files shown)


flying-sheep:
@ilan-gold what do you think: louvain 0.7.1 makes things crash, but 0.7.2 requires igraph≥0.9.5.

should I bump things until everything works together or disable louvain for the low-vers job in this backport?

ilan-gold:
> @ilan-gold what do you think: louvain 0.7.1 makes things crash, but 0.7.2 requires igraph≥0.9.5.

Sorry, isn't the igraph minimum version higher than 0.9.5? Why is this problematic?

flying-sheep:
ah my mistake, I read uv’s error wrong. What it really says is

> `louvain==0.7.2` depends on `igraph>=0.9.5,<0.10` and `scanpy[test]==1.11.2.dev29+g607bfb8f8.d20250527` depends on `igraph==0.10.8`

so louvain 0.7.2 introduced[^1] an upper bound that’s incompatible, which means that `louvain~=0.7` isn’t compatible with e.g. the minimum leiden version we support and needs to be bumped.

[^1]: in actuality that bound was of course there earlier, that’s why things crash. 0.7.2 just made the bound explicit.

---

I guess it would be valid to say “scanpy 0.11.x works with (louvain==0.7.2 and igraph==0.9.5) *or* (leiden==0.9.0 and igraph==0.10.8)”, but that’s getting too far into the weeds.

Let’s just define one baseline for low versions and that’s it.

ilan-gold:
The fix seems reasonable @flying-sheep.  Another option would be to introduce a temporary fix into the `min-vers.py` file but I think the number of people using _only_ louvain without leiden (or something else requiring a higher leiden bound) is probably vanishingly low.

PR #3653, opened 2025-05-28
feat: presets

### LINKED ISSUE #3730: Remove default from sc.pp.scale’s zero_center (so it never accidentally densifies)

See https://github.com/scverse/scanpy/issues/3719#issuecomment-3096454906

### LINKED ISSUE #4000: Scanpy 2.0 preview
- we should change our datasets and read_10x_h5 and so on to put IDs into var_names.
have .var['symbols'] as a convention for plotting functions so they only display var_names if no symbols exist.

https://www.biorxiv.org/content/10.1101/2024.04.04.588111v2

flying-sheep:
- [x] Closes #4000, closes #3730

Self-documenting presets work!

E.g.:
- code: https://github.com/scverse/scanpy/blob/3bae7ae591f954a5401e057be6cd754d71203f91/src/scanpy/_settings.py#L90-L96
- docs: https://icb-scanpy--3653.com.readthedocs.build/en/3653/generated/scanpy.Preset.highly_variable_genes.html


flying-sheep:
TODO: add `jaccard` to seurat preset https://github.com/scverse/scanpy/issues/3828

src/scanpy/_settings/presets.py
```diff
@@ -0,0 +1,200 @@
...
+            Preset.ScanpyV2Preview: PcaPreset(key_added="pca"),
+        }
+
+    @preset_property
+    def rank_genes_groups() -> Mapping[Preset, RankGenesGroupsPreset]:
+        """Correlation method for :func:`~scanpy.tl.rank_genes_groups`."""
+        return {
+            Preset.ScanpyV1: RankGenesGroupsPreset(method="t-test", mask_var=None),
+            Preset.ScanpyV2Preview: RankGenesGroupsPreset(
+                method="wilcoxon", mask_var=None
```
flying-sheep, line 172: `mask_var="highly_variable"` here?

flying-sheep:
@ilan-gold I asked your review because of scope:
- I’ll add some tests, but how thoroughly should it be tested? certainly not all 2.0 preview things
- I filed #4007 for a AdRef based API, which would be relevant here too
- I filed #4008 to maybe do more opinionated changes to after implementing #4007
- anything that should be done now instead of later?

src/scanpy/_settings/presets.py
ilan-gold: Maybe collect these sorts of questions and make a zulip/slack poll?

src/scanpy/preprocessing/_scale.py
```diff
@@ -70,7 +71,7 @@ def clip_array(
 def scale[A: _Array](
     data: AnnData | A,
     *,
-    zero_center: bool = True,
+    zero_center: bool | None = None,
```
ilan-gold: Just thinking out load - does it make sense to have a sentinel value for "this is going to be a new preset" instead of using `None`?

ilan-gold: In theory, this might also allow us to wrap these "replacement" ops in a decorator that just checks the signature for this value.  Could save a grand total of two lines of code as we progressively add more of these.

ilan-gold (commented):
This is a pretty harmless PR so nothing comes to mind as absolutely pressing.  The architecture overall makes sense.  

I would beef up the docs a bit to make it clear how to use this.

Why use `ScanpyV2Preview` instead of just `ScanpyV2`?

flying-sheep:
> Why use `ScanpyV2Preview` instead of just `ScanpyV2`?

Because when we release a new feature release before 2.0, I don’t want people to start relying on it keeping its behavior. I thought “preview” makes it clear that it’s not stable, but I’m open to suggestions.

src/scanpy/preprocessing/_scale.py
flying-sheep: I was thinking about that. We could even make its repr (e.g. in the default docs) display the current preset’s repr or so

src/scanpy/_settings/presets.py
flying-sheep: Definitely. The issue is that they’re hard to collect, I keep stumbling upon more.

src/scanpy/preprocessing/_scale.py
ilan-gold: That would be helpful for communicating as well which things are going to change in scanpy.  So yeah, I think this is definitely worth doing.

flying-sheep: OK, done! I
- added `| Default` everywhere and made the `repr` reflect the scanpy v1 default 
- removed the `| None` from preset-controlled parameter types when `None` means “the default” (without breaking backwards compat of course)

only issue is that we now have `_empty`, `None` AND `Default`. Any ideas?

---

looks like this

grafik

could alternatively reference the string repr (`True (preset='scanpy-v1')` or so)

and idk how to get rid of the `| Default` …

src/scanpy/_settings/presets.py
```diff
@@ -0,0 +1,216 @@
...
+    func: str
+    param: str
+
+    def __repr__(self) -> str:
+        import scanpy as sc
+
+        qualname = enum.Enum.__str__(sc.settings.preset)
+        params = getattr(sc.settings.preset, self.func)
+        value = getattr(params, self.param)
+        return f"{value!r} ({qualname})"
```
ilan-gold: I think the raw `qualname` is a little rough.  Maybe we could have something more human-interpretable? I guess the issue is that this is a general `repr` so maybe you don't want something like `'seurat' (this default has been marked for change to xyz in v2)`

src/scanpy/preprocessing/_scale.py
ilan-gold: Never have been totally clear what `_empty` was for but `Default` is definitely going to disappear in v2 and so will `_empty` hopefully?

ilan-gold (commented):
Modulo the repr (whcih I do think is worth discussion at the minmum if not a tweak), looks great!!!

src/scanpy/preprocessing/_scale.py
flying-sheep: How will we get rid of `_empty`? What’s your idea?

ilan-gold: > Never have been totally clear what _empty was for

Hence the question, was just opining that the ability to create a new set of defaults may make it obsolete, but that I don't know it's history

src/scanpy/_settings/presets.py
flying-sheep: This repr appears only in the docs and when users inspect a function at runtime. So it can be long-ish but not too long!

src/scanpy/preprocessing/_scale.py
flying-sheep: `_empty` is for “`None` means something other than ‘use the default’, so we need a sentinel that means that”

flying-sheep: I think I’ll just repurpose `Default` so it supports non-preset defaults.

E.g. for `read(cache_compression=…)`, `None` means “no compression” and `_empty` means `sc.settings.cache_compression`.

src/scanpy/plotting/_tools/scatterplots.py
```diff
@@ -961,8 +960,8 @@ def spatial(  # noqa: PLR0913
     *,
     basis: str = "spatial",
     img: np.ndarray | None = None,
-    img_key: str | None | Empty = _empty,
-    library_id: str | None | Empty = _empty,
+    img_key: str | None | Default = Default("'hires' | 'lowres'"),
+    library_id: str | None | Default = Default("uns['spatial'][key] if only one key"),
```
ilan-gold, line 964: This is awesome!

ilan-gold (approved):
Yes!

PR #3672, opened 2025-06-06
Settings docs overhaul

flying-sheep:
Preparation PR for presets

Compare:

Old | New
--- | ---
[Settings](https://scanpy.readthedocs.io/en/1.11.x/api/settings.html) | [Settings](https://icb-scanpy--3653.com.readthedocs.build/en/3653/api/settings.html?readthedocs-diff=true) (click here to see a diff)
[`scanpy._settings.ScanpyConfig`](https://scanpy.readthedocs.io/en/1.11.x/generated/scanpy._settings.ScanpyConfig.html) | [`scanpy.settings`](https://icb-scanpy--3653.com.readthedocs.build/en/3653/generated/scanpy.settings.html)
`set_figure_params` on both [`sc`](https://scanpy.readthedocs.io/en/1.11.x/generated/scanpy.set_figure_params.html) and [`sc.settings`](https://scanpy.readthedocs.io/en/1.11.x/generated/scanpy._settings.ScanpyConfig.set_figure_params.html) | `sc.settings.set_figure_params` deprecated in favor of [`sc.set_figure_params`](https://icb-scanpy--3653.com.readthedocs.build/en/3653/generated/scanpy.set_figure_params.html)
[`scanpy._settings.ScanpyConfig.verbosity`](https://scanpy.readthedocs.io/en/1.11.x/generated/scanpy._settings.ScanpyConfig.verbosity.html) | [`scanpy.settings.verbosity`](https://icb-scanpy--3653.com.readthedocs.build/en/3653/generated/scanpy.settings.verbosity.html) and [`scanpy.Verbosity`](https://icb-scanpy--3653.com.readthedocs.build/en/3653/generated/scanpy.Verbosity.html)
[`scanpy.logging.print_versions`](https://scanpy.readthedocs.io/en/1.11.x/generated/scanpy.logging.print_versions.html) deprecated | [`scanpy.logging.print_versions`](https://icb-scanpy--3653.com.readthedocs.build/en/3653/generated/scanpy.logging.print_versions.html) deprecated and moved to “Deprecated Functions”

### TODO
- [x] make `inherited-members` skipping work
- [ ] extract into own PR


docs/extensions/autosummary_skip_inherited.py
```diff
@@ -0,0 +1,55 @@
...
+    """Skip inherited members."""
+    # Skip `getdoc` property
+    if what == "method" and name == "getdoc":
+        return True
+
+    # find parent class
+    for frame, _ in walk_stack(None):
+        if frame.f_code.co_name == "_get_members" and frame.f_code.co_filename.endswith(
+            "/generate.py"
+        ):
```
ilan-gold, line 36: What condition is this checking? Not sure I understand the context of `generate.py` or `_get_members` here

src/scanpy/_singleton.py
```diff
@@ -0,0 +1,67 @@
...
+        if frame.name in {
+            "eval_config_file",  # Sphinx import
+            "generate_autosummary_docs",  # Autosummary generator
+            # "parse_generated_content",  # Autodoc parser
+            "get_object_members",  # Class level of autodoc
+            "import_object",  # Attr level of autodoc
+        }:
+            found = True
+        if frame.filename.endswith("/logging.py"):
+            logging = True
```
ilan-gold: What is the context here for finding this in the stack but also being in sphinx?

src/scanpy/_singleton.py
```diff
@@ -0,0 +1,67 @@
...
+
+class SingletonMeta(type):
+    def __new__(mcls, cls_name: str, *args, **kwargs):
+        cls = super().__new__(mcls, cls_name, *args, **kwargs)
+
+        # We do something differently when we are imported by autosummary.
+        if documenting():
+            props = {}
+            for name in dir(cls):
+                if name.startswith("_") or name in cls.__dict__:
```
ilan-gold: Why isn't the second half of this check looking for some sort of deprecated marker? I only see `cls.set_figure_params` as the only member of `cls.__dict__` that doesn't begin with `_` but also that would make sense to be filtered out.  However, I wouldn't say that it should be filtered out because it is in `__dict__` but rather because it is deprecated.

docs/extensions/autosummary_skip_inherited.py
flying-sheep: added a comment

src/scanpy/_singleton.py
flying-sheep: Hmm, the reason given in the docstring used to be true, but doesn’t seem to be anymore:
> (but not when e.g. a log message is accessing `settings.verbosity`)

So let’s try without it.

src/scanpy/_singleton.py
flying-sheep: yeah, let’s do the skipping somewhere else.

docs/extensions/autosummary_skip_inherited.py
```diff
@@ -0,0 +1,57 @@
...
+        return None
+
+    # return if it’s a member of the parent class
+    typ = parent
+    while typ is not type:
+        if name in typ.__dict__:
+            return None
+        typ = type(typ)
+
+    # skip since we know it’s not a member of the parent class
```
ilan-gold:
```suggestion
    # Skip this `name` because the `parent` `type` lacks the `name`, which indicates `name` does not belong to `parent`
```

docs/extensions/autosummary_skip_inherited.py
```diff
@@ -0,0 +1,57 @@
...
+            break
+    else:
+        return None
+
+    # return if it’s a member of the parent class
+    typ = parent
+    while typ is not type:
+        if name in typ.__dict__:
+            return None
+        typ = type(typ)
```
ilan-gold, line 51: Is the `while` loop really necessary? I would think if `parent` is something like
```python
class B:
    pass
parent = B
```
then we can just do a simply `if name in parent.__dict__` check.

So this might look something like

```python
if type(parent) is type and if name not in parent.__dict__:
    return True
return None
```
i.e., the `parent`'s type is `type` (so presumably it has members) and `name` is not a member.  Otherwise fall back.

Perhaps my understanding is flawed

flying-sheep: yes it is, because of metaclasses. it was like you said, but didn’t work for the metaclass properties.

ilan-gold: I had considered that, but didn't want to look foolish suggesting it and then being wrong.  Oh well.

```suggestion
        # Metaclasses need the while loop to trace back to their parents.
        # See `SingletonMeta` and the documentation of `settings` for an example.
        typ = type(typ)
```

ilan-gold (approved):
Just a suggestion

docs/extensions/autosummary_skip_inherited.py
flying-sheep: yeah, the only way `type(thing) is type` doesn’t apply is with metaclasses.

I guess I could have unrolled this to only allow one level of metaclass, but why not like this?

ilan-gold: I'm fine with it the way it is.   Just wanted to be clear about what was going on.

PR #3690, opened 2025-06-22
Include `scipy-stubs` in the optional dev dependencies

jorenham:
This adds [`scipy-stubs`](https://github.com/scipy/scipy-stubs) as an optional development dependency. It can help quite a bit with IDE supports, such as autocompletion and introspection. It requires no configuration, mypy plugins, and there's no runtime impact.  So even if you don't care about typing annotations, you'll basically improve DX for free :)


jorenham:
CI failures seem unrelated

flying-sheep:
thanks, good point!

PR #3700, opened 2025-07-03
(feat): `sc.get.aggregate` via `dask`

### LINKED ISSUE #3659: `sc.get.aggregate` with dask

New analysis tool: A simple analysis tool you have been using and are missing in `sc.tools`?

See #2578 for context.  A use-case here would be pseudobulking, but there are others.

ilan-gold:
See https://github.com/scverse/scanpy/issues/3723 for a follow-up where we would go algorithm-by-algorithm and implement column-wise operations as well

- [x] Closes #3659

### DIFF
--- a/docs/release-notes/3700.feature.md
+++ b/docs/release-notes/3700.feature.md
@@ -1,1 +1,1 @@
-Make {func}`scanpy.get.aggregate` `dask` compatible over all aggregations except median. {smaller}`I Gold`
+Make {func}`scanpy.get.aggregate` `dask` compatible with all aggregations except median. {smaller}`I Gold`
--- a/docs/release-notes/3700.feature.md
+++ b/docs/release-notes/3700.feature.md
@@ -1,1 +1,1 @@
-Make {func}`scanpy.aggregate.get` {mod}`dask` compatible over all aggregations except median. {smaller}`I Gold`
+Make {func}`scanpy.get.aggregate` `dask` compatible over all aggregations except median. {smaller}`I Gold`
--- a/docs/release-notes/3700.feature.md
+++ b/docs/release-notes/3700.feature.md
@@ -1,0 +1,1 @@
+Make {func}`scanpy.aggregate.get` {mod}`dask` compatible over all aggregations except median. {smaller}`I Gold`
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -393,7 +393,8 @@ def aggregate_dask(
         aggregated["var"] = aggredated_mean_var["var"]
         if has_mean:
             aggregated["mean"] = aggredated_mean_var["mean"]
-    # division must come after, not before, the summation for numerical precision.
+    # division must come after, not before, the summation for numerical precision
+    # i.e., we can't just call map blocks over the mean function.
     elif has_mean:
         group_counts = np.bincount(by.codes)
         aggregated["mean"] = (
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -396,6 +396,9 @@ def aggregate_dask(
         raise NotImplementedError(msg)
     has_mean, has_var = (v in funcs for v in ["mean", "var"])
     funcs_no_var_or_mean = funcs - {"var", "mean"}
+    # aggregate each row chunk individually,
+    # producing a #chunks × #categories × #features array,
+    # then aggregate the per-chunk results.
     aggregated = {
         f: data.map_blocks(
             partial(aggregate_chunk_sum_or_count_nonzero, func=func),
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from functools import partial, singledispatch
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Literal, TypedDict
 
 import numpy as np
 import pandas as pd
@@ -331,25 +331,30 @@ def _aggregate(
     *,
     mask: NDArray[np.bool_] | None = None,
     dof: int = 1,
-):
+) -> dict[AggType, np.ndarray]:
     msg = f"Data type {type(data)} not supported for aggregation"
     raise NotImplementedError(msg)
 
 
-def aggregate_dask_var(
+class MeanVarDict(TypedDict):
+    mean: DaskArray
+    var: DaskArray
+
+
+def aggregate_dask_mean_var(
     data: DaskArray,
     by: pd.Categorical,
     *,
     mask: NDArray[np.bool_] | None = None,
     dof: int = 1,
-):
+) -> MeanVarDict:
     mean = aggregate_dask(data, by, "mean", mask=mask, dof=dof)["mean"]
     sq_mean = aggregate_dask(fau_power(data, 2), by, "mean", mask=mask, dof=dof)["mean"]
     var = sq_mean - (mean**2)
     if dof != 0:
         group_counts = np.bincount(by.codes)
         var *= (group_counts / (group_counts - dof))[:, np.newaxis]
-    return var
+    return {"var": var, "mean": mean}
 
 
 @_aggregate.register(DaskArray)
@@ -360,9 +365,8 @@ def aggregate_dask(
     *,
     mask: NDArray[np.bool_] | None = None,
     dof: int = 1,
-):
+) -> dict[AggType, np.ndarray]:
     def aggregate_chunk_no_var(chunk: Array, block_info=None, *, func: AggType = func):
-        func = "sum" if func == "mean" else func
         subset = slice(*block_info[0]["array-location"][0])
         by_subsetted = by[subset]
         mask_subsetted = mask[subset] if mask is not None else mask
@@ -373,8 +377,8 @@ def aggregate_dask(
     if "median" in funcs:
         msg = "Dask median calculation not supported.  If you want a median-of-medians calculation, please open an issue."
         raise NotImplementedError(msg)
-    has_var = "var" in funcs
-    funcs_no_var = funcs - {"var"}
+    has_mean, has_var = (v in funcs for v in ["mean", "var"])
+    funcs_no_var_or_mean = funcs - {"var", "mean"}
     aggregated = {
         f: data.map_blocks(
             partial(aggregate_chunk_no_var, func=func),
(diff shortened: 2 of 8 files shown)


tests/test_aggregated.py
```diff
@@ -8,11 +8,13 @@
...
 
 import scanpy as sc
+from scanpy._compat import DaskArray
 from scanpy._utils import _resolve_axis, get_literal_vals
 from scanpy.get._aggregated import AggType
 from testing.scanpy._helpers import assert_equal
 from testing.scanpy._helpers.data import pbmc3k_processed
-from testing.scanpy._pytest.params import ARRAY_TYPES_MEM
+
+from .test_pca import ARRAY_TYPES
```
ilan-gold: This 100% warrants a discussion/refactor on its own.   We probably shouldn't even be testing the dask types that aren't the ones in this variable since they are effectively incorrectly chunked (i.e., not row-wise), or at least I'd say the behavior is undefined.

ilan-gold:
@Intron7 It won't let me request your review, but you're welcome to have a look

ilan-gold:
TODOs:

- [ ] CSC aggregation, i.e., aggregate by groups of features and then concatenate
- [ ] Similar to the above, try out adding (or reusing) some sort of `chunks` parameter that allows doing the aggregation on a subset of the requested groupby-categories and then concatenate at the end to potentially memory.

flying-sheep:
I added @Intron7 to https://github.com/orgs/scverse/teams/scanpy, now you can request his review (I also did so)

src/scanpy/get/_aggregated.py
```diff
@@ -330,13 +331,98 @@ def _aggregate(
...
+    mean = aggregate_dask(data, by, "mean", mask=mask, dof=dof)["mean"]
+    sq_mean = aggregate_dask(fau_power(data, 2), by, "mean", mask=mask, dof=dof)["mean"]
+    # TODO: If we don't compute here, the results are not deterministic under the process cluster for sparse.
+    if isinstance(data._meta, CSRBase):
+        sq_mean = sq_mean.compute()
+    var = sq_mean - fau_power(mean, 2)
+    if dof != 0:
+        group_counts = np.bincount(by.codes)
+        var *= (group_counts / (group_counts - dof))[:, np.newaxis]
+    return {"var": var, "mean": mean}
```
flying-sheep:
```suggestion
    return MeanVarDict(mean=mean, var=var)
```

src/scanpy/get/_aggregated.py
```diff
@@ -330,13 +331,98 @@ def _aggregate(
...
+    data: DaskArray,
+    by: pd.Categorical,
+    *,
+    mask: NDArray[np.bool_] | None = None,
+    dof: int = 1,
+) -> MeanVarDict:
+    mean = aggregate_dask(data, by, "mean", mask=mask, dof=dof)["mean"]
+    sq_mean = aggregate_dask(fau_power(data, 2), by, "mean", mask=mask, dof=dof)["mean"]
+    # TODO: If we don't compute here, the results are not deterministic under the process cluster for sparse.
+    if isinstance(data._meta, CSRBase):
```
flying-sheep, line 355: is this code CSR-specific or did you mean `CSBase`?

docs/release-notes/3700.feature.md
```diff
@@ -0,0 +1 @@
+Make {func}`scanpy.get.aggregate` `dask` compatible over all aggregations except median. {smaller}`I Gold`
```
flying-sheep: “compatible over”?

src/scanpy/get/_aggregated.py
```diff
@@ -330,13 +331,98 @@ def _aggregate(
...
+        msg = f"Got {type(data._meta)} meta in DaskArray but only csr_matrix/csr_array and ndarray are supported."
+        raise ValueError(msg)
+    if data.chunksize[1] != data.shape[1]:
+        msg = "Feature axis must be unchunked"
+        raise ValueError(msg)
+
+    def aggregate_chunk_sum_or_count_nonzero(
+        chunk: Array, *, func: Literal["count_nonzero", "sum"], block_info=None
+    ):
+        subset = slice(*block_info[0]["array-location"][0])
```
flying-sheep, line 388: what does a “`block_info`” look like? why do you extract the first element twice here, are these 1-element lists or … ?

src/scanpy/get/_aggregated.py
```diff
@@ -330,13 +331,98 @@ def _aggregate(
...
+    aggregated = {
+        f: data.map_blocks(
+            partial(aggregate_chunk_sum_or_count_nonzero, func=func),
+            new_axis=(1,),
+            chunks=((1,) * data.blocks.size, (len(by.categories),), (data.shape[1],)),
+            meta=np.array(
+                [],
+                dtype=np.float64
+                if func not in ["count_nonzero", "sum"]
+                else data.dtype,
```
flying-sheep: Can this info be centralized? this looks fragile because it’s an inline check based on the currently available functions.

tests/test_aggregated.py
```diff
@@ -387,15 +444,24 @@ def test_combine_categories(label_cols, cols, expected):
     pd.testing.assert_frame_equal(reconstructed_df, result_label_df)
 
 
-@pytest.mark.parametrize("array_type", ARRAY_TYPES_MEM)
+def xfail_dask_median(adata, metric):
+    if isinstance(adata.X, DaskArray) and metric == "median":
+        pytest.xfail("Median calculation not implemented for Dask")
```
flying-sheep: You’re not calling this function from a spot where you’ve run into a test failure, so it should *mark* tests as xfail instead (`request.applymarker` or so) and then let the test actually fail.

`pytest.xfail` is a very niche API that should basically mostly be called from plugins or `except` blocks.

also better move that function all the way up, you’re using it above and below.

tests/test_aggregated.py
```diff
@@ -194,6 +213,44 @@ def test_aggregate_incorrect_dim():
...
+
+@needs.dask
+@pytest.mark.anndata_dask_support
+@pytest.mark.parametrize(
+    ("func", "error_msg"),
+    [
+        pytest.param(to_csc, "only csr_matrix", id="csc"),
+        pytest.param(
+            to_bad_chunking, "Feature axis must be unchunked", id="bad_chunking"
+        ),
```
flying-sheep: We have a convention here that strings used as regexes are always raw strings.

```suggestion
        pytest.param(to_csc, r"only csr_matrix", id="csc"),
        pytest.param(
            to_bad_chunking, r"Feature axis must be unchunked", id="bad_chunking"
        ),
```

tests/test_qc_metrics.py
```diff
@@ -78,14 +78,27 @@ def test_segments_binary():
     "array_type", [*ARRAY_TYPES, pytest.param(sparse.coo_matrix, id="scipy_coo")]
 )
 def test_top_segments(request: pytest.FixtureRequest, array_type):
-    if "dask" in array_type.__name__:
-        reason = "DaskArray not yet supported"
+    if "dask" in array_type.__name__ and "1d_chunked" not in array_type.__name__:
```
flying-sheep, line 81: Not a fan of all the “stringly typed” things going on here.

maybe we should go f-a-u lite here and introduce a class with these things as flags for `array_type`.

I’d say that that’s out of scope for this PR, but you’re changing so many spots here that maybe we should extract and commit that bit before this PR.

flying-sheep (changes requested):
Awesome! I’m not 100% sure how it works though.

Could you explain how it works in respect to using the categories in the chunking? Are you making heterogeneous chunks to collect each subset according to a category?

src/scanpy/get/_aggregated.py
ilan-gold: Well, it's untested with anything besides `CSR` and we check for `CSR` so I would say it is as far as I know

docs/release-notes/3700.feature.md
ilan-gold: "compatible with?"

src/scanpy/get/_aggregated.py
ilan-gold: `dask` doesn't provide type info to my knowledge, but does have this in their [`map_blocks` docs](https://docs.dask.org/en/stable/generated/dask.array.map_blocks.html):

```
block_info  
{0: {'shape': (1000,),
     'num-chunks': (10,),
     'chunk-location': (4,),
     'array-location': [(400, 500)]},
 None: {'shape': (1000,),
        'num-chunks': (10,),
        'chunk-location': (4,),
        'array-location': [(400, 500)],
        'chunk-shape': (100,),
        'dtype': dtype('float64')}}
```

so 0 is the input array (and `None` the output) and we are interested in the location, specifically the first-axis index to subset `by` and `mask`.

tests/test_qc_metrics.py
ilan-gold: > I’d say that that’s out of scope for this PR, but you’re changing so many spots here that maybe we should extract and commit that bit before this PR.

I think it's FAU or bust, and yes, out-of-scope

src/scanpy/get/_aggregated.py
flying-sheep: I’d default to `CSBase` unless the code in question has an `else` branch for CSC.

docs/release-notes/3700.feature.md
flying-sheep: I think that’s how one says it, but maybe “compatible over” is also correct, it just sounded weird to my 2nd language ears.

src/scanpy/get/_aggregated.py
flying-sheep: maybe add a link to them then!

docs/release-notes/3700.feature.md
ilan-gold: No that's probably an "ilan"-ism :)

src/scanpy/get/_aggregated.py
ilan-gold: This really should be a TODO.  To be correct, we would need to handle the overflow potential of `sum` i.e., jumping from `uint16` to `uint32` or something because the summation gets bigger.  I wonder how `dask` or `fau` handles this internally?  For example https://github.com/scverse/fast-array-utils/blob/b3c7c9829b2fe4e6d23915a2c6bab1b025484b10/src/fast_array_utils/stats/_sum.py#L89-L98 actually seems wrong for this exact reason, no?

ilan-gold:
> Awesome! I’m not 100% sure how it works though.

The idea is that we go row_chunk-by-row_chunk and aggregate those individually before doing a sum-like operation over all of them to get the final full-dataset aggregation.  For example, for `sum`, the aggregation of rows 0-100 into 10 categories and 15 features produces a 10x15 matrix of that subset's summation aggregation, but the same is true of any subset.  If we do these repeatedly over subsets, we can then do a `sum` operation over these `10x15` matrices to get the final aggregation result.

So each `map_blocks` operation outputs a `1x10x15` matrix giving an intermediary `n_chunks x 10 x 15` matrix that is then summed over the first axis (i.e., the `.sum(axis=0)` call after `map_blocks`) to give the final result, a `10x15` matrix.

This architecture allows us to reuse our already existing implementation as a subroutine and paves the way hopefully for #3723 applied here as well where we just have an additional `concat` operation over the feature space.

Does this clarify things?

src/scanpy/get/_aggregated.py
```diff
@@ -330,13 +331,103 @@ def _aggregate(
...
+        res = _aggregate(chunk, by_subsetted, func, mask=mask_subsetted, dof=dof)[func]
+        return res[None, :]
+
+    funcs = set([func] if isinstance(func, str) else func)
+    if "median" in funcs:
+        msg = "Dask median calculation not supported.  If you want a median-of-medians calculation, please open an issue."
+        raise NotImplementedError(msg)
+    has_mean, has_var = (v in funcs for v in ["mean", "var"])
+    funcs_no_var_or_mean = funcs - {"var", "mean"}
+    aggregated = {
```
flying-sheep, line 403: maybe we can include sth. like this? feel free to edit this comment

```suggestion
    # aggregate each row chunk individually,
    # producing a #chunks × #categories × #features array,
    # then aggregate the per-chunk results.
    aggregated = {
```

src/scanpy/get/_aggregated.py
flying-sheep: First, I still stand by my comment. Please don’t bi-partition an extensible set by specifying one partition. This should be a lookup, either central or by defining and accessing an inline dict/switch-case that spells out all options.

regarding the rest: hmm, I guess I conceived fau to be more low-level, i.e. if you fear that something can overflow, set `dtype` manually, otherwise it preserves wherever possible, e.g. it doesn’t make sense to preserve int dtypes in `mean`, but fau doesn’t know what how big the numbers you `sum` up are so it makes no assumptions.

does that make sense or do you think it should be more opinionated?

ilan-gold: Fair point, the two things are separate.  I will add a TODO as well then.  Have a look and let me know if this is what you had in mind or if I misunderstood

flying-sheep (approved):
beautiful!

PR #3701, opened 2025-07-04
[Docs] Visualizing scanpy's codebase

ivanmilevtues:
This PR contains high-level diagrams for the scandata's codebase. You can see how they render in Github here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/scanpy/on_boarding.md

These diagrams are designed to help people quickly understand a codebase. A lot of scientists engage with the scanpy codebase and we want to reduce the time to get up-to-speed from hours to minutes. We believe that most people start with small improvements and componentes that they want to check, this is why we are generating a diagram. The diagram is designed to give people context of the full system and then they can dive deeper into the component of interest. I'd love to hear where do you stand on this and if such docs can be part of your on-boarding procedure.

Any feedback is more than welcome! We've just released a free GitHub Action that can automatically update the diagrams and this way keep them always up-to-date.

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

Just a note I did a PR in another repo of yours a while back, but it became a bit outdated with our new versions so I closed it: https://github.com/scverse/anndata/pull/2010

- [ ] Closes #

### DIFF
--- a/.codeboarding/AnnData_Data_Model.md
+++ b/.codeboarding/AnnData_Data_Model.md
@@ -1,0 +1,285 @@
+```mermaid
+
+graph LR
+
+    AnnData_Data_Model["AnnData Data Model"]
+
+    scanpy_preprocessing["scanpy.preprocessing"]
+
+    scanpy_tools["scanpy.tools"]
+
+    scanpy_plotting["scanpy.plotting"]
+
+    scanpy_readwrite["scanpy.readwrite"]
+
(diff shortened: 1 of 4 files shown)


Zethson:
Hi,

I don't think that we're interested at this time as the diagrams are easy to get out of date, would need to be verified, and might not be that useful.
We might get back to you at a later point. Thanks!

PR #3724, opened 2025-07-21
Catch all warnings

### LINKED ISSUE #3558: Scanpy’s tests should prevent DeprecationWarnings and FutureWarnings

Improved documentation or error message?

See title

flying-sheep:
- [x] Closes #3558

TODO:
- does this make sense? https://github.com/scverse/scanpy/blob/02ecd491d2616724b0990ad70589d7ec02f98e52/tests/test_highly_variable_genes.py#L698-L702

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -181,6 +181,7 @@ filterwarnings = [
     'ignore:\n*.*scipy\.sparse array:DeprecationWarning:networkx',
     # we want to see and eventually fix these
     'default::numba.core.errors.NumbaPerformanceWarning',
+    'default:.*TSNE.*random.*to.*pca:FutureWarning',      # we should set init=obsm["X_pca"] or so
 ]
 
 [tool.coverage.run]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -173,7 +173,7 @@ filterwarnings = [
     'error',
     # Umap warns when tensorflow isn’t installed.
     'ignore::ImportWarning:umap',
-    # seaborn causes some matplotlib warnings
+    # seaborn≤0.13.2 causes some matplotlib warnings
     'ignore::PendingDeprecationWarning:seaborn',
     # matplotlib<3.10.4 causes a Pillow warning
     'ignore:.*Pillow:DeprecationWarning',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -177,8 +177,6 @@ filterwarnings = [
     "ignore::PendingDeprecationWarning:seaborn",
     # matplotlib<3.10.4 causes a Pillow warning
     "ignore:.*Pillow:DeprecationWarning",
-    # leidenalg (≤0.10.0) keeps using the wrong number of backslashes
-    "ignore::SyntaxWarning:leidenalg",
     # we want to see and eventually fix these
     "default::numba.core.errors.NumbaPerformanceWarning",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -170,26 +170,13 @@ markers = [
     "anndata_dask_support: tests that require dask support in anndata",
 ]
 filterwarnings = [
-    # legacy-api-wrap: internal use of positional API
-    "error:The specified parameters:FutureWarning",
-    # When calling `.show()` in tests, this is raised
-    "ignore:FigureCanvasAgg is non-interactive:UserWarning",
-
-    # We explicitly handle the below errors in tests
-    "error:`anndata.read` is deprecated:FutureWarning",
-    "error:Observation names are not unique:UserWarning",
-    "error:The dtype argument is deprecated and will be removed:FutureWarning",
-    "error:The behavior of DataFrame\\.sum with axis=None is deprecated:FutureWarning",
-    "error:The default of observed=False is deprecated:FutureWarning",
-    "error:Series\\.__getitem__ treating keys as positions is deprecated:FutureWarning",
-    "error:The default value of 'ignore' for the `na_action` parameter in pandas\\.Categorical\\.map:FutureWarning",
-    "error:The provided callable.* is currently using:FutureWarning",
-    "error:The behavior of DataFrame concatenation with empty or all-NA entries is deprecated:FutureWarning",
-    "error:A value is trying to be set on a copy of a slice from a DataFrame",
-    "error:No data for colormapping provided via 'c'",
-    "error:\\n*The `scale` parameter has been renamed and will be removed",
-    "error:\\n*Passing `palette` without assigning `hue` is deprecated",
-    "error:\\n*Setting a gradient palette using color= is deprecated",
+    "error",
+    # Umap warns when tensorflow isn’t installed.
+    "ignore::ImportWarning:umap",
+    # seaborn causes some matplotlib warnings
+    "ignore::PendingDeprecationWarning:seaborn",
+    # we want to see and eventually fix these
+    "default::numba.core.errors.NumbaPerformanceWarning",
 ]
 
 [tool.coverage.run]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -177,6 +177,8 @@ filterwarnings = [
     "ignore::PendingDeprecationWarning:seaborn",
     # matplotlib<3.10.4 causes a Pillow warning
     "ignore:.*Pillow:DeprecationWarning",
+    # leidenalg (≤0.10.0) keeps using the wrong number of backslashes
+    "ignore::SyntaxWarning:leidenalg",
     # we want to see and eventually fix these
     "default::numba.core.errors.NumbaPerformanceWarning",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -170,15 +170,17 @@ markers = [
     "anndata_dask_support: tests that require dask support in anndata",
 ]
 filterwarnings = [
-    "error",
+    'error',
     # Umap warns when tensorflow isn’t installed.
-    "ignore::ImportWarning:umap",
+    'ignore::ImportWarning:umap',
     # seaborn causes some matplotlib warnings
-    "ignore::PendingDeprecationWarning:seaborn",
+    'ignore::PendingDeprecationWarning:seaborn',
     # matplotlib<3.10.4 causes a Pillow warning
-    "ignore:.*Pillow:DeprecationWarning",
+    'ignore:.*Pillow:DeprecationWarning',
+    # networkx 2.x warns about scipy.sparse changes
+    'ignore:\n*.*scipy\.sparse array:DeprecationWarning:networkx',
     # we want to see and eventually fix these
-    "default::numba.core.errors.NumbaPerformanceWarning",
+    'default::numba.core.errors.NumbaPerformanceWarning',
 ]
 
 [tool.coverage.run]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,7 +180,7 @@ filterwarnings = [
     # networkx 2.x warns about scipy.sparse changes
     'ignore:\n*.*scipy\.sparse array:DeprecationWarning:networkx',
     # h5py <3.9 uses deprecated np.product
-    'ignore:.*product.*deprecated.*NumPy:DeprecationWarning:h5py',
+    'ignore:.*product.*deprecated.*NumPy:DeprecationWarning',
     # we want to see and eventually fix these
     'default::numba.core.errors.NumbaPerformanceWarning',
     'default:.*TSNE.*random.*to.*pca:FutureWarning',      # we should set init=obsm["X_pca"] or so
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -179,6 +179,8 @@ filterwarnings = [
     'ignore:.*Pillow:DeprecationWarning',
     # networkx 2.x warns about scipy.sparse changes
     'ignore:\n*.*scipy\.sparse array:DeprecationWarning:networkx',
+    # h5py <3.9 uses deprecated np.product
+    'ignore:.*product.*deprecated.*NumPy:DeprecationWarning:h5py',
     # we want to see and eventually fix these
     'default::numba.core.errors.NumbaPerformanceWarning',
     'default:.*TSNE.*random.*to.*pca:FutureWarning',      # we should set init=obsm["X_pca"] or so
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -175,6 +175,8 @@ filterwarnings = [
     "ignore::ImportWarning:umap",
     # seaborn causes some matplotlib warnings
     "ignore::PendingDeprecationWarning:seaborn",
+    # matplotlib<3.10.4 causes a Pillow warning
+    "ignore:.*Pillow:DeprecationWarning",
     # we want to see and eventually fix these
     "default::numba.core.errors.NumbaPerformanceWarning",
 ]
--- a/src/scanpy/datasets/_datasets.py
+++ b/src/scanpy/datasets/_datasets.py
@@ -123,6 +123,8 @@ def burczynski06() -> AnnData:
     --------
     >>> import scanpy as sc
     >>> sc.datasets.burczynski06()
+    UserWarning: Variable names are not unique. To make them unique, call `.var_names_make_unique`.
+        utils.warn_names_duplicates("var")
     AnnData object with n_obs × n_vars = 127 × 22283
         obs: 'groups'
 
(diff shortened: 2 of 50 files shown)


src/scanpy/plotting/_tools/paga.py
```diff
@@ -1331,7 +1331,7 @@ def moving_average(a):
                 cmap=color_map_anno,
             )
             if show_yticks:
-                anno_axis.set_yticklabels(["", anno, ""], fontsize=ytick_fontsize)
+                anno_axis.set_yticks(range(3), [anno, "", ""], fontsize=ytick_fontsize)
```
flying-sheep, line 1334: I don’t think this has ever been visible, so it’s unclear if this fix is right.

(the other fix in this file *is* visually correct)

src/scanpy/preprocessing/_simple.py
```diff
@@ -565,56 +569,61 @@ def normalize_per_cell(
...
 
     """
-    if isinstance(data, AnnData):
-        start = logg.info("normalizing by total count per cell")
-        adata = data.copy() if copy else data
-        if counts_per_cell is None:
-            cell_subset, counts_per_cell = materialize_as_ndarray(
-                filter_cells(adata.X, min_counts=min_counts)
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", r".*sc\.pp\.normalize_total", FutureWarning)
```
flying-sheep, line 573: the only change to this whole thing is that I pulled the `catch_warnings` up and added this filter.

src/testing/scanpy/_pytest/fixtures/__init__.py
```diff
@@ -53,10 +53,22 @@ def showwarning(message, category, filename, lineno, file=None, line=None) -> No
         else:
             showwarning_orig(message, category, filename, lineno, file, line)
 
-    # make errors visible and the rest ignored
+    # make errors (except plt.show) visible and the rest ignored
```
flying-sheep: everywhere else I added an explicit `show=False` to avoid this warning but I don’t think that makes sense for doctests.

tests/test_highly_variable_genes.py
```diff
@@ -692,10 +695,17 @@ def test_dask_consistency(adata: AnnData, flavor, batch_key, to_dask):
     adata_dask = adata.copy()
     adata_dask.X = to_dask(adata_dask.X)
 
-    output_mem, output_dask = (
-        sc.pp.highly_variable_genes(ad, flavor=flavor, n_top_genes=15, inplace=False)
-        for ad in [adata, adata_dask]
-    )
+    with (
+        pytest.warns(UserWarning, match="n_top_genes.*normalized dispersions")
```
flying-sheep, line 704: we already test for this in `test_cellranger_n_top_genes_warning` above, so maybe this test should be changed to not raise this warning?

Then I’d just make it visible here (Using the `always` warning filter action) and leave it like that for now.

tests/test_preprocessing.py
```diff
@@ -247,28 +278,32 @@ def test_sample_copy_backed_error(tmp_path):
...
-    sc.pp.scale(adata, zero_center=zero_center, max_value=max_value)
-    with maybe_dask_process_context():
+    warn_ctx = pytest.warns(UserWarning, match=r"zero-centering.*densifies")
+    with warn_ctx if zero_center else nullcontext():
+        sc.pp.scale(adata, zero_center=zero_center, max_value=max_value)
+    with (
+        (
+            warn_ctx
+            if zero_center
+            and any(pat in array_type.__name__ for pat in ("sparse", "csc", "csr"))
```
flying-sheep, line 295: I made this more elegant for `count_matrix_format`, but to really fix that, we need fast-array-utils’ testing utils.

tests/test_rank_genes_groups.py
```diff
@@ -219,6 +223,7 @@ def test_log1p_save_restore(tmp_path):
     from anndata import read_h5ad
 
     pbmc = pbmc68k_reduced()
+    pbmc.X = pbmc.raw.X
```
flying-sheep, line 226: this makes it so log1p doesn’t complain about negative values.

tests/test_scaling.py
```diff
@@ -95,9 +98,18 @@ def test_scale(
     *, typ, container, zero_center, dtype, mask_obs, x, x_centered, x_scaled
 ):
     x = AnnData(typ(x, dtype=dtype)) if container == "anndata" else typ(x, dtype=dtype)
-    scaled = sc.pp.scale(
-        x, zero_center=zero_center, copy=container == "array", mask_obs=mask_obs
-    )
+    with warnings.catch_warnings():
+        # TODO: fix setting slices of sparse matrices in scale()
+        warnings.filterwarnings("always", category=sparse.SparseEfficiencyWarning)
```
flying-sheep, line 103: this is a case of having a real issue that wasn’t caught in a sea of warnings we ignored. Basically `scale()` does `csr_matrix()[:, some_rows] = ...` or so, which scipy complains about

src/scanpy/datasets/_datasets.py
```diff
@@ -123,6 +123,8 @@ def burczynski06() -> AnnData:
     --------
     >>> import scanpy as sc
     >>> sc.datasets.burczynski06()
+    UserWarning: Variable names are not unique. To make them unique, call `.var_names_make_unique`.
+        utils.warn_names_duplicates("var")
```
ilan-gold, line 127: I must be misunderstanding.  From the changes in `readwrite.py`, I would _not_ expect this warning to appear.  So what am I missing?

ilan-gold: Same goes for the other warnings that now appear in doctests

ilan-gold (commented):
I need to make another pass at this to be sure I understand what is going on.  Here's my understanding, this PR seems to do a few things:

1. Remove all warnings in `pyproject.toml` that don't apply anymore (presumably?)
2. Catch all warnings explicitly that weren't caught before (I can see this locally)
3. Clean up some `anndata<0.9` straggling functions/compat

Is this right?

I still get locally `SparseEfficiencyWarning` and `NumbaPerformanceWarning`.  Do you see this as well?

src/testing/scanpy/_pytest/fixtures/__init__.py
```diff
@@ -53,10 +54,17 @@ def showwarning(message, category, filename, lineno, file=None, line=None) -> No
...
+        r"FigureCanvasAgg is non-interactive, and thus cannot be shown",
+        UserWarning,
+    )
     # make errors visible and the rest ignored
+    action_map = defaultdict(lambda: "ignore", error="default")
     warnings.filters = [
-        ("default", *rest) for action, *rest in warnings.filters if action == "error"
-    ] + [("ignore", None, Warning, None, 0)]
+        (action_map[action], *rest) for action, *rest in warnings.filters
+    ]
```
flying-sheep, line 63: this is much better now, as it will preserve the warning filters more accurately: any action apart from “ignore” and “error” will normally cause a warning to be shown (once or more often), so this should keep the precedence intact while downgrading each warning one level.

ignoring the few warnings we let through or ignore in the rest of the test suite and displaying the vast majority of warnings we upgrade into errors in the rest of the test suite as doctest output.

src/scanpy/datasets/_datasets.py
flying-sheep: OK, so before this PR, we opted into turning some warnings into errors. Now we turn warnings into errors by default and opt-out of some.

The general idea of what tool I used where is
- we should globally ignore (in `pyproject.toml`) stuff that’s some 3rd party bug. then it’s localized with dependency versions and we’ll hopefully think of removing the entries when they’re obsolete
- we should use `pytest.warns` for warnings we expect to be raised here (so we get notified when they don’t get raised anymore)
- we should use `with catch_warnings(): filterwarnings('ignore', ...)` or `@pytest.mark.filterwarnings` for warnings that we expect but that might also go away eventually without that justifying tests breaking, and for warnings that the user doesn’t need to see (e.g. a second instance of a warning we just raised or some expected 3rd party warning that points out behavior that we handle correctly).

> Clean up some anndata<0.9 straggling functions/compat

yeah, that could be extracted to another PR, the rest all makes sense for this I think.

> From the changes in readwrite.py, I would not expect this warning to appear. So what am I missing?

the comment says “we raise this below/above”, meaning that without suppressing, the warning would be raised twice in one call.

similar for `datasets.visium` and `read_visium`, where we don‘t want the user to *also* see the `read_visium` warning when they call `datasets.visium` and get a warning for that.

ilan-gold: I think my question is simply "why not catch the unique variable names warning for datasets accessed through this file?"

pyproject.toml
```diff
@@ -170,26 +170,13 @@ markers = [
...
-    "error:\\n*The `scale` parameter has been renamed and will be removed",
-    "error:\\n*Passing `palette` without assigning `hue` is deprecated",
-    "error:\\n*Setting a gradient palette using color= is deprecated",
+    "error",
+    # Umap warns when tensorflow isn’t installed.
+    "ignore::ImportWarning:umap",
+    # seaborn causes some matplotlib warnings
+    "ignore::PendingDeprecationWarning:seaborn",
+    # we want to see and eventually fix these
+    "default::numba.core.errors.NumbaPerformanceWarning",
```
ilan-gold: I see this warning locally.

I tried looking at the CI output but now it runs for 34 minutes?

tests/test_qc_metrics.py
```diff
@@ -188,47 +188,45 @@ def test_dask_against_in_memory(adata, log1p):
...
 
 
 @pytest.mark.parametrize("cls", ARRAY_TYPES_MEM)
-def test_qc_metrics_format(cls):
-    adata_dense, init_var = adata_mito()
-    sc.pp.calculate_qc_metrics(adata_dense, qc_vars=["mito"], inplace=True)
-    adata = AnnData(X=cls(adata_dense.X), var=init_var.copy())
+def test_qc_metrics_format(cls, adata_mito: AnnData) -> None:
+    sc.pp.calculate_qc_metrics(adata_mito, qc_vars=["mito"], inplace=True)
+    adata = AnnData(X=cls(adata_mito.X), var=adata_mito.var.copy())
```
ilan-gold:
```suggestion
    var_clean = adata_mito.var.copy()
    sc.pp.calculate_qc_metrics(adata_mito, qc_vars=["mito"], inplace=True)
    adata = AnnData(X=cls(adata_mito.X), var=var_clean)
```
I assume the point of the test was to check if the calculated qc metrics for `var` are identical if you give it the same `var` twice, but with these changes we are giving the a copy of what was just run through `calculate_qc_metrics`

tests/test_highly_variable_genes.py
ilan-gold: I think that just because the warning is raised twice doesn't make it "wrong."  In this case, I would just leave it.

tests/test_get.py
```diff
@@ -407,11 +407,12 @@ def test_repeated_index_vals(dim, transform, func):
 )
 def shared_key_adata(request):
     kind = request.param
-    adata = sc.AnnData(
-        np.arange(50).reshape((5, 10)),
-        obs=pd.DataFrame(np.zeros((5, 1)), columns=["var_id"]),
-        var=pd.DataFrame(index=["var_id"] + [f"gene_{i}" for i in range(1, 10)]),
-    )
+    with pytest.warns(ImplicitModificationWarning):
+        adata = sc.AnnData(
```
ilan-gold: What is going on here? `sc.AnnData`? Why is there an `ImplicitModificationWarning`?

src/scanpy/readwrite.py
```diff
@@ -227,7 +228,13 @@ def read_10x_h5(
     with h5py.File(str(path), "r") as f:
         v3 = "/matrix" in f
     if v3:
-        adata = _read_10x_h5(path, _read_v3_10x_h5)
+        with warnings.catch_warnings():
+            if genome or gex_only:
+                # this will be thrown below by “adata.copy()”
+                warnings.filterwarnings(
+                    "ignore", r".*names are not unique", UserWarning
+                )
```
ilan-gold, line 236: Why would we catch this warning always? And why here and not e.g., for the `soft.gz` file in the other discussion?

src/scanpy/datasets/_datasets.py
flying-sheep: I was thinking that the warning is still valid, because the returned AnnData object still has duplicate vars.

We should have fixed that from the start, now we’re kinda locked in until scanpy 2: https://github.com/scverse/scanpy/issues/3729

pyproject.toml
flying-sheep: yeah, I’m debugging the remaining test failure, which means temporarily setting `PYTHONTRACEMALLOC`

tests/test_qc_metrics.py
flying-sheep: good point, we should use the .var before running this.

tests/test_get.py
flying-sheep: A lot of the tests use scanpy’s re-export of `AnnData`. Not a fan of that existing, maybe we can deprecate it.

> Why is there an ImplicitModificationWarning?

when passing `obs=pd.DataFrame()` without an explicit string index, that gets raised when converting the default RangeIndex to string.

but I guess passing `obs=dict(var_id=np.zeros(5))` here is the better idea.

src/scanpy/readwrite.py
flying-sheep: > Why would we catch this warning always?

see the comment on the line above the line you commented on: the warning does get raised again later in the same function, this prevents it from being raised twice.

> And why here and not e.g., for the soft.gz file in the other discussion?

I assume you mean `sc.datasets.burczynski06`? if so: because that only raises one warning.

ilan-gold: > see the comment on the line above the line you commented on: the warning does get raised again later in the same function, this prevents it from being raised twice.

Ah ok now I understand what you meant by that!

src/scanpy/datasets/_datasets.py
ilan-gold: Got it, yes now I get it

pyproject.toml
flying-sheep: OK, that helped figuring it out!

tests/test_get.py
```diff
@@ -407,12 +407,11 @@ def test_repeated_index_vals(dim, transform, func):
 )
 def shared_key_adata(request):
     kind = request.param
-    with pytest.warns(ImplicitModificationWarning):
-        adata = sc.AnnData(
-            np.arange(50).reshape((5, 10)),
-            obs=pd.DataFrame(np.zeros((5, 1)), columns=["var_id"]),
-            var=pd.DataFrame(index=["var_id"] + [f"gene_{i}" for i in range(1, 10)]),
-        )
+    adata = sc.AnnData(
```
ilan-gold, line 410: Why `sc.AnnData`?

src/scanpy/tools/_tsne.py
```diff
@@ -129,8 +125,6 @@ def tsne(  # noqa: PLR0913
         metric=metric,
         n_components=n_components,
     )
-    if metric != "euclidean" and (Version(sklearn.__version__) < Version("1.3.0rc1")):
-        params_sklearn["square_distances"] = True
```
ilan-gold, line 133: Did removing this cause `default:.*TSNE.*random.*to.*pca:FutureWarning` to appear or?

tests/test_highly_variable_genes.py
```diff
@@ -391,11 +393,16 @@ def test_compare_to_upstream(
...
         sc.pp.highly_variable_genes(pbmc, flavor=flavor, **params, inplace=True)
     elif func == "fgd":
-        sc.pp.filter_genes_dispersion(
-            pbmc, flavor=flavor, **params, log=True, subset=False
-        )
+        with pytest.warns(FutureWarning, match=r"sc\.pp\.highly_variable_genes"):  # noqa: PT031
+            # https://github.com/pandas-dev/pandas/issues/61928
+            warnings.filterwarnings(
+                "ignore", r"invalid value encountered in cast", RuntimeWarning
+            )
```
ilan-gold, line 400: If this is coming from our internal code, should we suppress it to outside users?

tests/test_get.py
flying-sheep: https://github.com/scverse/scanpy/pull/3724#discussion_r2222669324

src/scanpy/tools/_tsne.py
flying-sheep: No. Before removing this, the low-vers job raised a warning saying “square_distances does nothing and it will be removed”. So since it already does nothing in our minimum supported version, I just removed it.

The warning you’re referring to is interesting but completely unrelated: TSNE has an `init` parameter that apparently (unnoticed by us) changed behavior. The new behavior (I think) computes a PCA from the input, which is very unnecessary, given that we feed in a PCA in most cases. We should maybe fix that eventually, but who cares about TSNE anymore.

tests/test_highly_variable_genes.py
flying-sheep: it doesn’t come from us, it’s related to the pandas bug I link one line above your comment.

the pandas bug is valid in any case, and in very particular circumstances (i.e. only in the min-vers job and only in CI, I couldn’t reproduce it locally) pandas applying weird unnecessary casting to its Interval class it causes that warning.

maybe instead of working around pandas’ bug I should just globally catch that warning only in the min-vers job, but it feels wrong to catch a warning that we could feasibly trigger somewhere.

ilan-gold (approved):

PR #3735, opened 2025-07-24
Optimise batch-aware highly_variable_genes for Dask arrays

### LINKED ISSUE #3736: sc.pp.highly_variable_genes does not scale when `batch_key` is set to a large number of batches for dask arrays

The following doesn't scale well:

```python
# Create dummy dataset
adata_sub = sc.datasets.pbmc3k()
adata_sub.X = sparse.csr_matrix(adata_sub.X.astype('float32'))

sc.pp.normalize_total(adata_sub)
sc.pp.log1p(adata_sub)

n_batches = 100
adata = ad.concat([adata_sub] * n_batches, label='batch', keys=range(n_batches))
adata.obs_names_make_unique()

adata_dask = ad.AnnData(
    da.from_array(adata.X),
    obs=adata.obs,
    var=adata.var,
)

# Test highly_variable_genes
batch_key = 'batch'
sc.pp.highly_variable_genes(adata, n_top_genes=2000, batch_key=batch_key)
sc.pp.highly_variable_genes(adata_dask, n_top_genes=2000, batch_key=batch_key)
```
takes about 12 min for dask matrix, but only about 6s for csr_matrix.

mumichae:
`sc.pp.highly_variable_genes` does not scale well when run per batch (`batch_key` is set) for Dask arrays. This PR refactors the per batch runs to get similar runtime as the non dask array version.

- [x] Closes #3736
- [x] Tests included or not required because: no API changes


src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -437,14 +437,10 @@ def _highly_variable_genes_batched(
     flavor: Literal["seurat", "cell_ranger"],
     cutoff: _Cutoffs | int,
 ) -> pd.DataFrame:
-    sanitize_anndata(adata)
-    batches = adata.obs[batch_key].cat.categories
-    dfs = []
-    gene_list = adata.var_names
-    for batch in batches:
-        adata_subset = adata[adata.obs[batch_key] == batch]
+    def process_batch(batch_mask, adata, layer, gene_list, **hvg_kwargs):
```
flying-sheep: Please add types. You can type `hvg_kwargs` by doing

```py
class HvgArgs(TypedDict):
    cutoff: _Cutoffs | int
    n_bins: int,
    flavor: flavor: Literal["seurat", "cell_ranger"]
```

and then use it in functions:

```py
def ...(..., **hvg_kwargs: Unpack[HvgArgs]) -> ...:
```

you could also update `_highly_variable_genes_batched` to use it that way.

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -472,7 +469,33 @@ def _highly_variable_genes_batched(
...
+    sanitize_anndata(adata)
+    batches = adata.obs[batch_key].cat.categories
+    X = _get_obs_rep(adata, layer=layer)
+
+    per_batch_func = process_batch
+    if isinstance(X, DaskArray):
+        from dask import compute, delayed
+
+        per_batch_func = delayed(process_batch)
+        _set_obs_rep(adata, X.persist(), layer=layer)
```
flying-sheep: I still don’t know what `persist` is for, but @ilan-gold doesn’t like it I think. @ilan-gold, what are your objections again? Do they apply here?

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -472,7 +469,33 @@ def _highly_variable_genes_batched(
...
+            layer=layer,
+            gene_list=adata.var_names,
+            cutoff=cutoff,
+            n_bins=n_bins,
+            flavor=flavor,
+        )
+        for batch in batches
+    ]
+    if not isinstance(dfs[0], pd.DataFrame):
+        dfs = compute(*dfs)
```
flying-sheep: I think this is cleaner:

Do

```py
if is_dask := isinstance(X, DaskArray):
    from dask import delayed
```

above and then

```suggestion
    if is_dask:
        from dask import compute

        dfs = compute(*dfs)
```

maybe even do `dfs = (compute(df) for df in dfs)`?

if there are many, pandas might be able to use that more effectively, but maybe have dask schedule all the computes as it sees fit is better.

flying-sheep (commented):
This looks great! With a few tiny changes to the conditional import you mentioned, I’d say this is cleaner than before.

src/scanpy/preprocessing/_highly_variable_genes.py
ilan-gold: Also probably don't want this inlined

src/scanpy/preprocessing/_highly_variable_genes.py
ilan-gold: It requires that you have enough memory to do this, hence my general opposition.  

@mumichae In your specific use-case, are you reading data off-disk or are you creating  dask arrays on top of in-memory arrays to get the memory benefit from concatenation (I see in `adata = ad.concat([adata_sub] * n_batches, label='batch', keys=range(n_batches))` you are doing things in memory).  I would ask then why this shouldn't be something the user does outside this function?

ilan-gold (commented):
Generally this looks great!

src/scanpy/preprocessing/_highly_variable_genes.py
mumichae: The in-memory example was just for playing around and figuring out the bottleneck. But I also observed the same issue with on-disk arrays, which is what I use in real-world workflows.

With persist, I guess it's a tradeoff of memory vs runtime, and for my examples things scaled accordingly (w/o persist has lower memory usage but takes longer and vice versa). But even without persist, the current performance is nowhere near as bad as before my fix, so I think it should be fine.

I removed the persist statement from the code and will just call persist in my workflows. I agree that it's better to let the user decide how they want the matrix stored.

src/scanpy/preprocessing/_highly_variable_genes.py
mumichae: I refactored that part of the code, but also cleaned up some dataframe handling, as well as making the filter call part of the `_highly_variable_genes_single_batch` directly, which could be considered new functionality. Lmk if things look OK or you want to preserve more of the old code

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -428,56 +467,52 @@ def _nth_highest(x: NDArray[np.float64] | DaskArray, n: int) -> float | DaskArra
...
-                    min_cells=1,
-                    inplace=False,
-                )
-            )
+    func = _per_batch_func
+    if isinstance(X, DaskArray):
+        from dask import delayed
 
-        adata_subset = adata_subset[:, filt]
+        func = delayed(_per_batch_func)
```
flying-sheep, line 493: I don’t think this parameter exists!

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -538,6 +573,7 @@ def highly_variable_genes(  # noqa: PLR0913
     subset: bool = False,
     inplace: bool = True,
     batch_key: str | None = None,
+    _filter_genes: bool = False,
```
flying-sheep: this parameter is unused, what’s the plan here?

if this isn’t accidentally left in: if you want to introduce a new public parameter, it can’t have a name prefixed with `_`, if it’s for internal use only, it shouldn’t have documentation

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -428,56 +467,52 @@ def _nth_highest(x: NDArray[np.float64] | DaskArray, n: int) -> float | DaskArra
...
 
-        hvg = _highly_variable_genes_single_batch(
-            adata_subset, layer=layer, cutoff=cutoff, n_bins=n_bins, flavor=flavor
+    dfs = (
+        func(
+            _highly_variable_genes_single_batch,
+            adata=adata,
+            batch_mask=adata.obs[batch_key] == batch,
+            layer=layer,
+            gene_list=adata.var_names,
```
flying-sheep: I don’t think this parameter exists!

mumichae: Good point, I forgot to remove after refactoring.

```suggestion
```

src/scanpy/preprocessing/_highly_variable_genes.py
mumichae: I changed it to `filter_unexpressed_genes`. This parameter was introduced so that the gene filtering required by per-batch HVGs is possible. It also adds more functionality to the top-level `highly_variable_genes` function

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -538,6 +572,7 @@ def highly_variable_genes(  # noqa: PLR0913
     subset: bool = False,
     inplace: bool = True,
     batch_key: str | None = None,
+    filter_unexpressed_genes: bool = False,
```
flying-sheep: This parameter is still unused here, passing it to `highly_variable_genes` does nothing.

Can you please either remove it again or:

1. set the default to `None`, since it’s derived from `batch_key`.
2. explain in the parameter docs and a comment on the `else` branch where `_highly_variable_genes_batched` is called, that the parameter is always `True` if `batch_key` is `False`
3. raise an error if an incompatible combination is chosen (i.e. `batch_key="..."` and `filter_unexpressed_genes=False`) and add a test that this error is raised
4. actually pass it on to `_highly_variable_genes_single_batch` and add a test that checks that `highly_variable_genes(..., batch_key=None, filter_unexpressed_genes=True)` behaves differently from `highly_variable_genes(..., batch_key=None, filter_unexpressed_genes=False)`

flying-sheep (changes requested):

src/scanpy/preprocessing/_highly_variable_genes.py
mumichae: Ah, I meant to pass it to the `_highly_variable_genes_single_batch` in the non-batch aware mode. I applied your suggestions, but used a warning instead of an error in case of incompatible configuration, as I don't think it warrants an error. But open to changing it to an error if that makes more sense from a usability standpoint

mumichae:
@flying-sheep  I addressed your comments (thanks a lot for catching the issue btw, I completely missed it!)
But for some reasons I'm getting issues with the test upload. Any way you can rerun the github actions for the last commit? Or any input on what went wrong?

flying-sheep (approved):

src/scanpy/preprocessing/_highly_variable_genes.py
flying-sheep: I’m moving it into a new PR so this one can go into a patch

Issue #3718, opened 2025-07-18
Refactor `BasePlot` not to create a dataframe representation of the data


ilan-gold:
### What kind of feature would you like to request?

Additional function parameters / changed functionality / changed defaults?

### Please describe your wishes

See  #3717 for what prompted me to look at this code.

Currently `BasePlot` creates a in-memory copy as a dataframe of the main data of interest (obsm, X, layers etc.): https://github.com/scverse/scanpy/blob/0b82c934edeb640095df14e60725ed1fad6eebb1/src/scanpy/plotting/_baseplot_class.py#L148-L157

I believe this to be unnecessary as this dataframe is only ever used for `groupby` operations, for which we have a zero-copy solution in https://scanpy.readthedocs.io/en/latest/generated/scanpy.get.aggregate.html

Thus we should

- [ ] Refactor `Baseplot` not to create a copy
- [ ] Use the  https://scanpy.readthedocs.io/en/latest/generated/scanpy.get.aggregate.html for aggregation
- [ ] Ensure this doesn't affect performance
- [ ] Integrate #3700

dorzhey:
I'd like to work on this issue, will submit a PR soon.

ilan-gold:
wooo great thanks @dorzhey !

srikarjy:
Hi @ilan-gold @flying-sheep noticed this has bounced across a few milestones and doesn't currently show a linked PR. Is this still something you'd want a contribution on, given the _v2/holoviz plotting work in progress? Happy to pick it up if so, or take a pointer elsewhere if this one's lower priority now.

PR #3757, opened 2025-08-03
performance: Refactor BasePlot & subclasses to copy less

dorzhey:
- [x] Closes [#3718](https://github.com/scverse/scanpy/issues/3718)
- [x] [Tests][plotting] this is a pure refactor of plotting internals, no new behavior, and all existing plotting tests pass locally (the one failing heatmap test also fails on main in my environment).

- [X] [Release notes][] not necessary because: there are no user-facing API changes, only internal performance and memory optimizations.

### DIFF
--- a/src/scanpy/plotting/_baseplot_class.py
+++ b/src/scanpy/plotting/_baseplot_class.py
@@ -2,11 +2,14 @@
 
 from __future__ import annotations
 
+import warnings
 from collections.abc import Mapping
-from typing import TYPE_CHECKING, NamedTuple
+from typing import TYPE_CHECKING, NamedTuple, overload
 from warnings import warn
 
 import numpy as np
+import pandas as pd
+from anndata import AnnData
 from matplotlib import colormaps, gridspec
 from matplotlib import pyplot as plt
 
@@ -24,15 +27,14 @@ from ._anndata import (
 from ._utils import check_colornorm, make_grid_spec
 
 if TYPE_CHECKING:
-    from collections.abc import Sequence
+    from collections.abc import Iterable, Sequence
     from typing import Literal, Self
 
-    import pandas as pd
-    from anndata import AnnData
     from matplotlib.axes import Axes
     from matplotlib.colors import Colormap, Normalize
 
     from .._utils import Empty
+    from ..get._aggregated import AggType
     from ._utils import ColorLike, _AxesSubplot
 
     _VarNames = str | Sequence[str]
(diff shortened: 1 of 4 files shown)


dorzhey:
Ah, I failed some of the style checks. I apologize, pre-commit does not work on my environment for some reason, gives error with UTF encoding and tests fail on main

flying-sheep:
no worries, I fixed it for you

flying-sheep:
OK, I got rid of the “converting index to string” warnings that obscured the actual test failures, so you should be able to see which tests failed now: https://github.com/scverse/scanpy/actions/runs/16826467121/job/47664011892?pr=3757#step:6:7989

gkneighb (commented):
Thanks for tackling this @dorzhey — the move toward `sc.get.aggregate` is the right direction and matches what @ilan-gold suggested in #3718. A few things I noticed reading the diff:

## Correctness: `dotplot`'s `mean_only_expressed` path looks broken

In `_dotplot.py`, the new `mean_only_expressed and expression_cutoff > 0` branch is:

```python
mask = expression_cutoff < self._view.X
df_sum = self._agg_df("sum", mask=mask)
expr_counts = dot_size_df.values * group_sizes[:, None]
dot_color_df = df_sum.div(expr_counts).fillna(0)
```

But `_agg_df(..., mask=mask)` does `view.X = mask.astype(view.X.dtype)` and then aggregates — so:

- `df_sum` is the sum of the **mask** (= count of expressed cells per group/gene), not of `X`
- `expr_counts` = `dot_size_df * group_size` = fraction-expressed × group-size = count of expressed cells per group/gene

i.e. `dot_color_df = count_expressed / count_expressed = 1.0` wherever any cell expresses (and `0` otherwise). The actual expression values from `self._view.X` are never read in this branch.

The semantic on main is "mean expression value, restricted to cells where expression > cutoff":

```python
# main
self.obs_tidy.mask(~obs_bool).groupby(level=0, observed=True).mean().fillna(0)
```

A `sc.get.aggregate`-friendly equivalent would be to aggregate `X * mask` (sum of expressed values) and divide by the count of expressed cells:

```python
mask = self._view.X > expression_cutoff
df_sum = self._agg_df("sum", x_override=self._view.X * mask)  # see #2 below
df_count = self._agg_df("sum", x_override=mask.astype(self._view.X.dtype))
dot_color_df = df_sum.div(df_count).fillna(0)
```

Worth adding a unit test for this case — comparing the new code to the main implementation on a small fixture would have caught it.

Also: the `mean_only_expressed and expression_cutoff == 0` case is silently treated as "mean over all cells" rather than "mean over nonzero cells", which is the most common usage. That's a separate behavior regression in the same area.

## API: the `mask` parameter does double duty

`_agg_df`'s docstring says `mask` is "shape `(n_groups, n_vars)`" but every call site passes a `(n_obs, n_vars)` matrix, because the parameter actually overwrites `view.X`, not filters it. Renaming to `x_override` (or splitting into a separate method) would make this honest at the API surface.

## Minor

- `view = self._view.copy()` in `_agg_df` runs on every aggregation. For sparse `X` that copy could dominate the savings the refactor is going for. Worth skipping the copy when `mask is None`.
- `self._view = AnnData(X=obs_tidy.values, ...)` densifies and copies (`.values` materializes a contiguous ndarray). Per the comment, `_prepare_dataframe` is still needed, so the upstream copy isn't avoided — most of the wins are downstream of construction. Worth noting in the PR body so reviewers don't expect init-time savings.
- The PR currently shows as `CONFLICTING` with `main`; needs a rebase once the above is sorted.

The dotplot/matrixplot/violin layering looks much cleaner with `_scale_df` and `_agg_df` factored out — that's a nice cleanup independent of the perf wins.

PR #3771, opened 2025-08-22
fix: Ignore NaNs when calculating scatter plot dot size

### LINKED ISSUE #3770: Size calculation in `sc.pl.embedding` should ignore NaNs

Hi everyone!

When I use `sc.pl.embedding` with some basis and this basis has NaNs for some cells, it simply omits those cells in the plot. This is, because the basis gets passed to `matplotlib.pyplot.scatter`, which has this as its default behavior.
I like that a lot, because it allows me to store embeddings of only a subset of cells (padded by NaNs) in an anndata object.
However, when calculating the size for the dots in the scatterplot, the number of cells in the anndata is used instead of the number of dots that's actually plotted, leading to very small dots.

You can reproduce this as follows:
```python
import scanpy as sc
import numpy as np

m1 = np.random.randn(5,10)
m2 = np.full(shape=(5,10), fill_value=np.nan)
m = np.concatenate([m1, m2], axis=0)

adata = sc.AnnData(np.random.randn(10,10))
adata.obsm['X_umap'] = m

sc.pl.umap(adata[~np.isnan(adata.obsm['X_umap'].sum(axis=1))])
sc.pl.umap(adata)
```
In both cases, 5 cells are plotted, but the dots are smaller in the second UMAP.

I'll open a PR with a suggestion to improve this in a second.

MLubetzki:
- [x] Closes #3770 
- [ ] [Tests][] included or not required because: I ran the tests, but already had 11 failing tests on the main branch?

- [x] [Release notes][] not necessary because: Minor change


flying-sheep:
@MLubetzki thank you very much!

@Zethson this is a fix, it should go into the bugfix release milestone (and many thanks to you too for helping with triaging and reviews!)

PR #3772, opened 2025-08-25
ci: Check PRs for semantic labels

flying-sheep:
… and adapts towncrier fragments to match the labels

Also:
- cleans up CI (remove useless options, update actions)
- adds dependabot for automatic GH actions updates

Does not:
- check if PR title matches fragment (TODO? how to deal with breaking vs “!”)?


flying-sheep:
Ah OK, we’re not using the `publish` job for CI because of stuff like

> This branch had an error being deployed

got it. went back to duplicating the “build” job between the CI and publish workflow.

PR #3800, opened 2025-09-16
ci: Fix tests

flying-sheep:
See https://github.com/lmcinnes/umap/issues/1216

### DIFF
--- a/tests/test_neighbors.py
+++ b/tests/test_neighbors.py
@@ -6,17 +6,25 @@ from typing import TYPE_CHECKING
 import numpy as np
 import pytest
 from anndata import AnnData
+from packaging.version import Version
 from scipy import sparse
 from sklearn.neighbors import KNeighborsTransformer
 
 import scanpy as sc
 from scanpy import Neighbors
-from scanpy._compat import CSBase
+from scanpy._compat import CSBase, pkg_version
 
 if TYPE_CHECKING:
     from typing import Literal
 
     from pytest_mock import MockerFixture
+
+# https://github.com/lmcinnes/umap/issues/1216
+SKIPIF_UMAP_BROKEN = pytest.mark.skipif(
+    pkg_version("umap-learn") <= Version("0.5.9.post2")
+    and pkg_version("numba") >= Version("0.62.0rc1"),
+    reason="umap is broken with numba",
+)
 
 # the input data
 X = [[1, 0], [3, 0], [5, 6], [0, 4]]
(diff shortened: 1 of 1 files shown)


ilan-gold (approved):

simonwm:
@flying-sheep Fixing the test fixes the error message, but would it not make more sense to pin the numba version to <0.62? Or just monkey patch the issue in scanpy?
You reported the base issue already half a year ago and since the release of numba 0.62.0 everyone who is using fresh environments is getting wrong results possibly with many people not even noticing it.
Something like the following for `smooth_knn_dist` (and basically the same for `compute_membership_strengths`) would do the trick:
```
import numba
import umap.umap_
original_smooth_knn_dist = umap.umap_.smooth_knn_dist
# Recompile the original function without fastmath but otherwise identical parameters
patched_smooth_knn_dist = numba.njit(
	locals={
		"psum": numba.types.float32,
		"lo": numba.types.float32,
		"mid": numba.types.float32,
		"hi": numba.types.float32,
	})(
	original_smooth_knn_dist.py_func if hasattr(original_smooth_knn_dist, 'py_func') else original_smooth_knn_dist
	)
umap.umap_.smooth_knn_dist = patched_smooth_knn_dist
```
To get closer to the old fastmath performance one can allow for some fastmath optimizations, as e.g. @swap357 mentioned in https://github.com/numba/numba/issues/10519#issuecomment-4184503920

flying-sheep:
Back then it was reasonable to assume that the fix would be treated with a sense of urgency on the umap side, but that doesn’t seem to be the case. So in retrospect, yes.

I think the best we can do at this point is send a patch to github.com/conda-forge/conda-forge-repodata-patches-feedstock that makes umap incompatible with newer numba versions.

simonwm:
@flying-sheep I did not know about that conda-forge-repodata-patches option! So this way one can patch dependencies of packages the feedstock of which one is not a maintainer of? That would solve the issue quickly and at the root and achieve a permanent fix for all past releases for everyone. Much better than monkey patching this in scanpy! Awesome!

flying-sheep:
yeah, it of course doesn’t fix PyPI or future conda-forge releases, so you should file an issue here too: https://github.com/conda-forge/umap-learn-feedstock

simonwm:
Future releases *should* not be the issue, I *hope* that the next version will be 0.6.0 and include the fix...
But pypi is important, indeed... Patching this only for half the user base is only half the fun...

PR #3824, opened 2025-10-01
feat: add n_obs_aggregated column to aggregated AnnData output

### LINKED ISSUE #3822: Output numer of observations aggregated from `scanpy.get.aggregate`

The current `aggregate` function does not output how many things were aggregated to get the aggregation i.e., cells into metacells, replicates in conditions, etc.

Offhand, since `by` can be a list, I would think we want at least two `obs` columns added:

1. `n_obs_aggregated` represents the total number of observations that have been aggregated into a given row of the returned `AnnData` object
2. `f"n_{by[i]}_aggregated"` for each`by` counts how many of a given subgroup are present if `by` is a list.  So if you only `aggregate` by cell type, for example, i.e., `by="cell_type"`, this column would not be present because it would not make sense - that value is simply `n_obs_aggregated`.  But if it were`by=["patient", "cell_type"]` then you would have `n_obs_aggregated` is the number of cells present in each patient-celltype row, `n_cell_type_aggregated` is the number of cells of that cell type present in the row, nad `n_patient_aggregated` is the number of patients

If we do point 2., we should better settle on the naming convention because it would basically represent a breaking change if we were to alter it down the line (until scanpy 2.0)

RohanDisa:
This PR adds a new column `n_obs_aggregated` to the output of `sc.get.aggregate`, reporting the total number of observations aggregated into each group. This helps users track how many cells or replicates contributed to each aggregated row.

- [x] Closes #3822 
- [x] Tests included (unit tests for single-key and multi-key groupings, as well as empty group handling)
- [x] Release notes added (via `hatch run towncrier:create`)

**Notes / Questions:**
- I have implemented `n_obs_aggregated` (point 1).
- Regarding point 2 about `n_{by[i]}_aggregated` — could you clarify with an example what these columns should contain? When grouping by `["patient", "cell_type"]`, wouldn't `n_patient_aggregated` and `n_cell_type_aggregated` always be 1?


ilan-gold:
Wow @RohanDisa thanks for taking this on!

>Regarding point 2 about n_{by[i]}_aggregated — could you clarify with an example what these columns should contain? When grouping by ["patient", "cell_type"], wouldn't n_patient_aggregated and n_cell_type_aggregated always be 1?

Hmmm yes the way I wrote that is not helpful at all.  It should have been more like `n_obs_per_patient_aggregated` and `n_obs_cell_type_aggregated` so if you have 4 cells aggregated in a patient-cell type group, but there are 10 cells in the patient (across cell types) and 12 in the cell type (across patients), the 8 would be `n_obs_aggregated`, the 12 would have be `n_obs_cell_type_aggregated` and the 10 `n_obs_per_patient_aggregated`.

The idea would be to give a way to get the fraction of cells aggregated in a given cell-patient combination.  Perhaps this is actually a better metric. 

So I think then `fraction_of_cells` and `fraction_of_patients` might be a more helpful metric here i.e., .33 and .4 in the above example.

However, I think it's fine to leave this out actually and only add `n_obs_aggregated`.

tests/test_aggregated.py
```diff
@@ -544,3 +544,38 @@ def test_factors():
...
+
+
+def test_aggregate_adds_n_obs_aggregated_single_key(pbmc_adata):
+    result = sc.get.aggregate(pbmc_adata, by="louvain", func="mean")
+    # Check column exists
+    assert "n_obs_aggregated" in result.obs
+    # Counts should be positive
+    assert (result.obs["n_obs_aggregated"] > 0).all()
+    # Total counts should equal original n_obs
+    assert result.obs["n_obs_aggregated"].sum() == pbmc_adata.n_obs
```
ilan-gold: I think this is wrong, you'd want the per-louvain group counts

tests/test_aggregated.py
```diff
@@ -544,3 +544,38 @@ def test_factors():
...
+def test_aggregate_adds_n_obs_aggregated_multiple_keys(pbmc_adata):
+    pbmc_adata.obs["percent_mito_binned"] = pd.cut(
+        pbmc_adata.obs["percent_mito"], bins=5
+    )
+    result = sc.get.aggregate(
+        pbmc_adata, by=["louvain", "percent_mito_binned"], func="mean"
+    )
+    assert "n_obs_aggregated" in result.obs
+    # Still sums back to the total number of obs
+    assert result.obs["n_obs_aggregated"].sum() == pbmc_adata.n_obs
```
ilan-gold: Same on the count check

tests/test_aggregated.py
```diff
@@ -544,3 +544,38 @@ def test_factors():
...
+
+
+def test_aggregate_adds_n_obs_aggregated_multiple_keys(pbmc_adata):
+    pbmc_adata.obs["percent_mito_binned"] = pd.cut(
+        pbmc_adata.obs["percent_mito"], bins=5
+    )
+    result = sc.get.aggregate(
+        pbmc_adata, by=["louvain", "percent_mito_binned"], func="mean"
+    )
```
ilan-gold: It doesn't make sense to aggregate `percent_mito_binned` because it is not a categorical

tests/test_aggregated.py
```diff
@@ -544,3 +544,38 @@ def test_factors():
...
     res = sc.get.aggregate(adata, by=["a", "b", "c", "d"], func="sum")
     np.testing.assert_equal(res.layers["sum"], adata.X)
+
+
+def test_aggregate_adds_n_obs_aggregated_single_key(pbmc_adata):
+    result = sc.get.aggregate(pbmc_adata, by="louvain", func="mean")
+    # Check column exists
+    assert "n_obs_aggregated" in result.obs
+    # Counts should be positive
+    assert (result.obs["n_obs_aggregated"] > 0).all()
```
ilan-gold: No need to check this given the below check

tests/test_aggregated.py
```diff
@@ -544,3 +544,38 @@ def test_factors():
...
+
+def test_aggregate_n_obs_aggregated_no_empty_groups(pbmc_adata):
+    # Force a categorical with unused categories
+    pbmc_adata.obs["fake_group"] = pd.Categorical(
+        ["A"] * (pbmc_adata.n_obs - 1) + ["B"], categories=["A", "B", "C"]
+    )
+    result = sc.get.aggregate(pbmc_adata, by="fake_group", func="mean")
+    assert "n_obs_aggregated" in result.obs
+    # Only groups with data should appear
+    assert set(result.obs["fake_group"]) == {"A", "B"}
```
ilan-gold: No need to check these

tests/test_aggregated.py
```diff
@@ -544,3 +544,38 @@ def test_factors():
...
+    # Count check
+    assert result.obs["n_obs_aggregated"].sum() == pbmc_adata.n_obs
```
ilan-gold: Same on the count check

ilan-gold (changes requested):

RohanDisa:
Hey @ilan-gold, Right now I’ve written tests using synthetic toy AnnData objects where the group counts are known exactly. This makes it easy to validate n_obs_aggregated.

Do you think I should also add a test using the small PBMC test dataset (pbmc3k_parametrized_small) to ensure things behave correctly on a real dataset, or are the current toy-data tests sufficient?

ilan-gold:
> Hey @ilan-gold, Right now I’ve written tests using synthetic toy AnnData objects where the group counts are known exactly. This makes it easy to validate n_obs_aggregated.

I think that's fine actually.

ilan-gold:
@RohanDisa sorry about the unhelpful test failures previously but now you should be getting helpful ones :) they look simple enough

RohanDisa:
Hey @ilan-gold,

I wanted to provide some context about the CI failures in the aggregation tests. The issue is caused by the recent addition of the n_obs_aggregated column in aggregate():

group_sizes = pd.Series(categorical).value_counts().reindex(new_label_df.index)
new_label_df["n_obs_aggregated"] = group_sizes.values


This adds an extra column to .obs, which is why existing aggregation tests are failing with shape mismatches, e.g.:

[left]: (3, 1) [right]: (3, 2)


The last two tests I added specifically check for n_obs_aggregated and pass, but older tests were written before this column existed and expect the previous .obs shape.

Suggested solution:

To avoid breaking existing tests while keeping n_obs_aggregated functionality:

Make the addition of n_obs_aggregated optional via a flag, e.g., add_n_obs_aggregated: bool = False in aggregate().

Only add the column when this flag is True.

Update my two new tests to call aggregate(add_n_obs_aggregated=True).

Existing tests will continue to pass without changes.

This approach ensures backward compatibility and keeps the CI green.

Happy to implement this if you agree with the approach.

Just to clarify the other CI failures:

Test step failing – This seems to be caused by environment setup or dependency issues in the Hatch environment. My PR does not touch test setup beyond aggregate(), so I don’t think this is caused by my changes.

Benchmark workflow failing (igraph missing) – The benchmark jobs fail because igraph is not installed. My PR doesn’t use igraph, so this failure is unrelated.

ilan-gold:
@flying-sheep is the addition of a column to an output `obs` `pandas.DataFrame` a breaking change?

RohanDisa:
Hi @ilan-gold, just wanted to check in to see if there’s been any update or if there’s anything I can help with to move this forward. Thanks!

ilan-gold:
@RohanDisa Still waiting for @flying-sheep to respond but in any case, the tests in https://github.com/scverse/scanpy/actions/runs/18199982179/job/51816237544?pr=3824 are failing so could you fix those? Even if we put this behind a flag, the tests will still be necessary

RohanDisa:
Hey @ilan-gold,
So, I updated the code. The reason for the only two issues are: 
1. The benchmark job encountered an ImportError because the igraph package is missing. 
2. The ReadTheDocs build failed because of multiple unresolved Sphinx references to typing.Union. With fail_on_warning: true enabled, these warnings are treated as errors, causing the documentation build to fail even though the docs compile successfully otherwise.

Do let me know if any changes are needed on my side.

RohanDisa:
Hey @ilan-gold, @flying-sheep,
Just checking in to see if there’s anything else needed on my end.
Would love to get this merged if it looks okay to you guys!

ilan-gold:
@RohanDisa I'd still like to hear from @flying-sheep on whether this should go in a minor version or not, but if we don't hear, i.e., does outputting a new column in `obs` constitute a "feature" that would go into a minor instead of patch (like adding a new kwarg or something).

I will look at the unrelated failures (hopefully) soon

flying-sheep:
I think it’s not a breaking change, but it *is* a feature and should therefore not go into a patch version. So I think the current milestone is correct!

src/scanpy/get/_aggregated.py
```diff
@@ -295,6 +295,10 @@ def aggregate(  # noqa: PLR0912
 
     dim_df = getattr(adata, axis_name)
     categorical, new_label_df = _combine_categories(dim_df, by)
+
+    # Add number of obs aggregated into each group
+    group_sizes = pd.Series(categorical).value_counts().reindex(new_label_df.index)
+    new_label_df["n_obs_aggregated"] = group_sizes.values
```
ilan-gold:
```suggestion
    new_label_df["n_obs_aggregated"] = pd.Series(categorical).value_counts().reindex(new_label_df.index)
```

no?

tests/test_aggregated.py
```diff
@@ -516,7 +528,10 @@ def test_aggregate_obsm_labels():
     )
 
     expected = ad.AnnData(
-        obs=pd.DataFrame({"labels": pd.Categorical(list("abc"))}, index=list("abc")),
+        obs=pd.DataFrame(
+            {"labels": pd.Categorical(list("abc")), "n_obs_aggregated": [5, 3, 4]},
+            index=list("abc"),
+        ),
```
ilan-gold, line 536: Construct from `label_counts` variable please!
```suggestion
        obs=pd.DataFrame(
            {"labels": pd.Categorical([lc[0] for lc in label_counts]), "n_obs_aggregated": [lc[1] for lc in label_counts]},
            index=[lc[0] for lc in label_counts],
        ),
```

tests/test_aggregated.py
```diff
@@ -544,3 +559,69 @@ def test_factors():
...
+
+    for cluster, batch, expected_count in expected_combinations:
+        mask = (result.obs["cluster"] == cluster) & (result.obs["batch"] == batch)
+        actual_count = result.obs.loc[mask, "n_obs_aggregated"].values[0]
+        assert actual_count == expected_count, (
+            f"Expected {cluster}+{batch} to have {expected_count} obs, got {actual_count}"
+        )
+
+    # Total should equal original n_obs
+    assert result.obs["n_obs_aggregated"].sum() == 12
```
ilan-gold: I think these tests are redundant with the other ones
```suggestion
```

ilan-gold (commented):
Nice, we're hoping to do a pre-release within the next two weeks so just a few small things

ilan-gold (approved):

PR #3851, opened 2025-10-29
fix: dotplot scaling with swapped axes

### LINKED ISSUE #3708: Dotplot scaling with swapped axes

### What happened?

Maybe I am missing something here, but I was expecting that using the argument `swap_axes` in `scanpy.plotting.rank_genes_dotplot` would simply swap the x and y axes. However, that does not seem to be the case: In the first plot, I used the default, i.e., `swap_axes=False`, but the mean GEX in the groups is very different when setting `swap_axes=True` in the second plot - at first I thought it was a matter of using `standard_scale="group"`, instead, but that is not the case.

Image

Image

### Minimal code sample

```python
sc.pl.rank_genes_groups_dotplot(
    adata,
    groupby="cell_type",
    standard_scale="var",
    # swap_axes=True,
)
```

### Error output

```pytb

```

### Versions

<details>

```
llvmlite        0.44.0
kiwisolver      1.4.8
natsort 8.4.0
wcwidth 0.2.13
sphinxcontrib-jsmath    1.0.1
numba   0.61.0
python-dateutil 2.9.0.post0
zarr    2.18.3
defusedxml      0.7.1
leidenalg       0.10.2
h5py    3.13.0
setuptools      75.8.0
pyarrow 19.0.1
PyYAML  6.0.2
pytz    2025.1
pillow  11.1.0
cycler  0.12.1
igraph  0.11.8
texttable       1.7.0
asciitree       0.3.3
joblib  1.4.2
charset-normalizer      3.4.1
six     1.17.0
numcodecs       0.13.1
psutil  7.0.0
----    ----
Python  3.10.16 | packaged by conda-forge | (main, Dec  5 2024, 14:20:01) [Clang 18.1.8 ]
OS      macOS-15.5-arm64-arm-64bit
CPU     14 logical CPU cores, arm
GPU     No GPU found
Updated 2025-07-11 19:34
```

</details>

ilan-gold:
Despite the title of the corresponding issue, I took 

> at first I thought it was a matter of using standard_scale="group"

to mean that scaling was not the problem.

But it turns that that is indeed broken and actually both scaling methods were broken with swapped axes for a very explainable reason - the scaling operation was happening on the wrong axis when the axis was swapped because it was fixed.  Group scaling always happened on axis 0, for example, even though the group could be on axis 1 when the axes are swapped.

If you run the test suite on main without the changes, there should be errors for only the two swapped axes tests.  Arguably the tests should be pared down, but parametrizing is free and easy.  Happy to remove the extra ones.

- [x] Closes #3708


flying-sheep (approved):
lovely!

PR #3853, opened 2025-10-29
fix: scanpy benchmark uses `pip` for `igraph` + `setup_cache`

ilan-gold:
This PR fixes in one fell swoop our benchmarking issues on CI

1. We started getting 403s in https://github.com/scverse/scanpy/actions/runs/18912251783/job/53985928561 which I would guess were from repeated requests (in any case, there are clearly multiple downloads happening if that's what the progress bars are for, and I think they are).  So we move to `setup_cache` like in `anndata`.  There are a lot fewer downloading bars: https://github.com/scverse/scanpy/actions/runs/18915072587/job/53996072887?pr=3853
2. We move to `pip` for `igraph`: https://github.com/airspeed-velocity/asv/issues/1554

Separately, it looks like the benchmarking machine is out of disk space.

- [ ] Closes # Current benchmark CI is broken

- [x] [Release notes][] not necessary because: only benchmarking

### DIFF
--- a/benchmarks/benchmarks/preprocessing_counts.py
+++ b/benchmarks/benchmarks/preprocessing_counts.py
@@ -55,17 +55,12 @@ class PreprocessingCountsSuite:  # noqa: D101
     def peakmem_scrublet(self, *_):
         sc.pp.scrublet(self.adata, batch_key=self.adata.uns["batch_key"])
 
+    def time_hvg_seurat_v3(self, *_):
+        # seurat v3 runs on counts
+        sc.pp.highly_variable_genes(self.adata, flavor="seurat_v3_paper")
 
-# Can’t do seurat v3 yet: https://github.com/conda-forge/scikit-misc-feedstock/issues/17
-"""
-def time_hvg_seurat_v3(self):
-    # seurat v3 runs on counts
-    sc.pp.highly_variable_genes(adata, flavor="seurat_v3_paper")
-
-
-def peakmem_hvg_seurat_v3(self):
-    sc.pp.highly_variable_genes(adata, flavor="seurat_v3_paper")
-"""
+    def peakmem_hvg_seurat_v3(self, *_):
+        sc.pp.highly_variable_genes(self.adata, flavor="seurat_v3_paper")
 
 
 class FastSuite:
--- a/benchmarks/benchmarks/preprocessing_counts.py
+++ b/benchmarks/benchmarks/preprocessing_counts.py
@@ -7,69 +7,63 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+import anndata as ad
+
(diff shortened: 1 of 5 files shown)


benchmarks/benchmarks/preprocessing_counts.py
```diff
@@ -7,69 +7,63 @@
...
-def peakmem_scrublet(*_):
-    sc.pp.scrublet(adata, batch_key=batch_key)
+    def peakmem_scrublet(self, *_):
+        sc.pp.scrublet(self.adata, batch_key=self.adata.uns["batch_key"])
 
 
 # Can’t do seurat v3 yet: https://github.com/conda-forge/scikit-misc-feedstock/issues/17
 """
-def time_hvg_seurat_v3(*_):
+def time_hvg_seurat_v3(self):
```
flying-sheep: is the `*_` not necessary? I thought ASV passed all the parameters as method arguments, and will break if the function doesn’t use them?

If not, we can remove all the `*_`.

benchmarks/benchmarks/preprocessing_log.py
```diff
@@ -7,70 +7,68 @@
...
-def peakmem_pca(*_):
-    sc.pp.pca(adata, svd_solver="arpack")
-
-
-def time_highly_variable_genes(*_):
-    # the default flavor runs on log-transformed data
-    sc.pp.highly_variable_genes(adata, min_mean=0.0125, max_mean=3, min_disp=0.5)
+class PreprocessingSuite:  # noqa: D101
+    params = params
+    param_names = param_names
```
flying-sheep, line 33: let’s move them to inline here too, no?

benchmarks/benchmarks/preprocessing_log.py
```diff
@@ -7,70 +7,68 @@
...
 
+    def setup_cache(self):
+        """Without this caching, asv was running several processes which meant the data was repeatedly downloaded."""
+        for dataset in self.params[0]:
+            for layer in self.params[1]:
```
flying-sheep:
```suggestion
        for dataset, layer in product(*self.params):
```

flying-sheep (changes requested):
Awesome! Some nitpicks regarding consistency, otherwise this is great!

I wonder why the downloading thing happened. We should change the way scanpy downloads things to use a lock to prevent this.

benchmarks/benchmarks/preprocessing_log.py
ilan-gold: I think we need this for `skip_when`: https://github.com/scverse/scanpy/pull/3853/files#diff-bce8bef3873bfd48adcf73cbf91eeb740af78fc67908840df2781584a9aa929bR27

flying-sheep (approved):

PR #3860, opened 2025-10-30
ci: fix tests in 1.11 branch

flying-sheep:
`as_sparse_dask_array` now returns `csr_array`s, which scanpy 1.11 doesn’t support.

This fixes us accidentally trying to use it.

The `anndata[tests-min]` part should also go into `main`


flying-sheep:
OK, I’m pretty convinced this is right

PR #3863, opened 2025-10-30
feat: add `obsm` parameter to `normalize_total` and `pca`

### LINKED ISSUE #3846: Allow using `obsm` instead of `X`/`layers` in more functions

I want to make a UMAP plot using a smoothed and corrected representation with fewer genes from adata.obsm as the base. 

I need to run the following preprocessing steps in order to be able to make the plot:
sc.pp.normalize_total
sc.pp.pca
sc.tl.neighbours
sc.tl.umap

The normalize_total and pca functions take a layer parameter. However this parameter can only be used to refer to layers in anndata.layers. It is also impossible to move an obsm annotation to adata.layers or anndata.X if the shape does not match anndata.X. It is furthermore impossible, to tell the function directly to use anndata.obsm[obsm_entry] as it only takes full anndata objects as arguments (view_to_actual will throw an error it doesn't receive a whole anndata object).

According to this post https://github.com/scverse/scanpy/issues/1301#issuecomment-654741567 it was at least possible at one point to pass layers to the pca function directly. 

The only way I have found to circumvent this issue is to create a temporary adata object with the obsm annotation as anndata.X, run the 2 functions on the temp object's anndata.X and port the created fields to the main anndata object. 

I think it would be beneficial if it was possible to specify layers in obsm for these functions (all that take the layer parameter) to use, or to modify the layer parameter to also allow taking layers from adata.obsm.

flying-sheep:
- [x] Closes #3846

The two functions in #3846 (plus the experimental version) seem to be the only valid candidates!

Everything else that (isn’t plotting and) has a `layer` parameter is gene expression specific (like `highly_variable_genes`)

### DIFF
--- a/src/scanpy/experimental/pp/_normalization.py
+++ b/src/scanpy/experimental/pp/_normalization.py
@@ -96,6 +96,7 @@ def normalize_pearson_residuals(
     clip: float | None = None,
     check_values: bool = True,
     layer: str | None = None,
+    obsm: str | None = None,
     inplace: bool = True,
     copy: bool = False,
 ) -> AnnData | dict[str, np.ndarray] | None:
@@ -138,8 +139,8 @@ def normalize_pearson_residuals(
         adata = adata.copy()
 
     view_to_actual(adata)
-    x = _get_obs_rep(adata, layer=layer)
-    computed_on = layer if layer else "adata.X"
+    x = _get_obs_rep(adata, layer=layer, obsm=obsm)
+    computed_on = layer or obsm or "adata.X"
 
     msg = f"computing analytic Pearson residuals on {computed_on}"
     start = logg.info(msg)
@@ -148,7 +149,7 @@ def normalize_pearson_residuals(
     settings_dict = dict(theta=theta, clip=clip, computed_on=computed_on)
 
     if inplace:
-        _set_obs_rep(adata, residuals, layer=layer)
+        _set_obs_rep(adata, residuals, layer=layer, obsm=obsm)
         adata.uns["pearson_residuals_normalization"] = settings_dict
     else:
         results_dict = dict(X=residuals, **settings_dict)
--- a/src/scanpy/preprocessing/_normalization.py
+++ b/src/scanpy/preprocessing/_normalization.py
@@ -302,7 +302,7 @@ def normalize_total(  # noqa: PLR0912
     if inplace:
         if key_added is not None:
             adata.obs[key_added] = dat["norm_factor"]
-        _set_obs_rep(adata, dat["X"], layer=layer)
+        _set_obs_rep(adata, dat["X"], layer=layer, obsm=obsm)
 
     logg.info(
         "    finished ({time_passed})",
--- a/src/scanpy/preprocessing/_normalization.py
+++ b/src/scanpy/preprocessing/_normalization.py
@@ -148,6 +148,7 @@ def normalize_total(  # noqa: PLR0912
     max_fraction: float = 0.05,
     key_added: str | None = None,
     layer: str | None = None,
+    obsm: str | None = None,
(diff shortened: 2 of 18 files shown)


src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -191,7 +194,9 @@ def pca(  # noqa: PLR0912, PLR0913, PLR0915
     `.obsm['X_pca' | key_added]` : :class:`~scipy.sparse.csr_matrix` | :class:`~scipy.sparse.csc_matrix` | :class:`~numpy.ndarray` (shape `(adata.n_obs, n_comps)`)
         PCA representation of data.
     `.varm['PCs' | key_added]` : :class:`~numpy.ndarray` (shape `(adata.n_vars, n_comps)`)
-        The principal components containing the loadings.
+        The principal components containing the loadings *when `obsm=None`*.
+    `.uns['pca' | key_added]['components']` : :class:`~numpy.ndarray` (shape `(adata.obsm][obsm].shape[1], n_comps)`)
+        The principal components containing the loadings *when `obsm="..."`*.
```
flying-sheep: I think this is an OK alternative location for `.varm['PCs' | key_added]` when that’s impossible? Pretty complex but what can one do …

One way to make it a tad more consistent would be to use `'PCs'` instead of `'components'`.

src/scanpy/preprocessing/_normalization.py
```diff
@@ -265,10 +268,7 @@ def normalize_total(  # noqa: PLR0912
 
     view_to_actual(adata)
 
-    x = _get_obs_rep(adata, layer=layer)
-    if x is None:
-        msg = f"Layer {layer!r} not found in adata."
-        raise ValueError(msg)
```
flying-sheep, line 271: This was useless as `_get_obs_rep` already throws an error.

src/scanpy/tools/_ingest.py
```diff
@@ -119,16 +116,6 @@ def ingest(
...
     """
-    # anndata version check
-    anndata_version = pkg_version("anndata")
-    if anndata_version < ANNDATA_MIN_VERSION:
-        msg = (
-            f"ingest only works correctly with anndata>={ANNDATA_MIN_VERSION} "
-            f"(you have {anndata_version}) as prior to {ANNDATA_MIN_VERSION}, "
-            "`AnnData.concatenate` did not concatenate `.obsm`."
-        )
-        raise ValueError(msg)
```
flying-sheep, line 130: this is unrelated, but since I removed another outdated version check I thought I might as well.

src/scanpy/experimental/pp/_normalization.py
```diff
@@ -138,8 +139,8 @@ def normalize_pearson_residuals(
         adata = adata.copy()
 
     view_to_actual(adata)
-    x = _get_obs_rep(adata, layer=layer)
-    computed_on = layer if layer else "adata.X"
+    x = _get_obs_rep(adata, layer=layer, obsm=obsm)
+    computed_on = layer or obsm or "adata.X"
```
ilan-gold, line 143: What if there are multiple values present? I see there's an assertion in https://github.com/scverse/scanpy/blob/57660240015d53f9d622894120bcf143429debcf/src/scanpy/get/get.py#L423 but maybe it should be a proper error?

ilan-gold (approved):

flying-sheep:
OK! Refactored `_get_obs_rep` for a nice error message!

Sorry for the unrelated `_utils` changes, I got fed up that I couldn’t auto-import `_empty`, and then I ended up not using `_empty`. Just ignore all the added `_` prefixes.

ilan-gold (approved):

PR #3868, opened 2025-11-03
ci: adapt tests to igraph 1.0

### LINKED ISSUE #3865: igraph 1.0 release breaks some plotting tests
The daily CI failed on test for scanpy failed.  Please go to [the logs of the integration testing repo](https://github.com/scverse/integration-testing/actions/runs/18989184060) to review. @scverse/anndata

flying-sheep:
- [x] Closes #3865

- [x] [Release notes][] not necessary because: tests only

I don’t think we can do anything about it. igraph was very stable, so I hope no rng changes come after this 1.0 change.


tests/test_clustering.py
```diff
@@ -67,7 +67,7 @@ def test_leiden_random_state(adata_neighbors, flavor):
     adata_2 = sc.tl.leiden(
         adata_neighbors,
         flavor=flavor,
-        random_state=2,
+        random_state=3,
```
flying-sheep, line 70: leiden results for `random_state=1` and `random_state=2` were the same, so I changed this.

I interpret the test as “random state affects things”, not that no two random states generate the same result.

ilan-gold (commented):
Aren't these changes `igraph` version specific? Hence why the CI is breaing on min version?

flying-sheep:
Oh yeah, I didn’t check if we lower-bound igraph.

I guess we just skip the paga tests in the `low-vers` tests? It’s not like we touched the paga code in recent years.

ilan-gold (approved):

PR #3870, opened 2025-11-03
chore: add `csc`-in-`dask` tests

ilan-gold:
TODO: why wasn't https://github.com/scverse/anndata/pull/2087 merged into the patch release branch?

- [ ] Closes #

- [x] [Release notes][] not necessary because: test change


src/testing/scanpy/_pytest/params.py
```diff
@@ -78,7 +82,28 @@ def wrapper(a: np.ndarray) -> DaskArray:
...
         )
-        for wrapper, suffix in [(lambda x: x, ""), (_chunked_1d, "-1d_chunked")]
+        for wrapper, suffix in [
+            (lambda x: x, ""),
+            *(
+                ((_chunked_1d, "-1d_chunked"),)
+                if Version(version("anndata")) < Version("0.12.5")
+                else (
+                    (
+                        lambda func,
```
flying-sheep: Please make this a function and revert all the `array_type.__name__`→`request.node.name` changes.

I’d rather not search for character sequences in the node name, there can be too much stuff going in there. Also then we won’t have 8 indentation levels and an ugly lambda definition here.

tests/test_pca.py
```diff
@@ -137,7 +147,10 @@ def possible_solvers(
             svd_solvers = {"auto", "full", "tsqr", "randomized", "covariance_eigh"}
         case (dc, False) if id == "dask_array_dense-1d_chunked":
             svd_solvers = {"tsqr", "randomized"}
-        case (dc, True) if id == "dask_array_sparse-1d_chunked":
+        case (dc, True) if (
+            "dask_array_sparse-1d_chunked-csr" in id
+            or id == "dask_array_sparse-1d_chunked"
```
flying-sheep, line 154: how can there be both?

flying-sheep (changes requested):
Great work!

Only issue I see is that all the `array_type.__name__` checking is already fragile enough, so let’s make it better or keep it instead of making it more fragile.

tests/test_pca.py
ilan-gold: Considered adding a comment here but I will.  The right is the old version, the left is the new version (i.e., https://github.com/scverse/scanpy/pull/3870/files/216b21d91312b899e939db9636d9ab20e7c29d77#diff-920788c9da92683bdfd3ced42f26fa537bdfcb53161f2b23dde77a80ac2c20c5 )

flying-sheep: And why is the old version is still handled here? Shouldn’t all tests now use the new version (with some skipping CSC and some not)?

src/testing/scanpy/_pytest/params.py
flying-sheep: you could e.g. make this a higher-kinded function (function factory) that sets `__name__` explicitly, then you don’t have to duplicate things.

ilan-gold: That's what I just did locally :) Running tests now.  I'm trying more to think through the things you comment here before saying "how?" blindly.  Glad to know that process is yielding results :)))))

tests/test_pca.py
ilan-gold: That's probably a cleaner way, especially given the above function generation

flying-sheep (approved):

PR #3872, opened 2025-11-03
feat: support `csc` in `dask` arrays in `get.aggregate`

### LINKED ISSUE #3861: `sc.get.aggregate` with CSC dask

See #3723

ilan-gold:
- [x] Closes #3861


src/scanpy/get/_aggregated.py
```diff
@@ -376,10 +373,13 @@ def aggregate_dask(
...
     dof: int = 1,
 ) -> dict[AggType, DaskArray]:
-    if not isinstance(data._meta, CSRBase | np.ndarray):
+    if not isinstance(data._meta, CSBase | np.ndarray):
         msg = f"Got {type(data._meta)} meta in DaskArray but only csr_matrix/csr_array and ndarray are supported."
         raise ValueError(msg)
-    if data.chunksize[1] != data.shape[1]:
+    # i.e., if  data._meta is CSR/np.ndarray, this is 1 because we use row-chunking, but otherwise 0 and column chunking
+    unchunked_axis = int(isinstance(data._meta, CSRBase | np.ndarray))
+    chunked_axis = (unchunked_axis - 1) % 2
```
flying-sheep: no need to use a comment if you just use simpler code

```suggestion
    chunked_axis, unchunked_axis = (0, 1) if isinstance(data._meta, CSRBase | np.ndarray) else (1, 0)
```

src/scanpy/get/_aggregated.py
```diff
@@ -388,35 +388,50 @@ def aggregate_chunk_sum_or_count_nonzero(
     ):
         # See https://docs.dask.org/en/stable/generated/dask.array.map_blocks.html
         # for what is contained in `block_info`.
-        subset = slice(*block_info[0]["array-location"][0])
-        by_subsetted = by[subset]
-        mask_subsetted = mask[subset] if mask is not None else mask
+        if chunked_axis == 0:
+            # only subset the mask and by if we need to i.e., there is chunking along the same axis as by and mask
```
flying-sheep: I think this makes the comments’ scopes a bit clearer?

/edit: sheesh, GitHub is bad at displaying suggestions on a diff that spans deleted lines.

```suggestion
        # only subset the `mask` and `by` if we need to,
        # i.e. when there is chunking along the same axis as by and mask
        if chunked_axis == 0:
            # See https://docs.dask.org/en/stable/generated/dask.array.map_blocks.html
            # for what is contained in `block_info`.
```

src/scanpy/get/_aggregated.py
```diff
@@ -388,35 +388,50 @@ def aggregate_chunk_sum_or_count_nonzero(
...
     funcs = set([func] if isinstance(func, str) else func)
     if "median" in funcs:
         msg = "Dask median calculation not supported.  If you want a median-of-medians calculation, please open an issue."
         raise NotImplementedError(msg)
     has_mean, has_var = (v in funcs for v in ["mean", "var"])
     funcs_no_var_or_mean = funcs - {"var", "mean"}
-    # aggregate each row chunk individually,
-    # producing a #chunks × #categories × #features array,
+    # aggregate each row chunk or column chunk individually,
+    # producing a #chunks × #categories × #features or a #categories x #chunks array,
```
flying-sheep:
```suggestion
    # producing a #chunks × #categories × #features or a #categories × #chunks array,
```

src/scanpy/get/_aggregated.py
```diff
@@ -388,35 +388,50 @@ def aggregate_chunk_sum_or_count_nonzero(
...
                 dtype=np.float64
                 if func not in get_args(ConstantDtypeAgg)
                 else data.dtype,  # TODO: figure out best dtype for aggs like sum where dtype can change from original
             ),
-        ).sum(axis=0)
+        )
         for f in funcs_no_var_or_mean
     }
+    # If we have row chunking, we need to handle the extra axis by summing over all category x feature matrices.
+    # Otherwise, dask internally concatenates the #categories x #chunks arrays i.e., the column chunks are concatenated together toget a #categories x #features matrix.
```
flying-sheep:
```suggestion
    # If we have row chunking, we need to handle the extra axis by summing over all category × feature matrices.
    # Otherwise, dask internally concatenates the #categories × #chunks arrays i.e., the column chunks are concatenated together to get a #categories × #features matrix.
```

src/scanpy/get/_aggregated.py
flying-sheep:
```suggestion
    # producing a #chunks × #categories × #features or a #categories × #chunks array,
```

src/scanpy/get/_aggregated.py
flying-sheep: I also don’t get this part:

> Otherwise, dask internally concatenates the #categories × #chunks arrays

dask doesn’t do things implicitly right, so when would that happen?

flying-sheep (changes requested):
As usual, just some stylistic stuff. Great work!

I’m not super happy about the added complexity, but I also don’t know if we could do it better:

e.g. it doesn’t look like we can pull out everything depending on `{un,}chunked_axis` and pass that in from a function that just exists to handle that complexity, things are too interwoven. What do you think?

flying-sheep (approved):

flying-sheep:
what do you say to what I wrote in https://github.com/scverse/scanpy/pull/3872#pullrequestreview-3442230541?

ilan-gold:
> what do you say to what I wrote in https://github.com/scverse/scanpy/pull/3872#pullrequestreview-3442230541?

Ah yeah, I meant to reply to that.  I can look into it.  Let's leave this PR as-is.  I'll open a new one into this one fiddling around with it.  If it looks to be too complex, I may just punt and say "FAU when we can"

PR #3874, opened 2025-11-03
chore: switch to Python 3.12–3.14

### LINKED ISSUE #3873: SPEC 0 update for scanpy 1.12 (Python 3.12–3.14)
See https://github.com/scverse/anndata/pull/2143

flying-sheep:
- [x] Closes #3873

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -26,9 +26,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
-        with:
-          fetch-depth: 0
-          # no blob:none so ASV can compare commits
+        with: { fetch-depth: 0 } # no blob:none so ASV can compare commits
 
       - name: Fetch main branch for `asv run`’s hash
         run: git fetch origin main:main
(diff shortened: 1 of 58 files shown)


ilan-gold (approved):
Test failure should be relatively easy to fix, nice!

flying-sheep:
thanks!

any idea what’s going on here? Dask bug? AnnData bug? b8bf58c4bd01f11c9260d7e79ea9396f300aa93d

before that commit, the `low-vers` job failed because `tests/test_qc_metrics.py::test_top_segments[dask_array_dense]` XPASSed.

that happened because the Dask array passed into `top_segments` had only 1 chunk, which made our shape check pass. but WHY has it only one chunk?

/edit: I see, anndata 0.10 didn’t chunk, so now I’m making sure the thing is chunked on the scanpy side.

ilan-gold:
> any idea what’s going on here? Dask bug? AnnData bug? https://github.com/scverse/scanpy/commit/b8bf58c4bd01f11c9260d7e79ea9396f300aa93d

https://github.com/scverse/anndata/blob/0.10.x/src/anndata/tests/helpers.py#L611-L615 is probably it.  Our minimum version has this line for dense which probably only makes one chunk.  Does that make sense? Now we have something different, so that accounts for why the other tests did not fail: https://github.com/scverse/anndata/blob/main/src/anndata/tests/helpers.py#L867-L872

flying-sheep:
I see, this was uncovered since all these tests were skipped with `anndata==0.9.*`!

ilan-gold:
Exactly :) So once you removed the marker, the "problem" revealed itself

PR #3563, opened 2025-04-04
sparse array support

flying-sheep:
- [x] Closes #3423
- [x] Tests included or not required because:

- [ ] Release notes not necessary because:

PR #3882, opened 2025-11-07
chore: fix #3881’s confusing naming

flying-sheep:
- [x] Release notes not necessary because: unreleased

otherwise we have a different naming schema from anndata.

we should also integrate this with the tests in https://github.com/scverse/scanpy/pull/3563, but not in this PR, since this one is going to be backported to the 1.11.x branch


ilan-gold (approved):

flying-sheep:
actually I don’t feel like backporting this, let’s just get out 1.12

Issue #1508, opened 2020-11-25
Saving plots results in a prefixed filename, regardless of sc.settings.figdir


rbpatt2019:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of scanpy.
- [ ] (optional) I have confirmed this bug exists on the master branch of scanpy.

---

As far as I can tell, `sc.pl.violin` always prepends the string "violin" to all save filed, regardless of `sc.settings.figdir`. This complicates programatically setting the filename for outputs. I've also confirmed this behaviour for `sc.pl.scatter`, and suspect it holds true for other `sc.pl` methods.

As an example: I keep all results from my analysis in a `results/` directory, with each step having its own subdirectory. Currently, this doesn't seem possible. If I set `sc.settings.figdir = ./results`, then try to save to, say "qcmetrics/total_counts.png", the code fails as "violinqcmetrics/total_counts.png" does not exist.

A partial work around is to set `sc.settings.figdir` every time I want to save in a different directory, but this feels quite clunky to me. And it will still result in all my files being prepended with "violin" (or "scatter", etc.), which I'd like to avoid.

Is there any way to disable this prefix on save behaviour?

### Minimal code sample (that we can copy&paste without having any data)

```python
sc.settings.figdir = "./"
sc.pl.violin(adata, ["total_counts"], save="results/test_plot.png")
```

```pytb
FileNotFoundError: [Errno 2] No such file or directory: 'violinresults/test_plot.png'
```

#### Versions

<details>

-----
anndata     0.7.5
scanpy      1.6.0
sinfo       0.3.1
-----
IPython             7.19.0
PIL                 8.0.1
anndata             0.7.5
backcall            0.2.0
cffi                1.14.4
colorama            0.4.4
cycler              0.10.0
cython_runtime      NA
dateutil            2.8.1
decorator           4.4.2
flufl               NA
get_version         2.1
h5py                3.1.0
igraph              0.8.3
ipython_genutils    0.2.0
jedi                0.17.2
joblib              0.17.0
kiwisolver          1.3.1
legacy_api_wrap     1.2
leidenalg           0.8.3
llvmlite            0.35.0rc3
matplotlib          3.3.3
mpl_toolkits        NA
natsort             7.1.0
numba               0.51.2
numexpr             2.7.1
numpy               1.19.4
packaging           20.4
pandas              1.1.4
parso               0.7.1
pexpect             4.8.0
pickleshare         0.7.5
[...]

PR #3888, opened 2025-11-10
docs: generate 1.12.0rc1 release notes

flying-sheep:
- [x] Release notes not necessary because: compiles release notes

### DIFF
--- a/docs/release-notes/1.12.0rc1.md
+++ b/docs/release-notes/1.12.0rc1.md
@@ -34,6 +34,6 @@
 #### Bug fixes
 
 - Raise fewer redundant warnings, mainly in {mod}`scanpy.pl` and {mod}`scanpy.datasets` functions {smaller}`P Angerer` ({pr}`3724`)
-- Fix `swap_axes` argument used with `standard_scale` in `scanpy.pl.dotplot` {smaller}`I Gold` ({pr}`3851`)
+- Fix `swap_axes` argument used with `standard_scale` in {func}`scanpy.pl.dotplot` {smaller}`I Gold` ({pr}`3851`)
 - Allow operating on {attr}`~anndata.AnnData.obsm` arrays in {func}`scanpy.pp.normalize_total` and {func}`scanpy.pp.pca` {smaller}`P Angerer` ({pr}`3863`)
 - Fix {func}`~scanpy.pl.dpt_groups_pseudotime` {smaller}`P Angerer` ({pr}`3864`)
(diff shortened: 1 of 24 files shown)


docs/release-notes/1.12.0rc1.md
```diff
@@ -0,0 +1,39 @@
...
+- Add support for {class}`scipy.sparse.csr_array` and {class}`scipy.sparse.csc_array` {smaller}`P Angerer` ({pr}`3563`)
+- Added a new `compressed` parameter to {func}`~scanpy.read_10x_mtx` to support reading uncompressed matrix files produced by tools like STARsolo {smaller}`T Kuthuru` ({pr}`3564`)
+- Make {func}`scanpy.get.aggregate` {doc}`dask:index` compatible with all aggregations except median {smaller}`I Gold` ({pr}`3700`)
+- Add `filter_unexpressed_genes` parameter to {func}`scanpy.pp.highly_variable_genes` {smaller}`M Mueller` ({pr}`3767`)
+- Add column `"n_obs_aggregated"` to {func}`scanpy.get.aggregate` output to show the total number of observations aggregated per group {smaller}`R Disa` ({pr}`3824`)
+- Add `method='jaccard'` for generating connectivities in {func}`scanpy.pp.neighbors` {smaller}`J Pintar` ({pr}`3831`)
+- Add in `csc`-in-{doc}`dask:index` support for {func}`scanpy.get.aggregate` {smaller}`I Gold` ({pr}`3872`)
+- Raised warnings now always point at user code, not internal scanpy code {smaller}`P Angerer` ({pr}`3876`)
+
+#### Miscellaneous changes
```
Zethson, line 23: Why isn't the deprecation of the `save` parameter anywhere? This isn't a huge change but the issue has 130+ upvotes: https://github.com/scverse/scanpy/issues/1508#issuecomment-963421037 . Therefore, I think it deserves a spot in the changelog and maybe even a suggestion for how to do it right. WDYT?

Edit: Ah, likely because it's just a FutureWarning and we haven't removed it yet. Fair enough!

flying-sheep: It should be in there, you’re right. Would be if you had added the release note fragment :wink:

Zethson: I saw  you add one but this might've been a copy & paste issue: https://github.com/scverse/scanpy/pull/3675/commits/cbf494341d92173fcdfc19b6b908d13818eccf61

flying-sheep: I updated the notebooks submodule, which resulted in both your change and Ilan’s to come in.

I thought that splitting these changes up would be silly, so I added the Ilan part of the news. Then forgot to add/ask for yours!

PR #3890, opened 2025-11-11
feat: add dotplot variant

VladimirShitov:
I recently noticed a major issue with scanpy that somehow persisted for a very long time. `scanpy.plotting` is missing a `dogplot`, a well-known best practice in data visualisation, [supported](https://github.com/mwaskom/seaborn/blob/7001ebe72423238e99c0434a2ef0a0ebc9cb55c1/seaborn/miscplot.py#L33-L34) e.g., in seaborn. This PR closes this issue and makes sure scanpy keeps providing the best tools (and mood) to its users.

- [ ] Closes #
- [x] [Tests][] included

- [x] [Release notes][] not necessary because this is an easter egg

Here's an example output:
image

Images are compressed and size is reduced, so each of them now takes half the storage space as [this scanpy logo](https://github.com/scverse/scanpy/blob/main/docs/_static/img/Scanpy_Logo_RGB.png). 

The path to images is inferred dynamically, but perhaps we can change it later to a github link, similarly to seaborn.

Oh, and here is the signed consent from the data contributor:
image

Thank you very much, I'd be glad to incorporate any suggestions. I acknowledge Helmholtz Munich for providing enough budget for this little piece of software (it was spent on cheese for the data contributor).

And huge thanks to Daria Romanovskaya @dariarom94, who transferred the data from the contributor.


flying-sheep:
Great job!

The scanpy logo doesn’t end up in the package at all, just in the docs. That being said, scanpy is currently a 2.1MB download, so I’ll think about this. Maybe I can save space more easily elsewhere. /edit: no, but read on

You have the original pics, right? Can you try using an image format that isn’t older than you? Matplotlib uses Pillow, which [supports](https://pillow-wiredfool.readthedocs.io/en/stable/handbook/image-file-formats.html) JPEG 2000 and WebP, both of which run laps around JPEG.

src/scanpy/plotting/_easter_egg.py
```diff
@@ -0,0 +1,18 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+
+HERE = Path(__file__).parent
+
+
+def dogplot(*_, **__) -> None:
+    """Shows who's a good boy"""
+    pic = np.random.randint(1, 4)
+    pic_path = HERE / "dogplot_images" / f"doggo_{pic}.jpg"
```
flying-sheep: please use [`importlib.resources`](https://docs.python.org/3/library/importlib.resources.html) instead, something like:

```py
with read_binary("scanpy.plotting", "dogplot_images", f"doggo_{pic}.jpg") as f:
    img = plt.imread(f)
```

VladimirShitov: Thanks! Did it, using `importlib.resources.files` though

VladimirShitov:
Thanks @flying-sheep ! I reduced the image size a bit and used the webp format. The images now take 17 KB in total. In my humble opinion, any user with a computer younger than me can survive this overhead in exchange for a smile

flying-sheep:
I changed the PR title for more sneakiness

PR #3896, opened 2025-11-14
docs: Add `cruft` config and update authors

Zethson:
- Adds `.cruft.json` to allow us to enable template sync
- Adds a few authors that were missing. I would also be happy if this section were downsized to the few 3-4 contributors that actually developed most of it. Up to you!

This could be multiple PRs. Let me know if you want them split up but I wanted to be pragmatic.
I've also only added Phil in the authors field as I didn't want to add all of them. It should be fine because this is often manually fixed anyways.


flying-sheep:
thanks!

PR #3898, opened 2025-11-18
feat: allow `use_graph` and `neighbors_key` for metrics

flying-sheep:
- [ ] Closes #

### DIFF
--- a/src/scanpy/metrics/_common.py
+++ b/src/scanpy/metrics/_common.py
@@ -9,6 +9,7 @@ import numpy as np
 import pandas as pd
 
 from .._compat import CSRBase, DaskArray, SpBase, fullname, warn
+from .._utils import NeighborsView
 
 if TYPE_CHECKING:
     from typing import NoReturn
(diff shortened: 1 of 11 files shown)


ilan-gold (commented):
Is the title correct for this PR? An issue for it? It looks good just confused, not sure I'm totally clear

src/scanpy/metrics/_common.py
```diff
@@ -65,17 +66,22 @@ def __call__(self) -> np.ndarray:
...
-        raise NotImplementedError()
-    # Fix for anndata<0.7
-    if hasattr(adata, "obsp") and "connectivities" in adata.obsp:
-        return adata.obsp["connectivities"]
-    elif "neighbors" in adata.uns:
-        return adata.uns["neighbors"]["connectivities"]
-    else:
+        if neighbors_key is not None:
+            msg = "Cannot specify both `use_graph` and `neighbors_key`."
+            raise TypeError(msg)
```
ilan-gold, line 78: If we can't provide both, why don't we set `use_graph` up for deprecation? Or is it already deprecated?

ilan-gold (approved):

src/scanpy/metrics/_common.py
flying-sheep: - `use_graph` raised `NotImplementedError` before this PR
- they have different purposes: `use_graph` is just a key in `obsp`. `neighbors_key` requires some stuff in `.uns` to work, but exists to interact with this feature (e.g. define multiple `neighbors_key`s and use them in a loop.)

PR #3949, opened 2026-01-21
chore: add speed note to `leiden` warning

### LINKED ISSUE #3911: `leiden`: warn about speed of `igraph` flavor being **much** faster than `leidenalg` flavor

For 1.12, eventually changing default (at some point).

We should also warn for speed

ilan-gold:
- [x] Closes #3911

- [x] [Release notes][] not necessary because: Just a warning


src/scanpy/tools/_leiden.py
```diff
@@ -143,6 +143,8 @@ def leiden(  # noqa: PLR0912, PLR0913, PLR0915
             msg = "Do not pass in partition_type argument when using igraph."
             raise ValueError(msg)
     else:
+        msg = "The `igraph` implementation of leiden clustering is *orders of magnitude faster*. Consider switching to it."
```
Zethson:
```suggestion
        msg = "The `igraph` implementation of leiden clustering is *orders of magnitude faster*. Set the flavor argument to 'igraph' to use it."
```

Something like this and maybe we should also mention that igraph needs to be installed.

pyproject.toml
```diff
@@ -196,7 +196,8 @@ filterwarnings = [
     # https://github.com/matplotlib/matplotlib/pull/30589
     "ignore:.*'(oneOf|parseString|resetCache|enablePackrat|leaveWhitespace|setName|setParseAction|endQuoteChar|unquoteResults)'.*'(one_of|parse_string|reset_cache|enable_packrat|leave_whitespace|set_name|set_parse_action|end_quote_char|unquote_results)':DeprecationWarning:matplotlib",
     "ignore:.*'(parseAll)'.*'(parse_all)':DeprecationWarning",
-
+    # igraph vs leidenalg warning
+    "ignore:The `igraph` implementation of leiden clustering:UserWarning",
```
flying-sheep, line 200: why not do this more locally?

flying-sheep (commented):
Hmm, two warnings when `flavor=None` … I think it’s OK, but maybe it would be more elegant to just add the text to the `FutureWarning` when raising that one, and only raise the UserWarning when specifying `flavor="leidenalg"` manually?

ilan-gold:
> Hmm, two warnings when flavor=None … I think it’s OK, but maybe it would be more elegant to just add the text to the FutureWarning when raising that one, and only raise the UserWarning when specifying flavor="leidenalg" manually?

But I think we want people definitely using `igraph` either way - if you don't specify anything and if you do, either way you should be using `igraph`.

flying-sheep:
That’s why I suggested

> maybe it would be more elegant to just **add the [new] text** to the FutureWarning when raising that one [and raising the new text as a standalone UserWarning otherwise]


that way, the user would get
- max 1 warning
- warning text that talks about future behehavior when relying on default behavior
- warning text that talks about performance when not using igraph

PR #3950, opened 2026-01-22
docs: add inline example of pca_variance_ratio

marvinbuff:
Part of #1664

Add inline example for `sc.pl.pca_variance_ratio()`.

Done with @t-a-m-i and @polychromech.


flying-sheep:
sorry I missed this, we merged #4056 instead!

Issue #3944, opened 2026-01-12
[RFC] Refactoring Scanpy's Numba Backend: C-Style Kernel Layer & Biology-Optimized Sparse Matrices


krkawzq:
### What kind of feature would you like to request?

Additional function parameters / changed functionality / changed defaults?

## 1. Motivation: Current State of Scanpy's Numba Backend

This RFC proposes a **gradual refactoring** of Scanpy's Numba backend to replace the current ad-hoc, poorly optimized JIT implementations with a structured, high-performance kernel layer.

### The Problem: Chaotic Numba Usage

The current Numba usage in Scanpy is **disorganized and inefficient**. A prime example is [`scanpy/preprocessing/_highly_variable_genes.py`](https://github.com/scverse/scanpy/blob/main/src/scanpy/preprocessing/_highly_variable_genes.py):

```python
# Current Scanpy pattern - problematic
@njit(cache=True, parallel=False)  # ← Parallelism DISABLED
def _sum_and_sum_squares_clipped(indices, data, n_cols, clip_val, nnz):
    squared_batch_counts_sum = np.zeros(n_cols, dtype=np.float64)
    batch_counts_sum = np.zeros(n_cols, dtype=np.float64)
    for i in range(nnz):
        val = data[i]
        if val > clip_val[indices[i]]:
            val = clip_val[indices[i]]
        squared_batch_counts_sum[indices[i]] += val ** 2  # ← Race condition!
        batch_counts_sum[indices[i]] += val               # ← Race condition!
    return squared_batch_counts_sum, batch_counts_sum
```

**Critical Issues:**

| Problem | Description |
|---------|-------------|
| **Race Conditions** | Parallelism disabled (`parallel=False`) because race conditions on shared arrays were not properly handled via reduction or atomics |
| **No Optimization Hints** | Missing `assume()`, `vectorize()`, `likely()` → LLVM cannot optimize aggressively |
| **Pythonic Style** | High-level constructs prevent SIMD vectorization and loop unrolling |
| **Tight Coupling** | Business logic (string methods, flavor dispatch) mixed into JIT functions → Object mode fallbacks |
| **Unverified Compilation** | No CI checks for optimal parallel patterns or Python C-API callbacks |

This pattern repeats across Scanpy's codebase, leaving **1-2 orders of magnitude of performance** on the table.

---

### 2.1 The Kernel Layer

[...]

PR #3952, opened 2026-01-23
perf: try out new cliping numba kernel

ilan-gold:
- [ ] Trying out kernel from https://github.com/scverse/scanpy/issues/3944#issuecomment-3756147264


Intron7:
@ilan-gold I tried my old kernel that failed two years ago. It works now as well and might even be a bit faster than the proposed kernels, likely due to better memory coalescing per thread. Numba must have done some work on the parallel execution engine.

PR #3953, opened 2026-01-26
feat: add Harmony to scanpy

### LINKED ISSUE #3940: `scanpy.external.pp.harmony_integrate` incompatible with newly updated `harmonypy>=0.1.0`

`scanpy.external.pp.harmony_integrate` assumes the old harmonypy output format and applies Z_corr.T.
However, since `harmonypy>=0.1.0`, Z_corr is already returned in cells × PCs format, so the transpose is no longer needed. This leads to shape mismatches or incorrect embeddings.

Intron7:
After recent changes from Harmonypy its necessary to move HarmonyBatchcorrection to scanpy. This function is based on harmony-pytorch and rapids-singlecell


flying-sheep (changes requested):
Looks good! I fixed the docs and made the code a bit nicer, but the reference tests have been failing from the start. Any idea why? You said they work locally for you?

Intron7:
so 0.05 for L2 was way to low harmony-pytorch is at 0.0511. we were at 0.0501. I also ran harmoypy once and it was at 0.063. so I set L2 to 0.1

src/scanpy/preprocessing/_harmony_integrate.py
```diff
@@ -0,0 +1,140 @@
...
+    from anndata import AnnData
+    from numpy.typing import DTypeLike
+
+
+def harmony_integrate(  # noqa: PLR0913
+    adata: AnnData,
+    key: str | Sequence[str],
+    *,
+    basis: str = "X_pca",
+    adjusted_basis: str = "X_pca_harmony",
```
flying-sheep, line 22: these will become refs in 2.0, but we might release harmony before that.

src/scanpy/preprocessing/_harmony_integrate.py
```diff
@@ -0,0 +1,140 @@
...
+    sigma: float = 0.1,
+    n_clusters: int | None = None,
+    max_iter_harmony: int = 10,
+    max_iter_clustering: int = 200,
+    tol_harmony: float = 1e-4,
+    tol_clustering: float = 1e-5,
+    ridge_lambda: float = 1.0,
+    correction_method: Literal["fast", "original"] = "original",
+    block_proportion: float = 0.05,
+    tau: int = 0,
```
flying-sheep: theta, lambda, sigma, tau … we should probably
1. rename these to make sense
2. fix the docs:

  - `tau` “Discounting factor” – this one has a great description about what it does, let’s do that for the rest! e.g.
  - “Diversity penalty weight … default 2 for each batch variable” – “diversity” might be a good component of the param name, but what is it and what does “2” mean compared to smaller or bigger values?
  - “ridge regression regularization … default 1” – what does increasing and decreasing do, respectively?

src/scanpy/preprocessing/_harmony_integrate.py
```diff
@@ -0,0 +1,140 @@
...
+    random_state: int | None = 0,
```
flying-sheep, line 38: will change in #3983 (very soon, so we should probably wait for that PR)

tests/test_harmony.py
```diff
@@ -0,0 +1,166 @@
...
+def adata_reference() -> AnnData:
+    """Load reference data from harmonypy repository."""
+    paths = {
+        f: pooch.retrieve(
+            f"https://github.com/slowkow/harmonypy/raw/refs/heads/master/data/{name}",
+            known_hash=hash_,
+        )
+        for f, (name, hash_) in DATA.items()
+    }
+    dfs = {f: pd.read_csv(path, delimiter="\t") for f, path in paths.items()}
```
flying-sheep, line 64: CSV ugh. maybe do the `adata…_session` + `adata…` pattern so we copy it in memory instead of reading it from disk multiple times?

src/scanpy/preprocessing/_harmony_integrate.py
flying-sheep, line 1: shouldn’t this file go into `./_harmony/`?

flying-sheep (commented):
looks good! Some API considerations.

Please
- add parameter descriptions and
- rename the parameters (probably a good idea, this is how we do things in scanpy, but we might have to add a wrapper in .external for this)
- add a towncrier fragment

src/scanpy/preprocessing/_harmony_integrate.py
flying-sheep: OK, PR is merged, new APIs shouldn’t have `random_state` anymore at all

flying-sheep: fixed, let’s see if the tests still pass

flying-sheep:
OK, I got rid of `random_state`. Let’s see if the tests are robust against different k-means clustering.

Intron7:
> OK, I got rid of `random_state`. Let’s see if the tests are robust against different k-means clustering.

that is not a good way to do this. this test was designed to test if the math in fast and original is the same im going to revert it. We test the Kmeans in the overall test because those number were created with R

flying-sheep (approved):
OK, looks good! Anything else you want to change before we merge it?

Intron7:
@flying-sheep we can merge

src/scanpy/preprocessing/_harmony/core.py
```diff
@@ -0,0 +1,609 @@
...
+            row_sums = np.maximum(row_sums, 1e-12)
+            r_new /= row_sums
+
+            # Store back
+            r[cell_idx, :] = r_new
+
+            # Add new r contribution to o and e
+            r_new_sum = r_new.sum(axis=0)
+            o[b, :] += r_new_sum
+            e += pr_b * r_new_sum
```
ilan-gold, line 419: Can this be vectorized somehow over `b` so that we don't have to do `product(...)`? Maybe `scanpy.get.aggregate` for handling the summations by group? Suspicious of a big for loop like this...

Intron7: No, this can't be vectorized. It uses a Gauss-Seidel pattern to speed up convergence — each batch's update informs the next within a block. I tried vectorizing this in rsc and it didn't work (convergence breaks).

ilan-gold: Wow very interesting.  Yeah, I tried claude-coding a vectorized version over the weekend but it does change the intermediate penalty, right? Interestingly the tests passed.

Maybe we could mark here for which dataset the vectorized version fails on?

Intron7: This would change the algorithm and would lead to later or no convergance we should not do this. Easy datasets will still converage just slower and might not

ilan-gold:
@Intron7 In the paper they claim being able to run 1million cells / 800 batches in like 1 minute or something.  Should we expect the same performance here?

src/scanpy/preprocessing/_harmony/core.py
ilan-gold: Nice ok, makes sense!

AlbertoFabbri93:
Is this **not** merged in Scanpy 1.12.2?

ilan-gold:
No, new features come on the minor release.  We will have a release candidate soon, though. Sorry :/

dawe:
@ilan-gold It is not even in 1.12.3? I mean, I've just conda-installed it and harmony doesn't seem to be under sc.pp. It is in sc.external.pp, still suffering of the "transpose" issue #3940

Intron7:
it will be in 1.13

PR #3954, opened 2026-01-26
fix: regress_out for more threads than 64

Intron7:
Fix a potential bug for regress out where more than 64 threads leads to a hard crash


ilan-gold (commented):
Can you add a release note explaining this a bit? Maybe a link to something like how you discovered this?

Also, it this something we should be doing globally?

Also, what if this crshes before you reset?

flying-sheep:
@Intron7 could you please do that? I don’t want to have magic numbers without explanation in the code base.

PR #3966, opened 2026-02-05
ci: fix and deduplicate rank_gene_groups tests

### LINKED ISSUE #3965: Integration Testing CI Test Failure on python 3.13 with prerelease dependencies
The daily CI failed on test for scanpy failed.  Please go to [the logs of the integration testing repo](https://github.com/scverse/integration-testing/actions/runs/21695679673) to review. @scverse/anndata

flying-sheep:
- [x] Closes #3965

- [x] [Release notes][] not necessary because: tests only

This gets rid of potential numpy warnings by avoiding pickles in our test files.

What I did: 
1. converted test results from recarrays to regular arrays, and converted the “names” arrays to `np.uint8`s (manually checked that everything worked, it’s 40 numbers × 2 test methods
2. saved to `npz` with `allow_pickle=False`, which is now possible, and deleted the old files
3. restructured the tests without changing the logic. no clue why `[:7]` is done for wilcoxon in the first test and for both in the second 🤷


src/scanpy/metrics/_metrics.py
```diff
@@ -79,7 +79,8 @@ def confusion_matrix(
     mtx = _confusion_matrix(orig, new, labels=unique_labels)
     if normalize:
         sums = mtx.sum(axis=1)[:, np.newaxis]
-        mtx = np.divide(mtx, sums, where=sums != 0)
+        mtx = mtx.astype(np.float64)
+        np.divide(mtx, sums, where=sums != 0, out=mtx)
```
ilan-gold, line 83: `expect unitialized memory in output` i.e., from the warning in the "failing" test, I would think `out=` should have the result of `np.empty`, not `mtx`.  Or am I misinterpreting?

flying-sheep: After my change, what we do is to leave all values in `mtx` where `sums==0` as they are (and divide the others):

```py
mtx[sums != 0] /= sums
```

Before my change we replaced all values in `mtx` where `sums==0` with uninitialized memory (and divide the others):

```py
mtx_new = np.empty(mtx.shape)
mtx[sums != 0] = mtx_new / sums
```

ilan-gold: Oh wow! "If not provided or None, a freshly-allocated array is returned." from the numpy docs on `divide` - I didn't realize this was default behavior to return uninitialized memory. Ok then!

ilan-gold (approved):

src/scanpy/metrics/_metrics.py
flying-sheep: Yeah, some nuts behavior. Good they have that warning now.

PR #3983, opened 2026-02-23
feat: support np.random.Generator

### LINKED ISSUE #1131: Passing a RandomState instance can cause failures to save
```python
import scanpy as sc
import numpy as np

sc.tl.umap(pbmc, random_state=np.random.RandomState(10))
pbmc.write("tmp.h5ad")
```

```pytb
NotImplementedError: Failed to write value for uns/umap/params/random_state, since a writer for type <class 'numpy.random.mtrand.RandomState'> has not been implemented yet.

Above error raised while writing key 'uns/umap/params/random_state' of <class 'h5py._hl.files.File'> from /.
```

It looks like we'll have to not write this, or figure out how to represent it on disk.

<details>
<summary> Full traceback </summary>

```pytb
NotImplementedError                       Traceback (most recent call last)
~/github/anndata/anndata/_io/utils.py in func_wrapper(elem, key, val, *args, **kwargs)
    187         try:
--> 188             return func(elem, key, val, *args, **kwargs)
    189         except Exception as e:

~/github/anndata/anndata/_io/h5ad.py in write_not_implemented(f, key, value, dataset_kwargs)
    144     raise NotImplementedError(
--> 145         f"Failed to write value for {key}, "
    146         f"since a writer for type {type(value)} has not been implemented yet."

NotImplementedError: Failed to write value for uns/umap/params/random_state, since a writer for type <class 'numpy.random.mtrand.RandomState'> has not been implemented yet.

The above exception was the direct cause of the following exception:

NotImplementedError                       Traceback (most recent call last)
<ipython-input-2-1dd6b1c7e996> in <module>
      4 pbmc = sc.datasets.pbmc68k_reduced()
      5 sc.tl.umap(pbmc, random_state=np.random.RandomState(10))
----> 6 pbmc.write("tmp.h5ad")

~/github/anndata/anndata/_core/anndata.py in write_h5ad(self, filename, compression, compression_opts, force_dense, as_dense)
   1988             compression_opts=compression_opts,
   1989             force_dense=force_dense,
-> 1990             as_dense=as_dense,
   1991         )
   1992 

~/github/anndata/anndata/_io/h5ad.py in write_h5ad(filepath, adata, force_dense, as_dense, dataset_kwargs, **kwargs)
    110         write_attribute(f, "varp", adata.varp, dataset_kwargs=dataset_kwargs)
    111         write_attribute(f, "layers", adata.layers, dataset_kwargs=dataset_kwargs)
--> 112         write_attribute(f, "uns", adata.uns, dataset_kwargs=dataset_kwargs)
    113     if adata.isbacked:
    114         adata.file.open(filepath, "r+")

/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/functools.py in wrapper(*args, **kw)
    838                             '1 positional argument')
    839 
--> 840         return dispatch(args[0].__class__)(*args, **kw)
    841 
    842     funcname = getattr(func, '__name__', 'singledispatch function')

~/github/anndata/anndata/_io/h5ad.py in write_attribute_h5ad(f, key, value, *args, **kwargs)
    124     if key in f:
    125         del f[key]
--> 126     _write_method(type(value))(f, key, value, *args, **kwargs)
    127 
    128 

~/github/anndata/anndata/_io/h5ad.py in write_mapping(f, key, value, dataset_kwargs)
    284 def write_mapping(f, key, value, dataset_kwargs=MappingProxyType({})):
    285     for sub_key, sub_value in value.items():
--> 286         write_attribute(f, f"{key}/{sub_key}", sub_value, dataset_kwargs=dataset_kwargs)
    287 
    288 

/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/functools.py in wrapper(*args, **kw)
    838                             '1 positional argument')
    839 
--> 840         return dispatch(args[0].__class__)(*args, **kw)
    841 
    842     funcname = getattr(func, '__name__', 'singledispatch function')

~/github/anndata/anndata/_io/h5ad.py in write_attribute_h5ad(f, key, value, *args, **kwargs)
    124     if key in f:
    125         del f[key]
--> 126     _write_method(type(value))(f, key, value, *args, **kwargs)
    127 
    128 

~/github/anndata/anndata/_io/h5ad.py in write_mapping(f, key, value, dataset_kwargs)
    284 def write_mapping(f, key, value, dataset_kwargs=MappingProxyType({})):
    285     for sub_key, sub_value in value.items():
--> 286         write_attribute(f, f"{key}/{sub_key}", sub_value, dataset_kwargs=dataset_kwargs)
    287 
    288 

/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/functools.py in wrapper(*args, **kw)
    838                             '1 positional argument')
    839 
--> 840         return dispatch(args[0].__class__)(*args, **kw)
    841 
    842     funcname = getattr(func, '__name__', 'singledispatch function')

~/github/anndata/anndata/_io/h5ad.py in write_attribute_h5ad(f, key, value, *args, **kwargs)
    124     if key in f:
    125         del f[key]
--> 126     _write_method(type(value))(f, key, value, *args, **kwargs)
    127 
    128 

~/github/anndata/anndata/_io/h5ad.py in write_mapping(f, key, value, dataset_kwargs)
    284 def write_mapping(f, key, value, dataset_kwargs=MappingProxyType({})):
    285     for sub_key, sub_value in value.items():
--> 286         write_attribute(f, f"{key}/{sub_key}", sub_value, dataset_kwargs=dataset_kwargs)
    287 
    288 

/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/functools.py in wrapper(*args, **kw)
    838                             '1 positional argument')
    839 
--> 840         return dispatch(args[0].__class__)(*args, **kw)
    841 
    842     funcname = getattr(func, '__name__', 'singledispatch function')

~/github/anndata/anndata/_io/h5ad.py in write_attribute_h5ad(f, key, value, *args, **kwargs)
    124     if key in f:
    125         del f[key]
--> 126     _write_method(type(value))(f, key, value, *args, **kwargs)
    127 
    128 

~/github/anndata/anndata/_io/utils.py in func_wrapper(elem, key, val, *args, **kwargs)
    193                 f"Above error raised while writing key {key!r} of {type(elem)}"
    194                 f" from {parent}."
--> 195             ) from e
    196 
    197     return func_wrapper

NotImplementedError: Failed to write value for uns/umap/params/random_state, since a writer for type <class 'numpy.random.mtrand.RandomState'> has not been implemented yet.

Above error raised while writing key 'uns/umap/params/random_state' of <class 'h5py._hl.files.File'> from /.
```

</details>

### LINKED ISSUE #3371: Switch to `numpy.random.Generator`
Plan: [SPEC7](https://scientific-python.org/specs/spec-0007/) (switch to `rng: np.random.Generator | int` args, phase out `random_state: RandomState | int`)

- [rationale](https://github.com/scikit-learn/scikit-learn/issues/29315#issuecomment-2203877889)
- further context: https://github.com/scverse/scanpy/issues/2969#issuecomment-2408719632

Implementation:

- [x] #3372 
- [x] Implement transition tools and manually transistion in #943
- [ ] Decorator for functions that transition

flying-sheep:
- [x] Closes #3371

The idea is to make our code behave as closely as possible to how it did, but make it read like modern code:

1. add decorator that converts `random_state` into `rng` argument (deduplicating them and using 0 by default)
2. add helpers that allow old behavior (e.g. for APIs that only take `RandomState`) 
   - `_FakeRandomGen.wrap_global` and `_if_legacy_apply_global` to conditionally replace `np.random.seed` and other global-state-mutating functions
   - `_legacy_random_state` to get back a `random_state` argument for APIs that don’t take `rng`, e.g. `scikit-learn` stuff and `"random_state"` in `adata.uns[thing]["params"]`
3. after this PR: make #3653 change the default behavior when a preset is passed.

I also didn’t convert the `transformer` argument to `neighbors` (yet?), or deprecated stuff like `louvain` or the `external` APIs.

## Reviewing

First a short info abut how Generators work:

- they can `spawn` independent children (doing so advances their internal state once)
- all their other methods advance their internal state
- they are reproducible in the same environment (when initialized with the same seed of course) but make no reproducibility guarantee across versions
- the convention is to use `rng: SeedLike | RNGLike | None = None` for the argument, `None` meaning random initialization

Now questions to the reviewers:

- Should we store the new RNG in `adata.uns`? If no, this fixes #1131
- Should we keep `random_state` in the docs?
- How should we annotate that the default `rng` isn’t actually `None` but “a new instance of `_LegacyRandom(0)`” but people can pass `rng=None` to get the future default behavior?
- Should I handle passing `rng` to neighbors transformer?
- Did I miss other spots where `rng` can be passed?
- Did I miss any spots where we called `np.random.seed()`

## TODO:

- [x] add the decorator
- [x] add helpers for restarting (e.g. if 0 was passed, it’d be reused by the functions called in function body)
- [x] for the functions that called it: re-add the `np.random.seed` calls (maybe `if isinstance(rng, _FakeRandomGen): gen = legacy_numpy_gen(rng)` or so?)
  - [x] partially done, finish the work
- [x] add `spawn` to `_FakeRandomGen` (that does nothing) and use `spawn` for a tree structure
  - [x] https://github.com/numpy/numpy/issues/24086#issuecomment-3967486531
- [x] ingest


src/scanpy/preprocessing/_pca/_compat.py
```diff
@@ -18,21 +19,28 @@
     from sklearn.decomposition import PCA
 
     from ..._compat import CSBase
-    from ..._utils.random import _LegacyRandom
+    from ..._utils.random import RNGLike, SeedLike
 
 
+SCIPY_1_15 = pkg_version("scikit-learn") >= Version("1.5.0rc1")
```
selmanozleyen: did you mean SKLEARN here?

src/scanpy/preprocessing/_scrublet/sparse_utils.py
```diff
@@ -48,11 +46,10 @@ def subsample_counts(
...
-    random_seed: _LegacyRandom = 0,
+    rng: np.random.Generator,
 ) -> tuple[CSBase, NDArray[np.int64]]:
     if rate < 1:
-        random_seed = check_random_state(random_seed)
-        e.data = random_seed.binomial(np.round(e.data).astype(int), rate)
+        e.data = rng.binomial(np.round(e.data).astype(int), rate)
         current_totals = np.asarray(e.sum(1)).squeeze()
         unsampled_orig_totals = original_totals - current_totals
         unsampled_downsamp_totals = np.random.binomial(
```
selmanozleyen: I think you missed this line

selmanozleyen: maybe we can add this to the reprod tests?

src/scanpy/preprocessing/_pca/_compat.py
flying-sheep: great catch, no! `svds` comes from scipy indeed.

src/scanpy/preprocessing/_scrublet/sparse_utils.py
flying-sheep: another great catch, thank you!

src/scanpy/tools/_umap.py
```diff
@@ -170,22 +182,18 @@ def umap(  # noqa: PLR0913, PLR0915
 
     if a is None or b is None:
         a, b = find_ab_params(spread, min_dist)
-    adata.uns[key_uns] = dict(params=dict(a=a, b=b))
+    adata.uns[key_uns] = dict(params=dict(a=a, b=b, **meta_random_state))
```
selmanozleyen, line 182: can't meta_random_state be None?

src/scanpy/neighbors/__init__.py
```diff
@@ -844,10 +856,9 @@ def compute_eigen(
         self,
         *,
         n_comps: int = 15,
-        sym: bool | None = None,
```
selmanozleyen, line 847: this is a breaking change right? also docstring is not updated with it

selmanozleyen: plus asking out of curiosity. I'd imagine writing the implementations  `__init__.py`  wouldn't be good practice but maybe it's not the case so I'd like to know.

src/scanpy/preprocessing/_simple.py
```diff
@@ -1017,143 +1022,137 @@ def downsample_counts(
...
     geneptr = 0
     for count in sample:
         while count >= cumcounts[geneptr]:
             geneptr += 1
         col[geneptr] += 1
     return col
+
+
+@dataclass
+class sparse_as_csr[T: (np.ndarray | CSBase)](AbstractContextManager):  # noqa: N801
```
selmanozleyen, line 1143: why not `np.ndarray | CSBase` ?

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -241,21 +246,23 @@ def pca(  # noqa: PLR0912, PLR0913, PLR0915
...
         raise NotImplementedError(msg)
 
-    # check_random_state returns a numpy RandomState when passed an int but
     # dask needs an int for random state
-    if not isinstance(x, DaskArray):
-        random_state = check_random_state(random_state)
-    elif not isinstance(random_state, int):
-        msg = f"random_state needs to be an int, not a {type(random_state).__name__} when passing a dask array"
+    rng = np.random.default_rng(rng)
+    if not isinstance(rng, _FakeRandomGen) and not isinstance(
```
selmanozleyen: this throws 
```
AttributeError: 'numpy.random._generator.Generator' object has no attribute '_arg'
```


with `sc.pp.pca(adata, rng=np.random.default_rng(42))`

src/scanpy/tools/_umap.py
flying-sheep: yup, just found that out while updating the tests: should have been `{}` like everywhere else.

src/scanpy/neighbors/__init__.py
flying-sheep: Right, I must have forgot that this is a public API.

It does nothing, but that doesn’t prevent people from placebo-specifying it I guess.

src/scanpy/preprocessing/_simple.py
flying-sheep: The way this is now means “either exactly one or exactly the other” which means the type checker doesn’t get confused by the type guards here.

I guess `|` could mean “any subtype, incl. a subclass of both”?

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: huh, I was sure I fixed this before, but maybe that existed in multiple places. Should be `or`.

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -241,21 +246,23 @@ def pca(  # noqa: PLR0912, PLR0913, PLR0915
...
-        random_state = check_random_state(random_state)
-    elif not isinstance(random_state, int):
-        msg = f"random_state needs to be an int, not a {type(random_state).__name__} when passing a dask array"
+    rng = np.random.default_rng(rng)
+    if not isinstance(rng, _FakeRandomGen) or not isinstance(
+        rng._arg, int | np.random.RandomState
+    ):
+        # TODO: remove this error and if we don’t have a _FakeRandomGen,
+        #       just use rng.integers to make a seed farther down
+        msg = f"rng needs to be an int or a np.random.RandomState, not a {type(rng).__name__} when passing a dask array"
```
flying-sheep: reminder to self: do this TODO

src/scanpy/preprocessing/_simple.py
```diff
@@ -1017,143 +1022,137 @@ def downsample_counts(
...
+        rows = np.split(x.data, x.indptr[1:-1]) if isinstance(x, CSRBase) else x
         totals = stats.sum(x, axis=1)
-        under_target = np.nonzero(totals > counts_per_cell)[0]
-        for rowidx in under_target:
-            row = x[rowidx, :]
+        under_target = np.flatnonzero(totals > counts_per_cell)
+        for rowidx, sub_rng in zip(
+            under_target, rng.spawn(len(under_target)), strict=True
+        ):
             _downsample_array(
```
selmanozleyen: I think spawn is more for parallel independent streams. This is correct for reproducibility but it might be causing performance overhead for independence that is usually not needed. So far I saw usage of spawn in either when rng is being given to another function while having keeping the rng still independent in the current code flow or in parallel cases. Not in a sequential for loop. My argument goes like this: If this for loop was unrolled and the code was written sequentially would we still use `spawn`. I think no because in sparse multiply we don't do that for example.

In fact I ran an ai generated script and these were the results. For some reason `FakeRandomGen` is super slow. But also shows that spawn actually has an overhead.

```
python bench_downsample.py
Warming up numba JIT...
Done.

counts_per_cell=50  n_obs=500 n_vars=200 nnz=29840
  legacy (random_state=0)                20.07 ms
  real Generator (with spawn)             8.30 ms
  real Generator (no spawn)               4.53 ms
  --> spawn overhead: +83%   real vs legacy: -59%

counts_per_cell=100  n_obs=5000 n_vars=500 nnz=746289
  legacy (random_state=0)               399.50 ms
  real Generator (with spawn)            86.80 ms
  real Generator (no spawn)              53.74 ms
  --> spawn overhead: +62%   real vs legacy: -78%

counts_per_cell=50  n_obs=20000 n_vars=1000 nnz=5969861
  legacy (random_state=0)              2968.35 ms
  real Generator (with spawn)           320.06 ms
  real Generator (no spawn)             185.98 ms
  --> spawn overhead: +72%   real vs legacy: -89%
❯ python bench_downsample.py
Warming up numba JIT...
Done.

counts_per_cell=50  n_obs=500 n_vars=200 nnz=29840
  legacy (_FakeRandomGen)                18.96 ms
  real Generator (with spawn)             7.77 ms
  real Generator (no spawn)               4.22 ms
  --> spawn overhead: +84%   real vs legacy: -59%

counts_per_cell=100  n_obs=5000 n_vars=500 nnz=746289
  legacy (_FakeRandomGen)               412.81 ms
  real Generator (with spawn)            88.76 ms
  real Generator (no spawn)              54.64 ms
  --> spawn overhead: +62%   real vs legacy: -78%

counts_per_cell=50  n_obs=20000 n_vars=1000 nnz=5969861
  legacy (_FakeRandomGen)              2956.91 ms
  real Generator (with spawn)           322.48 ms
  real Generator (no spawn)             187.00 ms
  --> spawn overhead: +72%   real vs legacy: -89%
 ~/p/scanpy │ pa/rng ⇣1 *1 !2 ?2                ✔ │ 28s │ scanpy Py │ 16:30:20 
```

the script: https://gist.github.com/selmanozleyen/2046c849bb10f541642cea2ac7daa0db

ilan-gold:
> they can spawn independent children (doing so advances their internal state once)

Is this True? I was looking into your comment and found this:

```python
# High quality entropy created with: f"0x{secrets.randbits(128):x}"
import numpy as np
entropy = 0x3034c61a9ae04ff8cb62ab8ec2c4b501
rng = np.random.default_rng(entropy)
# Generate a random number as the first step
first_random = rng.uniform()
rng = np.random.default_rng(entropy)
# Now, spawn as the first step
child_rng1, child_rng2 = rng.spawn(2)
# And then do RNG
assert first_random == rng.uniform()
```

Also, reading through the linked thread, it seems the consensus was `spawn` does _not_ advance internal state, although maybe I'm misunderstanding what you mean by "internal state" (you even say in your comment "but this means that when users pass a Generator, its state isn’t advanced at all (since spawn doesn’t do that)").  Do you mean the internal state of the children?

flying-sheep:
@ilan-gold yeah I'm sure! I was confused: they have two parts of internal state. Spawning advances the SeedSequence. I can do a writeup when I'm back working

ilan-gold:
Ok @flying-sheep That makes a load of sense - I assumed as much was going on (how else could what I posted be true unless they were literally tracking the number of children created) and it turns out, that is exactly what is happening.  Like literally:

```
SeedSequence(
    entropy=64076961259285389890164002958222865665,
    n_children_spawned=2,
)
```
after spawning so the entropy is unchanged.

src/scanpy/tools/_umap.py
```diff
@@ -142,6 +148,12 @@ def umap(  # noqa: PLR0913, PLR0915
         UMAP parameters.
 
     """
+    rng_init, rng_umap = np.random.default_rng(rng).spawn(2)
+    meta_random_state = (
+        dict(random_state=rng._arg) if isinstance(rng, _FakeRandomGen) else {}
```
ilan-gold: Why make `_arg` private?

src/scanpy/_utils/random.py
```diff
@@ -76,28 +85,55 @@ def set_igraph_random_state(
...
+                np.random.set_state(arg.get_state(legacy=False))
+                return _FakeRandomGen(arg, state)
+            np.random.seed(arg)
+        return _FakeRandomGen(arg, np.random.RandomState(np.random.get_bit_generator()))
+
+    def spawn(self, n_children: int) -> list[Self]:
+        """Return `self` `n_children` times.
+
+        In a real generator, the spawned children are independent,
+        but for backwards compatibility we return the same instance.
```
ilan-gold:
```suggestion
        but for backwards compatibility we return the same instance so that its internal state is advanced by each child.
```

src/scanpy/tools/_draw_graph.py
```diff
@@ -111,6 +117,13 @@ def draw_graph(  # noqa: PLR0913
 
     """
     start = logg.info(f"drawing single-cell graph using layout {layout!r}")
+    rng = np.random.default_rng(rng)
+    meta_random_state = (
+        dict(random_state=rng._arg) if isinstance(rng, _FakeRandomGen) else {}
+    )
+    rng = _if_legacy_apply_global(rng)
```
ilan-gold: Is this not a behavior change? For example `if init_pos in adata.obsm` branch + `if layout == "fa"` would never set the global seed?

src/scanpy/preprocessing/_utils.py
```diff
@@ -3,27 +3,21 @@
...
+    dims: tuple[int, ...], nsamp: int, *, rng: np.random.Generator
 ) -> NDArray[np.int64]:
     """Randomly sample indices from a grid, without repeating the same tuple."""
-    idx = sample_without_replacement(
-        np.prod(dims), nsamp, random_state=random_state, method=method
-    )
+    if isinstance(rng, _FakeRandomGen):
+        from sklearn.random_projection import sample_without_replacement
+
+        idx = sample_without_replacement(np.prod(dims), nsamp, random_state=rng._arg)
```
ilan-gold: Again seems like `_arg` should be public

src/scanpy/preprocessing/_simple.py
```diff
@@ -1017,143 +1022,137 @@ def downsample_counts(
...
-    np.random.seed(random_state)
+    rng = _if_legacy_apply_global(rng)
     cumcounts = col.cumsum()
+    total = np.int_(cumcounts[-1])
+    sample = rng.choice(total, target, replace=replace)
+    sample.sort()
+    return _downsample_array_inner(col, cumcounts, sample, inplace=inplace)
+
+
+# TODO: can/should this be parallelized?
```
ilan-gold, line 1125: Seems like a good candidate!

src/scanpy/preprocessing/_pca/_compat.py
```diff
@@ -18,21 +19,28 @@
...
+    rng: SeedLike | RNGLike | None = None,
 ) -> tuple[NDArray[np.floating], PCA]:
     """Sparse PCA for scikit-learn <1.4."""
-    random_state = check_random_state(random_state)
-    np.random.set_state(random_state.get_state())
-    random_init = np.random.rand(np.min(x.shape))
+    rng = np.random.default_rng(rng)
+    # this exists only to be stored in our PCA container object
+    random_state_meta = _legacy_random_state(rng)
+    [rng_init, rng_svds] = rng.spawn(2)
```
ilan-gold: I see you have "we use spawn to pass sub-generators to each subtask, so we can add subtasks without affecting subsequent calls e.g." in your comment on the numpy issue but why do we want this property of independent subtasks?

ilan-gold (commented):
> Should we store the new RNG in adata.uns? If no, this fixes https://github.com/scverse/scanpy/issues/1131

No opinion, seems moderately unrelated. I guess with the new structure it's a bit meaningless because seeds have no impact if you're passing in a `Generator` and you can't serialize a `Generator`

> Should we keep random_state in the docs?

No

> How should we annotate that the default rng isn’t actually None but “a new instance of _LegacyRandom(0)” but people can pass rng=None to get the future default behavior?

I noticed there is no default docstring for `rng` like for `random_state`.  Why not? It seems like the long term goal is to unify public APIs.  Is there a way to create a default docstring that generalizes but can refer to an old version of the pakcage/docs as a disclaimer for "default" behavior? Maybe not. It's hard for me to hold everything in my head.

> Should I handle passing rng to neighbors transformer?

It seems like we don't do it now unless it's a self-constructed transformer, so I probably wouldn't. Is this what you mean?

flying-sheep:
> you can't serialize a Generator

Not in AnnData, but being serialized in general is an explicit feature of them.

> I noticed there is no default docstring for rng like for random_state. Why not? It seems like the long term goal is to unify public APIs.

great point, I’ll add that, and that’ll make documenting this easy.

> It seems like we don't do it now unless it's a self-constructed transformer

I think we do it when the user passed a transformer class or when using an explicitly transformer that supports it (sklearn?)

My point was: sklearn doesn’t support `rng` yet, should we pre-empt the possible change and somehow handle both `random_state` and `rng` even though there probably won’t be transformers that accept `rng` for a while.

src/scanpy/tools/_umap.py
flying-sheep: I guess to distinguish it from all the attributes inherited from `Generator`, but I guess it’s not necessary.

It’s a private class, so once we’ve established that an object has this class, accessing its fields might as well be public.

src/scanpy/tools/_draw_graph.py
flying-sheep: this kind of comment is exactly why I was looking forward to more eyes on this PR, thanks!

src/scanpy/preprocessing/_simple.py
flying-sheep: yeah, I just left it alone to not cause any more changes in this PR as there were.

src/scanpy/preprocessing/_pca/_compat.py
flying-sheep: See my comment there: https://github.com/numpy/numpy/issues/24086#issuecomment-3968272805

I’m torn between these two arguments:
- when libraries change from under us, things change anyway, so whenever you update your packages and expect your script to produce the exact same result even when passing a seed, and
- Some steps aren’t impactful enough to have qualitative changes, and spawning allows to pass `rng` to more things that we previously didn’t pass it to (see example in that comment)

selmanozleyen:
```diff
 [rng_root] = rng.spawn(1)
-[rng_eigsh] = rng_root.spawn(1)
+rng_eigsh, rng_other_stuff = rng_root.spawn(2)
 random_init = rng_root.uniform(...)
 eigsh(A, k, v0=random_init, rng=rng_eigsh)
+do_thing(rng_other_stuff)
```

i think in your example it only matters if you don't know `eigsh` still works under the hood in parallel after it returns or not. But we know it doesn't right? So continueing with `rng_other_stuff` or `rng_eigsh` doesn't matter.

I think it only matters for either:
- abstraction reasons: when you don't know or don't want to know what the function you call does with the rng, how long it keeps it etc. 
- or writing parallel code

flying-sheep: > So continueing with `rng_other_stuff` or `rng_eigsh` doesn't matter.

No, but only 
- because we’re calling things in this order
- we’re using `spawn` above

If we don’t call `spawn` at all, we can’t change *anything* without changing all subsequent calls as we’re just passing one generator down the call tree and generating a single random byte more or less changes things.

So please @ilan-gold @selmanozleyen: what strategy do you think we should use here? Just do that and tell people reproducibility is only guaranteed on the exact patch version they’re on? Or have some kind of defensive strategy to reduce this. If so which one, I’m all ears!

selmanozleyen: > If we don’t call spawn at all, we can’t change anything without changing all subsequent calls as we’re just passing one generator down the call tree and generating a single random byte more or less changes things.

Like wasn't this always the case for older versions? I think it is a very strong guarantee otherwise. Who guarantees exact randomness across versions except numpy itself maybe? 

> No, but only:
>     because we’re calling things in this order
>     we’re using spawn above

I see but for me putting an independent rng to the `eigsh` is only meaningful if it means my results only changes in a separable and visible way if `eigsh` changes, rest will still be same. 

For example if I return 

```python
return { 
   'eigen': eigsh(A, k, v0=random_init, rng=rng_eigsh)
   'thing': do_thing(rng_other_stuff)
}
```

Then I see what you want. But if I do

```python
return do_thing(rng=rng_other_stuff, eigen_results)
```
then I don't see the point because the results depend on the outcome of a call that was supposed to be independent. We might as well use rng_root there, it makes no difference because anyone passing an rng wouldn't expect their rng not to change.

also see my unrolling loops argument, for sequential python loops I really think it is redundant: https://github.com/scverse/scanpy/pull/3983#discussion_r2864943289

src/scanpy/preprocessing/_simple.py
flying-sheep: > For some reason FakeRandomGen is super slow.

that should be caused by the legacy rng being slow, not anything we do, but might make sense to check.

src/scanpy/preprocessing/_pca/_compat.py
flying-sheep: > Like wasn't this always the case for older versions? I think it is a very strong guarantee otherwise

it was, this is why we only have the chance to change it right now, while adding support for `np.random.Generator` (unless we don’t care about preserving reproducibility between releases at all, then we can change it whenever, but also don’t need to)

> Who guarantees exact randomness across versions except numpy itself maybe?

nobody lol I think we’re basically the only ones. @ivirshup cared about this.

but in numpy this guarantee only holds for legacy RNG, so we won’t be able to promise it anyway once we switch to `rng` (in theory, in practice I don’t think numpy has broken rng stability yet)

> also see my unrolling loops argument

Yeah if we don’t guarantee reproducibility across versions we can get rid of it. The way it is, we could e.g. promise stability across patch releases, and still parallelize this in a patch release.

selmanozleyen: > The way it is, we could e.g. promise stability across patch releases, and still parallelize this in a patch release.

ah I see, thats why you did it that way

ilan-gold: Overall, it sounds like spawning is relatively low-cost and provides us with more stability for little tradeoff. I would be in favor of spawning then.  I understand now why we want to do it, and I like it architecturally since it allows us to be clear about where we use randomness and where other libraries might.

ilan-gold:
> I think we do it when the user passed a transformer class or when using an explicitly transformer that supports it (sklearn?)

Could you check this out? My impression looking at our code was that there was no mechanism for "updating" a user-passed-in transformer i.e., giving it an RNG (state). So I'm not sure even what this would look like. You would use https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator.set_params or something?

src/scanpy/preprocessing/_simple.py
flying-sheep: OK, new numbers. `random_state` means legacy RNG, `rng` means new RNG; `spawn` means spawning one rng per cell, `reuse` means reusing the same rng.

Seems like my changes did introduce a little bit of overhead for the default / legacy case (rightmost column). I wonder if that comes from the fact that the rng doesn’t happen inside of the numba code anymore, or if the all the wrapping is that noticable (it really shouldn’t be).

Otherwise the results match yours: spawning takes a less than 2× the time of reusing and the new generator is faster than the old one.

```
For scanpy commit af57cffc <main>:
================= ============= ============= ====================== ======================
--                                              layer / param3                             
----------------- -------------------------------------------------------------------------
     dataset       rng / spawn   rng / reuse   random_state / spawn   random_state / reuse 
================= ============= ============= ====================== ======================
 pbmc68k_reduced       n/a           n/a               n/a                7.56±0.02ms      
      pbmc3k           n/a           n/a               n/a                 119±0.5ms       
================= ============= ============= ====================== ======================
```

```
For scanpy commit 93a9d90e <pa/rng>:

================= ============= ============= ====================== ======================
--                                              layer / param3                             
----------------- -------------------------------------------------------------------------
     dataset       rng / spawn   rng / reuse   random_state / spawn   random_state / reuse 
================= ============= ============= ====================== ======================
 pbmc68k_reduced    20.8±0.2ms    11.1±0.5ms           n/a                 17.4±0.9ms      
      pbmc3k         92.4±3ms     54.3±0.5ms           n/a                  137±1ms        
================= ============= ============= ====================== ======================
```

flying-sheep:
@ilan-gold ah yeah I assumed we supported passing a transformer class, but we only support passing an instance.

src/scanpy/tools/_umap.py
```diff
@@ -147,12 +147,7 @@ def umap(  # noqa: PLR0913, PLR0915
         UMAP parameters.
 
     """
-    rng_init, rng_umap = np.random.default_rng(rng).spawn(2)
```
ilan-gold: Sorry, why did we undo this? I thought I had said here it seems like a good idea: https://github.com/scverse/scanpy/pull/3983#discussion_r2917224348

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -308,8 +296,13 @@ def pca(  # noqa: PLR0912, PLR0913, PLR0915
...
 
-                if not isinstance(rng, _LegacyRng) or rng.arg != 0:
-                    msg = f"Ignoring random_state={_legacy_random_state(rng)} when using a sparse dask array"
+                if not rng_is_default:
+                    dbg = (
+                        f"random_state={_legacy_random_state(rng)!r}"
+                        if isinstance(rng, _LegacyRng)
+                        else f"rng={rng!r}"
+                    )
+                    msg = f"Ignoring {dbg} when using a sparse dask array"
```
ilan-gold: Hmmm, is it so much that it is igorned? I think it is that the algorithm is non-random, which at least should be communicated, but I would probably just opt for no message.

src/scanpy/preprocessing/_pca/_compat.py
```diff
@@ -54,11 +60,13 @@ def rmat_op(v: NDArray[np.floating]):
         rmatmat=rmat_op,
     )
 
-    u, s, v = svds(linop, solver=solver, k=n_pcs, v0=random_init)
-    # u_based_decision was changed in https://github.com/scikit-learn/scikit-learn/pull/27491
-    u, v = svd_flip(
-        u, v, u_based_decision=pkg_version("scikit-learn") < Version("1.5.0rc1")
+    random_init = rng.uniform(size=np.min(x.shape))
+    kw: Any = (
+        dict(rng=rng) if SCIPY_1_15 else dict(random_state=_legacy_random_state(rng))
```
ilan-gold, line 65: Same here, what happened to `spawn`?

src/scanpy/tools/_umap.py
flying-sheep: When discussing with Selman, I realized that the only stable pattern would be

```py
[root_rng] = rng.spawn(1)  # advance rng’s seed sequence by 1
rng_task1, rng_task2 = root_rng.spawn(2)  # advance root_rng’s seed sequence by 2, can be changed
del rng, root_ rng  # make sure they aren’t used anymore
```

which isn’t what we used and is a bit complex for little gain – numpy has no stability guarantee anyway, so what this would allow is to add and use a `rng_task3` which would probably change results anyway, as there is no place where we create completely independent results using the RNGs (like, predict two different things and then only store them without using them further)

At this point I have no opinion here anymore. @selmanozleyen and @ilan-gold please align on something and I’ll do it. (not trying to sound peeved here, just ready to move on)

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: hmm, maybe for random, yeah, but as a user I’d like to know if `algorithm` (warning on the next line) is being ignored.

src/scanpy/tools/_umap.py
selmanozleyen: I think I agree with current version. If the return value is strongly coupled with multiple rng's there is little benefit IMO (unless it was computed parallel). Because the initialization is also going to effect the core algo anyway.

ilan-gold: I don't really get why this is better than just spawning two tasks out of `rng`

selmanozleyen: I think it should be done only when there is clear benefit. That's just my opinion for clarity. Otherwise I would be confused if rng_init and rng_umap are supposed to be very independent or not. I don't think its better or worse though

src/scanpy/preprocessing/_simple.py
selmanozleyen: But current version doesn't reuse the rng right? Maybe we can create performance issue to create rng per thread. But this will limit reproducibility when n_jobs are the same.

selmanozleyen (approved):
I think adding TODO's and issues will clarify the spawn usages are for possible parallel implementations. 

Also we need to clarify if we will create rng per core or independent component.

src/scanpy/neighbors/__init__.py
```diff
@@ -220,21 +229,23 @@ def neighbors(  # noqa: PLR0913
...
-        if ignored := {
+        ignored = {
             p.name
             for p in signature(neighbors).parameters.values()
-            if p.name in {"use_rep", "knn", "n_pcs", "metric_kwds", "random_state"}
+            if p.name in {"use_rep", "knn", "n_pcs", "metric_kwds"}
             if params[p.name] != p.default
-        }:
+        }
+        if meta_random_state.get("random_state") != 0:  # rng or random_state was passed
```
ilan-gold: I think
```suggestion
        if meta_random_state.get("random_state") != 0:  # rng was not passed in or random_state was passed
```
Right? I think that's the behavior of `_accepts_legacy_random_state`?

src/scanpy/neighbors/__init__.py
```diff
@@ -220,21 +229,23 @@ def neighbors(  # noqa: PLR0913
...
+            ignored.add("rng/random_state")
+        if ignored:
             warn(
                 f"Parameter(s) ignored if `distances` is given: {ignored}",
                 UserWarning,
             )
-            random_state = 0
```
ilan-gold, line 237: Why did we get rid of this?

flying-sheep:
I tried to parallelize downsample and ran into a bunch of numba limitations and bugs: https://github.com/scverse/scanpy/pull/4004

src/scanpy/neighbors/__init__.py
flying-sheep: I now do `meta_random_state.pop` so it doesn’t get stored in the metadata as intended

flying-sheep:
Since it’s impossible to parallelize downsampling with `replacement=False` using numba right now, I just removed the spawning. If we merge #4004, it’ll change things anyway, no matter how it looks.

src/scanpy/preprocessing/_simple.py
flying-sheep: see main thread: I just removed it for now as we can’t handle this forwards-compatibly

PR #4013, opened 2026-03-26
perf: "two-pass" seurat hvg via `scanpy.get.aggregate`

ilan-gold:
An idea that popped into my head for disk-bound datasets but likely also normal ones.  This should, in theory, greatly improve on-disk access and produce speed ups for disk bound data by reducing the amount of i/o in the worst case, unordered scenario (while, I would guess, leaving in-memory datasets untocuhed or maybe improved thanks to memory access + more efficient mean/var).

Dependent on #4143

- [ ] Closes #


ilan-gold:
The old seurat_v3 (on `main`) literally timed out with dask: https://github.com/scverse/scanpy/pull/4013/checks?check_run_id=83417211191 so the performance benefit is at least ~5x (since the timeout is 60s and this branch does 13s)

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -172,17 +167,56 @@ def _highly_variable_genes_seurat_v3(  # noqa: PLR0912, PLR0915
...
+        )
+        mean_global, var_global = (
+            aggregated_mean_var.layers[l] for l in ["mean", "var"]
+        )
+        if isinstance(mean_global, DaskArray):
+            import dask.array as da
+
+            mean_global, var_global = da.compute(mean_global, var_global)
+            aggregated_mean_var.layers["mean"] = mean_global
+            aggregated_mean_var.layers["var"] = var_global
```
flying-sheep: this seems a bit verbose for what it is, don’t we have a helper for that or am I thinking f-a-u?

flying-sheep (approved):
Looks very straightforward, nice idea!

src/scanpy/preprocessing/_highly_variable_genes.py
ilan-gold: What aspect of it is verbose?

flying-sheep: Creating the intermediates. I think I got confused searching for where they are used after, only to realize they aren’t. But maybe that’s just me.

Would this work or can they be non-ndarrays?

```suggestion
        aggregated_mean_var.layers["mean"], aggregated_mean_var.layers["var"] = materialize_as_ndarray(
            *(aggregated_mean_var.layers[l] for l in ["mean", "var"])
        )
```

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -172,17 +167,51 @@ def _highly_variable_genes_seurat_v3(  # noqa: PLR0912, PLR0915
...
+            index=adata.obs_names, data={"__hvg_v3_batch_info__": batch_info}
+        ),
+    )
+    if batch_key is not None:
+        aggregated_mean_var = aggregate(
+            adata_agg, by="__hvg_v3_batch_info__", func=["mean", "var"]
+        )
+        aggregated_mean_var.layers["mean"], aggregated_mean_var.layers["var"] = (
+            materialize_as_ndarray(aggregated_mean_var.layers[l])
+            for l in ["mean", "var"]
```
flying-sheep: hm, I thought that  `dask.compute(a, b)` does something different than `dask.compute(a), dask.compute(b)`.

I guess it takes a tuple instead of multiple args 

```suggestion
            materialize_as_ndarray(
                tuple(aggregated_mean_var.layers[l] for l in ["mean", "var"])
            )
```

ilan-gold:
Thanks for the catch @flying-sheep !

PR #4015, opened 2026-03-27
fix: use fast-array-utils’ threadsafe njit

flying-sheep:
- [ ] Closes #

@ilan-gold ported scanpy’s njit wrapper to fast-array-utils where I finally fixed it so it works as intended


ilan-gold (changes requested):
Looks we still got an old (broken) import in there judging by the tests!

ilan-gold (approved):
I added the benchmark label. As long as nothing regresses horribly, I think we're good

PR #4025, opened 2026-03-31
docs(plotting): add inline plot example for `highest_expr_genes`

haoyu-haoyu:
## Summary

Add `.. plot::` directive examples to the docstrings of two plotting functions that were missing inline examples:

- **`sc.pl.highest_expr_genes`**: default usage and `n_top=10` variant
- **`sc.pl.highly_variable_genes`**: default usage and `log=True` variant

These are two of the remaining unchecked items from #1664.

## Pattern

Follows the established pattern from existing inline examples (e.g., `pca_loadings`, `rank_genes_groups_heatmap`):
- First block loads data with `sc.datasets.pbmc3k_processed()`
- Subsequent blocks demonstrate parameter variations
- Uses `:context: close-figs` directive option

Ref #1664

### DIFF
--- a/src/scanpy/plotting/_preprocessing.py
+++ b/src/scanpy/plotting/_preprocessing.py
@@ -41,6 +41,23 @@ def highly_variable_genes(  # noqa: PLR0912
         A string is appended to the default filename.
         Infer the filetype if ending on {{`'.pdf'`, `'.png'`, `'.svg'`}}.
 
+    Examples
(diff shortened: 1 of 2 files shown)


flying-sheep:
hi, thank you for your contribution! @Ekin-Kahraman had the same idea for hvg, but uses the usual preprocessing, so I’ll merge both PRs.

PR #4032, opened 2026-04-05
docs: add inline example plot to sc.pl.highly_variable_genes

Ekin-Kahraman:
Adds an `Examples` section to the `sc.pl.highly_variable_genes` docstring with two inline plot blocks:

1. Minimal pipeline: load PBMC 3k, normalise, log-transform, compute HVGs, plot
2. Same data with `log=True` to demonstrate the parameter

Uses `sc.datasets.pbmc3k()` and `:context: close-figs` following the pattern from merged contributions to the #1664 checklist.

Checklist item from #1664.

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -51,7 +51,7 @@ jobs:
     name: Check for release notes
     runs-on: ubuntu-latest
     needs: check-milestone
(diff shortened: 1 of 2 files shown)


flying-sheep:
Hi thank you for your contribution!

Please next time don’t delete the PR template, but fill it out instead.

also looks like @haoyu-haoyu had the same idea a bit earlier in #4025, but you use the right preprocessing. https://icb-scanpy--4032.com.readthedocs.build/en/4032/api/generated/scanpy.pl.highly_variable_genes.html

Issue #2387, opened 2022-12-27
`sc.external.pp.magic`: what is `n_jobs`'s default _really_?


alexlenail:
[scanpy docs](https://scanpy.readthedocs.io/en/stable/generated/scanpy.external.pp.magic.html): 

> n_jobs : [Optional](https://docs.python.org/3/library/typing.html#typing.Optional)[[int](https://docs.python.org/3/library/functions.html#int)] (default: None)
Number of threads to use in training. All cores are used by default.

[magic docs](https://magic.readthedocs.io/en/stable/api.html):

> n_jobs (integer, optional, default: 1) – The number of jobs to use for the computation. If -1 all CPUs are used. If 1 is given, no parallel computing code is used at all, which is useful for debugging.

scanpy code: 

https://github.com/scverse/scanpy/blob/536ed15bc73ab5d1131c0d530dd9d4f2dc9aee36/scanpy/external/pp/_magic.py#L164

https://github.com/scverse/scanpy/blob/536ed15bc73ab5d1131c0d530dd9d4f2dc9aee36/scanpy/_settings.py#L82

I'm guessing the scanpy docs are wrong when they say "All cores are used by default." ?

jpmcga:
^^I've empirically confirmed this; been seeing multi-fold speedups when I set `n_jobs=-1` corresponding to the number of cores, versus running with the default.  From what I can tell, running without specifying `n_jobs` results in 1 core being used.

flying-sheep:
Yeah, we definitely need to improve this. #2390 is also relevant here

PR #4037, opened 2026-04-07
feat: allow exponentiation post agg for log-fold-change in `rank_genes_groups`

### LINKED ISSUE #864: rank_genes_groups logfoldchange different than seurat
Hi!
I've noticed that the function `rank_genes_groups` calculate logFC differently than seurat. 
https://github.com/theislab/scanpy/blob/5800db45dde0c2060ad0a9bed8ea931bd41da936/scanpy/tools/_rank_genes_groups.py#L207-L208

https://github.com/theislab/scanpy/blob/5800db45dde0c2060ad0a9bed8ea931bd41da936/scanpy/tools/_rank_genes_groups.py#L223

Thus the equation is
`log(exp(mean(values))`

while in Seurat is
https://github.com/satijalab/seurat/blob/96d07d80bc4b6513b93e9c10d8a9d57ae7016f9f/R/differential_expression.R#L175-L179  

thus
`log(mean(exp(values)))`

I was thus wondering if this was intended, since it leads to different logFC values.

ilan-gold:
My three questions here:

1. What to do about t-tests? Do people expect fold change of the input to the t-test (as is currently done) or fold change of the exponent?
2. How do we test this? Dummy groups? Wait for `illico` to compare? I added a very simple test.
3. Does the name `exp_post_agg` make sense? `exp_post_mean`?

- [x] Closes #864


flying-sheep (commented):
Regarding 1: I didn’t look at the math, but I assume to get both accurate logfoldchanges and correct t-test results, the t-test needs the means of the data, and the logfoldchanges need the means of e^data. So we’d have to calculate means twice I think. If @eroell doesn’t chime in I’ll re-learn the math …

Regarding 3: maybe `lfc_estim_fast` or so?

ilan-gold:
> Regarding 1:

I think my question is less about math than expectations.  Do people expect https://github.com/scverse/scanpy/issues/864 to be applied to _all_ tests that report LFC or just wilcoxon?

> Regarding 3 `lfc_estim_fast`

I like this but I wonder if being more explicit about the math is nicer.  @eroell thoughts?

flying-sheep:
> Do people expect #864 to be applied to all tests that report LFC or just wilcoxon?

I think it would be confusing if the option only applied to one of them, but of course I don’t know what people expect.

ilan-gold:
That's a pretty good reason! Nothing in the issue is focused on wilcoxon, I guess!

ilan-gold:
> Regarding 3: maybe lfc_estim_fast or so?

Looking back at things, I think this setting was created for a time when the code looked very different.  It is only faster for `t-test` to do `exp_post_agg`.

flying-sheep:
I think you’re right, but `expm1` is a more expensive operation than the others, so doing it on the full data instead of just the means might do something. Probably not a huge factor, but let’s see.

flying-sheep:
huh, I didn’t expect it to have zero impact. OK then, let’s see what @eroell says

eroell:
> What to do about t-tests? Do people expect fold change of the input to the t-test (as is currently done) or fold change of the exponent?

I think the ` log(mean(exp(values)))` as log fold change would also apply here

> Does the name exp_post_agg make sense? exp_post_mean?

How about `mean_in_log_space`? `True` would be the current setting, and `False` would be the new one

eroell:
Also, a side comment on my understanding of the issue: my agreement with the issue there is that I also consider a log fold change to compute the mean of a group in "count space".

Since data seems to lie around generally in log space prior to this function call, the issue mentions `log(mean(exp(values)))`. The innermost `exp` is only there since the original values experienced a log1p already to become the `values` in this expression

From this point of view comes also my suggestion for the argument name of calling the current version log space.

For the tests, if a `logfoldchange` is needed for a test, I'd go with the new one in count space.

Sorry for the slow response to this issue, thank you so much for taking it up!

src/scanpy/tools/_rank_genes_groups.py
```diff
@@ -511,9 +523,12 @@ def rank_genes_groups(  # noqa: PLR0912, PLR0913, PLR0915
     corr_method: _CorrMethod = "benjamini-hochberg",
     tie_correct: bool = False,
     layer: str | None = None,
+    exp_post_agg: bool | Default = Default(
```
eroell, line 526: `mean_in_log_space`?

eroell:
Regarding the unchanged benchmark, of which I did not get into the weeds;

Does this test numerically computed logfoldchange values, or rather rankings of genes generated by them?

ilan-gold:
> Does this test numerically computed logfoldchange values, or rather rankings of genes generated by them?

Only speed

> mean_in_log_space

I like this!

thanks for the reply!

flying-sheep (approved):
looks great! maybe add a test that checks that the wilcoxon results are affected in an expected way, but that’s pretty clear from the code so not necessary. We’ll have to change the reference tests for 2.0 anyway

ilan-gold:
> maybe add a test that checks that the wilcoxon results are affected in an expected way

The results should not be affected, just the LFC.  Right @eroell ?

zboldyga:
hey @ilan-gold so as per the convo in #4090 , I reviewed this PR and #4038 . On the 5 initial datasets I benchmarked against, the illico port is numerically identical to the existing scanpy implementation (yay). However, there are a few things I think are worth addressing. I will provide more details tomorrow (I'm US PT), but initially I see:

- some options are not passing through properly (e.g. mean_in_log_space)
- on the latest 1.12.1 release, by default all threads are used (and I think a specified number of threads might have been ignored). on the main branch I think that has now changed, and it defaults to 1, which I'd argue isn't ideal because the user has to specify a value or it runs substantially slower. I recall someone added a relevant feature recently, I think to specify number of threads, but it seems reverting to default of 1 was perhaps an unintended side effect. 
- some minor code suggestions
- there's a performance issue I noticed -- the overall speed of scanpy with illico is actually substantially slower than illico because of calls like basic_stats (like 30x slower). This can be vectorized, and a few other performance optimizations. I did this locally and got it much closer to illico's speed. But this is probably worthy of it's own PR?

I can jump in and contribute on any of this, or at least give you clarity on exactly what the items I see are. Given this is an existing set of PRs, with moving pieces, how would you prefer I assist?

Maybe for the smaller items I can code review in here, and for the last item w/ perf opt I can stage a PR on top of #4038 to outline what this looks like?

ilan-gold:
> on the latest 1.12.1 release, by default all threads are used (and I think a specified number of threads might have been ignored). on the main branch I think that has now changed, and it defaults to 1, which I'd argue isn't ideal because the user has to specify a value or it runs substantially slower. I recall someone added a relevant feature recently, I think to specify number of threads, but it seems reverting to default of 1 was perhaps an unintended side effect.

I _think_ we need to hook this up to the scanpy setting for `n_jobs` but @flying-sheep would know better.

> some minor code suggestions

all ears!

> there's a performance issue I noticed -- the overall speed of scanpy with illico is actually substantially slower than illico because of calls like basic_stats (like 30x slower). This can be vectorized, and a few other performance optimizations. I did this locally and got it much closer to illico's speed. But this is probably worthy of it's own PR?

Absolutely!!!! This would be super helpful!

> I can jump in and contribute on any of this, or at least give you clarity on exactly what the items I see are. Given this is an existing set of PRs, with moving pieces, how would you prefer I assist?

Clear, well-scoped PRs into existing PRs (if open) would be best! So 

> Maybe for the smaller items I can code review in here, and for the last item w/ perf opt I can stage a PR on top of https://github.com/scverse/scanpy/pull/4038 to outline what this looks like?

would be great! Feel free to jump in with a review on #4038 as well!

ilan-gold:
> some options are not passing through properly (e.g. mean_in_log_space)

fixed

flying-sheep:
> on the main branch I think that has now changed, and it defaults to 1, which I'd argue isn't ideal 

> I think we need to hook this up to the scanpy setting for n_jobs

@zboldyga pointed out that we just did that for `rank_genes_groups` in #4082, but `settings.n_jobs` defaults to 1 which should be changed.

Please direct discussion about this to #2387!

ilan-gold:
> there's a performance issue I noticed -- the overall speed of scanpy with illico is actually substantially slower than illico because of calls like basic_stats (like 30x slower). This can be vectorized, and a few other performance optimizations. I did this locally and got it much closer to illico's speed. But this is probably worthy of it's own PR?

Last thing: have a look at `scanpy.get.aggregate` for the calculation of basic stats. It has a numba-fied version of grouped mean-var calculation as of #4062 which should be at least as fast as this loop, but might (definitely should) be faster.

flying-sheep:
> > some options are not passing through properly (e.g. mean_in_log_space)
>
> fixed

this is exactly why I advocate for not having defaults in internal APIs that are basically just “pass through arguments from a public API and do part of an algorithm”.

do you mind removing them from `compute_statistics`?

ilan-gold:
Great call on internal defaults!

zboldyga:
btw I plan to continue working on this tomorrow -- will follow your suggestions @ilan-gold

PR #4040, opened 2026-04-08
fix: copy size array before reorder in multi-color scatter plots

### LINKED ISSUE #4024: Per-point size array is cumulatively reordered across multi-color subplots

Ekin-Kahraman:
## Summary

Fixes #4024.

- **Root cause**: In `embedding()`, the per-point `size` array was reordered in-place (`size = np.array(size)[order]`) inside the per-color loop. Each iteration's z-order permutation was applied to the already-reordered array from the previous iteration, causing cumulative corruption of marker sizes across subplots.
- **Fix**: Use a loop-local variable `_size` so each iteration reorders from the original `size` array. This matches the behavior of `color_source_vector`, `color_vector`, and `coords`, which are all freshly derived each iteration.

## Changes

- `src/scanpy/plotting/_tools/scatterplots.py`: Replace in-place mutation of `size` with a loop-local `_size` variable; update all downstream references within the loop body.
- `tests/test_plotting.py`: Add `test_scatter_size_not_mutated_across_panels` regression test that verifies the input size array is not modified after plotting with multiple color keys.

## Test plan

- [ ] `test_scatter_size_not_mutated_across_panels` passes — confirms the size array is preserved
- [ ] Existing scatter/embedding tests continue to pass (no behavioral change for single-panel plots or scalar sizes)

### DIFF
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -1875,6 +1875,25 @@ def test_umap_mask_no_modification():
     pd.testing.assert_series_equal(pbmc.obs["louvain"], data_copy)
 
 
+def test_scatter_size_not_mutated_across_panels():
+    """Per-point size array must not be cumulatively reordered across subplots.
+
+    Regression test for https://github.com/scverse/scanpy/issues/4024
+    """
+    pbmc = pbmc3k_processed()
+    rng = np.random.default_rng(0)
+    sizes = rng.uniform(10, 200, size=pbmc.n_obs)
+    sizes_original = sizes.copy()
+
+    axes = sc.pl.umap(
+        pbmc, color=["louvain", "n_genes"], size=sizes, show=False
+    )
+    plt.close()
+
+    # The input array must not be modified
+    np.testing.assert_array_equal(sizes, sizes_original)
+
+
 def test_string_mask(tmp_path, check_same_image):
     """Check that the same mask given as string or bool array provides the same result."""
     pbmc = pbmc3k_processed()
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -1875,38 +1875,6 @@ def test_umap_mask_no_modification():
     pd.testing.assert_series_equal(pbmc.obs["louvain"], data_copy)
 
 
-def test_scatter_size_not_mutated_across_panels():
-    """Per-point size array must not be cumulatively reordered across subplots.
-
-    Regression test for https://github.com/scverse/scanpy/issues/4024
-    Uses 3 panels (categorical + two continuous) to exercise cumulative reorder.
-    """
-    pbmc = pbmc3k_processed()
-    rng = np.random.default_rng(0)
-    sizes = rng.uniform(10, 200, size=pbmc.n_obs)
-    sizes_original = sizes.copy()
-
-    axes = sc.pl.umap(
-        pbmc,
-        color=["louvain", "n_genes", "n_counts"],
-        size=sizes,
-        show=False,
-    )
-
-    # The input array must not be modified
-    np.testing.assert_array_equal(sizes, sizes_original)
-
-    # Each panel must plot the correct per-point sizes (just reordered by
-    # z-order, not cumulatively scrambled).  Sorting makes the comparison
-    # independent of z-ordering.
-    expected_sorted = np.sort(sizes)
-    for ax in axes:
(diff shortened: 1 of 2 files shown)


tests/test_plotting.py
```diff
@@ -1875,6 +1875,23 @@ def test_umap_mask_no_modification():
...
+    """Per-point size array must not be cumulatively reordered across subplots.
+
+    Regression test for https://github.com/scverse/scanpy/issues/4024
+    """
+    pbmc = pbmc3k_processed()
+    rng = np.random.default_rng(0)
+    sizes = rng.uniform(10, 200, size=pbmc.n_obs)
+    sizes_original = sizes.copy()
+
+    axes = sc.pl.umap(pbmc, color=["louvain", "n_genes"], size=sizes, show=False)
```
JhonatanFelix: The variable `axes` is not getting used afterwards and giving problems with the `ruff check`, you can remove it and just call directly the `sc.pl.umap(...)` or call it as `_ = sc.pl.umap(...)`.

edit: I made a small [PR](https://github.com/Ekin-Kahraman/scanpy/pull/1) just to take off the variable.

JhonatanFelix (commented):
Nice fix! 

I think that you could strengthen the regression test on `test_scatter_size_not_mutated_across_panels` because currently it cheacks that the caller's original `sizes` array is unchanged, but the bug in #4024 is really that later subplots receive the wrong per-point sizes.

Would it make sense to asset the plotted sizes from the matplotliv artists for each axis, so the test proves that all pannels get the same size mapping?? Another thing is that the issue's reproduction used three panels (`["louvain", "n_genes", "n_counts"]`), which might exercise the cumulative reorder more clearly than the current 2 panels case

Ekin-Kahraman:
Thanks for the review @JhonatanFelix! Both great suggestions — I've pushed an update that:

1. Uses 3 color keys (`["louvain", "n_genes", "n_counts"]`) matching the original issue reproduction, so the test exercises the cumulative reorder across a categorical + two continuous panels.
2. Extracts the actual plotted sizes from each axis's `PathCollection` via `get_sizes()` and asserts that the sorted sizes match the expected array for every panel, not just checking that the caller's input is unchanged.

flying-sheep:
Hi, thanks! The test still doesn’t do anything I’m afraid. If I temporarily disable your fix by doing

```diff
-size_plot = size[order] if isinstance(size, np.ndarray) else size
+size = size_plot = size[order] if isinstance(size, np.ndarray) else size
```

the test still passes.

flying-sheep:
actually #4023 came first with this fix, so I’m closing this in favor of it.

PR #4055, opened 2026-04-10
feat: create `scanpy2` extra for dependencies

### LINKED ISSUE #4051: New scanpy 2.0 optional dependency extra
I think we definitely need:

1. `igraph`
2. `scikit-misc`

And may potentially need:

1. `illico`

I think a new option like `scanpy2` could help smooth this out.  We could then raise an error if you, for example, turn on the 2.0 presets without having one of these installed.

`scikit-misc` is definitely the most worrisome default dependncy given the issues we have had with it in the past.  This one may be worth an scverse reimplementation.  Or python bindings to https://github.com/thisisamirv/loess-rs (disclaimer, I just googled "loess rust")

ilan-gold:
I noticed that the scanpy 2 settings are not tested in general - this seems like an oversight maybe? Or? Marking this as a draft because I'm not sure this is the best way to check this behavior.

- [x] Closes #4051

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,6 +99,7 @@ scanorama = [ "scanorama" ]
 scrublet = [ "scikit-image>=0.23.1" ]
 # highly_variable_genes method 'seurat_v3'
 skmisc = [ "scikit-misc>=0.5.1" ]
+scanpy2 = [ "igraph>=0.10.8", "leidenalg>=0.10.1", "scikit-misc>=0.5.1" ]
 
 [dependency-groups]
 dev = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,7 +97,7 @@ scanorama = [ "scanorama" ]
 scrublet = [ "scikit-image>=0.23.1" ]
 # highly_variable_genes method 'seurat_v3'
 skmisc = [ "scikit-misc>=0.5.1" ]
-scanpy2 = [ "igraph>=0.10.8", "leidenalg>=0.10.1", "scikit-misc>=0.5.1" ]
+scanpy2 = [ "igraph>=0.10.8", "scikit-misc>=0.5.1" ]
 
 [dependency-groups]
 dev = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -202,7 +202,7 @@ lint.per-file-ignores."src/scanpy/tools/_sim.py" = [ "N" ]
 # D*: No need for docstrings for all test modules and test functions
 # PLR0913: Test may use many fixtures
 lint.per-file-ignores."tests/**/*.py" = [ "D100", "D101", "D103", "PLR0913" ]
-lint.allowed-confusables = [ "×", "–", "’", "α" ]
+lint.allowed-confusables = [ "×", "–", "‘", "’", "α" ]
 lint.external = [ "PLR0917" ]
 lint.flake8-bugbear.extend-immutable-calls = [ "scanpy._settings.Default" ]
 lint.flake8-type-checking.exempt-modules = []
--- a/src/scanpy/_settings/__init__.py
+++ b/src/scanpy/_settings/__init__.py
@@ -105,7 +105,6 @@ class SettingsMeta(SingletonMeta, type):
 
     @preset.setter
     def preset(cls, preset: Preset | str) -> None:
-
         new_preset = Preset(preset)
         new_preset.check_deps()
         cls._preset = new_preset
--- a/src/scanpy/_settings/__init__.py
+++ b/src/scanpy/_settings/__init__.py
@@ -105,11 +105,9 @@ class SettingsMeta(SingletonMeta, type):
 
     @preset.setter
     def preset(cls, preset: Preset | str) -> None:
-        from .presets import _check_scanpy_v2_deps
 
         new_preset = Preset(preset)
-        if new_preset is Preset.ScanpyV2Preview:
-            _check_scanpy_v2_deps()
+        new_preset.check_deps()
         cls._preset = new_preset
 
     @property
--- a/src/scanpy/_settings/__init__.py
+++ b/src/scanpy/_settings/__init__.py
@@ -105,7 +105,12 @@ class SettingsMeta(SingletonMeta, type):
 
     @preset.setter
     def preset(cls, preset: Preset | str) -> None:
-        cls._preset = Preset(preset)
+        from .presets import _check_scanpy_v2_deps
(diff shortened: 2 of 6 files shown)


src/scanpy/_settings/presets.py
```diff
@@ -128,6 +130,30 @@ def get(self: Preset) -> NT:
...
+
+    from packaging.requirements import Requirement
+
+    dist_to_module = {d: m for m, ds in packages_distributions().items() for d in ds}
+    missing = [
+        r.name
+        for r in map(Requirement, requires("scanpy"))
+        if r.marker
+        and r.marker.evaluate({"extra": "scanpy2"})
+        and find_spec(dist_to_module.get(r.name, r.name.replace("-", "_"))) is None
```
flying-sheep: why not just this? (can also make `dist_to_module` a set with a different variable name then)

```suggestion
        and r.name in dist_to_module
```

`packages_distributions` already gives all installed distributions and we want distribution names, so no need to look at modules at all.

pyproject.toml
```diff
@@ -99,6 +99,7 @@ scanorama = [ "scanorama" ]
 scrublet = [ "scikit-image>=0.23.1" ]
 # highly_variable_genes method 'seurat_v3'
 skmisc = [ "scikit-misc>=0.5.1" ]
+scanpy2 = [ "igraph>=0.10.8", "leidenalg>=0.10.1", "scikit-misc>=0.5.1" ]
```
flying-sheep: we should push for [PEP 771](https://peps.python.org/pep-0771/) to become accepted.

we can’t put `igraph` into `dependencies` (GPL), and having `pip install scanpy` install a version where `sc.tl.leiden(...)` just raises an error isn’t very elegant.

tests/test_settings.py
```diff
@@ -17,3 +21,21 @@ def test_resets(_attempt: int) -> None:
 def test_set_figure_params_warns() -> None:
     with pytest.warns(FutureWarning, match=r"scanpy\.set_figure_params"):
         sc.settings.set_figure_params()
+
+
+def test_preset_scanpy_v2_preview_checks_deps() -> None:
+    _dist_to_module = {d: m for m, ds in packages_distributions().items() for d in ds}
```
flying-sheep: why use `_leading_underscore` names for local variables?

src/scanpy/_settings/__init__.py
```diff
@@ -105,7 +105,12 @@ def preset(cls) -> Preset:
 
     @preset.setter
     def preset(cls, preset: Preset | str) -> None:
-        cls._preset = Preset(preset)
+        from .presets import _check_scanpy_v2_deps
+
+        new_preset = Preset(preset)
+        if new_preset is Preset.ScanpyV2Preview:
+            _check_scanpy_v2_deps()
```
flying-sheep: maybe this could be moved into the `Preset` class?

flying-sheep (changes requested):
This is great! One issue with the dependency check, otherwise I like it very much!

pyproject.toml
ilan-gold: From a quick glance, I don't see anyone raising the license issue in the attached thread to the PEP. Is that the appropriate location to voice our support? I think this is a real use-case

flying-sheep: I commented!

src/scanpy/_settings/__init__.py
```diff
@@ -105,7 +105,10 @@ def preset(cls) -> Preset:
 
     @preset.setter
     def preset(cls, preset: Preset | str) -> None:
-        cls._preset = Preset(preset)
+
```
flying-sheep:
```suggestion
```

src/scanpy/_settings/presets.py
```diff
@@ -217,6 +223,31 @@ def override(self, preset: Preset) -> Generator[Preset, None, None]:
...
+                dist_to_module = {
+                    d: m for m, ds in packages_distributions().items() for d in ds
+                }
+                missing = [
+                    r.name
+                    for r in map(Requirement, requires("scanpy"))
+                    if r.marker
+                    and r.marker.evaluate({"extra": "scanpy2"})
+                    and r.name in dist_to_module
+                ]
```
flying-sheep:
```suggestion
                dists = {
                    d for m, ds in packages_distributions().items() for d in ds
                }
                missing = [
                    r.name
                    for r in map(Requirement, requires("scanpy"))
                    if r.marker
                    and r.marker.evaluate({"extra": "scanpy2"})
                    and r.name in dists
                ]
```

src/scanpy/_settings/presets.py
```diff
@@ -217,6 +223,31 @@ def override(self, preset: Preset) -> Generator[Preset, None, None]:
...
+                if missing:
+                    missing_str = ", ".join(f"’{m}’" for m in missing)
```
flying-sheep:
```suggestion
                    missing_str = ", ".join(f"‘{m}’" for m in missing)
```

tests/test_settings.py
```diff
@@ -17,3 +21,21 @@ def test_resets(_attempt: int) -> None:
 def test_set_figure_params_warns() -> None:
     with pytest.warns(FutureWarning, match=r"scanpy\.set_figure_params"):
         sc.settings.set_figure_params()
+
+
+def test_preset_scanpy_v2_preview_checks_deps() -> None:
+    dist_to_module = {d: m for m, ds in packages_distributions().items() for d in ds}
+    _scanpy2_deps_missing = any(
+        find_spec(dist_to_module.get(r.name, r.name.replace("-", "_"))) is None
```
flying-sheep:
```suggestion
    dists = {d for m, ds in packages_distributions().items() for d in ds}
    _scanpy2_deps_missing = any(
        r.name in dists
```

flying-sheep (approved):

flying-sheep:
huh, I didn’t notice that you undid my requested change here: https://github.com/scverse/scanpy/pull/4055#discussion_r3092691504

Why?

ilan-gold:
It didn't work (hence the failing tests) - I just went back to the thing I had before.  I suppose I could have thought harder and baked the lookup into `dists = { }` so that `r.name in dists` worked as expected, but I just figured "revert."  Sorry

flying-sheep:
I fixed it in #4069, the reason it didn’t work was due to a lack of canonicalization of names. I asked you to do it just with distribution names because I don’t want the complexity of their relationship to import packages to play a role at all.

PR #4056, opened 2026-04-11
docs: add inline example plot to sc.pl.pca_variance_ratio

Ekin-Kahraman:
Part of #1664. Adds rendered plot examples to `sc.pl.pca_variance_ratio` showing:
1. Default variance ratio plot on PBMC 3k
2. Log-scale variant

Follows the same pattern as #4032 (merged).


flying-sheep:
thanks!

Issue #3991, opened 2026-02-26
seaborn is broken with pandas 3: Violin plots squished, points misaligned


jaro-slamecka:
### Please make sure these conditions are met

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of scanpy.
- [ ] (optional) I have confirmed this bug exists on the main branch of scanpy.

### What happened?

Hi, I'd appreciate any ideas about an issue with violins being squished and points not lining up. This started after a recent update of my mamba/conda environment but worked fine before. I tried searching for similar issues here and also on Seaborn's page but don't have any leads and can't say I'm well-versed with plotting in Python yet.

Thanks!

<img width="30650" height="7441" alt="Image" src="https://github.com/user-attachments/assets/6e048a25-3279-4530-ba9d-c7cf7c46a800" />

### Minimal code sample

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "scanpy@git+https://github.com/scverse/scanpy.git@main",
#   "anndata==0.12.6"
# ]
# ///
#
# This script automatically imports the development branch of scanpy to check for issues

import scanpy as sc
import anndata as ad

adata = sc.datasets.pbmc68k_reduced()
sc.pl.violin(adata, ["n_genes", "n_counts", "percent_mito", "S_score"], groupby="bulk_labels", rotation=90)
```

### Error output

```pytb

```

### Versions

<details>

```
scanpy	1.12
anndata	0.12.6
----	----
ipykernel	6.29.5
cloudpickle	3.1.2
natsort	8.4.0
matplotlib	3.10.8
session-info2	0.4
colorama	0.4.6
PyQtWebEngine	5.15.7
psutil	7.2.2
ipywidgets	8.1.8
sphinxcontrib-applehelp	2.0.0
executing	2.2.1
pyzmq	27.1.0
beautifulsoup4	4.14.3
stack_data	0.6.3
scikit-learn	1.8.0
PyYAML	6.0.3
sphinxcontrib-serializinghtml	1.1.10
six	1.17.0
prompt_toolkit	3.0.52
joblib	1.5.3
platformdirs	4.9.2
numba	0.63.1
debugpy	1.8.20
seaborn	0.13.2
sphinxcontrib-qthelp	2.0.0
typing_extensions	4.15.0
packaging	26.0
pyxdg	0.28
spyder-kernels	3.1.3
jupyter_client	7.4.9
setuptools	82.0.0
[...]

PR #4057, opened 2026-04-11
docs: add inline example plot to sc.pl.rank_genes_groups_stacked_violin

Ekin-Kahraman:
Part of #1664. Converts existing doctest to rendered plot showing top marker genes per group on PBMC 68k reduced.


flying-sheep:
Thanks! This one doesn’t look right because of a seaborn bug, see #3991: https://icb-scanpy--4057.com.readthedocs.build/en/4057/api/generated/scanpy.pl.rank_genes_groups_stacked_violin.html

PR #4058, opened 2026-04-11
docs: add inline example plot to sc.pl.rank_genes_groups_violin

Ekin-Kahraman:
Part of #1664. Shows violin distribution of top-ranked genes on PBMC 68k reduced.


flying-sheep:
thank you for all these!

PR #4059, opened 2026-04-11
docs: add inline example plot to sc.pl.correlation_matrix

Ekin-Kahraman:
Part of #1664. Shows group correlation matrix with dendrogram on PBMC 68k reduced.


flying-sheep:
thank you!

PR #4062, opened 2026-04-15
perf: `numba` based aggregations for sparse data

ilan-gold:
Now that I know #4013 won't be hurt if we use the acceleration in this PR in-memory, I'm redoing #4041 against `main` with a standalone benchmark.

- [ ] Closes #

### DIFF
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -88,12 +88,13 @@ class Aggregate:
         """
         return self._sum(data=self.data, power_of_2=power_of_2)
 
-    def _sum(self, *, data: np.ndarray | CSBase, power_of_2: bool = False) -> np.ndarray:
+    def _sum(
+        self, *, data: np.ndarray | CSBase, power_of_2: bool = False
+    ) -> np.ndarray:
 
         if isinstance(data, np.ndarray):
             return utils.asarray(
-                self.indicator_matrix
-                @ (_power(data, 2) if power_of_2 else data)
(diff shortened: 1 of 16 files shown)


src/scanpy/get/_aggregated.py
```diff
@@ -97,10 +112,7 @@ def mean(self) -> Array:
         Array of mean.
 
         """
-        return (
-            utils.asarray(self.indicator_matrix @ self.data)
-            / np.bincount(self.groupby.codes)[:, None]
-        )
```
flying-sheep, line 103: do you think it would make sense to avoid re-executing sum by having basically `_sum_mean` and `_sum_mean_var` and using that in `aggregate_array`?

Or is `sum` so fast that re-executing it is fine?

flying-sheep (approved):
awesome! do these kernels come from other work (annbatch?) or did you make them just now?

ilan-gold:
> awesome! do these kernels come from other work (annbatch?) or did you make them just now?

Just made them on the spot! Saw a problem (un-parallelized aggregation used in two-pass seurat HVG #4013  causing a performance regresion) and a solution (parallel kernels in numba).

src/scanpy/get/_aggregated.py
ilan-gold: I think for now, this is probably fine.  But it's a good point, no doubt!  This PR was just focused on the current implementation.  The perf difference between having 2 sum calls vs 1 in `mean_var` wasn't that huge so it probably is "very fast" as you say

PR #4070, opened 2026-04-17
perf: Combat perf improvements

ilaykav:
- [ ] Closes # - Perf Improvement
- [x] [Tests][] included or not required because: all tests pass,  Added combat benchmark
- [x] [Release notes][] not necessary because: internal perf improvement

Devloped on venv, ran benchamrk locally, main changes:
- Replaced np.dot(x, np.ones(...)) with numpy broadcasting ([:, np.newaxis]) 
- Extract `.values` from DataFrames before computation loops
Benchmarked locally on 50K cells × 500 genes, 2-3 speedup on pp.combat


benchmarks/benchmarks/tools.py
```diff
@@ -44,3 +44,28 @@ def time_rank_genes_groups(self) -> None:
...
+        # assign cells to 3 batches deterministically
+        np.random.seed(0)
+        adata.obs["batch"] = np.random.choice(["A", "B", "C"], size=adata.n_obs)
+        adata.write_h5ad("adata_combat.h5ad")
+
+    def setup(self) -> None:
+        self.adata = ad.read_h5ad("adata_combat.h5ad")
+
+    def time_combat(self) -> None:
+        sc.pp.combat(self.adata, key="batch")
```
flying-sheep: since combat is in `.pp`, the suite should go into one of the `preprocessing_*` files, not the `benchmarks/tools.py`, which is for `.tl`

ilaykav:
~1.5x-2x times faster on 3K-20K cells, measured locally on my Intel i7-8665U python 3.14

benchmarks/benchmarks/tools.py
```diff
@@ -44,3 +44,28 @@ def time_rank_genes_groups(self) -> None:
...
+
+    def setup_cache(self) -> None:
+        import numpy as np
+
+        adata = pbmc3k()
+        sc.pp.highly_variable_genes(adata, n_top_genes=500)
+        adata = adata[:, adata.var["highly_variable"]].copy()
+        sc.pp.scale(adata, max_value=10)
+        # assign cells to 3 batches deterministically
+        np.random.seed(0)
```
flying-sheep: we don’t use that API anymore, we use methods on `np.random.Generator` instances like `np.random.default_rng()`

src/scanpy/preprocessing/_combat.py
```diff
@@ -105,28 +105,30 @@ def _standardize_data(
 
     design = _design_matrix(model, batch_key, batch_levels)
 
+    # use numpty .values extration only once to avoid pandas overhead
+    design_arr = design.values
```
flying-sheep: `.values` is deprecated, this should be `.to_numpy()` both in the comment and the code (also in the rest of your changes)

src/scanpy/preprocessing/_combat.py
```diff
@@ -105,28 +105,30 @@ def _standardize_data(
...
         print(f"Found {np.sum(var_pooled == 0)} genes with zero variance.")
-    stand_mean = np.dot(
-        grand_mean.T.reshape((len(grand_mean), 1)), np.ones((1, int(n_array)))
-    )
-    tmp = np.array(design.copy())
+    stand_mean = grand_mean[:, np.newaxis]
+    tmp = design_arr.copy()
     tmp[:, :n_batch] = 0
-    stand_mean += np.dot(tmp, b_hat).T
+    stand_mean = stand_mean + np.dot(tmp, b_hat).T
```
flying-sheep: why this change? this should mean exactly the same, right?

flying-sheep (changes requested):
hi, thanks! there are a few superficial issues.

About the general idea, why do you think that `to_numpy()` influences anything? For pandas arrays that already wrap underlying numpy arrays, the method should have basically no performance impact.

src/scanpy/preprocessing/_combat.py
ilaykav: Sadly Numpy broadcast's syntax for += (\_\_iadd\_\_) doesn't support changing the result dimensions..
https://stackoverflow.com/questions/73468760/different-dimension-numpy-array-broadcasting-issue-with-operator

src/scanpy/preprocessing/_combat.py
ilaykav: Thanks, will fix

benchmarks/benchmarks/tools.py
ilaykav: Will fix 👍

benchmarks/benchmarks/tools.py
ilaykav: Thanks, will fix

ilaykav:
You're right, the broadcasting itself (and the memory save from removing the np.ones allocations, not the .to_numpy()) is what actually made the impact, dropping the `.to_numpy()` change

src/scanpy/preprocessing/_combat.py
flying-sheep: got it! so this is where the array becomes 2D now!

flying-sheep (approved):

flying-sheep:
thank you!

PR #4073, opened 2026-04-17
feat: add `keep_sparse` parameter for aggregate and optimize it

### LINKED ISSUE #4074: `sc.get.aggregate` memory leak for Dask array

### What happened?

When working with large data (>8M cells) and trying to generate many pseudobulks (>100k), I noticed that the memory usage explodes, which defeats the purpose of using dask over loading everything in memory.

### Minimal code sample

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
# ]
# ///
#
# This script automatically imports the development branch of scanpy to check for issues

import scanpy as sc
import anndata as ad

adata = ad.experimental.read_lazy(<large file with >8M cells>)
adata.obs = adata.obs.to_memory()

adata.obs['group'] = adata.obs[['donor_id', 'cluster']].astype(str).agg('-'.join, axis=1)
pb_data = sc.get.aggregate(adata, 'group', 'sum')

pb_data.layers['sum'].compute()
```

### Error output

```pytb
There appear to be 2 leaked semaphore objects to clean up at shutdown
```

### Versions

<details>

```
scanpy  1.12.1
----    ----
wrapt   2.1.2
stack_data      0.6.3
fast-array-utils        1.4.1
matplotlib      3.10.8
anndata 0.12.10
scikit-learn    1.8.0
packaging       26.1
parso   0.8.6
h5py    3.16.0
traitlets       5.14.3
cycler  0.12.1
six     1.17.0
pyarrow 23.0.1
psutil  6.1.1
legacy-api-wrap 1.5
typing_extensions       4.15.0
MarkupSafe      3.0.3
prompt_toolkit  3.0.52
scipy   1.16.3
Deprecated      1.3.1
PyYAML  6.0.3
ipython 9.12.0
executing       2.2.1
pytz    2026.1.post1
numcodecs       0.15.1
msgpack 1.1.2
sparse  0.18.0
colorama        0.4.6
tblib   3.2.2
natsort 8.4.0
joblib  1.5.3
threadpoolctl   3.6.0
setuptools      82.0.1
decorator       5.2.1
pure_eval       0.2.3
python-dateutil 2.9.0.post0
jedi    0.19.2
pyparsing       3.3.2
asttokens       3.0.1
asciitree       0.3.3
pillow  12.2.0
xarray  2026.4.0
numba   0.65.0
Pygments        2.20.0
cloudpickle     3.1.2
kiwisolver      1.5.0
llvmlite        0.47.0
numpy   2.4.4
session-info2   0.4.1
zarr    2.18.7
pandas  2.3.3
Jinja2  3.1.6
wcwidth 0.6.0
dask    2024.7.1
toolz   1.1.0
----    ----
Python  3.12.13 | packaged by conda-forge | (main, Mar  5 2026, 16:50:00) [GCC 14.3.0]
OS      Linux-4.18.0-553.33.1.el8_10.x86_64-x86_64-with-glibc2.28
CPU     128/128 logical CPU cores, x86_64
GPU     No GPU found
```

</details>

mumichae:
Adjusted the dask call to use dask.delayed, which has massively improved the memory footprint on my end. I also threw in some little adjustments for making use of sparsity and returning sparse matrices, since even aggregated data can be sparse.

- [x] Closes #4074

- [x] [Release notes][] not necessary because:


ilan-gold:
> Adjusted the dask call to use dask.delayed, which has massively improved the memory footprint on my end. I also threw in some little adjustments for making use of sparsity and returning sparse matrices, since even aggregated data can be sparse.

Are you sure the `dask.delayed` is responsible for the reduced memory footprint? It seems like the later would be more responsible for that. This would be quite surprising to me if `dask.delayed` helped with this.  Do you have a sense for why?

src/scanpy/get/_aggregated.py
```diff
@@ -92,17 +107,26 @@ def _sum(self, data: ArrayT):
         (agg_sum_csr if isinstance(data, CSRBase) else agg_sum_csc)(
             self.indicator_matrix, data, out
         )
+        if keep_sparse and isinstance(data, CSBase):
+            return type(data)(out)  # convert to sparse type of input
```
ilan-gold, line 111: Interesting, I would have assumed aggregations of `sum` would produce _something_ for all features - you have features with literally 0 value in a feature across some categories? Maybe with pseudobulk, I could see that.

Maybe we should try out a kernel that creates a `coo`-like data structure instead as new category-feature combinations are seen which is then csr-ed instead of doing this sparsification.  Or category-by-category the sparse-ification happens instead (i.e., allocate a buffer up front per-category, and then sparse-ify in `numba`). Or a mix between the two - allocate enough memory for a max-sized coo-matrix, fill it up as far as you need, then drop the unused allocation.  At the end, concatenate all the results and call `tocsr` before returning.

I would think sparsifying to be a relatively expensive operation that doesn't do much for a lot of situations (where you don't have such small categories so that feature is non-zero when summed).

ilan-gold (commented):
Ok I see why we need `delayed` now - there is 3D CSR matrix so our old trick of adding a dimension i.e, `return res[None, :] if unchunked_axis == 1 else res` and then summing over hte first dimension doesn't work.  It's possible we can work around that.  Let's see, otherwise this PR makes a lot of sense.  Thanks @mumichae !

flying-sheep:
@ilan-gold you mean “there is *no* 3D CSR matrix”, right?

@mumichae needs a release note mentioning the new parameter and optimization

ilan-gold:
> @ilan-gold you mean “there is no 3D CSR matrix”, right?

Yes @flying-sheep but potentially she could add a `sparse` extra for 3d sparse matrices: https://github.com/dask/dask/blob/ba5045e001d67decd6a83a0f50f666a5e3845250/dask/array/chunk_types.py#L123-L126C25 which are supported by dask: https://sparse.pydata.org/en/latest/examples/dask_example/#create-arrays

PR #3184, opened 2024-08-01
Add key_added to umap, tsne, and pca

flying-sheep:
- [x] Closes #1861
- [x] Tests included or not required because:

- [ ] Release notes not necessary because:

PR #4076, opened 2026-04-20
feat: `base` without `X_`

### LINKED ISSUE #3188: Don't prepend `"X_"`

flying-sheep:
- [x] Closes #3188

TODO:

- https://github.com/scverse/scanpy/blob/e6e9647386ad8eed937997cdf4f1a34bfb784a72/src/scanpy/plotting/_tools/scatterplots.py#L858
- https://github.com/scverse/scanpy/blob/e6e9647386ad8eed937997cdf4f1a34bfb784a72/src/scanpy/plotting/_utils.py#L1034
- …


src/scanpy/tools/_utils.py
```diff
@@ -51,30 +51,30 @@ def _choose_representation(
...
         f"You’re trying to run this on {adata.n_vars} dimensions of `.X`, "
-        "if you really want this, set `use_rep='X'`.\n         "
+        "if you really want this, set `use_rep=’X’`.\n         "
         "Falling back to preprocessing with `sc.pp.pca` and default params."
     )
     warn(msg, UserWarning)
     n_pcs_pca = n_pcs if n_pcs is not None else settings.N_PCS
     pca(adata, n_comps=n_pcs_pca)
-    return adata.obsm[settings.preset.pca.key_added or "X_pca"]
+    return adata.obsm[_pca_keys(adata)[0]]
```
ilan-gold: Not sure what the intention is here but it seems to be failing

ilan-gold (commented):
Are these the only two places that this is the case anymore i.e., prepend `_X`? It seems to have been done for `PCA`, but not UMAP: https://github.com/scverse/scanpy/blob/d96d91de3162f29d901194ac56fd732459389784/src/scanpy/tools/_umap.py#L147 or am I misreading this?

docs/release-notes/4076.feat.md
```diff
@@ -0,0 +1 @@
+Add preset support to {func}`scanpy.tl.diffmap`’s and {func}`scanpy.tl.draw_graph`’s `key_added` parameter {smaller}`P Angerer`
```
ilan-gold: Relnote needs updating

src/scanpy/tools/_ingest.py
```diff
@@ -554,10 +557,10 @@ def to_adata_joint(
...
-        if "X_umap" in self._obsm:
-            adata.uns["umap"] = self._adata_ref.uns["umap"]
-        if "X_pca" in self._obsm:
-            adata.uns["pca"] = self._adata_ref.uns["pca"]
-            adata.varm["PCs"] = self._adata_ref.varm["PCs"]
+        if keys := _existing_preset_keys(self._adata_ref, _umap_keys):
+            adata.uns[keys[1]] = self._adata_ref.uns[keys[1]]
+        if keys := _existing_preset_keys(self._adata_ref, _pca_keys):
+            adata.varm[keys[1]] = self._adata_ref.varm[keys[1]]
+            adata.uns[keys[2]] = self._adata_ref.uns[keys[2]]
```
ilan-gold: I'll just pick this one out but would it make sense to have all the second-argument callables return a `TypedDict` that has standardized locations (i.e., keys) like `uns` and `varm` so that we can do `adata.uns[keys["uns"]] = adata.uns[keys[1]] = self._adata_ref.uns[keys["uns"]]` for example?

ilan-gold (commented):
My only real comment is both concrete and exploratory - I think stronger types make sense here regardless of what I say below (after all, it should be nicer than indexing into a tuple with an integer?).

Furthermore, if we have more strongly typed returns, we could also investigate differing key names more cleanly.  So instead of the present controlling one central key i.e., `adata.varm["pca"]` and `adata.obsm["pca"]` being identical, it could control all the keys (maybe?).  This is really high-level, but my motivating thought here is that the different keys currently have the nice property of being a bit more readable. For example, `adata.varm["PCs"]`, while a little clunky, represents "principal components in gene space" meaningfully (compared to `adata.varm["pca"]`).

Thoughts?

src/scanpy/tools/_ingest.py
flying-sheep: true, better than tuples for sure

flying-sheep:
> the different keys currently have the nice property of being a bit more readable

I agree, `varm["PCs"]` is more informative than `varm["pca"]`. That does apply only
- to the default (I think it’s cleaner to directly use `key_added` when specified)
- to PCA (the others don’t have a `varm` key)

So the question boils down to:
> should we leave differing keys when calling `sc.pp.pca()` while ScanpyV2Preview is enabled?

Or did I miss something?

ilan-gold:
> to PCA (the others don’t have a varm key)

This applies to diffmap too, for example:

https://github.com/scverse/scanpy/pull/4076/changes#diff-f25e128747ab8dc77cf527f9e972df2e5ddb32aa14f037d6d4b2d9219a99f397R70-R76

i.e., `evals` really mean "eigenvalues" which is different than `X_diffmap` in `obsm`

flying-sheep:
right, and e.g. neighbors also adds things like `adata.obsp['distances' | f'{key_added}_distances']` which means there’s no guarantee that `key_added` is used verbatim everywhere. (But also the actual reason for it doing so is that it adds two `obsm` entries)

Problem with PCA is that the semantics already happened in 1.11: https://github.com/scverse/scanpy/pull/3184

I think it would get really complicated for users to keep backwards compatibility but also re-introduce `PCs` somehow/

ilan-gold:
> Problem with PCA is that the semantics already happened in 1.11: https://github.com/scverse/scanpy/pull/3184
> I think it would get really complicated for users to keep backwards compatibility but also re-introduce PCs somehow/

I guess my question was perhaps more to understand and not prescribe - depending on how hard we lean in to exported refpaths, things like `adata[pc_loadings]` becomes feasible if `pc_loadings = A.varm["pca"]`.

src/scanpy/tools/_dpt.py
```diff
@@ -142,12 +156,14 @@ def dpt(
...
             "    adata.var['xroot'] = adata[root_cell_name, :].X"
         )
-    if "X_diffmap" not in adata.obsm:
+    if not diffmap_key and not _existing_preset_keys(adata, "diffmap"):
         logg.warning(
             "Trying to run `tl.dpt` without prior call of `tl.diffmap`. "
             "Falling back to `tl.diffmap` with default parameters."
         )
-        _diffmap(adata, neighbors_key=neighbors_key, rng=_LegacyRng(0))
+        from ._diffmap import diffmap
```
ilan-gold: This is a circular import (well, not exactly) but in any case, the `diffmap` from `_diffmap.py` actually comes from this file

src/scanpy/_keys.py
```diff
@@ -0,0 +1,126 @@
...
+) -> _EmbeddingKeys:
+    if key_added_ext is not None:
+        msg = "Passing `key_added_ext` is deprecated, use `key_added`'s template functionality instead."
+        warn(msg, category=FutureWarning)
+        suffix = key_added_ext
+    else:
+        suffix = layout
+    if key_added is None:
+        return _EmbeddingKeys("draw_graph", f"X_draw_graph_{suffix}")
+    formatted = key_added.format(layout=suffix)
```
ilan-gold, line 97: Do you think the failure message/exception here is clear enough if `layout` is not present?

ilan-gold (approved):

src/scanpy/tools/_dpt.py
flying-sheep: Yeah, I did this to make use of the un-underscored version’s default handling. Reorganized a bit, still circular though.

src/scanpy/_keys.py
flying-sheep: There is none! `"".format(foo=1) == ""`.

So you can just do `sc.tl.draw_graph(adata, key_added="x")` and it works like any other function.

PR #4081, opened 2026-04-21
feat: scverse-misc settings, set default `n_jobs=4`

flying-sheep:
I think having a handful of jobs is a better default than -1 for number of cores, no?

- [x] step towards #2387

- [x] [Release notes][] not necessary because: internal change only

TODO:
- [x] simplify logging stuff. `settings.logpath` can probably just be a deprecated alias to `Path(settings.logfile.name)`, and all the side effects can happen in [model validators](https://pydantic.dev/docs/validation/latest/concepts/validators/#model-validators)


flying-sheep:
OK, only issue is that the autosummary tables on [the settings page](https://icb-scanpy--4081.com.readthedocs.build/en/4081/api/settings.html) don’t have individual settings docstrings anymore.

src/scanpy/_settings/__init__.py
```diff
@@ -34,324 +46,185 @@
...
-        return wrapped
+def _is_run_from_ipython() -> bool:
+    """Determine whether we're currently in IPython."""
+    import builtins
 
-    return decorator
+    return getattr(builtins, "__IPYTHON__", False)
 
 
 # `type` is only here because of https://github.com/astral-sh/ruff/issues/20225
```
ilan-gold:
```suggestion
```

src/scanpy/_settings/__init__.py
```diff
@@ -34,324 +46,185 @@
...
+        "no_gate",
+        "?",
+    })
+    """Categories that are omitted in plotting etc."""
 
-    @n_jobs.setter
-    @_type_check_arg2(int)
-    def n_jobs(cls, n_jobs: int) -> None:
-        cls._n_jobs = n_jobs
+    _frameon: bool
```
ilan-gold: Why can't this and those below be initialized directly here? Why the post init?

ilan-gold (commented):
Nice `logpath` / `logfile` simplification!

src/scanpy/_settings/__init__.py
flying-sheep: great point, I think I misinterpreted the docs, [which say](https://pydantic.dev/docs/validation/latest/concepts/models/#private-model-attributes):

> Attributes whose name has a leading underscore are not treated as fields by Pydantic, and are not included in the model schema. Instead, these are converted into a “private attribute” which is not validated or even set during calls to `__init__`, `model_validate`, etc.

which I took at face value, even though the following example demonstrate that only applies *unless you use `pydantic.PrivateAttr`, which does* allow to create private attributes with a `default` or `default_factory`.

src/scanpy/_settings/verbosity.py
```diff
@@ -67,6 +68,7 @@ def level(self) -> int:
         m = getLevelNamesMapping()
         return m[_VERBOSITY_TO_LOGLEVEL[self.name]]
 
+    @deprecated("Use `scanpy.settings.override(verbosity=...)` instead")
```
ilan-gold, line 71: Just saw that this was marked for 1.13.  If it's a dev change, why not put it in the current series so people get this warning now?

flying-sheep: Because the `override` method is the feature that makes this deprecation possible.

I don’t like seeing deprecations without replacements.

I guess we could backport this PR, but as said, override *is* a feature.

flying-sheep:
test failures should only be https://github.com/scverse/scanpy/pull/4106

src/scanpy/_settings/verbosity.py
ilan-gold: Ok, fair enough!

ilan-gold (approved):

PR #4082, opened 2026-04-23
fix: limit Numba threads in Wilcoxon path of rank_genes_groups

### LINKED ISSUE #2390: Call `numba.set_num_threads` when `n_jobs` is specified (e.g. in `scanpy.tl.rank_genes_group`)
- [x] Additional function parameters / changed functionality / changed defaults?

Hi ScanPy team

I emailed @ivirshup but others should be involved I think. 

This function would be useful if we could specify the number of threads to use: https://scanpy.readthedocs.io/en/stable/generated/scanpy.tl.rank_genes_groups.html

Based on the number of items in the "groupby" field, we could use a basic split-merge approach here: each thread would take several of these items, the calculations are entirely independent of one another, and then when each is completed we would join + concatenate the results. 

I'm happy to help write up a PR help (or participate), but I'd like to hear if this is something you'd be willing to prioritize. (It's related to a project whereby Fabian is the PI.)

Best, Evan

JhonatanFelix:
## Summary

This PR limits Numba thread usage during the wilcoxon path of `scanpy.tl.rank_genes_groups` by temporarily applying `scanpy.settings.n_jobs` via `numba.set_num_threads(.....)` and restoring the previous value afterward.

## Implementation

- added a private `_numba_thread_limit(....)` context manager in `scanpy.tools._rank_genes_groups`
- used `numba.get_num_threads()` / `numba.set_num_threads(...)`
- clamped the requested value to the valid range
- restored the previous thread count in a `finally` block
- applied the limit only when `method == "wilcoxon"`

## Required checks

- [x] Closes #2390 

Added some focused tests for the new numba thread-limiting behavior in `rank_genes_groups`:
  - temporary restoration of the previous numba thread count
  - restoration on exception
  - clamping to the configured numba maximum
  - applying the thread limit for `method="wilcoxon"` using `scanpy.settings.n_jobs`
  - not applying the thread limit for unrelated methods
 

- [x] [Release notes][] not necessary because: It is a small internal behavior fix, it does not change the public API or user workflow.


flying-sheep:
Looks great! Please merge https://github.com/JhonatanFelix/scanpy/pull/1

JhonatanFelix:
Done! @flying-sheep

flying-sheep:
wonderful, thank you!

PR #4087, opened 2026-04-24
style: add NPY lints to prevent use of legacy `np.random`

flying-sheep:
One more for https://github.com/scverse/anndata/issues/1241

Basically to move to `np.random.Generator`:
- `rand` and `randn` are almost aliases for `random` and `standard_normal`, except that they have a worse API (`f(*shape)` instead of the more common `f(shape)` or `f(..., *, size=shape)`)
- `randint` is renamed to `integers`
- the rest has the same name

There were a few leftover `np.random.seed` calls:
- some in the tests could just be removed, they did nothing?
- one was left in the tests (`test_notebooks.py::test_pbmc`) where it basically tests that the backwards compat defaults work as expected
- I didn’t port some of the old `sim` code last time, now I did. I wonder if anyone uses it.

- [ ] Closes #

- [x] [Release notes][] not necessary because: linting only

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -167,6 +167,7 @@ lint.select = [
     "ICN",    # Follow import conventions
     "ISC",    # Implicit string concatenation
     "N",      # Naming conventions
+    "NPY",    # Numpy-specific rules
     "PD",     # Pandas
     "PERF",   # Performance
     "PIE",    # Syntax simplifications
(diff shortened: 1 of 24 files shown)


ilan-gold (commented):
Modulo the failing CI, this seems reasonable

ilan-gold (approved):

flying-sheep:
That was a bug in `_delegate`! The only method that was really renamed between `RandomState` and `Generator` is `randint` → `integers`, and I messed up doing that replacement properly. (the more sensible replacements for `rand` and `randn` already exist on `RandomState`)

PR #4088, opened 2026-04-24
Backport PR #4082 on branch 1.12.x (fix: limit Numba threads in Wilcoxon path of rank_genes_groups)

flying-sheep:
Backport PR #4082


zboldyga:
hey @flying-sheep I was just doing some benchmarking of another change I'm suggesting to the library for wilcoxon DE and I noticed the default behavior of the library has changed (as compared with published v 1.12.0 -- it defaults to using a single thread for wilcoxon DE now when n_jobs is unspecified). I think it's this PR

I believe the prior default was to use all available threads, which seems to make more sense in the default case?

PR #4105, opened 2026-05-07
chore: SPEC0 2026 Q2 updates

flying-sheep:
can’t drop zarr 2 yet since that would mean depending on anndata ≥0.12, and it’s not even time to drop 0.11 yet, I think: 0.11.0 released in November 2024, and all the current SPEC0 deprecations happend latest in June 2024.

what do you think @ilan-gold?

- [ ] Closes #

- [x] [Release notes][] not necessary because: dep updates


ilan-gold (commented):
Makes sense, that's the point of all this stuff, to keep us consistent and smooth!

ilan-gold (approved):

flying-sheep:
yeah, the gains from dropping anndata 0.11 aren’t big anyway. I think it’s great that we’re in an era of very few anndata compat hacks, because there are changes upcoming!

PR #4108, opened 2026-05-07
add backends

Intron7:
## Summary

- Add Scanpy backend dispatch for RAPIDS single-cell across preprocessing, tools, and `sc.get.aggregate`.
- Require `scverse-backends>=0.0.3` and exercise its lazy entry-point discovery and automatic backend-parameter signature injection.
- Rebase the branch onto current `main`, preserving upstream API changes alongside dispatch decorators.

## Validation

- `uv run --group test pytest -q tests/test_backends.py tests/test_package_structure.py::test_sig_conventions`
- `uv run --group test pytest -q tests/test_aggregated.py tests/test_pca.py tests/test_embedding.py tests/test_rank_genes_groups.py`


src/scanpy/_settings/__init__.py
```diff
@@ -170,6 +170,37 @@ def model_post_init(self, context: object) -> None:
...
+    def backend(self, backend: str) -> None:
+        from .._backends import settings as backend_settings
+
+        backend_settings.backend = backend
+
+    def use_backend(self, backend: str):
+        """Temporarily set the active computational backend."""
+        from .._backends import settings as backend_settings
+
+        return backend_settings.use_backend(backend)
```
flying-sheep: this should be removed, `sc.settings.override(backend=...)` exists, and if it doesn’t work for some reason, we should fix that

src/scanpy/_settings/__init__.py
```diff
@@ -170,6 +170,37 @@ def model_post_init(self, context: object) -> None:
...
+        """Return canonical names of installed computational backends."""
+        from .._backends import settings as backend_settings
+
+        return backend_settings.available_backends()
+
+    def get_backend(self, name: str):
+        """Look up an installed computational backend by name or alias."""
+        from .._backends import settings as backend_settings
+
+        return backend_settings.get_backend(name)
```
flying-sheep: Why does this exist? Does the user ever need to access the backend object?

src/scanpy/_backends.py
```diff
@@ -0,0 +1,42 @@
...
+from __future__ import annotations
+
+from scverse_backends import BackendDispatcher
+
+dispatcher = BackendDispatcher(
+    entrypoint_group="scanpy.backends",
+    host_name="scanpy",
+    trusted_backends={
+        "rapids_singlecell": {
+            "aliases": ["cuda", "rapids", "rapids-singlecell"],
```
flying-sheep: Didn’t you just rename it to this? I like the convention of using kebap-case to refer to things that aren’t Python parameters / variables.

```suggestion
        "rapids-singlecell": {
            "aliases": ["cuda", "rapids", "rapids_singlecell"],
```

tests/test_backends.py
```diff
@@ -0,0 +1,196 @@
...
+    value = "rapids_singlecell.backends.scanpy:ScanpyBackend"
+    dist = FakeDistribution()
+
+    @staticmethod
+    def load():
+        """Load the fake backend entry point."""
+        return FakeRapidsBackend
+
+
+DISPATCHED_FUNCTIONS = [
```
flying-sheep: I don’t want to keep this in sync. Can’t the decorator just populate a list that we can access here?

One thing to note is that it probably can’t just put its function argument in the list, as other decorators might be applied after this one.

pyproject.toml
```diff
@@ -67,6 +67,7 @@ dependencies = [
     "pynndescent>=0.5.13",
     "scikit-learn>=1.6",
     "scipy>=1.15",
+    "scverse-backends>=0.0.3,<0.1",
```
flying-sheep: this has no dependencies right?

flying-sheep (commented):
looks good! I didn’t look at the test yet (I will once you mark it ready for review)

pyproject.toml
Intron7: No None. I'll also remove the pin

src/scanpy/_backends.py
```diff
@@ -0,0 +1,57 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+dispatcher = BackendDispatcher(
+    entrypoint_group="scanpy.backends",
+    host_name="scanpy",
+    trusted_backends={
+        "rapids-singlecell": {
+            "aliases": ["cuda", "rapids", "rapids_singlecell"],
```
grst, line 17: Does it make sense to have that many aliases that mean the same thing? Could be confusing. 
I think aliases are mostly useful for backwards-compatibility.

PR #4118, opened 2026-05-11
Rank refactor 1

zboldyga:
- [ ] Closes #
- [ x] [Tests][] included or not required because: behavior conserved


zboldyga:
@ilan-gold here's a proof of concept for the stats speedup. 

The stats are the biggest performance improvement remaining on the scanpy side of the illico integration -- here's the total scanpy illico time before vs. after the patch.

Note that this is only relevant to vs_rest mode (all other cells). Using an individual group as a reference was already fine.

## vs_rest

| dataset        | n_cells | n_groups | n_genes | original (s) | latest (s) | speedup |
|---             | ---:    | ---:     | ---:    | ---:         | ---:       | ---:    |
| adamson16      | 62,673  | 89       | 20,616  | 65.38        | 30.31      | **2.16×** |
| wessels23      | 30,636  | 184      | 16,775  | 48.56        | 15.33      | **3.17×** |
| sunshine23     | 38,431  | 194      | 23,570  | 69.19        | 21.60      | **3.20×** |
| norman19       | 111,545 | 238      | 22,608  | 224.67       | 47.98      | **4.68×** |
| nadig25hepg2   | 133,757 | 1,819    | 9,623   | 1,838.24     | 72.96      | **25.20×** |
| replogle22k562 | 308,646 | 1,972    | 8,563   | 4,025.46     | 137.68     | **29.24×** |

I used aggregate as you mentioned.

That said, two additional points:

1. You'll notice that I preserved some prior stats computation logic for the t tests in: _compute_rest_stats_for_t_test , which means we now have different computations for t test and wilcoxon . There are two issues with significant digits / numerical stability that prevent changes to this without changing the outputs:
-- using 'aggregate' has a variance calculation that is numerically unstable, e.g. it can suffer from significant digit cancellation
-- X[~self.groups_masks_obs[g]] is inefficient, but the alternative used for the sped-up variant for wilcoxon is numerically unstable for the variance computation needed for t test.

There's a better algorithm for calculating variance that doesn't have these issues: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm . I did an implementation of this (didn't commit), and in my initial tests it was roughly the same speed as this current approach using 'aggregate' (possibly 20% faster but too early to say). I would need to think more carefully about where this fits in scanpy, e.g. it might be best as a util in get alongside aggregate, or a replacement for aggregate. So that in itself is a separate issue, perhaps it needs to be addressed before we can finish this basic stats speedup work... e.g. with that in place, I can simplify this code a bit more, and we avoid numerical stability issues.

(note that this issue already exist in 'aggregate').

2. The _RankGenes class uses side effects and it makes the code complex (e.g. functions that act on the class's self objects, it's unclear what functions require/return). I think it would be helpful to refactor this to a more pure functional approach, which also seems in line with other scanpy code I reviewed. It should be straightforward, but I wanted to check with you whether this is something you'd accept, and when and off which branch to implement this?

Thoughts on these 2 points and the current PR?

ilan-gold:
So to your points

1. I would wholesale replace the variance calculation if possible in `aggregate`.  There's a comment in the dask implementation about how strangely unstable the variance calculation can be so presumably that is less a dask thing than a "data big enough to require dask" thing i.e, triggering instability due to more summation ops, in which case, I think the calculation can be considered "wrong" and replaced.  The online algorithm looks to be compatible with a single-pass approach to calculating mean at the same time i.e., only one data access and reuse the value.  I would just replace our current sparse numba kernel for mean var calculation with one that does this.  That seems possible, right?  Does this appear in dense data as well? Presumably.
2. Yeah, this was a separate issue and predates me.  I specifically didn't open that can of worms in my other PR but yes, I think we would absolutely take a refactor of this.

zboldyga:
am still working this one, simplifying and limiting the scope properly. will push some more changes tomorrow most likely to wrap it up.

src/scanpy/tools/_rank_genes_groups.py
```diff
@@ -230,64 +247,118 @@ def __init__(
...
+        mask_all = self.grouping_mask.to_numpy().all()
+
+        # Compute rest means without another pass
+        # TODO: can any cells not be part of a group? (will mask_all ever be False)
+        # If not, remove fallback
+        sum_g = self.means * n_arr
+        sum_total = sum_g.sum(0) if mask_all else np.asarray(X.sum(axis=0)).ravel()
+        self.means_rest = (sum_total - sum_g) / n_r
+
+        # TODO: if `aggregate` exposed `sum_of_squares` (an additive
```
zboldyga: @ilan-gold thoughts on this?

Basically we need to compute the variance for the 'rest' sets somehow for t test. I kept it as old code for now, to minimize surface area. So it's slow, and mean_var does suffer from the same (rare) 'catastrophic cancellation' situation.

We could defer this to another PR to limit scope now

I figured I'd bring to your attention because exposing the sum_of_square values in aggregate enables us to do a quicker var computation here

My impression is it didn't seem like aggregate itself would be something we'd want to manipulate to allow the vs_rest logic, so it seems like for the vs_rest stats we're doing the computation in here (and exposing sum of squares reduces computation)

Additionally, we could also abstract some of the chan stuff into a small util and use that both in aggregate and here?

ilan-gold: If we implemented Welford's algorithm inside https://github.com/scverse/fast-array-utils/ for `mean_var`, would that take care of the catastrophic cancellation potential? Then it seems like we wouldn't have to worry about the numeric stability here.

So if I have that right, I would think that the route forward would be then, as you say, put that in a separate PR there and then we'll automatically start benefitting from it everywhere in the scanpy codebase.

zboldyga: @ilan-gold So the problem with t test computations in the vs_rest case is that we need the group variances as well as the 'rest' variances. If we use mean_var for rest, it means we have to do O(n_groups * rest_nnz) work just to create each 'rest' slice of X. e.g. X[~target_group] is a slow operation since that results in the majority of X being re-materialized for every group.

So basically t test computation is very slow with this approach of doing X[~target_group] and computing mean_vars each time -- 99% of the work in producing the results is in this looping over X per-group.

Performance-wise, I think there's a good case here for re-using the group level variances to produce the 'rest' variances, it should be way faster. And Chan's algo is a way to do that. And actually, now that I've been thinking this through again, I think I see a way to do this in our rank genes code without changing the API of aggregate. I will draft this up shortly to demonstrate w/ code!

That said, I do think this will raise the question of modularizing the Chan's algo logic. Because I'll basically be re-applying Chan's here for the 'rest' variance computations, just on the group axis.

This should also be quite a bit more friendly to larger datasets. I will include a quick benchmark result to clarify the speedup (but note that we can't demonstrate numerical stability until 'aggregate' is also numerically stable).

... And while it has no relevance here with the proposed path, mean_vars should really also have a more stable variance computation. It's not common someone would run into numerical stability issues, but I'm sure it will happen in some datasets, especially with a generic tool like that... So for fixing that, we can likely reuse a good chunk of the work we're doing here to patch-up that library separately.

zboldyga: @ilan-gold so the latest commit is the proof of concept for this. speedup is same as magnitude of speedup on wilcoxon in this PR (t-test took e.g. 30 seconds on a real dataset, 3 seconds with this fix). 

What we could do here is extract the chan algo into a util and use it in both aggregate and here?

ilan-gold: > What we could do here is extract the chan algo into a util and use it in both aggregate and here?

I think that would make sense roughly.  It seems that the combination step could be refactored out, absolutely.  I would encourage aggressive use of `numba` here if possible.  Overall, this is a very cool idea to reuse the means/vars for the adaptation of chan's, very cool.

I'd propose then the following, if this makes sense:

1. Make an initial PR for `scanpy.get.aggregate` to be able to do Welford's mean/var for in-memory objects.  I am not sure about the complexity of chan's algorithm for in-memory objects or the utility over just computing the groups themselves in parallel i.e., run Welford's per-group. Regardless, I would make sure the chan primitive defined [here for `dask`](https://github.com/scverse/scanpy/pull/4123/changes#diff-81a0e8b2798185f027173b3500116077f804fa7104435c0b64d931df3e829a90R472-R483) can work for in-memory objects well enough so you can reuse it here effectively. You can probably just make a PR into that PR.  This would give us `dask` for free, I _think_, in this PR as well
2. Once that is merged, Update this PR to use that clean `comb` function, just a drop-in.  
3. `numba`fy `_derive_rest_stats` for in-memory arrays which seems very amenable to parallelization in the number of groups.
4. Reuse the clean chan `comb` step to accerate dask arrays for the rest calculation via `map_blocks` or similar in this PR

Does this make sense? I am trying to break things down into steps so that we can get the wins elsewhere! The main one I am unsure about is step 4 but that's also why I think proceeding from the root (defining the utility like you mention) and moving outward is a good idea.

zboldyga: at an initial read through I think this makes sense. I'll come back to this tomorrow morning and double-check, and jump into the relevant PRs to help wrap it all up.

zboldyga:
@ilan-gold been reviewing and crunching runtimes again this morning, will come back to it later today with a small additional commit and confirm completion of this stats PR

zboldyga:
benchmarks with Pts=True and vs_rest (~worst case usage)

### Stats computation time (`_basic_stats`)

| Dataset | cells × genes | groups | Method | Before | After | Speedup |
| --- | ---: | ---: | --- | ---: | ---: | ---: |
| wessels23 | 30,636 × 16,775 | 184 | t-test | 56.6 s | 1.7 s | **33×** |
| wessels23 | 30,636 × 16,775 | 184 | wilcoxon (illico) | 64.0 s | 0.63 s | **102×** |
| replogle22k562 | 308,646 × 8,563 | 1,972 | t-test | ~12,400 s\* | 9.2 s | **~1,350×** |
| replogle22k562 | 308,646 × 8,563 | 1,972 | wilcoxon (illico) | ~8,250 s\* | 4.8 s | **~1,740×** |

### Full run time (whole `rank_genes_groups` call)

| Dataset | groups | Method | Before | After | Speedup |
| --- | ---: | --- | ---: | ---: | ---: |
| wessels23 | 184 | t-test | 58.2 s | 3.3 s | **18×** |
| wessels23 | 184 | wilcoxon (illico) | 78.4 s | 15.1 s | **5.2×** |
| replogle22k562 | 1,972 | wilcoxon (illico) | ~8,380 s (~2.3 h)\* | 135 s | **~62×** |

zboldyga:
unsure of source of remaining errors, might be flaky existing tests. will look more tomorrow

src/scanpy/tools/_rank_genes_groups.py
```diff
@@ -516,60 +652,81 @@ def compute_statistics(  # noqa: PLR0912
...
+            mean_in_log_space=mean_in_log_space,
+        )
 
-            if n_genes_user is not None:
-                self.stats[group_name, "names"] = self.var_names[global_indices]
 
-            self.stats[group_name, "scores"] = scores[global_indices]
+def _build_stats_dataframe(
+    rg,
+    results,
```
ilan-gold: Types

src/scanpy/tools/_rank_genes_groups.py
```diff
@@ -516,60 +652,81 @@ def compute_statistics(  # noqa: PLR0912
...
+        if df is None:
+            idx = pd.MultiIndex.from_tuples([(group_name, first_col)])
+            df = pd.DataFrame(columns=idx)
+
+        if n_genes_user is not None:
+            df[group_name, "names"] = rg.var_names[global_indices]
+        df[group_name, "scores"] = scores[global_indices]
+
+        if pvals is not None:
+            df[group_name, "pvals"] = pvals[global_indices]
```
ilan-gold, line 732: This means we can get rid of that `copy_pvals` argument inside `_illico_results_to_iter`?

src/scanpy/tools/_rank_genes_groups.py
```diff
@@ -232,64 +293,138 @@ def __init__(
...
+
+        Grouped by ``codes`` (values ``0 .. n_parts-1``). Returns ``(mean, var,
+        nnz)`` of shape ``(n_parts, n_genes)``, zero-filled for partitions with
+        no cells. ``var`` is ``None`` unless ``need_var``; ``nnz`` is ``None``
+        unless ``self.comp_pts``.
+        """
+        n_genes = x_used.shape[1]
+        mean = np.zeros((n_parts, n_genes))
+        var = np.zeros((n_parts, n_genes)) if need_var else None
+        nnz = np.zeros((n_parts, n_genes)) if self.comp_pts else None
```
ilan-gold: Preallocating here to account for the fact that `aggregate` drops categories with no members and `rank_genes_groups` doesn't? It feels like that should be rectified, maybe for scanpy 2.0.  I don't really get the category-dropping thing TBH in general and think it's bad practice 

cc @flying-sheep

ilan-gold (commented):
I'd like to get a benchmark in here at some point for `rank_genes_groups`, just for `illico` because the others are probably way too slow.

Overall, really like where this PR is headed

zboldyga:
@ilan-gold design question:

right now when any cells don't have a group assigned (None or NaN in the groupby column, leading to a category code of -1 after sanitization):

- we include those cells in the vs_rest populations
- we don't compute stats specifically for those cells

This is existing behavior. IMO it's not exactly clear in the docs what should happen with unassigned cells wrt 'rest' treatment. https://scanpy.readthedocs.io/en/stable/generated/scanpy.tl.rank_genes_groups.html "if 'rest', compare each group to the union of the rest of the group[s]". 

And I don't see any gate in scanpy that would prevent cells from being unassigned to a group, or warn about these cells.

So maybe there is some unintended behavior here for some users who are thinking they're eliminating cells from analysis by not assigning them to a group? Perhaps users should be forced to assign all cells to a group (even if they manually make an Unknown group or something like this).

At minimum I think it should be made more clear what the behavior here is. Perhaps also a warning to indicate that x count of cells are being included in vs_rest due to unassigned group?

src/scanpy/tools/_rank_genes_groups.py
zboldyga: yes it seems aggregate should be changed to not drop empty groups. Just a more typical behavior

ilan-gold: Yup, this was a mistake/concept that started in `anndata` but needs to be rooted out downstream

ilan-gold:
> And I don't see any gate in scanpy that would prevent cells from being unassigned to a group, or warn about these cells.

I don't follow this, what do you mean?

> So maybe there is some unintended behavior here for some users who are thinking they're eliminating cells from analysis by not assigning them to a group? Perhaps users should be forced to assign all cells to a group (even if they manually make an Unknown group or something like this).

@zboldyga I would be open to breaking this for scanpy 2.0 - we have machinery for this via `Preset` settings.  I would think `groups="all"` would refer to non-NA groups or that as you're hinting at here, everything must be assigned.  It may help to know what users expect.  At the moment, `aggregate` excludes `NA` and I am not sure I see a reason not to do that here TBH although the simplicity of "everything must be assigned" is appealing.

> At minimum I think it should be made more clear what the behavior here is. Perhaps also a warning to indicate that x count of cells are being included in vs_rest due to unassigned group?

Yes, as long as we're clear, it doesn't really matter what we do on some level

zboldyga:
hey @ilan-gold it looks like this work wasn't merged but the PR was closed? ... Do you want me to adjust this to target main?

PR #4122, opened 2026-05-14
fix float32 precision loss in aggregate variance: cast to float64 before squaring

zboldyga:
- [ ] Closes #

While working on the wilcoxon illico integration, I ran into numerical stability issues with the 'aggregate'  function.

Upon further inspection, the issues primarily stem from float32 squaring before the mean_of_sq - sq_of_mean variance calculation. It seems there was some code to cast to float64 e.g. in the csc and csr cases, but then the next line cast it right back to float32 ```value = data.data[j]```. And other cases like the Dask and dense paths were not using float64.

I'll add a writeup from AI in another message below about the error levels / empirical findings.

...

So this small fix solves a large part of the numerical stability problem. This is a sufficient-enough fix to enable the stats computation code in the wilcoxon illico work to be completed using 'aggregate', which speeds it up an OOM.

That said, 'aggregate' has further issues -- the variance calculation used here is known to have a 'catastrophic cancellation' scenario when the mean/var ratio is high, e.g. 10,000+. I will open another PR for this soon, but it's a much bigger change, probably requiring more testing and review.

**Note:** This uses float64 so it's a bit slower. I tested a few perturb-seq datasets and it was about 30% slower on the Dask path. But only about 5% slower on the CSR path, and ~10% on CSC. However, I am fairly sure once we add the welford's online approach (or similar) this will speed up.

@ilan-gold tagging you here since this resolves the issue in wilcoxon-illico work / unblocks completion of my PR on stats computation speedup

### DIFF
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -139,7 +139,15 @@ class Aggregate[ArrayT: np.ndarray | CSBase]:
         if isinstance(self.data, np.ndarray):
             mean_ = self.mean()
             # sparse matrices do not support ** for elementwise power.
-            mean_sq = self._sum(_power(self.data, 2)) / group_counts[:, None]
+            # Cast to float64 before squaring so float32 input doesn't
+            # lose precision at the squaring step (errors there feed
+            # directly into the `mean_sq - sq_mean` variance calculation).
+            data_f64 = (
+                self.data
+                if self.data.dtype == np.float64
+                else self.data.astype(np.float64)
+            )
+            mean_sq = self._sum(_power(data_f64, 2)) / group_counts[:, None]
             sq_mean = mean_**2
             var_ = mean_sq - sq_mean
         else:
@@ -372,7 +380,10 @@ def aggregate_dask_mean_var(
     dof: int = 1,
 ) -> MeanVarDict:
     mean = aggregate_dask(data, by, "mean", mask=mask, dof=dof)["mean"]
-    sq_mean = aggregate_dask(fau_power(data, 2), by, "mean", mask=mask, dof=dof)["mean"]
+    # Cast to float64 before squaring; float32 squaring at chunk level
+    # would lose precision before the variance calculation. Cast is lazy for dask.
+    data_f64 = data if data.dtype == np.float64 else data.astype(np.float64)
+    sq_mean = aggregate_dask(fau_power(data_f64, 2), by, "mean", mask=mask, dof=dof)["mean"]
     # TODO: If we don't compute here, the results are not deterministic under the process cluster for sparse.
     if isinstance(data._meta, CSRBase):
         sq_mean = sq_mean.compute()
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -383,7 +383,9 @@ def aggregate_dask_mean_var(
     # Cast to float64 before squaring; float32 squaring at chunk level
     # would lose precision before the variance calculation. Cast is lazy for dask.
     data_f64 = data if data.dtype == np.float64 else data.astype(np.float64)
-    sq_mean = aggregate_dask(fau_power(data_f64, 2), by, "mean", mask=mask, dof=dof)["mean"]
+    sq_mean = aggregate_dask(fau_power(data_f64, 2), by, "mean", mask=mask, dof=dof)[
+        "mean"
+    ]
     # TODO: If we don't compute here, the results are not deterministic under the process cluster for sparse.
     if isinstance(data._meta, CSRBase):
         sq_mean = sq_mean.compute()
--- a/src/scanpy/get/_kernels.py
+++ b/src/scanpy/get/_kernels.py
@@ -68,7 +68,6 @@ def mean_var_csr(
             for j in range(start_obs, end_obs):
                 col = data.indices[j]
                 value = np.float64(data.data[j])
-                value = data.data[j]
                 mean[cat_num, col] += value
                 var[cat_num, col] += value * value
 
@@ -102,7 +101,6 @@ def mean_var_csc(
 
             if cat != -1:
                 value = np.float64(data.data[j])
-                value = data.data[j]
                 mean[cat, col] += value
                 var[cat, col] += value * value
 


zboldyga:
Lengthy AI writeup that attempts to explain the underlying stability issue with the float32 squaring in this data regime 😅 

# Why float32 squaring is problematic for scanpy data

## The math

Variance is computed as `var = Σx²/n − μ²`. This is **a subtraction
of two nearly-equal large numbers** when variance is small relative to
the mean — and that's where precision dies.

Specifically: each input to that subtraction (`Σx²/n` and `μ²`) is
correct to roughly the *relative* precision of the floating-point
format used to compute it. The subtraction itself doesn't lose
precision — but its *result*, which is much smaller than either
operand, inherits the inputs' absolute error and so its relative
error is amplified by:

```
amplification factor  =  mean² / var
```

### Concrete numbers

| storage dtype | relative precision | digits in `Σx²/n` and `μ²` |
|---|---:|---:|
| float32 | ε ≈ 1.2e-7 | ~7 digits |
| float64 | ε ≈ 2.2e-16 | ~16 digits |

For a ribosomal gene with `mean²/var ≈ 5,000`:

- log₁₀(5,000) ≈ 3.7 → the subtraction loses ~3.7 digits.
- float32 inputs: 7 digits − 3.7 ≈ **3.3 digits** left in var.
  Relative error ≈ 5,000 × 1.2e-7 ≈ **6e-4**.
- float64 inputs: 16 digits − 3.7 ≈ **12.3 digits** left.
  Relative error ≈ 5,000 × 2.2e-16 ≈ **1e-12**.

That's the 9-orders-of-magnitude difference between the two regimes.

The fix is one cast — `value.astype(np.float64)` before squaring —
so the inputs to the cancellation step are at float64's 16-digit
precision instead of float32's 7-digit precision. The cancellation
amplifier still exists; it just has much less to amplify.

## Why scanpy's typical data lives in this regime

scanpy stores log1p-normalized counts as **float32 by default**,
with values mostly in `[0, ~10]`. In that range, common gene
categories routinely have small per-cell-type variance relative to
mean — pushing `mean²/var` high:

| gene category | log1p mean | per-cell-type var | `mean²/var` | digits left in var (float32) |
|---|---:|---:|---:|---:|
| Random / noisy genes | ~1 | ~1 | ~1 | ~7 |
| Housekeeping (ACTB, EEF1A1) | ~4 | ~0.05 | ~300 | ~4.5 |
| Mitochondrial (MT-CO*) | ~5 | ~0.04 | ~600 | ~4.2 |
| Ribosomal (RPL/RPS) | ~5 | ~0.005 | ~5,000 | ~3.3 |
| Abundant lncRNA (MALAT1) | ~6 | ~0.002 | ~18,000 | ~2.7 |

Pseudobulk-of-pseudobulks workflows (the documented use case for
`sc.get.aggregate`) compress within-group variance further, pushing
typical workloads into the 10³–10⁵ ratio range. At those ratios,
float32 inputs leave **2–4 digits** of usable precision in var; the
fix restores **~11–13 digits**.

## What this looks like empirically

On the wessels23 RPS15 pseudobulk scenario (real data,
`mean²/var ≈ 4,000`):

| | predicted | measured |
|---|---:|---:|
| float32 inputs (before fix) | ~5e-4 | **2.5e-5** |
| float64 inputs (after fix) | ~1e-12 | **5e-7** |

Predictions overshoot observed values by 10–100× because real
log1p-normalized data has *correlated* (integer-quantized) rounding
errors rather than random-walk. The order-of-magnitude shape — that
the cost of float32 inputs scales with `mean²/var` — matches the
data exactly.

The 5e-7 residual after the fix isn't kernel error anymore — it's
the float32-precision floor of the *input values themselves*, which
is the best any algorithm can do given float32-stored data.

zboldyga:
And some evidence of the resolution via float64 usage (AI writeup again here, lengthy data).

The gist is that this was impacting the results enough that it would have caused a regression if aggregated was used in wilcoxon-illico stats computation. And I'd say it is likely the float32 squaring would cause quite tangible issues when dealing with other use cases like pseudobulk-related things, manipulated data that is not standard log1p counts, or maybe newer datasets with higher library sizes.

## Reference Note
"Correct" is defined as two-pass shifted-mean variance computed at
float64. That reference was cross-checked against arbitrary-precision
`decimal.Decimal` at 50 digits: agreement to **9.5e-16** (machine
precision). All error numbers below are *relative* errors against
that reference.

## Numerical impact

### Synthetic adversarial scenarios

Each row is a single `sc.get.aggregate` call on synthetic float32
input with the noted mean / variance / N. Numbers shown are the
max relative error across the (group, gene) output grid vs the
reference. Lower is better; 1e-16 is the machine-precision floor.

| scenario | mean | var | N | current main | this PR |
|---|---:|---:|---:|---:|---:|
| housekeeping (5 groups) | 6.5 | 0.01 | 5,000 | 4.6e-6 | **6.2e-11** |
| ribosomal (184 groups) | 8.5 | 0.002 | 200 | 3.5e-4 | **1.6e-10** |
| clonal population (5 groups) | 7 | 5e-4 | 100 | 8.4e-4 | **2.6e-10** |

Reading: on the ribosomal scenario, the kernel returned variances that
were off by ~0.035% **relatively** before the fix — meaningful given
downstream consumers (e.g., t-test scores) propagate variance error
through `1/sqrt(var)`. After the fix, errors are below 1e-9.

### Real pseudobulk workloads

The same measurement on real perturb-seq pseudobulk-of-pseudobulks
data (the documented use case for `sc.get.aggregate`). "Worst gene"
means the (gene, kind) cell with the largest `mean²/var` ratio, where
the cancellation issue bites hardest. Sparse-CSR backend.

| dataset | worst gene | mean²/var | current main | this PR |
|---|---|---:|---:|---:|
| adamson16 | RPL8 | 11,635 | 1.7e-5 | **2.8e-7** |
| frangieh21 | EEF1A1 | 12,964 | 2.3e-5 | **4.4e-7** |
| tian21crispri | MALAT1 | 18,728 | 2.1e-5 | **7.0e-7** |
| wessels23 | RPS15 | 4,237 | 2.5e-5 | **7.0e-7** |

Reading: routine biological workflows (housekeeping/ribosomal/abundant
genes) reliably trigger the bug. Errors are ~10⁻⁵ before the fix —
not catastrophic in absolute terms, but enough to propagate into
downstream ULP-level rank flips. Fix reduces them by ~30–80×.

### Cross-backend confirmation

The bug is in the squaring step, which appears identically across all
four `aggregate` backends. Verified by running the same wessels23
pseudobulk scenario through each path, with float32 input cast to the
backend's native form (`sparse.csr_matrix`, `sparse.csc_matrix`,
`np.ndarray`, `dask.array`):

| backend | current main | this PR | improvement |
|---|---:|---:|---:|
| csr | 2.48e-5 | 5.16e-7 | 48× |
| csc | 2.48e-5 | 5.16e-7 | 48× |
| dense | 2.48e-5 | 5.16e-7 | 48× |
| dask | 2.48e-5 | 5.16e-7 | 48× |

Reading: identical errors across all four backends — same root cause,
same fix.

ilan-gold:
> That said, 'aggregate' has further issues -- the variance calculation used here is known to have a 'catastrophic cancellation' scenario when the mean/var ratio is high, e.g. 10,000+. I will open another PR for this soon, but it's a much bigger change, probably requiring more testing and review.

I started one for dask, will post now.  I think we should just replace the algorithm IMO

Update: https://github.com/scverse/scanpy/pull/4123

zboldyga:
sweet ok I'll close this and am happy to review or contribute on that one as needed

ilan-gold:
> sweet ok I'll close this and am happy to review or contribute on that one as needed

Your review would be extremely welcome!!!!

PR #4126, opened 2026-05-15
perf: memory usage for dask pca

ilan-gold:
This should give a very nice memory performance by not having the intermediate product of `x.T @ x` in memory.  I am not sure whether the upper-triangular performance benefit is real or not - It seems to appear real on the tahoe dataset but I'm not 100% sure given how large the dataset is and how reliant this pipeline is on i/o.

The failing test here is a warning because the kernel is not parallel so I see two options there:

1. silence in `fau`
2. Use `@njit(nogil=True, nopython=True)` manually (since this is only ever run from dask anyway)

- [ ] Closes #

### DIFF
--- a/docs/release-notes/4126.perf.md
+++ b/docs/release-notes/4126.perf.md
@@ -1,0 +1,1 @@
+Improve memory usage of {func}`scanpy.pp.pca` when run on sparse {class}`dask.array.Array` {smaller}`I Gold`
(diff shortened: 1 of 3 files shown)


flying-sheep (approved):
Very cool! Regarding `njit`, yeah, definitely just use the one from numba directly and silence the Ruff warning.

PS: We should probably add `nogil` as an option to FAU’s `njit` …

ilan-gold:
> PS: We should probably add nogil as an option to FAU’s njit …

Separate discussion, but yeah, probably.  The only thing is thread safety - I know Severin uses mutable dask blocks (i.e. mutating the block) for speed reasons but this is a big no-no.  It might be worth ensuring none of our kernels do that.

PR #4132, opened 2026-05-19
feat(plotting): add `ncols` to wrap multi-panel violin plots (#1552)

### LINKED ISSUE #1552: requesting ncols parameter in sc.pl.violin()
It would be nice to have a way to split the violin plots from `sc.pl.violin()` into a specific number of rows, similar to the `ncols` parameter in `scanpy.pl.umap()`

Hore01:
## Summary

Adds an `ncols` parameter to `sc.pl.violin` so multi-key violins can wrap into a grid instead of one giant horizontal row. Mirrors the `ncols` convention already used by `sc.pl.umap` / `sc.pl.embedding`.

Closes #1552.

## What changes

Two code paths in `violin()` get the new parameter:

- **`multi_panel=True` (seaborn `catplot` path)**, forwards `ncols` to seaborn's `col_wrap=`. The existing stripplot and rotation loops used `g.axes[0]` 2D indexing, which `col_wrap` flattens, so I swapped them to `g.axes_dict`, same semantics either way, but compatible with both shapes.
- **`groupby` set, multiple keys (legacy `setup_axes` path)** — when `ncols` is provided, builds a gridspec figure via `_panel_grid` (already used by `sc.pl.embedding`). The import is local to the function to dodge the `_tools/__init__.py` ↔ `_anndata.py` circular-import (which I hit on the first try — the package's `_tools/__init__.py` already imports `ranking`
  from `_anndata.py`, so a module-level import would cycle).

`ncols=None` (default) flows through the unchanged code, so every existing `test_violin` image baseline stays byte-identical.

## Why hard-code `hspace`/`wspace`

`embedding` exposes both; doing the same here would add API surface for no gain on the issue as filed. Defaults `(0.25, 0.1)` match `embedding`. Easy follow-up if anyone asks.

## Tests

Added one new subtest in `test_violin`:

```python
with subtests.test("ncols"):
    sc.pl.violin(pbmc, ["n_genes", "percent_mito", "n_counts"],
                 multi_panel=True, ncols=2, show=False)
    assert len(plt.gcf().axes) == 3
    save_and_compare_images("violin_ncols_wrap")
```

The `assert len(...) == 3` is cheap insurance; `tol=40` would happily mask an off-by-one in panel count, and an earlier draft of this patch had exactly that bug (the original `Path B` only made one axis when `groupby is None` because the data is melted into a single column, which is why I ended up wiring the wrap into `multi_panel=True` instead). The baseline at
`tests/_images/violin_ncols_wrap/expected.png` was generated locally and visually checked.

Full `test_violin` (all four subtests) passes locally.

## Refs

- Issue #1552 (open since 2020, `good first issue`)
- Pattern: `_panel_grid` in `src/scanpy/plotting/_tools/scatterplots.py`
- Convention: `docs/release-notes/<PR>.feat.md`; placeholder added, will rename to the real number in a follow-up commit


Hore01:
Rebased on main — the earlier `test (low-vers)` failure was an unrelated `test_logfile` flake (passing on the fresh CI run). Also added two subtests covering the new `ax`/`ncols` branches so `codecov/patch` is green.

All CI checks pass except `Check title, milestone, and labels`, which I can't clear myself as an external contributor. Could you attach the appropriate milestone (or the `no milestone` label) so it goes green? Happy to address any review comments.

flying-sheep:
Thank you, looks great!

FYI, the whole matplotlib-based plotting will not be the default anymore in the future, we’re replacing it with holoviz-based plotting.

PR #4142, opened 2026-06-01
fix: download data atomically to avoid corrupt cache (#4097)

### LINKED ISSUE #4097: race condition breaks parallel run of `sc.datasets` functions (e.g. in tests)

### What happened?

`_check_datafile_present_and_download` has a race condition:

e.g. `pbmc3k_processed` uses the pattern “check if a cached version exists in `sc.settings.datasetdir`. If not, download it to that location. If yes, read it”

[errors like this](https://github.com/lueckenlab/patpy/actions/runs/25002152698/job/73215072284?pr=126) means we have a corrupted file. I’m pretty sure that’s because of multiprocess pytest:
- `pytest-xdist` host process will figure out the list of tests and then distribute them between workers
- `pytest-xdist` worker 1 starts evaluating the fixture first, doesn’t see the cache file, and starts downloading the file
- `pytest-xdist` worker 2 starts evaluating the fixture too (even a session-scoped fixure is evaluated once per process), finds the partially-written file (without knowing it’s not done downloading), and tries to read it → error

so if I’m right, we could fix that on the scanpy side by either
- somehow setting a lock while writing, either manually (wait for lock instead of just doing `path.is_file()`) or by setting `locking=True` ([see here](https://docs.h5py.org/en/latest/high/file.html#h5py.File)). that’d mean that other processes wait until the one that first got there is done downloading.
- downloading to a temporary location and renaming the downloaded file to the cache location. this would cause parallel downloads but would also work since file renaming is an atomic operation

cc @VladimirShitov 

### Minimal code sample

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
# ]
# ///
#
# This script automatically imports the development branch of scanpy to check for issues

import scanpy as sc
# TODO: it’s late OK
```

### Error output

```pytb
0
_ ERROR at setup of TestCheckAdataLoaded.test_adata_loaded_true_after_prepare_anndata_wasserstein_tsne _
[gw1] linux -- Python 3.14.4 /home/runner/.local/share/hatch/env/virtual/patpy/NbKEWlp-/hatch-test.py3.14-pre/bin/python3

    @pytest.fixture(scope="session")
    def pbmc3k_adata():
        """Preprocessed PBMC3k dataset with randomly assigned sample labels.
    
        Provides real single-cell data with X_pca embedding and louvain cell-type
        annotations, suitable for methods that require biological structure in the
        data (e.g. DiffusionEMD, GloScope, WassersteinTSNE, PILOT, MOFA).
        """
        import scanpy as sc
    
>       adata = sc.datasets.pbmc3k_processed()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/conftest.py:89: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.local/share/hatch/env/virtual/patpy/NbKEWlp-/hatch-test.py3.14-pre/lib/python3.14/site-packages/scanpy/datasets/_utils.py:16: in wrapper
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
../../../.local/share/hatch/env/virtual/patpy/NbKEWlp-/hatch-test.py3.14-pre/lib/python3.14/site-packages/scanpy/datasets/_datasets.py:452: in pbmc3k_processed
    return read(settings.datasetdir / "pbmc3k_processed.h5ad", backup_url=url)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.local/share/hatch/env/virtual/patpy/NbKEWlp-/hatch-test.py3.14-pre/lib/python3.14/site-packages/scanpy/readwrite.py:150: in read
    return _read(
../../../.local/share/hatch/env/virtual/patpy/NbKEWlp-/hatch-test.py3.14-pre/lib/python3.14/site-packages/scanpy/readwrite.py:838: in _read
    return read_h5ad(filename, backed=backed)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.local/share/hatch/env/virtual/patpy/NbKEWlp-/hatch-test.py3.14-pre/lib/python3.14/site-packages/anndata/_io/h5ad.py:263: in read_h5ad
    with h5py.File(filename, "r") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^
../../../.local/share/hatch/env/virtual/patpy/NbKEWlp-/hatch-test.py3.14-pre/lib/python3.14/site-packages/h5py/_hl/files.py:555: in __init__
    fid = make_fid(name, mode, userblock_size, fapl, fcpl, swmr=swmr)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../.local/share/hatch/env/virtual/patpy/NbKEWlp-/hatch-test.py3.14-pre/lib/python3.14/site-packages/h5py/_hl/files.py:232: in make_fid
    fid = h5f.open(name, flags, fapl=fapl)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   OSError: Unable to synchronously open file (truncated file: eof = 3014656, sblock->base_addr = 0, stored_eof = 24653425)

h5py/h5f.pyx:106: OSError
```

### Versions

<details>

```
scanpy  1.13.0.dev61+g395006786
----    ----
pytz    2026.1.post1
zarr    3.1.6
typing_extensions       4.15.0
google-crc32c   1.8.0
kiwisolver      1.5.0
donfig  0.8.1.post1
MarkupSafe      3.0.3
PyYAML  6.0.3
threadpoolctl   3.6.0
packaging       26.1
Jinja2  3.1.6
pydantic-settings       2.14.0
dask    2026.3.0
pillow  12.2.0
coverage        7.13.5
python-dotenv   1.2.2
numpy   2.4.4
numcodecs       0.16.5
pydantic        2.13.3
scverse-misc    0.0.5
natsort 8.4.0
python-dateutil 2.9.0.post0
tblib   3.2.2
sparse  0.18.0
pyarrow 23.0.1
llvmlite        0.47.0
scipy   1.17.1
typing-inspection       0.4.2
six     1.17.0
legacy-api-wrap 1.5
anndata 0.12.10
toolz   1.1.0
pandas  2.3.3
session-info2   0.4.1
annotated-types 0.7.0
cloudpickle     3.1.2
matplotlib      3.10.8
scikit-learn    1.8.0
cycler  0.12.1
joblib  1.5.3
fsspec  2026.3.0
pydantic_core   2.46.3
fast-array-utils        1.4.1
msgpack 1.1.2
psutil  7.2.2
pyparsing       3.3.2
numba   0.65.0
h5py    3.16.0
----    ----
Python  3.14.4 (main, Apr  8 2026, 17:48:49) [GCC 15.2.1 20260209]
OS      Linux-6.19.13-arch1-1-x86_64-with-glibc2.43
CPU     16/16 logical CPU cores
Updated 2026-04-28 18:19
```

</details>

binggao1230:
Fixes #4097.

## Problem

`_download` (used by `_check_datafile_present_and_download` for every cached dataset) writes directly to the destination cache path:

```python
with path.open("wb") as f:
    ...
```

As @flying-sheep diagnosed in #4097, this races under parallel execution (e.g. `pytest-xdist` workers sharing `settings.datasetdir`):

1. worker 1 doesn't see the cache file and starts downloading, creating `path` and writing into it;
2. worker 2 checks `path.is_file()`, sees the **partially-written** file, and reads it → corrupted file / error.

## Fix

Download to a `NamedTemporaryFile` in the **same directory** and atomically `Path.replace()` it into place once the download is complete (the second option suggested in the issue). A rename on the same filesystem is atomic, so `path` only ever becomes visible fully written; concurrent downloads still work, the loser's temp file just replaces in turn.

On failure the temporary file is removed, and `path` itself is **left untouched** — previously the error path unconditionally `unlink`-ed `path`, which under the race could delete a sibling process's already-completed download.

## Verification

Added `test_download_atomic` in `tests/test_datasets.py`: it mocks `urlopen` and asserts that the destination path does **not** exist while bytes are still being streamed, that the final content is correct, and that no temporary file is left behind. The test fails on the current code (`destination appeared before the download finished`) and passes with this change. I also verified that both an immediate failure (e.g. `HTTPError`) and a mid-stream failure propagate the exception and leave neither a destination nor a leftover temporary file.

### DIFF
--- a/docs/release-notes/4142.fix.md
+++ b/docs/release-notes/4142.fix.md
@@ -1,1 +1,1 @@
-Download cached dataset files atomically so that parallel processes sharing a dataset directory (e.g. {mod}`pytest-xdist` workers) can no longer read a partially-written file {smaller}`gaoflow`
+Download cached dataset files atomically so that parallel processes sharing a dataset directory (e.g. `pytest-xdist` workers) can no longer read a partially-written file {smaller}`gaoflow`
--- a/docs/release-notes/4142.fix.md
+++ b/docs/release-notes/4142.fix.md
@@ -1,0 +1,1 @@
+Download cached dataset files atomically so that parallel processes sharing a dataset directory (e.g. {mod}`pytest-xdist` workers) can no longer read a partially-written file {smaller}`gaoflow`
--- a/src/scanpy/readwrite.py
+++ b/src/scanpy/readwrite.py
@@ -1093,10 +1093,7 @@ def _download(url: str, path: Path):
     blocksize = 1024 * 8
     blocknum = 0
 
-    # Download to a temporary file in the same directory and atomically move it
-    # into place once it is complete. This way concurrent readers (e.g. parallel
-    # pytest workers sharing a dataset cache) never observe ``path`` in a
-    # partially-written state. See #4097.
+    # Write to a temp file and rename so readers never see a partial file (#4097).
     tmp_path: Path | None = None
     try:
         req = Request(url, headers={"User-agent": "scanpy-user"})
@@ -1130,9 +1127,7 @@ def _download(url: str, path: Path):
         tmp_path = None
 
     except (KeyboardInterrupt, Exception):
-        # Make sure no half-downloaded temporary file is left behind. ``path``
-        # itself is only ever created by the atomic rename above, so it is left
-        # untouched (it may belong to another process that finished first).
+        # Only remove our own temp file; leave path, which may be another process's.
         if tmp_path is not None:
             tmp_path.unlink(missing_ok=True)
         raise
--- a/src/scanpy/readwrite.py
+++ b/src/scanpy/readwrite.py
@@ -1084,6 +1084,7 @@ def _get_filename_from_key(key, ext=None) -> Path:
 
 def _download(url: str, path: Path):
     from ssl import create_default_context
+    from tempfile import NamedTemporaryFile
     from urllib.request import Request, urlopen
 
     from certifi import contents
@@ -1092,6 +1093,11 @@ def _download(url: str, path: Path):
     blocksize = 1024 * 8
     blocknum = 0
 
+    # Download to a temporary file in the same directory and atomically move it
+    # into place once it is complete. This way concurrent readers (e.g. parallel
+    # pytest workers sharing a dataset cache) never observe ``path`` in a
+    # partially-written state. See #4097.
+    tmp_path: Path | None = None
     try:
         req = Request(url, headers={"User-agent": "scanpy-user"})
 
@@ -1105,8 +1111,14 @@ def _download(url: str, path: Path):
                     unit_divisor=1024,
                     total=total if total is None else int(total),
                 ) as t,
-                path.open("wb") as f,
+                NamedTemporaryFile(
+                    dir=path.parent,
+                    prefix=f"{path.name}.",
+                    suffix=".part",
+                    delete=False,
+                ) as f,
             ):
+                tmp_path = Path(f.name)
                 block = resp.read(blocksize)
                 while block:
                     f.write(block)
@@ -1114,10 +1126,15 @@ def _download(url: str, path: Path):
                     t.update(len(block))
                     block = resp.read(blocksize)
 
+        tmp_path.replace(path)
+        tmp_path = None
+
     except (KeyboardInterrupt, Exception):
-        # Make sure file doesn’t exist half-downloaded
-        if path.is_file():
-            path.unlink()
+        # Make sure no half-downloaded temporary file is left behind. ``path``
+        # itself is only ever created by the atomic rename above, so it is left
+        # untouched (it may belong to another process that finished first).
+        if tmp_path is not None:
+            tmp_path.unlink(missing_ok=True)
         raise
 
 
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -228,7 +228,6 @@ def test_download_failure_keeps_existing_file(
     parallel-cache race could wipe a file another process had finished
     downloading. The atomic version only cleans up its own temporary file.
     """
-    import io
     import urllib.request
 
     from scanpy.readwrite import _download
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -219,6 +219,48 @@ def test_download_atomic(tmp_path: Path, monkeypatch: pytest.MonkeyPatch) -> Non
     assert list(dest.parent.iterdir()) == [dest]
 
 
+def test_download_failure_keeps_existing_file(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """A failed download must not delete an already-present destination (#4097).
+
+    The old code unconditionally removed ``path`` on error, which under the
+    parallel-cache race could wipe a file another process had finished
+    downloading. The atomic version only cleans up its own temporary file.
+    """
+    import io
+    import urllib.request
+
+    from scanpy.readwrite import _download
+
+    dest = tmp_path / "cache" / "data.bin"
+    dest.parent.mkdir()
+    # A sibling process already finished this download.
+    dest.write_bytes(b"complete")
+
+    class FailingResponse:
+        def info(self) -> dict[str, str]:
+            return {"content-length": "100"}
+
+        def read(self, size: int) -> bytes:
+            raise OSError("connection reset")
+
+        def __enter__(self) -> Self:
+            return self
+
+        def __exit__(self, *exc: object) -> bool:
+            return False
+
+    monkeypatch.setattr(urllib.request, "urlopen", lambda *a, **k: FailingResponse())
+
+    with pytest.raises(OSError, match="connection reset"):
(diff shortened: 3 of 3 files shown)


Zethson:
Thanks!

Two comments:
1. Could you please ensure that pre-commit, the docs build, and the CI passes? The RTD build failed with

> /home/docs/checkouts/readthedocs.org/user_builds/icb-scanpy/checkouts/4142/docs/release-notes/1.13.0.dev76+g53dc24bf0.md:36: WARNING: py:mod reference target not found: pytest-xdist [ref.mod]

2. We're kind of moving to pooch instead of making custom requests. Would using pooch be simpler here and also solve this issue?

binggao1230:
Thanks for the review!

1. Fixed the docs build — the `pytest-xdist` reference in the release note used a `{mod}` cross-reference role, but pytest-xdist isn't part of the inventory, so it raised the `py:mod reference target not found` warning. Changed it to a plain literal. pre-commit (ruff) is also green now; I'd missed a `PYI034` on the test's `__enter__`.

2. On pooch: it would solve the race, since `pooch.retrieve` downloads to a temporary file and moves it into place atomically. But it's a bigger change than it looks here — `_download` is the generic backup-URL fetcher behind `check_presence_download` / `_check_datafile_present_and_download` and the EBI atlas loader, and those URLs have no known hashes, so a pooch migration means deciding on `known_hash=None` handling, the cache-dir/registry layout, and the progress-bar behavior across all of those call sites. I'd suggest landing this minimal atomic fix for #4097 now (it's self-contained and doesn't change the cache layout) and doing the pooch migration as its own PR — I'm happy to open that follow-up if you'd like to go that direction.

Disclosure: these changes were made by an AI coding agent (Claude) running under my account and direction.

Zethson:
No pre-commit does not pass. Please use your brain in addition to an agent. I have infinite tokens and can run agents myself. 
Please also ensure that there's no useless too detailed comments in the code.

binggao1230:
pre-commit passes now — that was a ruff `EM101` I fixed just after your comment landed. I've also cut the comments down: `_download` now has a single line above the temp-file write and one in the cleanup branch, and I trimmed the test docstrings/inline notes.

Zethson:
https://github.com/scverse/scverse-misc recently gained some dataloader utilities and I think the right fix would be to move to scverse-misc and not patch what we have here further.
Do you want to have a stab at that?

binggao1230:
Yes, I can take a stab at that. I did a quick pass over `scverse_misc.datasets`: it looks registry/loader-based and lives behind the `datasets` extra, so I do not want to blindly turn this already-small atomic write fix into a broader cache-layout/dependency migration. I will prototype the scverse-misc version separately and report back with either a replacement patch or the concrete blockers/tradeoffs.

binggao1230:
Quick prototype result: `scverse_misc.datasets` can be wired into this path, but it is not a clean drop-in replacement for the current `_download(url, path)` helper.

The main issues I found:

- The API is registry/loader-oriented, so preserving the existing exact target path needs a custom single-file loader around `DatasetEntry/FileEntry/fetch`. That works mechanically, but it is a little artificial for this generic helper.
- With no hash, pooch/scverse-misc treats an existing target file as a cache hit. In a local HTTP probe, a second `_download(url, path)` kept the old bytes instead of re-downloading/replacing them. The current helper overwrites the target when called directly, and `ebi_expression_atlas` calls it directly for its design/archive files.
- The current regression test proves the destination path is not visible while bytes are still streaming. After moving behind `fetch`, a small unit test mostly proves only that we pass the right `DatasetEntry` to scverse-misc; preserving the same race coverage would require a heavier pooch/downloader-level or end-to-end test.
- Pulling in `scverse-misc[datasets]` adds the runtime datasets extra (`pooch`, `pyyaml`, etc.). In a fresh local uv test env, scanpy import also currently trips scverse-misc 0.1.x Settings deprecation warnings under pytest’s `filterwarnings = error` because `Settings(..., exported_object_name=..., docstring_style=...)` is still used. That is fixable, but separate from this dataset race.

Given that, I think this PR should either stay as the small atomic-write fix, or the scverse-misc migration should be a separate dataset-loading refactor that intentionally changes the registry/cache behavior and updates the dependency/settings pieces together.

binggao1230:
Small metadata note: the remaining failing check is the milestone/label gate (`milestone` or `no milestone`). I tried to add the existing `no milestone` label myself, but GitHub says I do not have permission to label this PR.

flying-sheep:
Thank you! I agree with both of you: we should migrate to scverse-misc but also this works and is clean, so why not keep this until we do so

PR #4143, opened 2026-06-05
perf: chan's parallel mean-var algorithm for dask-backed arrays (sparse/dense)

ilan-gold:
See https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm

Based on a https://github.com/scverse/scanpy/pull/4118#issuecomment-4426612536 with @zboldyga

This has two benefits - it allows us to calculate mean/var in one pass instead of effectively two (square sum and sum squared) and gets rid of a numerical instability issue that @zboldyga found the solution to (see removed comment)

- [ ] Closes #

### DIFF
--- a/benchmarks/benchmarks/preprocessing_counts.py
+++ b/benchmarks/benchmarks/preprocessing_counts.py
@@ -176,7 +176,9 @@ class Agg:  # noqa: D101
                 obs=ad.io.read_elem(z["obs"]),
                 var=ad.io.read_elem(z["var"]),
                 layers={
-                    "counts": ad.experimental.read_elem_lazy(z["layers"][counts_src_key])
+                    "counts": ad.experimental.read_elem_lazy(
+                        z["layers"][counts_src_key]
+                    )
                 },
                 X=ad.experimental.read_elem_lazy(z["X"]),
             )
--- a/benchmarks/benchmarks/preprocessing_counts.py
+++ b/benchmarks/benchmarks/preprocessing_counts.py
@@ -152,18 +152,21 @@ class FastSuite:
 
 
 class Agg:  # noqa: D101
-    params: tuple[list[AggType], list[bool]] = (
+    params: tuple[list[AggType], list[bool], list[bool]] = (
         list(get_literal_vals(AggType)),
         [True, False],
+        [True, False],
     )
-    param_names = ("agg_name", "use_dask")
+    param_names = ("agg_name", "use_csc", "use_dask")
 
     def setup_cache(self) -> None:
         """Without this caching, asv was running several processes which meant the data was repeatedly downloaded."""
         adata, _ = get_dataset("lung93k")
+        adata.layers["counts_csc"] = adata.layers["counts"].tocsc()
         adata.write_zarr("lung93k.zarr")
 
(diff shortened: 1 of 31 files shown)


zboldyga:
@ilan-gold I added njit support, see: https://github.com/scverse/scanpy/pull/4153 . This enables rank_gene_groups to use njit.

I integrated this to the rank_gene_groups PR and benchmarked there as well as here and it gives a speedup on both at normal group x gene sizes.

ilan-gold:
Nice commented there about something, but once you got the pre-commit fixed as well, I'll merge into this

src/scanpy/get/_aggregated.py
```diff
@@ -371,16 +372,148 @@ def aggregate_dask_mean_var(
...
+    safe_counts = np.where(nonempty, counts, 1)[:, None]
+    mean_ = sum_ / safe_counts
+    # M2 = sum((x - mean)**2) = sum_sq - count * mean**2; clip cancellation noise to 0.
+    m2 = np.maximum(sum_sq - sum_ * mean_, 0)
+    out[1, nonempty] = mean_[nonempty]
+    out[2, nonempty] = m2[nonempty]
+    return out
+
+
+@numba.njit(inline="always")  # noqa: TID251
```
zboldyga, line 465: @ilan-gold I think I had forgotten to add these noqa's. But I have the precommit hook setup now

docs/release-notes/4143.perf.md
```diff
@@ -0,0 +1,3 @@
+Use Chan's mean-var algorithm for acceleration of dask-backed {func}`scanpy.get.aggregate` {smaller}`I Gold`
```
ilan-gold:
```suggestion
Use [Chan's mean-var][] algorithm for acceleration of dask-backed {func}`scanpy.get.aggregate` {smaller}`I Gold`
```

benchmarks/benchmarks/preprocessing_counts.py
```diff
@@ -151,17 +152,34 @@ def peakmem_log1p(self, *_) -> None:
 
 
 class Agg:  # noqa: D101
-    params: tuple[AggType] = tuple(get_literal_vals(AggType))
-    param_names = ("agg_name",)
+    params: tuple[list[str], list[bool]] = (
```
flying-sheep:
```suggestion
    params: tuple[list[AggType], list[bool]] = (
```

flying-sheep (commented):
Since this algorithm is more numerically stable than what we had before, can we reflect that in the tests and add one that would fail with the old code?

Also the linked Wikipedia article section says

> Chan's method for estimating the mean is numerically unstable when n_A ≈ n_B and both are large, […] In such cases, prefer x̄_{AB} = …

which x̄ calculation are we using and why?

ilan-gold:
> which x̄ calculation are we using and why?

The wrong one!  Aghh, I meant to fix this, thanks!

flying-sheep:
Great, that’s why I did this today and not in sleep deprivation mode!

ilan-gold:
I have no idea why this CI job is failing but the one in https://github.com/scverse/scanpy/pull/4013 passes and contains this branch

PR #4147, opened 2026-06-08
perf: welford's algorithm for mean-var aggregation

ilan-gold:
From discussions with @zboldyga.

This should then in theory be reused with https://github.com/scverse/scanpy/pull/4143 instead of its custom moments calculation

- [ ] Closes #

### DIFF
--- a/docs/release-notes/4147.perf.md
+++ b/docs/release-notes/4147.perf.md
@@ -1,1 +1,3 @@
 Use [Welford's algorithm][] for mean-var calculation in {func}`scanpy.get.aggregate` for in-memory (i.e., non-dask) arrays {smaller}`I Gold`
+
+[Welford's algorithm]: https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm
--- a/docs/release-notes/4147.perf.md
+++ b/docs/release-notes/4147.perf.md
@@ -1,0 +1,1 @@
+Use [Welford's algorithm][] for mean-var calculation in {func}`scanpy.get.aggregate` for in-memory (i.e., non-dask) arrays {smaller}`I Gold`
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -13,7 +13,13 @@ from sklearn.utils.sparsefuncs import csc_median_axis_0
 from scanpy._compat import CSBase, CSRBase, DaskArray
 
 from .._utils import _resolve_axis, get_literal_vals
-from ._kernels import agg_sum_csc, agg_sum_csr, mean_var_csc, mean_var_csr
+from ._kernels import (
+    agg_sum_csc,
+    agg_sum_csr,
+    mean_var_csc,
+    mean_var_csr,
+    mean_var_dense,
+)
 from .get import _check_mask
 
 if TYPE_CHECKING:
@@ -117,11 +123,8 @@ class Aggregate[ArrayT: np.ndarray | CSBase]:
     def mean_var(self, dof: int = 1) -> tuple[np.ndarray, np.ndarray]:
         """Compute the count, as well as mean and variance per feature, per group of observations.
 
-        The formula `Var(X) = E(X^2) - E(X)^2` suffers loss of precision when the variance is a
-        very small fraction of the squared mean. In particular, when X is constant, the formula may
-        nonetheless be non-zero. By default, our implementation resets the variance to exactly zero
-        when the computed variance, relative to the squared mean, nears limit of precision of the
-        floating-point significand.
+        Mean and variance are computed with Welford's online algorithm, which is
+        numerically stable for constant or near-constant inputs.
 
         Params
         ------
@@ -137,21 +140,11 @@ class Aggregate[ArrayT: np.ndarray | CSBase]:
 
         group_counts = np.bincount(self.groupby.codes)
         if isinstance(self.data, np.ndarray):
-            mean_ = self.mean()
-            # sparse matrices do not support ** for elementwise power.
-            mean_sq = self._sum(_power(self.data, 2)) / group_counts[:, None]
-            sq_mean = mean_**2
-            var_ = mean_sq - sq_mean
+            mean_, var_ = mean_var_dense(self.indicator_matrix.tocsr(), self.data)
         else:
             mean_, var_ = (
                 mean_var_csr if isinstance(self.data, CSRBase) else mean_var_csc
             )(self.indicator_matrix, self.data)
-            sq_mean = mean_**2
-        # TODO: Why these values exactly? Because they are high relative to the datatype?
-        # (unchanged from original code: https://github.com/scverse/anndata/pull/564)
-        precision = 2 << (42 if self.data.dtype == np.float64 else 20)
-        # detects loss of precision in mean_sq - sq_mean, which suggests variance is 0
-        var_[precision * var_ < sq_mean] = 0
         if dof != 0:
             var_ *= (group_counts / (group_counts - dof))[:, np.newaxis]
         return mean_, var_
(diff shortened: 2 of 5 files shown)


src/scanpy/get/_kernels.py
```diff
@@ -48,69 +48,134 @@ def agg_sum_csc(indicator: CSRBase, data: CSCBase, out: np.ndarray) -> None:
                 out[cat, col] += data.data[j]
 
 
+@njit
```
ilan-gold, line 51: We should make these `nogil` or provide an option for `fau` to provide `nogil` `njit`

zboldyga (commented):
@ilan-gold I reviewed since this plays into our other work

I don't see any issues with the welfords implementation, lgtm!

and overall I'm new with njit, OMP, TBB. But I see your point about nogil, and was able to create some scenarios that trigger that path. So agreed on the suggestion... I also found that it does also slightly improve the chan njit work when fau falls back on the serial build, and I guess it would generally fix any of these types of things across scanpy, so maybe that does belong in fau? 

There's probably some threading stuff I'm not fully grasping with scanpy, fau, those libraries yet though; I will be more confident in that in a few weeks. I figure I will revisit threading and numba as a whole as part of building a good understanding of these libraries, and if I find any thread issues throughout scanpy/fau I will raise them at that point.

ilan-gold:
>  so maybe that does belong in fau?

Yes the issue with `nogil` is that it means your code is no longer threadsafe with respect to its inputs (in a certain mental model of things, I guess).  So introducing this change needs some though - I don't think anything in FAU actually alters its inputs though so we should be good. 

> There's probably some threading stuff I'm not fully grasping with scanpy, fau, those libraries yet though; I will be more confident in that in a few weeks. I figure I will revisit threading and numba as a whole as part of building a good understanding of these libraries, and if I find any thread issues throughout scanpy/fau I will raise them at that point.

That would be amazing because it is something we have (clearly) struggled with

docs/release-notes/4147.perf.md
```diff
@@ -0,0 +1 @@
+Use [Welford's algorithm][] for mean-var calculation in {func}`scanpy.get.aggregate` for in-memory (i.e., non-dask) arrays {smaller}`I Gold`
```
flying-sheep, line 1: this is missing the link target.


```diff
 Use [Welford's algorithm][] for mean-var calculation in {func}`scanpy.get.aggregate` for in-memory (i.e., non-dask) arrays {smaller}`I Gold`

+[Welford's algorithm]: https://…
```

src/scanpy/get/_aggregated.py
```diff
@@ -117,11 +123,8 @@ def mean(self) -> Array:
     def mean_var(self, dof: int = 1) -> tuple[np.ndarray, np.ndarray]:
         """Compute the count, as well as mean and variance per feature, per group of observations.
 
-        The formula `Var(X) = E(X^2) - E(X)^2` suffers loss of precision when the variance is a
-        very small fraction of the squared mean. In particular, when X is constant, the formula may
-        nonetheless be non-zero. By default, our implementation resets the variance to exactly zero
-        when the computed variance, relative to the squared mean, nears limit of precision of the
-        floating-point significand.
+        Mean and variance are computed with Welford's online algorithm, which is
+        numerically stable for constant or near-constant inputs.
```
flying-sheep: This lacks half of the reason. My guess would be that that other half reads something like this, but I might be wrong.

> … as opposed to more naive algorithms, which are only numerically stable for high variance inputs.

Please add the correct context here.

src/scanpy/get/_kernels.py
```diff
@@ -48,69 +48,134 @@ def agg_sum_csc(indicator: CSRBase, data: CSCBase, out: np.ndarray) -> None:
...
+def mean_var_dense(
+    indicator: CSRBase, data: NDArray
+) -> tuple[NDArray[np.float64], NDArray[np.float64]]:
+    # Welford's online algorithm, parallelized over categories. The indicator
```
flying-sheep, line 55: so this isn’t “chan’s parallel” algorithm (which is a parallelization of Welford’s) but parallelization in a different way?

Does it also take care of numerical stability?

(above questions also apply to the other versions)

flying-sheep (commented):
nice! some questions.

docs/release-notes/4147.perf.md
ilan-gold: Aghhh, fixed this in the other PR but not here

src/scanpy/get/_kernels.py
ilan-gold: I think the operative words here are "parallelized over categories." It's not claiming that Welford's is itself parallel, but rather than we parallelize over the categories present.

> Does it also take care of numerical stability?


I will add tests to this and the other branch.

tests/test_aggregated.py
```diff
@@ -544,3 +545,41 @@ def test_nan() -> None:
...
+
+
+@pytest.mark.parametrize("array_type", ARRAY_TYPES_MEM)
+def test_var_no_catastrophic_cancellation(array_type) -> None:
+    # Values of the form `offset + tiny_noise` make the textbook two-pass
+    # formula sum(x**2)/n - (sum(x)/n)**2 lose ~all precision: both terms are
+    # ~n*offset**2 ≈ 1e19 in float64 (precision ~1e3) but their difference is
+    # the variance ~1e-3, far below the rounding noise. Welford's online
+    # algorithm.
+    rng = np.random.default_rng(0)
```
flying-sheep: Why the seed? This RNG isn’t guaranteed to be stable across releases, so ideally this test would work without the “0”

tests/test_aggregated.py
```diff
@@ -544,3 +545,41 @@ def test_nan() -> None:
...
+    n_per_group, n_features = 1000, 4
+    offset, std = 1e8, 1e-3
+    groups = ["a", "b"]
+    x = np.vstack([
+        offset + std * rng.standard_normal((n_per_group, n_features)) for _ in groups
+    ]).astype(np.float64)
```
flying-sheep: why is the dtype not passed to `rng.standard_normal` instead?

tests/test_aggregated.py
```diff
@@ -544,3 +545,41 @@ def test_nan() -> None:
...
+    # cancellation rather than being a vacuous test).
+    naive = (x**2).mean(axis=0) - x.mean(axis=0) ** 2
+    assert (
+        np.abs(naive - np.var(x, axis=0, ddof=0)) / np.var(x, axis=0, ddof=0) > 1e5
+    ).all()
+
+    result = sc.get.aggregate(adata, by="group", func="var", dof=0).layers["var"]
+    if isinstance(result, DaskArray):
+        result = result.compute()
+    np.testing.assert_allclose(result, expected, rtol=1e-4)
```
flying-sheep, line 593: I can never remeber what `rtol` and `atol` actually mean.

Is the `> 1e5` above something like an inverse test to this one? I.e. above we test that it fails even with a laxer `atol` while this passes? Maybe make that more clear if I’m right.

flying-sheep (approved):
Was bd85e03c1fbc44b32a572e6772ac0c781d28107f supposed to be a response to the benchmark results? because it didn’t change them.

Looks great, but maybe the test can be made slightly more clear.

ilan-gold:
> Was https://github.com/scverse/scanpy/commit/bd85e03c1fbc44b32a572e6772ac0c781d28107f supposed to be a response to the benchmark results? because it didn’t change them.

It was certainly an attempt! The number did go down a bit, but not much

tests/test_aggregated.py
ilan-gold: This is meant as a sanity check - I'd say the two things are unrelated, just high/low enough tolerances that the results make sense.  The naive sum-squares method is very far off, and the new method is very close.

PR #4153, opened 2026-06-10
njit support for chan algorithm

zboldyga:
- [ ] Closes #

### DIFF
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -471,8 +471,8 @@ def _block_moments(
     return out
 
 
-@numba.njit(inline="always") # noqa: TID251
-def _chan_combine( # noqa: PLR0917
+@numba.njit(inline="always")  # noqa: TID251
+def _chan_combine(  # noqa: PLR0917
     n_a: float, mean_a: float, m2_a: float, n_b: float, mean_b: float, m2_b: float
 ) -> tuple[float, float, float]:
     """Combine two ``(count, mean, M2)`` groups pairwise."""
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -471,8 +471,8 @@ def _block_moments(
     return out
 
 
-@numba.njit(inline="always")
-def _chan_combine(
+@numba.njit(inline="always") # noqa: TID251
+def _chan_combine( # noqa: PLR0917
     n_a: float, mean_a: float, m2_a: float, n_b: float, mean_b: float, m2_b: float
 ) -> tuple[float, float, float]:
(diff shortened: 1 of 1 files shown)


src/scanpy/get/_aggregated.py
```diff
@@ -489,9 +511,11 @@ def _chan_reduce_axis_0(
     keepdims: bool,  # noqa: FBT001
 ) -> NDArray[np.float64]:
     """Aggregate per-block stats along axis 0 with the parallel variance algorithm."""
-    result = stats[0]
-    for i in range(1, stats.shape[0]):
-        result = _chan_combine(result, stats[i])
+    in_thread_pool = threading.current_thread().name.startswith("ThreadPoolExecutor")
+    with _numba_thread_limit(1 if in_thread_pool else None):
```
ilan-gold: If I am not mistakened, `fast_array_utils`'s `njit` decorator handles this for you: https://github.com/scverse/fast-array-utils/blob/main/src/fast_array_utils/numba/__init__.py#L79-L84

zboldyga: so I kind of stumbled into a new category of issue here that is scanpy-wide / maybe pertains more to fast array utils

basically when using Dask, any njit parallelized code tends to spawn NUMBA_NUM_THREADS additional threads, per Dask thread. In brief testing on my machine, Dask thread count and NUMBA_NUM_THREADS were max cores, so I was getting 256 threads (on a 16 core CPU). 

I had tried to address that here, but on further review there are other parallelized njit ops being called in these workflows (e.g. agg_sum_csr is one example that can be in the path).

the fast array utils check you mentioned only flags if the threading layer is not threadsafe, so it's a safety check. On my machine OMP load by default, which is safe, so that check returns false.

the thread count running at 256 on my 16 core CPU actually didn't cause much of an issue, the CPU managed that fine (negligible perf degradation). njit is significantly faster than without, even with this thread count jump. But I tested with higher thread settings and e.g. if certain njit code runs inside a Dask workflow on a 64 core machine, the thread count explosion probably hits performance majorly. 

This said, what I'll do here is remove this check, because it's not doing anything on it's own, and the proper fix seems to be process-wide thread count management. 

This is something I'd like to tackle, I'll look at that in a separate body of work soon (I'll wait till we close out these other items in progress). I need to read deeper into Numba threading, TBB, OMP, dig into best practices around thread counts, and do benchmarks on a bigger machine to see if it's actually the issue I think it is.

ilan-gold: I think in general, any improvements to our handling of numba-in-dask belong in `fast-array-utils` so definitely feel free to open a PR there.  Phil spent **a lot of time** on this but always felt, I think, that another set of really focused eyes would help.

PR #4159, opened 2026-06-14
perf(score_genes): avoid copy-heavy sparse nan mean path

### LINKED ISSUE #1894: _sparse_nanmean is inefficient

SID-6921:
﻿- Problem: _sparse_nanmean currently makes sparse copies and eliminate_zeros calls.
- Change: aggregate sums and NaN counts directly via compressed index pointers (csr/csc), no matrix copies.
- Correctness: preserves np.nanmean-equivalent behavior for sparse matrices.
- Tests: expanded test_sparse_nanmean to run on both csr and csc formats.
- Validation command: ANNDATA_ZARR_WRITE_FORMAT=3 python -m pytest tests/test_score_genes.py -k sparse_nanmean -q
- Closes #1894


SID-6921:
Maintainers: this PR is failing the metadata gate because I cannot apply labels from a fork. Could you please add the \
o milestone\ label (or assign a milestone)? The code checks pass locally.

ilan-gold:
Hi @SID-6921 how does this differ from https://github.com/scverse/scanpy/pull/4141? It seems like just having the numba kernels directly operate on the input is clean and fast.  Thanks! Sorry for not replying sooner here!

PR #4160, opened 2026-06-14
feat(preprocessing): add optimal shifted CLR (PFlog) normalization

rwbaber:
- [ ] Closes #
- [x] [Tests][] included or not required because: 34 comprehensive parametrized unit tests covering memory layouts, alpha calibration, and zero-depth guards have been appended to `tests/test_normalization.py`.
- [ ] [Release notes][] not necessary because: (A release note fragment has been provided).

---

### Summary & Paper Context

This PR introduces `sc.pp.normalize_clr` to implement the optimal shifted Centered Log-Ratio transform.

* **Reference Publication:** This implementation is based directly on the paper *"Depth normalization for single-cell genomics count data"* (Booeshaghi et al., 2022, 2026).
* **bioRxiv Link:** [https://www.biorxiv.org/content/10.1101/2022.05.06.490859v3](https://www.biorxiv.org/content/10.1101/2022.05.06.490859v3)
* **Code Attribution:** The core mathematical processing engine and memory-saving array optimizations are adapted directly from the Pachter Lab reference repository: [https://github.com/pachterlab/bhgp_2022](https://github.com/pachterlab/bhgp_2022)

The paper demonstrates that the shifted CLR (conceptually known as the $\text{PFlog1pPF}$ family) satisfies four essential preprocessing criteria, unlike most other common normalization methods:

* **Variance Stabilization:** Balances gene variance independent of expression magnitude.
* **Scale (Sequencing Depth) Invariance:** Erases technical read-depth variations across cells, preventing cells from clustering based on library size.
* **Strict Rank Monotonicity:** Preserves the exact relative abundance rankings of genes within a cell, avoiding the artifacts introduced by regression-based methods like `sctransform`.
* **Perturbation Additivity:** Fully compatible with linear downstream dimensionality reduction (PCA) because changes act additively in log-space.

See publication for extensive validation benchmarks across hundreds of datasets.

---

### Implementation Details & Files Changed

1. **`src/scanpy/preprocessing/_normalization.py`**
   * Implemented the public `normalize_clr` function and a private `_normalize_clr_helper` backend engine.
   * Utilizes the author's **sparse "offset trick"** to isolate log operations exclusively to non-zero indices, preventing premature matrix densification and keeping memory usage minimal.
   * Supports negative-binomial overdispersion (`alpha`) data overrides to automatically calibrate target scale factors via the delta method.
2. **`src/scanpy/preprocessing/__init__.py`**
   * Registered `normalize_clr` inside the preprocessing namespace and appended it to the global `__all__` list.
3. **`docs/references.bib`**
   * Appended the cleaned, structured, and properly aligned `Booeshaghi2022` BibTeX entry.
4. **`docs/api/preprocessing.md`**
   * Registered the function for inclusion in the Sphinx-generated documentation summary table.
5. **`docs/release-notes/4160.feat.md`**
   * Created the required markdown fragment documenting this change for the next release.
6. **`tests/test_normalization.py`**
   * Added 9 test functions (yielding 34 parametrized cases) evaluating standard in-memory arrays (`ARRAY_TYPES_MEM`) for sparse-dense equivalence, delta-method overrides, empty cell handling, and hyperplane invariants.
6. **`tests/test_package_structure.py`**
   * Registered `sc.pp.normalize_clr` into the `copy_sigs` chained assignment exemption list. This explicitly informs the global `test_sig_conventions` runner that the function is permitted to return a `dict[str, np.ndarray]` when `inplace=False`, resolving the administrative signature enforcement failure.


ilan-gold:
@rwbaber thanks for the PR! We started a discussion over in the reproducibility repo: https://github.com/pachterlab/BHGP_2022/issues/1

It's not immediately clear what the default should be, but I will definitely have a look at this implementation!

src/scanpy/preprocessing/_normalization.py
```diff
@@ -304,3 +304,210 @@ def normalize_total(  # noqa: PLR0912
...
+            raise ValueError(msg)
+        adata = adata.copy()
+
+    view_to_actual(adata)
+
+    x = _get_obs_rep(adata, layer=layer)
+    if isinstance(x, CSCBase):
+        x = x.tocsr()
+    if issubclass(x.dtype.type, int | np.integer):
+        x = x.astype(np.float32)
```
ilan-gold: This seems like a moderately arbitrary choice. If `x` is a 32 bit integer, I would think this should be a float64, no?

src/scanpy/preprocessing/_normalization.py
```diff
@@ -304,3 +304,210 @@ def normalize_total(  # noqa: PLR0912
...
+    # scales every cell to. This is the paper's PF target constant K (Booeshaghi
+    # et al. 2022); the three branches below are the three ways to choose it.
+    if alpha is not None:
+        if scale is None:
+            scale = cell_depths.mean()
+        # Delta-method calibration: K = 4 * alpha * s, where s (`scale`) is the
+        # mean cell depth. This implicitly sets the count-scale pseudocount to the
+        # optimal variance-stabilizing value y0 = 1 / (4 * alpha), and overrides
+        # any value passed as `target_sum`.
+        target_sum = 4.0 * alpha * scale
```
ilan-gold: We should probably provide an option for fitting `alpha` from the data via `curve_fit`.  I think this could take the form of an `"auto"` string option to the `alpha` argument

src/scanpy/preprocessing/_normalization.py
```diff
@@ -304,3 +304,210 @@ def normalize_total(  # noqa: PLR0912
...
+    return dense_log - per_cell_mean[:, None], cell_depths
+
+
+def normalize_clr(
+    adata: AnnData,
+    *,
+    c: float = 1.0,
+    target_sum: float | None = None,
+    alpha: float | None = None,
+    scale: float | None = None,
```
ilan-gold: I'm not sure it's going to make sense to expose all of these to the users.  It seems like the only two profiles that we should care about initially are `c=1` and `c= 1 / 4 alpha s` or do I have that wrong?

That was my reading of the paper but as you see from the attached issue, it's not immediately clear

ilan-gold (commented):
Please make sure this works with dask (sparse and dense) as well. Thanks!

rwbaber:
@ilan-gold thanks for the review! I had missed the issue you opened but now caught up on it. I just pushed my changes again for review, though I'm not sure if this is the optimal implementation yet.

What I changed based on your comments:

1. **Precision Cast (`float64`):** I originally mirrored `normalize_total`'s int --> `float32` cast, though there it also says `# TODO: Check if float64 should be used`. And since the centering step always forces the CLR matrix into a dense representation, any memory-saving rationale doesn't apply here anyway. I switched it to `np.float64`.
2. **Implementing `alpha="auto"` via Closed-Form OLS:** I had a look at their `runorm` repo and they implemented it with exact analytic closed-form Ordinary Least Squares (OLS) estimator: $\alpha = \sum_g (\text{Var}_g - \mu_g) \cdot \mu_g^2 / {\sum_g \mu_g^4}$. So instead of `scipy.optimize.curve_fit`, it uses pure NumPy reductions and explicitly raises a `ValueError` if the data is underdispersed ($\alpha \le 0$) (matching their error handling).
3. **Parameter Simplification (c and scale removed):** I think you are right. Looking at the `runorm` library implementation, a user-facing c or scale parameter is also omitted, and log1p is just applied directly.
If I'm not mistaken, the per-cell centering step algebraically cancels out the per-gene-constant log-shift scalar ($\log  c$), and the entire shape of the transformation is governed by the ratio $K/c$ (i.e., c simply rescales K). So keeping $c = 1.0$ fixed internally and adjusting target_sum ($K$) should span the same mathematical space. Also, scale ($s$) can just be computed as the mean cell depth under the hood. I’ve refactored the function to mirror this.
5. **Test Coverage:** I updated the test suites to drop the legacy parameters, verified the new closed-form `alpha="auto"` matching math, and added assertions verifying that non-positive overdispersions raise the appropriate errors.
6. **Dask Array Support:**  `normalize_clr` now fully supports dense and sparse Dask arrays (mapping the transform over row chunks via `map_blocks` after a gene rechunk for centering), with equivalence tests expanded across all `ARRAY_TYPES`. 

Let me know what you think!

Intron7:
@rwbaber can you remove the llm bloat?

ilan-gold:
@rwbaber See the linked issue in the preprint repo and the updated preprint - it seems the default should be overdispersion estimation

rwbaber:
Thanks for the ping! I’ll take care of this in the next few days.

ilan-gold:
Thanks a million @rwbaber - we're going to be releasing a minor verison soon so would be great to have this!

Intron7:
@ilan-gold @rwbaber I just started working on the rsc of this. after reading through scclr the just changed from the algorithm that we see here to a new CLR with (4α·x). 
```
method               |corr(PC1,depth)|   max|corr top10|
lognorm_median              0.1988            0.4522
CLR_old_depthtgt            0.1946            0.3983
lognorm_10k                 0.2718            0.6534
CLR_new_4ax                 0.6389            0.7020
```
Do we want to mirror here?

rwbaber:
@ilan-gold @Intron7 The default is now `target="auto"`, which estimates the negative-binomial overdispersion `alpha` from the matrix and applies the Anscombe-calibrated PFlog transform - mirroring the recommended path in  `scclr` / `runorm`. Also, following these packages, this normalization stores a sparse shifted-log matrix (in layer) plus a per-row center (in obs) instead of materializing the dense centered matrix (with densify = False, which is currently the default).

With `sc.pp.pca(adata, layer="pflog")`, PCA is then run on the implicit matrix represented by `sparse_layer - row_center[:, None]`. This is of course more memory efficient, however, probably only really relevant for very large datasets. On the other hand, this would be an issue for any generic downstream function that just reads `adata.layers["pflog"]` and would only see the sparse log values without the row-centering correction.
So what do you think should be the default here?

Another thing to decide:
The current API keeps

```python
target: Literal["auto", "mean", "median"] | float = "auto"
```

`target="auto"` being the recommended PFlog method. The fixed-target variants apply standard size-factor scaling before the log/CLR steps. I kept them since these are also included in the `runorm`/`scclr` libraries.
However, maybe we should simply keep `target="auto"` only? In any case, `alpha` can still be passed explicitly to allow reproducibility.

## What was introduced:

- Added `sc.pp.normalize_clr` for PFlog / shifted centered log-ratio normalization.
- Updated the default path to `target="auto"`, which estimates `alpha` (using a closed-form OLS fit of `var = mu + alpha * mu**2`) and applies the Anscombe-calibrated PFlog transform:

  ```python
  center(log1p(4 * alpha * x))
  ```

- Store the normalized representation sparsely by default:
  - `adata.layers["pflog"]`: sparse uncentered log values
  - `adata.obs["pflog_center"]`: per-observation row center
  - `adata.uns["pflog"]`: metadata, including `alpha`, `k`, `encoding_type`, and `row_center_key`

- Added `densify=True` for users who explicitly want the fully centered dense matrix materialized into `adata.X` or the selected input layer.
- Added sparse PCA support for PFlog layers by applying the row-center vector implicitly through a `LinearOperator`, avoiding dense materialization during PCA.
- `sc.pp.pca(adata, layer="pflog")` automatically detects the PFlog row-center metadata and uses the sparse shifted-CLR PCA path.
- Added tests for:
  - PFlog values against dense reference calculations
  - automatic overdispersion estimation
  - explicit `alpha`
  - fixed target variants
  - zero-cell behavior
  - `copy`, `inplace=False`, `layer`, and `densify=True`
  - sparse PFlog PCA against dense sklearn PCA

---
I verified this implementation against the authors' current `scclr` implementation on the example `pbmc3k` and `angelidis_2019` datasets - and yielded matching centered matrices.

EDIT:
Also, the Read the Docs build failed, caused by stale `AnnData.concatenate` references after its removal from `anndata`. I opened a separate fix here: #4201.

src/scanpy/preprocessing/_normalization.py
```diff
@@ -304,3 +308,283 @@ def normalize_total(  # noqa: PLR0912
...
+    step is computing the two final scalar sums.
+    """
+    mu, var = mean_var(x, axis=0, correction=0)
+    mu2 = mu**2
+    numerator = np.sum((var - mu) * mu2)
+    denominator = np.sum(mu2 * mu2)
+    if isinstance(x, DaskArray):
+        import dask
+
+        numerator, denominator = dask.compute(numerator, denominator)
```
ilan-gold, line 336: Why do we need to compute this here?

src/scanpy/preprocessing/_normalization.py
```diff
@@ -304,3 +308,283 @@ def normalize_total(  # noqa: PLR0912
...
+        target_sum = 4.0 * alpha * mean_depth
+        resolved_target = "auto"
+    else:
+        alpha = None
+        if target == "mean":
+            target_sum = mean_depth
+        elif target == "median":
+            depths = (
+                cell_depths.compute()
+                if isinstance(cell_depths, DaskArray)
```
ilan-gold: I think we should simply error out with dask rather than load everything into memory for median.  We can throw an error

src/scanpy/preprocessing/_normalization.py
```diff
@@ -304,3 +308,283 @@ def normalize_total(  # noqa: PLR0912
...
+
+    if isinstance(x, DaskArray):
+        if alpha is not None:
+            x = x * scale
+        log_values = x.map_blocks(
+            _log1p_sparse_block, dtype=np.float64, meta=x._meta.astype(np.float64)
+        )
+    else:
+        if alpha is not None:
+            x = x * scale
```
ilan-gold:
```suggestion
```

src/scanpy/preprocessing/_normalization.py
```diff
@@ -304,3 +308,283 @@ def normalize_total(  # noqa: PLR0912
...
+        else:
+            target_sum = float(target)
+        if not target_sum > 0:
+            msg = f"`target` must resolve to a positive depth, got {target_sum}."
+            raise ValueError(msg)
+        resolved_target = str(target)
+        x = axis_mul_or_truediv(
+            x, cell_depths / target_sum, op=truediv, axis=0, allow_divide_by_zero=False
+        )
+
```
ilan-gold, line 431:
```suggestion
   if alpha is not None:
        x = x * scale
```

src/scanpy/preprocessing/_normalization.py
```diff
@@ -304,3 +308,283 @@ def normalize_total(  # noqa: PLR0912
...
+    if isinstance(x, DaskArray):
+        if alpha is not None:
+            x = x * scale
```
ilan-gold:
```suggestion
```

src/scanpy/preprocessing/_normalization.py
```diff
@@ -304,3 +308,283 @@ def normalize_total(  # noqa: PLR0912
...
+        np.asarray(row_center.compute()).ravel()
+        if isinstance(row_center, DaskArray)
+        else np.asarray(row_center).ravel()
+    )
+    if not isinstance(log_values, DaskArray | CSBase):
+        log_values = sparse.csr_matrix(log_values)  # noqa: TID251
+
+    dense_x = _densify_shifted_clr(log_values, row_center) if densify else None
+    row_center_key = f"{key_added}_center"
+    metadata = dict(
```
ilan-gold, line 562: Do we need to be tracking this? I agree tracking is good, but I'd rather not do this ad-hoc

src/scanpy/preprocessing/_normalization.py
rwbaber: Because the subsequent conditional (`if denominator == 0.0`) and type cast (`alpha = float(numerator / denominator)`) need concrete scalar values at that point, not lazy Dask scalar objects.
I ran a quick test, and it looks like Dask can actually evaluate scalar truthiness and float conversion implicitly without throwing an error, but it does so sequentially for each line. This triggers separate computations and duplicates the upstream work.
Calling `dask.compute(numerator, denominator)` explicitly together merges both reductions into a single optimized pass. And this doesn't materialize the count matrix, it only executes the reduction graph down to the two final scalar sums.

src/scanpy/preprocessing/_normalization.py
rwbaber: Agreed. `target="median"` now raises for Dask input and asks the user to pass one of the others.

src/scanpy/preprocessing/_normalization.py
rwbaber: Yes, I'd say so: when densify=False, the layer stores uncentered sparse values rather than the final matrix, so the idea is that for PCA, this metadata explicitly declares that the layer has special storage semantics (encoding_type) and where the offset vector (row_center_key) is. But I adjusted it now with the params sub-dictionary that follows the existing Scanpy pattern (like uns["object"]["params"]).
Or what would you suggest?

ilan-gold:
@rwbaber I have been thinking a bit about this and I’m beginning to mull over the following (this isn’t me requesting, just asking for feedback):

1. We add a `clr` flavor to `pca` which handles doing the `clr` on-the-fly
2. The `clr` function itself always densifies

I think the downside would be clr-specific kwargs to `pca`. The upside would be no “ancillary” storage needed.

At the moment, I am not aware of where else this transform would be used in a way where the dense-ification needs be delayed.

And if it does, I’m wondering if perhaps for scanpy 2.0 (or sooner), `log1p` should just be adapted to handle the sparse part.

Just thinking out loud - the paper only benchmarks the transform for pca goes my thinking.

rwbaber:
I have thought about this a bit more as well, and I think I agree with the general direction @ilan-gold.

My current thinking is that `normalize_clr` should probably default to `densify=True`. I think for most "standard" dataset sizes, the dense centered matrix should not cause OOM errors, and it keeps the function behavior least surprising: after normalization, `.X` or the selected layer actually contains the normalized matrix.

Otherwise, it could get messy with downstream functions that expect a normalized matrix there (e.g. `sc.pp.highly_variable_genes`, `sc.tl.rank_genes_groups`, or external tools like celltypist). This is also why I agree that the current `densify=False` storage is awkward. Storing the row-centering vector in `.obs` and the uncentered sparse log matrix in `.layers`/`.X` could easily lead to misuse, since downstream functions would see a matrix that looks usable but is not the final centered CLR matrix.

So, also thinking out loud, a middle ground might be:

- `normalize_clr(..., densify=True)` as the default: write the actual dense centered CLR/PFlog matrix, so it behaves like a normal normalization function.
- `normalize_clr(..., densify=False)` as an opt-in for large datasets that would otherwise run into OOM errors: do not store an encoded sparse matrix, only estimate or store (if user-defined) the PFlog parameter (`alpha`) in `.uns`.
- `pca(..., flavor="clr")` or similar: read raw counts plus the stored/explicit `alpha`, and apply the implicit sparse CLR matrix internally for PCA to write `X_pca`.

Admittedly, `normalize_clr(densify=False)` would not do very heavy lifting for being its own function, but it would keep the familiar user flow of “normalize first, then PCA”, rather than asking users to skip `normalize_clr` and know that PCA itself should perform the transform.
In that case, the docs should of course be very clear that this would basically only be for anything that rides on X_pca (neighbors, UMAP, Leiden clustering, PAGA, ...) and otherwise only leaves the raw counts.

So the possible refactor would be: make `densify=True` the default, keep `densify=False` for the OOM/large-data path but store only `alpha`, and move the sparse implicit-matrix logic into a CLR-aware PCA path. That would also let us drop the current `encoding_type` / `row_center_key` layer representation entirely.

ilan-gold:
Sorry @rwbaber I was on vacation.  I do not think we should be exposing `densify=False` then because I think a lot of people will just blindly pass in whatever is in `X` - I would rather just hold that feature for later once it is clearer what this function will be used for outside of PCA. Otherwise, the plan seems reasonable.

rwbaber:
No worries, @ilan-gold, I was away for a while too. I also think it's reasonable for `normalize_clr` to always produce the complete matrix and to leave the memory-efficient approach for a separate PR. As you suggested, that could potentially be contained entirely in `pca` as a CLR flavor or something similar, once the intended API is clearer.

I have therefore reduced this PR to implementing `normalize_clr`, and removed all CLR-specific PCA changes. `scanpy.pp.pca` is now unchanged relative to `main`.

`normalize_clr` now always writes the complete centered PFlog matrix to `.X` or the selected layer. I removed the `densify` option and the sparse shifted-log plus separate row-centering representation. This avoids exposing an intermediate matrix that looks like normalized data but could be interpreted incorrectly by downstream functions.

I simplified the normalization API further, although I would be interested in your opinion on this part:

- I removed `target="mean"`, `"median"`, and numeric targets: These previously selected a fixed proportional-fitting scale \(K\), giving `center(log1p(K * x / cell_depth))`. I don't really see in which scenario these fixed-target shifted-CLR variants would be more appropriate than the PFlog method recommended by the paper (the auto overdispersion estimation). I originally included the fixed-target variants because they are exposed by `scclr`/`runorm`. Though, I'm not sure whether they are mainly retained there for benchmarking and compatibility.

With those fixed-target modes removed, the `target` parameter is of course obsolete, so instead:

- `alpha=None` estimates the dataset-level negative-binomial overdispersion and applies the Anscombe-calibrated transform.
- An explicit positive `alpha` applies the same transform using a supplied value, which remains useful for reproducibility or reusing an estimate.

The resulting public interface is essentially:

```python
sc.pp.normalize_clr(
    adata,
    alpha=None,
    layer=None,
    inplace=True,
    copy=False,
)
```

Do you think keeping the API focused on the recommended Anscombe-calibrated PFlog method makes sense, or should `normalize_clr` also expose the fixed mean/median/numeric-\(K\) shifted-CLR variants? I can reintroduce them if the broader functionality is desirable.

Also, Dask support remains: Dask arrays with either dense or sparse raw-count blocks produce a lazy Dask array containing the final centered values in dense output blocks.

PR #4168, opened 2026-06-22
docs: add inline example for pca_scatter

joeydufourd:
Adds an inline rendered example for `sc.pl.pca_scatter`, addressing part of #1664.

`pca_scatter` is a backwards-compatible alias for `pca`, so this PR replaces the plain alias with a small wrapper that keeps behavior unchanged while giving the alias its own docstring and plot example.

- [x] Addresses part of #1664 
- [x] [Tests][] included or not required because: this is a documentation-only change adding an inline example for `sc.pl.pca_scatter`.
- [x] [Release notes][] not necessary because: this only adds a documentation example and does not change user-facing behavior.

### DIFF
--- a/src/scanpy/plotting/_tools/__init__.py
+++ b/src/scanpy/plotting/_tools/__init__.py
@@ -102,7 +102,34 @@ def pca_overview(adata: AnnData, **params):
 
 
 # backwards compat
-pca_scatter = pca
+@_doc_params(scatter_bulk=doc_scatter_embedding, show_save_ax=doc_show_save_ax)
+def pca_scatter(adata: AnnData, **params):
+    """Scatter plot in PCA coordinates.
+
+    This is a backwards-compatible alias for :func:`~scanpy.pl.pca`.
+
+    Parameters
+    ----------
+    adata
+        Annotated data matrix.
+    {scatter_bulk}
+    {show_save_ax}
+
+    Returns
+    -------
+    If `show==False` a :class:`~matplotlib.axes.Axes` or a list of it.
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        import scanpy as sc
+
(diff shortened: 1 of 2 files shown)


joeydufourd:
I checked the failing Read the Docs log. The build appears to fail because existing Sphinx warnings are treated as errors, mostly unresolved `scanpy.settings.*` references, rather than because of the new `pca_scatter` example.

I also checked locally:

`bash
python -m compileall src/scanpy/plotting/_tools/__init__.py
git diff --check`

Both passed, and the branch is clean.

The “Check title, milestone, and labels” job also seems to fail because the PR has no label/milestone, which I don’t think I can set as an external contributor. Could a maintainer add the appropriate label/milestone?

src/scanpy/plotting/_tools/__init__.py
```diff
@@ -102,7 +102,34 @@ def pca_overview(adata: AnnData, **params):
...
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        import scanpy as sc
+
+        adata = sc.datasets.pbmc3k_processed()
+        sc.pl.pca_scatter(adata, color="louvain")
+    """
```
ilan-gold:
```suggestion
    Parameters
    ----------
    adata
        Annotated data matrix.
    {scatter_bulk}
    {show_save_ax}

    Returns
    -------
    If `show==False` a :class:`~matplotlib.axes.Axes` or a list of it.
```

In the interest of pushing people towards the new method, maybe no example, but thanks for the docs!

ilan-gold (approved):
I would remove the example because we don't want people using this! If `pl.pca` needs an example, let's put it there, thanks :)

joeydufourd:
Thanks! Removed the example to avoid encouraging use of pca_scatter.

PR #4170, opened 2026-06-22
feat: add networkit flavor for parallel Leiden clustering

amalia-k510:
This PR adds NetworKit's ParallelLeiden as a new flavor option for `scanpy.tl.leiden`, enabling multithreaded Leiden community detection via `scanpy.tl.leiden(adata, flavor="networkit")`. The motivation behind is that scanpy's current Leiden backends (igraph, leidenalg) are single-threaded. For atlas-scale datasets (500k+ cells), clustering becomes a bottleneck. NetworKit's ParallelLeiden is a C++ parallel implementation available via pip with no additional compilation, making it the lowest-friction path to parallel Leiden in scanpy.


src/scanpy/_utils/__init__.py
```diff
@@ -98,6 +99,17 @@ def ensure_igraph() -> None:
     raise ImportError(msg)
 
 
+def ensure_network() -> None:
```
ilan-gold: Where is this used?
```suggestion
def ensure_networkit() -> None:
```

src/scanpy/_utils/__init__.py
```diff
@@ -98,6 +99,17 @@ def ensure_igraph() -> None:
...
+    if importlib.util.find_spec("netowrkit"):
```
ilan-gold:
```suggestion
    if importlib.util.find_spec("networkit"):
```

src/scanpy/tools/_leiden.py
```diff
@@ -185,6 +185,33 @@ def leiden(  # noqa: PLR0913
             "MutableVertexPartition",
             leidenalg.find_partition(g, partition_type, seed=seed, **clustering_args),
         )
+    elif flavor == "networkit":
+        from types import SimpleNamespace
+
+        import networkit
+
+        seed = int(rng.integers(np.iinfo(np.int64).max))
+        networkit.setSeed(seed, useThreadId=True)
```
ilan-gold, line 201: Despite this, things aren't deterministic, right? Just making sure, but if that's the case, please comment why we do this (helps with determinism even if it's not perfect maybe?)

src/scanpy/tools/_leiden.py
```diff
@@ -185,6 +185,33 @@ def leiden(  # noqa: PLR0913
...
+        # only undirected for Parallel Leiden
+        g = _utils.get_networkit_from_adjacency(adjacency, weighted=use_weights)
+        iterations = n_iterations if n_iterations > 0 else 3
```
ilan-gold, line 208: Where did the number 3 come from?

ilan-gold (commented):
Nice! Would be best to run the installation check :)

Also please add an extra to our `pyproject.toml`.

Let's try to use a heuristic about warning users about large inputs to `igraph` and the faster alternative

ilan-gold:
Also can you point this branch at https://github.com/networkit/networkit/pull/1422 (and then `main` after merge followed by the correct min version when released) and use this method (also in your notebook) to make sure it works properly?

src/scanpy/tools/_leiden.py
amalia-k510: I checked the the NetworKit's header file for the ParallelLeiden and found that default for ParallelLeiden.iterations is set to three; hence why I decided to use it here as well.
`explicit ParallelLeiden(const Graph &graph, int iterations = 3, bool randomize = true, double gamma = 1);`

ilan-gold: Can we document that? Also, why isn't it the default then for python?

src/scanpy/tools/_leiden.py
```diff
@@ -263,6 +301,20 @@ def _validate_flavor(
...
     return flavor
+
+
+def _maybe_suggest_networkit(g: Graph) -> None:
+    """Encourage users toward the parallel NetworKit backend on large graphs."""
+    _networkit_edge_heuristic = 500000
+    if g.ecount() < _networkit_edge_heuristic:
+        return
+    if importlib.util.find_spec("networkit") is None:
+        return
```
ilan-gold, line 323: Would still warn even if there is no `networkit` installed :)

Do you have a benchmarking notebook as a gist somewhere?

amalia-k510: I think this is already handled. It returns early if `find_spec("networkit") is None`, so the hint only fires when networkit's actually installed. Might've been from before I pushed the guard? Let me know if you're still seeing it warn without networkit and I'll take another look.

src/scanpy/_utils/__init__.py
amalia-k510: It was a type. I fixed. I renamed to `ensure_networkit`, and it's called at the top of the `elif flavor == "networkit":` branch in `_leiden.py` before the import.

src/scanpy/tools/_leiden.py
amalia-k510: Right, it won't be fully reproducible but, from what I understand, we can get close to it. Seeding controls which nodes the algorithm tries to move and in what order, but when two threads race to modify neighbors at the same time, the OS decides who wins, and, from what I understand, that part can't be seeded. It gets you most of the way to full reproducibility, just not the last part. I've added a comment to clarify.

src/scanpy/tools/_leiden.py
amalia-k510: I checked and turns out that `iterations=3` is already the default on the Python side for ParallelLeiden. I could drop it, but I left it explicit so the value is obvious at the call site and clearly maps to scanpy's n_iterations.

src/scanpy/tools/_leiden.py
amalia-k510: Also, I tried to create gist, and it is erroring out on the notebook. I think the file is too large with the embedded outputs. Want me to push it to the PR branch, strip the outputs and re-upload to a gist, or just share it another way?

ilan-gold: Discussed IRL, but notebook with a cell for downloading the data works

tests/test_clustering.py
ilan-gold, line 55: We should probably update this warning to use `_maybe_suggest_networkit` if `networkit` is detected

ilan-gold, line 174: Can we add `networkit` variants to both of these tests? I would expect fairly concordant results.

src/scanpy/tools/_leiden.py
```diff
@@ -185,8 +187,45 @@ def leiden(  # noqa: PLR0913
...
+    elif flavor == "networkit":
+        from types import SimpleNamespace
+
+        _utils.ensure_networkit()
+        import networkit
+
+        # Seeding controls which nodes are tried and in what order, but not
+        # which thread wins when two move neighboring nodes simultaneously.
+        # So runs are statistically reproducible for a fixed thread count,
+        # but not bit-for-bit identical.
```
ilan-gold, line 199: Oh wow, this is an amazing find. Can we add a unit test for this maybe? I think I mentioned adding a unit test independently of this anyway. I would also mention this in the docstring.

ilan-gold (commented):
Nice! Just small stuff. Can you add your notebook as well? Really great to see this!

PR #4178, opened 2026-06-25
fix: masked cells were accidentally added to group 0 + mask ignored in aggregate computations in `get.aggregate`

zboldyga:
- [ ] Closes #

### DIFF
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -160,10 +160,10 @@ class Aggregate[ArrayT: np.ndarray | CSBase]:
         """
         medians = []
         for group in range(len(self.groupby.categories)):
-            mask = self.groupby.codes == group
+            group_mask = self.groupby.codes == group
             if self.mask is not None:
-                mask = mask & self.mask
-            group_data = self.data[mask]
+                group_mask = group_mask & self.mask
+            group_data = self.data[group_mask]
             if isinstance(group_data, CSBase):
                 if group_data.format != "csc":
                     group_data = group_data.tocsc()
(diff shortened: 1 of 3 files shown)


src/scanpy/get/_aggregated.py
```diff
@@ -555,19 +557,30 @@ def _combine_categories(
...
     categorical: pd.Categorical,
     *,
     mask: NDArray[np.bool] | None = None,
 ) -> sparse.coo_array:
-    # TODO: why is this float64.  This is a scanpy 2.0 problem maybe?
-    mask = (
-        np.broadcast_to(1.0, len(categorical)) if mask is None else mask.astype("uint8")
-    )
-    # can’t have -1s in the codes, but (as long as it’s valid), the value is ignored, so set to 0 where masked
-    codes = np.where(mask, categorical.codes, 0)
```
zboldyga, line 702: @ilan-gold here's where we were setting mask to 0 for cells that were masked, meaning they appear to be in group 0 (which is a real group with other cells)

zboldyga: @ilan-gold additionally, 'mask' was not being leveraged in the various computations. The net is that group 0 would erroneously include data for any masked cells.

I figured the best way to solve this is to drop the masked cells altogether, so each computation does not have to independently apply a mask

There are a few other similar issues resolved here (n_obs was wrong, the total count used for certain calculations was wrong)

ilan-gold (approved):
Thanks a bunch! No idea why this wasn't tested yikes!

zboldyga:
@ilan-gold lgtm thanks for the tidy-ups!

PR #4179, opened 2026-06-25
feat: add Array API support via array-api-compat

amalia-k510:
This adds Array API support to scanpy's preprocessing pipeline, where I am planning to use JAX as the test case. The goal is for someone to put a JAX array into `adata.X` and run the usual pipeline without the array getting silently pulled to CPU partway through. It builds on the recent `fast-array-utils` work, which means a lot of the heavy lifting (sums, means, variances) already works across backends.


flying-sheep (changes requested):
OK, please remember the comments from the fast-array-utils PR, e.g. we agreed to handle singledispatch functions like this: https://github.com/scverse/fast-array-utils/blob/febaf245ecd6da849c2331638219f649935d5a21/src/fast_array_utils/stats/_power.py#L38-L39

We also need tests. There’s `MAP_ARRAY_TYPES` which should gain an entry, which will make many many tests run with the type you add. We can discuss how to best exclude certain tests, for starters you can just add it and make some of the tests fail – that’ll show you where things aren’t working yet (and might or might not be fixed in this PR)

Please also add a release note (`hatch run towncrier:create 4179.feat.md`)

src/scanpy/_utils/__init__.py
```diff
@@ -733,12 +782,36 @@ def _[T: (DaskArray, np.ndarray)](
...
 
 
-@axis_nnz.register(CSBase)
-def _(x: CSBase, /, axis: Literal[0, 1]) -> np.ndarray:
-    return x.count_nonzero(axis=axis)
+@axis_nnz.register(HasArrayNamespace)
+def _(x: HasArrayNamespace, /, axis: Literal[0, 1]) -> Any:
+
+    xp = get_namespace(x)
+    return xp.count_nonzero(x, axis=axis)
```
flying-sheep: this handles numpy with equal performace too no?

src/scanpy/_utils/__init__.py
```diff
@@ -733,12 +782,36 @@ def _[T: (DaskArray, np.ndarray)](
...
+
+else:
+    # older scipy versions don’t have any way to get the nnz of a sparse array
+    @axis_nnz.register(CSBase)
+    def _(x: CSBase, /, axis: Literal[0, 1]) -> np.ndarray:
+        if isinstance(x, _CSArray):
+            from scipy.sparse import csc_array, csr_array  # noqa: TID251
+
+            x = (csr_array if x.format == "csr" else csc_array)(x)
+        return x.getnnz(axis=axis)
```
flying-sheep: looks like you got a merge wrong, this change reverts an intentional simplification.

src/scanpy/_utils/__init__.py
```diff
@@ -757,6 +830,17 @@ def check_nonnegative_integers(x: _SupportedArray, /) -> bool | DaskArray:
     raise NotImplementedError
 
 
+@check_nonnegative_integers.register(HasArrayNamespace)
+def _check_nonnegative_integers_array_api(x: HasArrayNamespace, /) -> bool:
```
flying-sheep: should this go below the `np.ndarray` case? I think `_check_nonnegative_integers_array_api.dispatch(np.ndarray)` will return the first match, no?

src/scanpy/metrics/_common.py
```diff
@@ -93,11 +92,17 @@ def _resolve_vals(val: pd.DataFrame | pd.Series) -> NDArray: ...
...
-def _resolve_vals(val: object) -> NoReturn:
+def _resolve_vals(val: object):
     msg = f"Unsupported type {type(val)}"
     raise TypeError(msg)
 
 
+@_resolve_vals.register(HasArrayNamespace)
+def _resolve_vals_array_api(val: HasArrayNamespace) -> NDArray:
+    # Moran's I / Geary's C use numba kernels, so convert at the boundary
+    return np.asarray(val)
```
flying-sheep: same here, should go below the `np.ndarray` case right?

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -398,19 +399,25 @@ def _highly_variable_genes_single_batch(
     if n_removed:
         x = x[:, filt].copy()
 
+    from .._compat import get_namespace  ### double check
```
flying-sheep: what does `### double check` mean?

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -398,19 +399,25 @@ def _highly_variable_genes_single_batch(
...
+        elif isinstance(x, HasArrayNamespace):
+            xp = get_namespace(x)
+            x = xp.expm1(x)
         else:
             x = np.expm1(x)
 
     mean, var = materialize_as_ndarray(stats.mean_var(x, axis=0, correction=1))
     # now actually compute the dispersion
-    mean[mean == 0] = 1e-12  # set entries equal to zero to small value
+    # allocating a fresh writiable array = jax issue
```
flying-sheep: typo, also I don’t understand what that means specifically. Is the `=` supposed to mean “leads to”? Also: what kind of issue?

src/scanpy/preprocessing/_normalization.py
```diff
@@ -15,12 +15,16 @@
...
-def _compute_nnz_median(counts: np.ndarray | DaskArray) -> np.floating:
+def _compute_nnz_median(
+    counts: np.ndarray | DaskArray | HasArrayNamespace,
+) -> np.floating:
     """Given a 1D array of counts, compute the median of the non-zero counts."""
     if isinstance(counts, DaskArray):
         counts = counts.compute()
+
     counts_greater_than_zero = counts[counts > 0]
     median = np.median(counts_greater_than_zero)
```
flying-sheep: no median for array api?

src/scanpy/preprocessing/_scale.py
```diff
@@ -177,13 +190,21 @@ def scale_array[A: _Array](
...
+            )
+            x = x.astype(np.float64)
+    else:
+        xp = get_namespace(x)
+        if xp.isdtype(x.dtype, "integral"):
+            logg.info(
+                "... as scaling leads to float results, integer "
+                "input is cast to float, returning copy."
+            )
+            x = xp.astype(x, xp.float64)
```
flying-sheep: assign the message before this if/else block so it’s not duplicated

src/scanpy/preprocessing/_simple.py
```diff
@@ -355,6 +356,15 @@ def log1p(
     return log1p_array(data, copy=copy, base=base)
 
 
+@log1p.register(HasArrayNamespace)
```
flying-sheep: again specificity/order. please check all of these in case I forgot one!

src/scanpy/_compat.py
```diff
@@ -66,6 +67,13 @@ def pkg_metadata(package: str) -> PackageMetadata:
     return metadata(package)
 
 
+def get_namespace(x):
```
flying-sheep: please add the appropriate return type

src/testing/scanpy/_pytest/__init__.py
```diff
@@ -17,6 +18,10 @@
 if TYPE_CHECKING:
     from collections.abc import Generator, Iterable, Mapping
 
+if find_spec("jax"):
+    import jax
+
+    jax.config.update("jax_enable_x64", True)  # noqa: FBT003
```
flying-sheep, line 26: can you add a short comment what that does?

flying-sheep (changes requested):

src/scanpy/_utils/__init__.py
amalia-k510: `xp.count_nonzero` resolves to a wrapper over `np.count_nonzero`. The difference between the two in negligible. I decided to keep it explicit due to the code consistency as that is what I have been doing for other fuctions. I can drop it, if you'd rather have less code.

src/scanpy/preprocessing/_highly_variable_genes.py
amalia-k510: Leftover note to myself, sorry. The import was real but I forgot to move it to the module level; moved it to the `_compat` import and deleted the inline one.

src/scanpy/preprocessing/_normalization.py
amalia-k510: median isn't in the array API spec (it's not in the statistical functions list), but `array-api-compat` namespace expose it, so `xp.median` works for the backends we actually support.

src/scanpy/preprocessing/_highly_variable_genes.py
amalia-k510: Fixed the typo, and rewrote the comment to actually explain it: JAX arrays are immutable, so the old in-place `mean[mean == 0] = 1e-12` was leading to a TypeError. So I have switch to `np.where` returns a fresh array instead

src/scanpy/_compat.py
```diff
@@ -66,6 +68,13 @@ def pkg_metadata(package: str) -> PackageMetadata:
     return metadata(package)
 
 
+def get_namespace(x) -> ModuleType:
```
flying-sheep, line 71: Did I forget something or why aren’t we using types from here? https://array-api.readthedocs.io/en/latest/index.html

pyproject.toml
```diff
@@ -154,6 +157,7 @@ test-min = [
 [tool.hatch]
 version.source = "vcs"
 version.raw-options.version_scheme = "release-branch-semver"
+metadata.allow-direct-references = true
```
flying-sheep, line 160: why?

flying-sheep (commented):
Looking good! Just nitpicks left in the code that’s there.

One big change we should do is updating the array type metadata so it lists Array API support.

The docs currently have [one central table](https://scanpy.readthedocs.io/en/stable/api/index.html#array-type-support) and each function has its own little info box.

Since you know what’s going to be supported, you should update the metadata by adding an `aa` tag and filling out the support here (so far, `np` means numpy, `sp` means scipy sparse, `da` means dask, `da[...]` means “... in dask”, and `sp[csc]` means “scipy sparse with csc format”):

https://github.com/scverse/scanpy/blob/fc0b41901c34d23419a420f5eb24c497513a4dbb/docs/conf.py#L178-L210

If you want, you can also update [the Sphinx code](https://github.com/scverse/scanpy/blob/fc0b41901c34d23419a420f5eb24c497513a4dbb/docs/extensions/array_support.py), but you don’t need to learn how to do that if you don’t want to. I’ll happily do that myself.

pyproject.toml
amalia-k510: No reason anymore. I used for `fast-array-utils` latest version, but since it was updated, I no longer use this line of code. I ended up removing it.

src/scanpy/_compat.py
amalia-k510: I think it is just a leftover one that I forgot to change. I ended up switching to using `HasArrayNamespace` from `fast-array-utils` for x, since that's what we already dispatch on everywhere. So my question is, would it be better to stick to `fast-array-utils` or switch to a proper `ArrayNamespace` instead?

flying-sheep: None of that, we should
1. remove this function and directly use the one from `array_api_compat` instead.
2. add `types-array-api` next to `scipy-stubs`

tests/test_pca.py
```diff
@@ -154,9 +154,9 @@ def possible_solvers(
             svd_solvers = {"arpack", "covariance_eigh"}
         case (type() as dc, False) if issubclass(dc, CSBase):
             svd_solvers = {"arpack", "randomized"}
-        case (helpers.asarray, True):
-            svd_solvers = {"auto", "full", "arpack", "randomized", "covariance_eigh"}
-        case (helpers.asarray, False):
+        case (helpers.asarray | helpers.as_dense_jax_array, True):
+            svd_solvers = {"auto", "full", "arpack", "randomized"} | SKLEARN_ADDITIONAL
```
flying-sheep, line 159: this doesn’t exist and breaks the tests

flying-sheep (changes requested):

src/scanpy/tools/_rank_genes_groups.py
```diff
@@ -878,6 +879,11 @@ def rank_genes_groups(  # noqa: PLR0912, PLR0913, PLR0915
     >>> sc.pl.rank_genes_groups(adata)
 
     """
+    from scanpy import settings
+
+    # rank_genes_groups uses numba kernels internally, so need convert at entry.
+    if isinstance(adata.X, HasArrayNamespace) and not isinstance(adata.X, np.ndarray):
+        adata.X = np.asarray(adata.X)
```
flying-sheep, line 886: This isn’t the right approach, as it modifies an object the user passed. Why not do this in `_RankGenes.__init__`?

tests/test_pca.py
flying-sheep: Why did you resolve this? `SKLEARN_ADDITIONAL` still doesn’t exist and breaks the tests.

flying-sheep:
@amalia-k510 what’s the state here? Tests and checks are failing, and there’s a pending review comment from me.

src/scanpy/preprocessing/_simple.py
```diff
@@ -379,6 +381,15 @@ def log1p_array(x: np.ndarray, *, base: Number | None = None, copy: bool = False
     return x
 
 
+@log1p.register(HasArrayNamespace)
+def log1p_array_api(x, *, base: Number | None = None, copy: bool = False):
+    xp = array_namespace(x)
+    result = xp.log1p(x)
+    if base is not None:
+        result = result / float(np.log(base))
```
flying-sheep, line 389: why not `xp`?

src/scanpy/preprocessing/_scale.py
```diff
@@ -74,6 +88,49 @@ def clip_array(
...
+    if isinstance(x, np.ndarray | CSBase | DaskArray):
+        if not np.issubdtype(x.dtype, np.integer):
+            return x
+        logg.info(msg)
+        return x.astype(np.float64)
+    xp = array_namespace(x)
+    if not xp.isdtype(x.dtype, "integral"):
+        return x
+    logg.info(msg)
+    return xp.astype(x, xp.float64)
```
flying-sheep, line 106: do we need a numpy branch here? this should function identically when using just the array api version right? or does it do an extra copy in that case?

if you condense this to one branch, please remove the helper and move it back into the original function.

src/scanpy/preprocessing/_scale.py
```diff
@@ -54,6 +62,12 @@ def _(x: DaskArray, *, max_value: float, zero_center: bool = True) -> DaskArray:
     )
 
 
+@clip.register(HasArrayNamespace)
+def _(x, *, max_value: float, zero_center: bool = True):
+    xp = array_namespace(x)
+    return xp.clip(x, min=-max_value if zero_center else None, max=max_value)
```
flying-sheep, line 68: why the `-max_value`, are the semantics of Array API different than the others?

src/scanpy/_utils/__init__.py
```diff
@@ -772,6 +824,16 @@ def _check_nonnegative_integers_in_mem(x: _MemoryArray, /) -> bool:
...
+    xp = array_namespace(x)
+    if bool(xp.any(x < 0)):
+        return False
+    if xp.isdtype(x.dtype, "integral"):
+        return True
+    return not bool(xp.any((x % 1) != 0))
```
flying-sheep, line 834: maybe we can add a shortcut branch for “if unsigned integer then return `True` before checking individual elements”

I don’t know why we don’t do that for the numpy case either.

tests/test_rank_genes_groups.py
```diff
@@ -21,7 +22,10 @@
 from testing.scanpy._helpers import random_mask
 from testing.scanpy._helpers.data import pbmc68k_reduced
 from testing.scanpy._pytest.marks import needs
-from testing.scanpy._pytest.params import ARRAY_TYPES, ARRAY_TYPES_MEM
+from testing.scanpy._pytest.params import (
+    ARRAY_TYPES,
+    ARRAY_TYPES_MEM,
+)
```
flying-sheep, line 28: there’s something wrong with how you ran ruff. please just run it via `prek`, then it should be configured properly.

```suggestion
from testing.scanpy._pytest.params import ARRAY_TYPES, ARRAY_TYPES_MEM
```

docs/conf.py
```diff
@@ -191,11 +191,11 @@
     "pp.highly_variable_genes": (["np", "sp", "da"], ["da[sp[csc]]"]),
     "pp.log1p": (["np", "sp", "da"], []),
```
flying-sheep, line 192: please make sure that you mark *all* functions that are now supported.

also please check in the tests that these functions preserve the data type. you can do that only for jax if it’s too hard, we plan to update the tests with a way to make it easier to get the expected type in the future.

flying-sheep (changes requested):
looks good! I merged your anndata fix and removed the xfail branches.

Only some small issues left, except for the part about documenting and testing what you changed: https://github.com/scverse/scanpy/pull/4179#discussion_r4047099304

src/scanpy/preprocessing/_simple.py
amalia-k510: I think I accidentally mistyped it. It should have been `xp.log(base)`

tests/test_rank_genes_groups.py
amalia-k510: Kept it multi-line here. I think when I was merging, I might have deleted `as_dense_jax_array` as I needed  in here too, which pushes it over the line length, so ruff wraps it.

src/scanpy/preprocessing/_scale.py
amalia-k510: Semantics are the same. I tested against `clip_array` with numpy and they give identical results. It uses inline `xp.clip` instead of `clip_array` because `clip_array` is numba-based and mutates in place: numba can't handle a JAX array (throws a `TypingError`), and the in-place writes fail since JAX arrays are immutable. So the array-api path uses `xp.clip` to do the same thing.

tests/test_rank_genes_groups.py
flying-sheep: My point is that your PR currently does

```diff
-from testing.scanpy._pytest.params import ARRAY_TYPES, ARRAY_TYPES_MEM
+from testing.scanpy._pytest.params import (
+    ARRAY_TYPES,
+    ARRAY_TYPES_MEM,
+)
```

The suggestion above is to revert that. I fixed similar formatting-only changes but missed this one.

Might be that you first added an import then removed it again, and Ruff’s magic comma keeps it multi-line, which means that your PR now introduces a change where none is needed.

src/scanpy/preprocessing/_scale.py
flying-sheep: Got it, must have missed it in the numba code!

tests/test_highly_variable_genes.py
```diff
@@ -661,6 +662,13 @@ def test_subset_inplace_consistency(
     - for dask arrays and non-dask arrays
     - for both with and without batch_key
     """
+    if array_type is as_dense_jax_array and flavor in {"seurat_v3", "seurat_v3_paper"}:
+        request.applymarker(
+            pytest.mark.xfail(
+                reason="seurat_v3 not jax compatible (numba aggregation + DLPack readonly limitation)",
+                strict=False,
+            )
+        )
```
flying-sheep, line 671: I fixed the DLPack thing in anndata. What does “numba aggregation” mean?

amalia-k510: Got it! I just addressed this issue.

flying-sheep: never do `strict=False` please

flying-sheep (approved):

flying-sheep:
actually what’s your reply to this? https://github.com/scverse/scanpy/pull/4179#discussion_r4047099304

> also please check in the tests that these functions preserve the data type. you can do that only for jax if it’s too hard, we plan to update the tests with a way to make it easier to get the expected type in the future.

PR #4190, opened 2026-06-30
ci: fix benchmarks

flying-sheep:
Looks like `setup()` only reliably runs *after* `setup_cache` if there are parameters.

So I’m returning something from `setup_cache` to introduce a “parameter”.

### DIFF
--- a/benchmarks/benchmarks/preprocessing_counts.py
+++ b/benchmarks/benchmarks/preprocessing_counts.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 
 from inspect import signature
 from itertools import product
-from pathlib import Path
 from typing import TYPE_CHECKING
 
 import anndata as ad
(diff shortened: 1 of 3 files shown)


benchmarks/benchmarks/preprocessing_counts.py
```diff
@@ -24,12 +25,12 @@
     from ._utils import Dataset, KeyCount
 
 
-def cache_adata(dataset: Dataset, layer: KeyCount) -> None:
+def make_adata(dataset: Dataset, layer: KeyCount) -> ad.AnnData:
     """Without this caching, asv was running several processes which meant the data was repeatedly downloaded."""
```
flying-sheep:
```suggestion
```

this comment makes no sense here anymore. note to self: commit this after removing the benchmark label.

ilan-gold (approved):

PR #4192, opened 2026-07-01
fix: aggregate dask multifunc

zboldyga:
@ilan-gold there was a pre-existing bug in aggregate's dask code that causes a failure when calling with multiple function types. it works for just mean and var, but if adding anything else it failed. simple fix. this is a blocker for the stats PR

- [ ] Closes #


ilan-gold:
@zboldyga Great, thanks!

PR #4196, opened 2026-07-04
docs: document that variance/std use ddof=1 (sample correction)

AxelMcKenna:
- [ ] Closes #
- [x] [Tests][] included or not required because: docs-only change, no behavior is added or modified.

---

`sc.pp.scale` computes gene-wise variance and standard deviation with Bessel's correction (`ddof=1`), via `mean_var(x, axis=0, correction=1)` in `scale_array` and `scale_array_masked`. This convention isn't stated anywhere in the docstring, so a reimplementation using `numpy.std()` defaults (`ddof=0`, population variance) silently diverges from `sc.pp.scale`'s output - negligible for large `n_obs`, but non-negligible for small datasets (e.g. ~0.02% on a ~2700-cell PBMC3k-sized dataset, enough to shift where `max_value` clipping takes effect).

This PR adds a `.. note::` to the `scale` docstring stating the correction used. No behavior change.


ilan-gold (approved):
@AxelMcKenna If you wanted to make a PR to support other options, feel free!

PR #4199, opened 2026-07-06
feat: Allow using accessors in `sc.get.aggregate`

flying-sheep:
- [ ] Closes #

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -142,8 +142,7 @@ pygments_dark_style = "native"
 katex_prerender = shutil.which(NODEJS_BINARY) is not None
 
 intersphinx_mapping = dict(
-    # Needs latest until `.acc` is released in 0.13
-    anndata=("https://anndata.scverse.org/en/latest/", None),
+    anndata=("https://anndata.scverse.org/en/stable/", None),
     bbknn=("https://bbknn.readthedocs.io/en/latest/", None),
     cuml=("https://docs.rapids.ai/api/cuml/stable/", None),
     cycler=("https://matplotlib.org/cycler/", None),
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 import shutil
 import sys
 from datetime import datetime
@@ -126,6 +127,7 @@ nb_execution_mode = "cache"
 nb_execution_excludepatterns = [
     f"{d}{'/*' * n}" for d in ["tutorials", "how-to"] for n in (1, 2, 3)
 ]
+nb_execution_show_tb = bool(os.environ.get("READTHEDOCS"))
 nb_merge_streams = True
 
 ogp_site_url = "https://scanpy.scverse.org/en/stable/"
@@ -140,7 +142,8 @@ pygments_dark_style = "native"
 katex_prerender = shutil.which(NODEJS_BINARY) is not None
 
 intersphinx_mapping = dict(
-    anndata=("https://anndata.scverse.org/en/stable/", None),
+    # Needs latest until `.acc` is released in 0.13
+    anndata=("https://anndata.scverse.org/en/latest/", None),
     bbknn=("https://bbknn.readthedocs.io/en/latest/", None),
     cuml=("https://docs.rapids.ai/api/cuml/stable/", None),
     cycler=("https://matplotlib.org/cycler/", None),
--- a/src/scanpy/_settings/presets.py
+++ b/src/scanpy/_settings/presets.py
@@ -5,7 +5,7 @@ import inspect
 import re
 from contextlib import contextmanager
 from dataclasses import dataclass
-from functools import cached_property, partial, wraps
+from functools import cache, cached_property, partial, wraps
 from importlib.metadata import distributions, requires
 from typing import TYPE_CHECKING, Literal, NamedTuple
 
@@ -17,7 +17,10 @@ from .._utils._doctests import doctest_needs
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Generator, Mapping
+    from collections.abc import Set as AbstractSet
     from typing import Self
+
+    from packaging.utils import NormalizedName
 
 
 __all__ = [
@@ -312,14 +315,18 @@ class Preset(enum.StrEnum):
                 return self
 
 
+@cache
+def dist_names() -> AbstractSet[NormalizedName]:
+    return dict.fromkeys(canonicalize_name(d.name) for d in distributions()).keys()
+
+
 def _missing_scanpy2_deps() -> list[Requirement]:
-    dist_names = {canonicalize_name(d.name) for d in distributions()}
     return [
         r
         for r in map(Requirement, requires("scanpy") or ())
         if r.marker
         and r.marker.evaluate({"extra": "scanpy2"}, "requirement")
-        and canonicalize_name(r.name) not in dist_names
+        and canonicalize_name(r.name) not in dist_names()
     ]
 
 
--- a/src/scanpy/_settings/presets.py
+++ b/src/scanpy/_settings/presets.py
@@ -6,7 +6,7 @@ import re
 from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import cache, cached_property, partial, wraps
-from importlib.metadata import distributions, requires
+from importlib.metadata import distributions, requires, version
 from typing import TYPE_CHECKING, Literal, NamedTuple
 
 from packaging.requirements import Requirement
@@ -315,19 +315,28 @@ class Preset(enum.StrEnum):
                 return self
 
 
-@cache
-def dist_names() -> AbstractSet[NormalizedName]:
-    return dict.fromkeys(canonicalize_name(d.name) for d in distributions()).keys()
-
-
 def _missing_scanpy2_deps() -> list[Requirement]:
     return [
         r
         for r in map(Requirement, requires("scanpy") or ())
-        if r.marker
-        and r.marker.evaluate({"extra": "scanpy2"}, "requirement")
-        and canonicalize_name(r.name) not in dist_names()
+        if (
+            r.marker
+            and r.marker.evaluate({"extra": "scanpy2"}, "requirement")
+            and not _req_satisfied(r)
+        )
     ]
+
+
+def _req_satisfied(req: Requirement) -> bool:
+    return (
+        canonicalize_name(req.name) in dist_names()
+        and version(req.name) in req.specifier
+    )
+
+
+@cache
+def dist_names() -> AbstractSet[NormalizedName]:
+    return dict.fromkeys(canonicalize_name(d.name) for d in distributions()).keys()
 
 
 for postprocess in preset_postprocessors:
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -246,8 +246,7 @@ def aggregate(
     ),
     func: AggType | Iterable[AggType],
     *,
-    # TODO: allow str once `.resolve` supports accs
-    acc: LayerAcc | MultiAcc | None = None,
+    acc: LayerAcc | MultiAcc | str | None = None,
     mask: NDArray[np.bool] | AdRef[Idx2D | int, AnnData] | str | None = None,
     dof: int = 1,
     # old API
--- a/src/scanpy/get/_aggregated.py
+++ b/src/scanpy/get/_aggregated.py
@@ -13,6 +13,7 @@ from scipy import sparse
 from sklearn.utils.sparsefuncs import csc_median_axis_0
 
 from .._compat import CSBase, CSRBase, DaskArray, warn
+from .._settings import Preset, settings
 from .._utils import _resolve_axis, get_literal_vals
 from .._utils._doctests import doctest_needs
 from ._kernels import (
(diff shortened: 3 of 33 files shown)


src/scanpy/get/_aggregated.py
```diff
@@ -278,42 +369,59 @@ def aggregate(  # noqa: PLR0912
...
+            acc = A.X
+        if not isinstance(acc, LayerAcc | MultiAcc):
+            msg = (
+                "`acc` must be a `LayerAcc` (e.g. `A.X`, `A.layers[...]`) or "
+                f"`MultiAcc` (e.g. `A.obsm[...]`, `A.varm[...]`), was {acc!r}"
+            )
+            raise TypeError(msg)
+        data = adata[acc]
+        dim_df = pd.DataFrame({
+            ref.idx if isinstance(ref.idx, str) else str(ref): adata[ref]
```
flying-sheep: Idk how to do this better:
- only simplify `A.obs[k]` to string? If so, should we allow to mix strings and `AdRef`s in `by=[…]`?
- simplify unless there are duplicate `.idx`s?
- don’t stringify at all and use `AdRefs` as `df.columns`?

ilan-gold: Is the question here about the `isinstance` check? Like, what to do about people who pass in the index as a `by` column? 

This line of code seems inoffensive other than I am not sure why `ref.idx` would be anything other than a `str`

flying-sheep: the question is about “we now accept `AdRef`s, what to use as dataframe columns when they’re used”

> I am not sure why `ref.idx` would be anything other than a `str`

```pycon
>>> type(ad.acc.A.obsm["y"][:, 0].idx)
int
```

ilan-gold: > the question is about “we now accept AdRefs, what to use as dataframe columns when they’re used”

But these columns should really just be `ref.idx` since `ad.acc.A.obsm["y"][:, 0]` isn't a valid `by` unless I'm missing something.

flying-sheep: why would it not be?

```py
import numpy as np
import scanpy as sc
from anndata import acc, AnnData

adata = AnnData(np.random.rand(3, 4), obsm=dict(thing=np.array(["x", "y", "y"])))
sc.get.aggregate(adata, acc.A.obsm["thing"][:, 0], ())
# AnnData object with n_obs × n_vars = 2 × 4
#    obs: "A.obsm['thing'][:, 0]", 'n_obs_aggregated'
```

ilan-gold: I guess my comment was more based on the current state of things, which enforces the `by` to columns in the `DataFrame` of `obs` or `var`. I don't have a great solution for stringifying "first columns of `obsm` key `thing`" aside from "only allow `obs` and `var`" or `str(ref)` as you have

ilan-gold: To clarify, if you leave it as-is, I think that's fine.  If you revert to the old behavior, I think that's also perfectly reasonable (and would remove this question)

ilan-gold: I'm leaving towards what you have now, at least in terms of functionality.  Being able to do one-versus-rest type aggregations, perhaps over  a neighborhood graph, could be cool, if it's supported i.e., vec of 1s and 0s (neighbors and not neighbors).  Not sure if people will really do this though, but just a thought.

tests/test_aggregated.py
```diff
@@ -548,10 +544,97 @@ def test_aggregate_obsm_labels() -> None:
...
+        adata = adata.T.copy()
+
+    if attr is None:
+        old = sc.get.aggregate(adata, by, ["sum", "mean"], axis=axis)
+    elif attr == "layers":
+        old = sc.get.aggregate(adata, by, ["sum", "mean"], axis=axis, layer="test")
+    elif attr == "obsm":
+        old = sc.get.aggregate(adata, by, ["sum", "mean"], axis=axis, obsm="test")
+    else:
+        old = sc.get.aggregate(adata, by, ["sum", "mean"], axis=axis, varm="test")
```
ilan-gold:
```suggestion
    old = sc.get.aggregate(adata, by, ["sum", "mean"], axis=axis, **({} if attr is None else {attr: "test"}))
```

ilan-gold (commented):
I don't see any tests for the added behavior i.e., around `by=A.obsm["foo"][:, 1]`

tests/test_aggregated.py
flying-sheep: Huh, so confused. Your suggestion is what I typed!

src/scanpy/get/_aggregated.py
flying-sheep: what do you think of the third option from my initial comment in this thread?

src/scanpy/get/_aggregated.py
```diff
@@ -199,50 +216,124 @@ def _power(x: Array, power: float) -> Array:
...
-
     .. array-support:: get.aggregate
 
     Params
     ------
     adata
         :class:`~anndata.AnnData` to be aggregated.
     by
-        Key of the column to be grouped-by.
+        References to the column(s) to be grouped-by.
```
ilan-gold: Not really columns anymore though...

tests/test_aggregated.py
```diff
@@ -548,10 +545,154 @@ def test_aggregate_obsm_labels() -> None:
...
+
+    adata = sc.datasets.blobs()
+    adata.obs["blobs"] = adata.obs["blobs"].astype(str)
+    adata.obsm["thing"] = np.column_stack([
+        adata.obs["blobs"].astype(int).to_numpy(),
+        np.zeros(adata.n_obs, dtype=int),
+    ])
+
+    result = sc.get.aggregate(adata, by=A.obsm["thing"][:, 0], func=["sum", "mean"])
+    expected = sc.get.aggregate(adata, by="blobs", func=["sum", "mean"])
```
ilan-gold, line 721: Does it make sense to check the names of the `result` i.e., `result.obs_names`?

ilan-gold: Also the `obs` of `result` should contain a certain column derived from `by`, right?

flying-sheep: yeah but: https://github.com/scverse/scanpy/pull/4199#discussion_r3535983364

src/scanpy/get/_aggregated.py
ilan-gold: Oh missed this

> don’t stringify at all and use AdRefs as df.columns?

Is this serializable? What if someone tries to write their results?

flying-sheep: Hmm, the problem is that stringifying it introduces one more format that isn’t really supported. We have

1. the actual `AdRef` object
2. the simple string version (supported via `A.resolve`, e.g. `"obs.some_col"`. Can only represent a subset of possible refs)
3. the `repr` (not supported as conversion format anywhere)


We could fix that by doing something like
- here, if passing a collection of `AdRef`s, serialize to `repr` always (instead of the conditional `.idx` we have now)
- in `A.resolve`, also support the repr format (`if input.startswith('A.'): …`)

then people can just use `A.resolve` to get back the input `AdRef`s from the `obs` column names.

Or we serialize/deserialize `AdRef`s in `AnnData` …

ilan-gold: I think we kind of landed on "just stringify it" (and use whatever comes out) but let me know if there was more you wanted it.

docs/conf.py
```diff
@@ -140,7 +142,8 @@
 katex_prerender = shutil.which(NODEJS_BINARY) is not None
 
 intersphinx_mapping = dict(
-    anndata=("https://anndata.scverse.org/en/stable/", None),
+    # Needs latest until `.acc` is released in 0.13
+    anndata=("https://anndata.scverse.org/en/latest/", None),
```
ilan-gold:
```suggestion
    anndata=("https://anndata.scverse.org/en/stable/", None),
```

src/scanpy/get/_aggregated.py
```diff
@@ -199,50 +208,87 @@ def _power(x: Array, power: float) -> Array:
...
+    by: (
+        str
+        | Collection[str | AdRef[Idx2D | int, AnnData]]
+        | AdRef[Idx2D | int, AnnData]
+    ),
     func: AggType | Iterable[AggType],
     *,
-    axis: Literal["obs", 0, "var", 1] | None = None,
-    mask: NDArray[np.bool] | str | None = None,
+    # TODO: allow str once `.resolve` supports accs
```
ilan-gold: Should be released now

src/scanpy/get/get.py
```diff
@@ -400,42 +416,96 @@ def var_df(
...
+        if isinstance(acc, LayerAcc) and dim == "var":
+            data = data.T
+        return data
+
     # https://github.com/scverse/scanpy/issues/1546
     if not isinstance(use_raw := choices.get("use_raw", False), bool):
         msg = f"use_raw expected to be bool, was {type(use_raw)}."
         raise TypeError(msg)
-    assert choices.keys() <= {"layer", "use_raw", "obsm", "obsp"}
+    assert choices.keys() <= {"layer", "use_raw", "obsm", "obsp", "varm", "varp"}
```
ilan-gold, line 496: So we can `_get_arr` over `obsp` and `varp` with `choices` but not with `acc` (due to the `LayerAcc | MultiAcc` guard)?

src/scanpy/_settings/presets.py
```diff
@@ -312,14 +315,18 @@ def check(self) -> Self:
...
+
 def _missing_scanpy2_deps() -> list[Requirement]:
-    dist_names = {canonicalize_name(d.name) for d in distributions()}
     return [
         r
         for r in map(Requirement, requires("scanpy") or ())
         if r.marker
         and r.marker.evaluate({"extra": "scanpy2"}, "requirement")
-        and canonicalize_name(r.name) not in dist_names
+        and canonicalize_name(r.name) not in dist_names()
```
ilan-gold: Shouldn't we start checking versions here as well given the explicit reliance in this PR on newer `anndata`?

flying-sheep: good point!

src/scanpy/get/get.py
flying-sheep: fixed, I just forgot.

src/scanpy/get/get.py
```diff
@@ -434,37 +436,45 @@ class _Rep(TypedDict, total=False):
...
     adata: AnnData,
-    acc: LayerAcc | MultiAcc | Collection[LayerAcc | MultiAcc] | None = None,
+    acc: ArrAcc | str | Collection[ArrAcc | str] | None = None,
     *,
     dim: Literal["obs", "var"] | None = None,
     **choices: Unpack[_Rep],
 ) -> Any:
     """Get a 2D array aligned with `dim`, via an `anndata.acc` accessor or old-style choices."""
-    if _collection_of(acc, LayerAcc | MultiAcc):
+    if _collection_of(acc, (LayerAcc, MultiAcc, str)):
```
ilan-gold: Why no `GraphAcc` here?

src/scanpy/get/_aggregated.py
```diff
@@ -199,50 +208,86 @@ def _power(x: Array, power: float) -> Array:
...
+        | AdRef[Idx2D | int, AnnData]
+    ),
     func: AggType | Iterable[AggType],
     *,
-    axis: Literal["obs", 0, "var", 1] | None = None,
-    mask: NDArray[np.bool] | str | None = None,
+    acc: LayerAcc | MultiAcc | str | None = None,
+    mask: NDArray[np.bool] | AdRef[Idx2D | int, AnnData] | str | None = None,
     dof: int = 1,
+    # old API
```
ilan-gold, line 254: If the `ScanpyV2Preivew` is on, in general, does it make sense to raise a `DeprecationWarning`? I am not sure we have run into this issue yet, but probably should be consistent.  I feel like blanket warning that these arguments will disappear in scanpy 2 is a bit aggressive right now

src/scanpy/_settings/presets.py
```diff
@@ -313,16 +316,29 @@ def check(self) -> Self:
...
+            and r.marker.evaluate({"extra": "scanpy2"}, "requirement")
+            and not _req_satisfied(r)
+        )
     ]
 
 
+def _req_satisfied(req: Requirement) -> bool:
+    return (
+        canonicalize_name(req.name) in dist_names()
+        and version(req.name) in req.specifier
```
ilan-gold, line 333: If this doesn't happen, as the current CI failures show, we got a kind of useless error: 

```
ImportError: Setting preset to <Preset.ScanpyV2Preview: 'scanpy-v2-preview'> requires optional dependencies that are not installed: ‘anndata’. Install them with: pip install `scanpy[scanpy2]`
```

while the solution is correct, the problem announced (anndata is not installed) is not correct so maybe add a version identifier to the error message?

src/scanpy/get/_aggregated.py
```diff
@@ -278,74 +327,81 @@ def aggregate(  # noqa: PLR0912
...
-    elif axis == 1:
-        # i.e., all of `varm`, `obsm`, `layers` are None so we use `X` which must be transposed
-        data = data.T
-
-    dim_df = getattr(adata, axis_name)
-    categorical, new_label_df = _combine_categories(dim_df, by)
+    if settings.preset is Preset.ScanpyV2Preview and any(
+        v is not None for v in (axis, layer, obsm, varm)
+    ):
+        msg = "Use `acc` instead of `layer`, `obsm`, and `varm`."
```
ilan-gold:
```suggestion
        msg = "`acc` will replace `layer`, `obsm`, and `varm` arguments in scanpy 2."
```

ilan-gold: Something stronger

ilan-gold (approved):
I <3 accessors

PR #4205, opened 2026-07-08
perf: illico + numba threads

zboldyga:
- [ ] Closes #

@ilan-gold this is a small PR that does a few things:

- illico previously wasn't getting a thread value passed down (e.g. n_jobs), it was defaulting to 1. that is fixed
- there was a preexisting bug with the _numba utility for resolving thread count, it resolved -1 to 1 thread. the only place it was used was in rank_genes_groups, which passed n_jobs, so this conflicted with scanpy docs. This now properly sets to the numba thread count
- we also wrap the illico call with the _numba utility so numba thread count restores upon completion (I think illico modifies numba thread count, this helps prevent side effects).


zboldyga:
@ilan-gold while not high impact, I also saw some inconsistencies with the way n_jobs is used throughout scanpy. I preferred not to open that can of worms here. Might just be something that is most easily solved with clear docs in each use case anyway.

ilan-gold (approved):

PR #4207, opened 2026-07-12
fix: guard bw conversion in pl.spatial against missing image

steps-re:
## Summary

`sc.pl.spatial(..., bw=True)` raises `TypeError: 'NoneType' object is not subscriptable` when called without an image (no `img` passed and no image available in `.uns['spatial']`, e.g. general spatial coordinates without a background image).

`_check_img` runs the grayscale conversion unconditionally whenever `bw=True`:

```python
if bw:
    img = np.dot(img[..., :3], [0.2989, 0.5870, 0.1140])
```

If `img` is still `None` at that point, indexing it crashes. The existing test suite already documents this gap: `test_manual_equivalency_no_img` skips the `bw` case with the comment "Has no meaning when there is no image" instead of asserting it works.

## Fix

Guard the conversion on `img is not None`, matching how the rest of `_check_img` and `_check_na_color` treat a possibly-missing image.

## Test plan

- [x] Added `test_spatial_bw_no_img`, exercising `sc.pl.spatial(..., bw=True)` on data with no image anywhere, confirming it no longer raises.
- [x] Full `tests/test_plotting_embedded/` suite passes locally (117 passed, 5 skipped, no change to skip count).
- [x] `ruff check` clean on changed files.
- [x] Added a towncrier fragment at `docs/release-notes/4207.fix.md` (will rename if the PR number differs).

### DIFF
--- a/docs/release-notes/4207.fix.md
+++ b/docs/release-notes/4207.fix.md
@@ -1,0 +1,1 @@
+Fix {func}`scanpy.pl.spatial` raising `TypeError` when `bw=True` is passed without an image {smaller}`M German`
--- a/src/scanpy/plotting/_tools/scatterplots.py
+++ b/src/scanpy/plotting/_tools/scatterplots.py
@@ -1382,7 +1382,7 @@ def _check_img(
         )  # Throws StopIteration Error if keys not present
     if img is None and spatial_data is not None and img_key is not None:
         img = spatial_data["images"][img_key]
-    if bw:
+    if bw and img is not None:
         img = np.dot(img[..., :3], [0.2989, 0.5870, 0.1140])
     return img, img_key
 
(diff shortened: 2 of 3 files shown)


steps-re:
Heads up on the red CI: the only failing test is `test_sim.py::test_sim_toggleswitch` (`ValueError: operands could not be broadcast together with shapes (100,2) (200,2)`), which lives in `sc.tl.sim` / `_sim.py`. This PR only touches the `bw` image guard in `scatterplots.py`, so that failure is orthogonal to the change here. The plotting suite and the new `test_spatial_bw_no_img` regression test pass, and codecov is green. Happy to rebase if a fix for the sim test lands separately.

ilan-gold:
https://scanpy.readthedocs.io/en/latest/api/generated/scanpy.pl.spatial.html this function is deprecated so I think we're probably not going to fix this unless another maintainer wants to see this through.

steps-re:
makes sense, closing this out. no point carrying a fix for a deprecated function, and it has gone conflicted against main since anyway.

leaving the diff and the repro in the thread in case anyone hits the same crash on pl.spatial before it goes away. thanks for the look.

PR #4208, opened 2026-07-13
fix: PCA centering in ingest

Intron7:
Ingest centers its PCA against the query data not the fit data. I also made PCA sparse


tests/test_ingest.py
```diff
@@ -69,6 +70,52 @@ def test_representation(adatas):
...
+    sc.pp.pca(adata_ref, n_comps=3, mask_var="selected")
+    sc.pp.neighbors(adata_ref, n_neighbors=3, n_pcs=2)
+    ing = sc.tl.Ingest(adata_ref)
+
+    if as_sparse:
+
+        def fail_toarray(*args, **kwargs) -> None:
+            pytest.fail("PCA ingest must not densify the query matrix")
+
+        monkeypatch.setattr(type(adata_new.X), "toarray", fail_toarray)
```
flying-sheep, line 100: a bit fragile, but I can’t think of a better way …

flying-sheep (approved):

flying-sheep:
just add a relnote

Intron7:
@flying-sheep already done.

flying-sheep:
huh strange, I wouldn’t have said that If I had seen it. Sorry!

PR #4232, opened 2026-07-22
feat: update and speedup harmony

Intron7:
This also adds compatibility for https://github.com/scverse/rapids-singlecell/issues/734 Its also speeds up scanpy harmony to be faster than harmonypy==2.0 by roughly 1.5x


src/scanpy/preprocessing/_harmony/__init__.py
```diff
@@ -35,14 +35,16 @@ def harmony_integrate(  # noqa: PLR0913
...
+    correction_method: Literal["fast"] = "fast",
     block_proportion: float = 0.05,
     rng: SeedLike | RNGLike | None = None,
 ) -> None:
     """Integrate different experiments using the Harmony algorithm :cite:p:`Korsunsky2019,Patikas2026`.
 
-    This CPU implementation is based on the harmony-pytorch & rapids_singlecell version,
-    using NumPy for efficient computation.
+    This CPU implementation was originally based on the harmony-pytorch
+    and rapids-singlecell implementations, using NumPy for efficient computation.
```
flying-sheep: So it’s not based on these anymore? Why mention them then? Also we need to start by describing what the function *is*, not what it was.

Please rephrase this paragraph:

1. one sentence describing what the first line means (why would one want to “integrate different experiments”, what does it mean to do that?)
2. What makes this implementation special? Attribution to the implementations this is based on can go here or at the end.
3. (optional) a `.. note` explaining how this function’s behavior changed compared to previous scanpy versions if that’s super relevant.

src/scanpy/preprocessing/_harmony/core.py
```diff
@@ -87,9 +86,17 @@ def __post_init__(
...
+                "ridge_lambda must be a finite positive number when "
+                f"dynamic_lambda=False, got {self.ridge_lambda}."
+            )
+            raise ValueError(msg)
 
         # Process batch keys
-        self.batch_codes, self.n_batches = _get_batch_codes(batch_df, batch_key)
+        self.batch_codes, self.n_levels = _get_batch_codes(batch_df, batch_key)
+        self.n_batches = int(self.n_levels.sum())
+        self.n_covariates = int(self.n_levels.size)
```
flying-sheep: So like this? Can’t be 2D right?

```suggestion
        self.n_covariates = len(self.n_levels)
```

src/scanpy/preprocessing/_harmony/core.py
```diff
@@ -222,43 +226,81 @@ def _correct(
...
+    return codes, n_levels
+
+
+def _get_theta_array[S: np.generic](
+    theta: float | Sequence[float],
+    n_levels: NDArray[np.integer],
+    dtype: np.dtype[S],
+) -> NDArray[S]:
+    """Normalize scalar, per-variable, or per-category theta values."""
+    levels = np.atleast_1d(n_levels).astype(np.int64, copy=False)
```
flying-sheep: Why the `atleast_1d`? Isn’t `n_levels` the guaranteed-1D array from `_get_batch_codes` or is there another possible source?

tests/test_harmony.py
```diff
@@ -156,54 +162,191 @@ def test_harmony_integrate_reference(
...
+def test_harmony_batch_keys_are_nonempty_and_complete() -> None:
+    obs = pd.DataFrame({"batch": ["a", None]})
+    with pytest.raises(ValueError, match="contains missing values"):
+        _get_batch_codes(obs, "batch")
+    with pytest.raises(ValueError, match="at least one column"):
+        _get_batch_codes(obs, [])
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_harmony_multikey_correction_matches_dense_design(
```
flying-sheep, line 246: Can you elaborate on what this test does and why it has all this manual implementation in there?

tests/test_harmony.py
```diff
@@ -372,3 +515,34 @@ def test_compute_lambda_kb_zero_denom(dtype: type[np.floating]) -> None:
...
+        sigma=0.1,
+        n_clusters=2,
+        max_iter_harmony=10,
+        max_iter_clustering=4,
+        tol_clustering=1e-3,
+        tol_harmony=1e-2,
+        block_proportion=0.05,
+        rng=734,
+        alpha=0.2,
+        batch_prune_threshold=1e-5,
```
flying-sheep: These are all the same as above. What’s the rationale for specifying them all?

So we can change the defaults without this test breaking? Are these different from the defaults and match `harmonypy` instead?

Maybe define a helper `partial` that has all these set, with a name and/or comment explaining what it is.

flying-sheep (commented):
Looks great, thanks a lot!

Some clarifying questions, and the docstring needs work, otherwise awesome!

flying-sheep:
By the way: I format the text so the line breaks are after `.` or `,` if appropriate, that way diffs are shorter when just changing one sentence.

PR #4252, opened 2026-07-26
fix(plotting): keep boolean colors stable after subsetting

### LINKED ISSUE #4219: Support dictionaries for Boolean colors in 'adata.uns'

Currently, Boolean colors are assigned positionally:

```python
adata.uns["boolean_colors"] = ["red", "blue"]
```

This is ambiguous when only one value is present. If all observations are `True` or all are `False`, the remaining value receives the first color.

It would be useful to support an explicit mapping:

```python
adata.uns["boolean_colors"] = {
    False: "red",
    True: "blue",
}
```

This would keep Boolean colors consistent after filtering or subsetting.

daveringelberg:
Fixes #4219
Release notes are included in `docs/release-notes/4219.feat.md`

## Summary

This draft PR adds support for mappings stored in `adata.uns[f"{key}_colors"]`, so boolean colors can stay tied to `True`/`False` after subsetting.

Existing list-based color behavior is preserved.

## Testing

Added tests for:
- mapping palettes stored in `adata.uns`
- boolean mappings after subsetting to only `True` or only `False`
- boolean mappings passed through the explicit `palette=` argument
- regression test for #4219 covering boolean color mappings in `adata.uns` after subsetting.

Local checks run:
- `ruff check`
- `ruff format --check`
- `python -m py_compile`
- `git diff --check`

I could not run `pytest` locally due to local environment issues, so I am opening this as a draft for CI and maintainer feedback.


flying-sheep:
Looks like it mostly works except for a small discrepancy in expected values (alpha channel I suppose):

```
FAILED tests/plotting/legacy/test_plotting_utils.py::test_get_palette_bool_mapping_from_palette_argument - AssertionError: assert {'False': '#f...: '#0000ffff'} == {'False': '#f...e': '#0000ff'}
  
  Differing items:
  {'True': '#0000ffff'} != {'True': '#0000ff'}
  {'False': '#ff0000ff'} != {'False': '#ff0000'}
```

But this is a lot of added complexity.

Don’t you think it makes more sense to just add both colors in a list if the categorical is boolean?

daveringelberg:
Thanks, that makes sense. I updated the draft to follow the simpler list-based approach for boolean categoricals instead of adding dict support. The implementation now keeps the boolean category order fixed as False/True, so list-based colors remain stable after subsetting.

flying-sheep:
nice, much simpler!

PR #3571, opened 2025-04-08
`normalize_total` with numba

Intron7:
Fixes #3135

This also removes the deprecated functions out

PR #4256, opened 2026-07-27
fix: use the non-zero median as `target_sum` on the sparse path

### LINKED ISSUE #4251: normalize_total(target_sum=None) picks a different target for CSR than for dense — and can zero out the data

### What happened?

`normalize_total(target_sum=None)` documents its target as the "median of total counts for observations", but `_normalize_total_helper` computes it two different ways depending on the input layout ([`src/scanpy/preprocessing/_normalization.py`](https://github.com/scverse/scanpy/blob/main/src/scanpy/preprocessing/_normalization.py), line numbers from `main`):

```python
if isinstance(x, CSRBase):
    ...
    if target_sum is None:
        target_sum = np.median(counts_per_cell)            # L105 — median over ALL cells
else:
    counts_per_cell = stats.sum(x, axis=1)
    ...
    if target_sum is None:
        target_sum = _compute_nnz_median(counts_per_cell)  # L117 — median over POSITIVE cells
```

So when any cell has zero total counts, CSR and dense inputs normalize the same data to different targets. CSC is `.tocsr()`-converted first (L267–268), so it follows the CSR rule.

The branches disagree only when a zero-total cell shifts the median — ties between the middle order statistics hide it (row sums `[0, 10, 10, 20]` give 10 either way), which is probably why this has gone unnoticed.

**The severe case:** when at least half the cells are empty, the CSR branch's median is `0`, and dividing by it wipes out the matrix. This is a silent data-destruction path — it returns all zeros, with only two `RuntimeWarning`s to show for it:

```
case 2: row sums [0, 0, 10] -> X after normalize_total
  dense: [0. 0.  0. 0.  4. 6.]     # correct: target = 10, populated cell untouched
    csr: [0. 0.  0. 0.  0. 0.]     # target = 0 -> the only populated cell is destroyed
```

Empty cells are common enough in practice (an unfiltered raw matrix, an aggressive gene subset, a `.raw` slice) that this is reachable without doing anything unusual. `normalize_total` already anticipates them — it warns `"Some cells have zero counts"` and takes care to divide only positive cells. It just doesn't exclude them when *choosing* the target on the CSR path.

### Which rule looks like the intended one?

The positive-cell median, for three reasons:

1. **It is the historical behaviour.** In 1.11.1 and earlier there was a single code path and it excluded zeros on both branches — `np.median(counts_greater_than_zero, axis=0)` and `np.median(counts_per_cell[cell_subset])`, with `cell_subset = counts_per_cell > 0`. The split arrived with the numba CSR kernel in #3571 (merged 2025-05-19; backported to 1.11.x as #3636), which preserved the old rule for the non-CSR branch as `_compute_nnz_median` and gave the new CSR branch a plain `np.median`. That reads like an oversight rather than a decision — neither the PR nor its release note mentions a semantic change.
2. **`_compute_nnz_median`'s own docstring** says "compute the median of the non-zero counts".
3. **It cannot produce a zero target** as long as one cell is non-empty.

Suggested fix — one line, restoring the pre-1.11.2 semantics:

```diff
     if target_sum is None:
-        target_sum = np.median(counts_per_cell)
+        target_sum = _compute_nnz_median(counts_per_cell)
```

Happy to open that PR with a regression test if you agree on the direction. Whichever rule you prefer, it would help to state it in the `target_sum` docstring — "median of total counts for observations" doesn't currently settle it.

### Affected versions

Introduced in **1.11.2** (2025-05-28) and present in every release since, including the latest **1.12.3**, plus current `main`. **1.11.1** and earlier are unaffected. Verified by reading `_normalization.py` at tags `1.11.0`, `1.11.1`, `1.11.2`, `1.11.5`, `1.12.1`, `1.12.3` and `main`, and by running the script below against both `1.12.1` and `main`.

### Minimal code sample

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
# ]
# ///
#
# Runs against scanpy's development branch.
#
# `normalize_total(target_sum=None)` picks its target over ALL cells for CSR
# input but over POSITIVE cells only for dense input, so the same data
# normalizes to two different targets when a zero-total cell is present.

import anndata as ad
import numpy as np
import scanpy as sc
import scipy.sparse as sp

# Case 1 — row sums [0, 10, 20, 30]. The zero-total cell shifts the median.
dense = np.array([[0.0, 0.0], [4.0, 6.0], [8.0, 12.0], [12.0, 18.0]], dtype=np.float32)

print("case 1: row sums [0, 10, 20, 30] -> per-cell totals after normalize_total")
for label, X in (("dense", dense.copy()), ("csr", sp.csr_matrix(dense))):
    a = ad.AnnData(X=X)
    sc.pp.normalize_total(a, target_sum=None)
    print(f"  {label:>5}: {np.asarray(a.X.sum(axis=1)).ravel()}")
print("  expected: dense target = median(10, 20, 30)    = 20")
print("            csr   target = median(0, 10, 20, 30) = 15")

# Case 2 — row sums [0, 0, 10]. The CSR median is 0, so the divide wipes the
# matrix: the only populated cell is destroyed.
dense2 = np.array([[0.0, 0.0], [0.0, 0.0], [4.0, 6.0]], dtype=np.float32)

print("\ncase 2: row sums [0, 0, 10] -> X after normalize_total")
for label, X in (("dense", dense2.copy()), ("csr", sp.csr_matrix(dense2))):
    a = ad.AnnData(X=X)
    sc.pp.normalize_total(a, target_sum=None)
    out = a.X if isinstance(a.X, np.ndarray) else a.X.toarray()
    print(f"  {label:>5}: {out.ravel()}")
print("  expected: dense keeps the populated cell (target = 10)")
print("            csr   returns all zeros        (target = 0)")
```

Output on `main`:

```
case 1: row sums [0, 10, 20, 30] -> per-cell totals after normalize_total
  dense: [ 0. 20. 20. 20.]
    csr: [ 0. 15. 15. 15.]

case 2: row sums [0, 0, 10] -> X after normalize_total
  dense: [0. 0. 0. 0. 4. 6.]
    csr: [0. 0. 0. 0. 0. 0.]
```

### Error output

```pytb
# No exception — the destructive case is silent apart from these:
.../scanpy/preprocessing/_normalization.py:119: RuntimeWarning: divide by zero encountered in divide
  counts_per_cell = counts_per_cell / target_sum
.../scanpy/preprocessing/_normalization.py:119: RuntimeWarning: invalid value encountered in divide
  counts_per_cell = counts_per_cell / target_sum
```

### Versions

<details>

```
scanpy      1.12.1
anndata     0.12.19
numpy       2.4.6
scipy       1.18.0
pandas      2.3.3
h5py        3.16.0
numba       0.65.1
scikit-learn 1.8.0
Python      3.12.11 (main, Jun  4 2025, 17:36:43) [Clang 20.1.4]
OS          Linux-5.15.0-176-generic-x86_64-with-glibc2.35

Also reproduced on scanpy @ main (git), Python 3.13.
Source-verified on 1.11.2 - 1.12.3 and main; see "Affected versions".
```

</details>

JOhnsonKC201:
- [x] Closes #4251
- [x] Tests added

## Problem

With `target_sum=None`, `normalize_total` produces different results for sparse
and dense input when the matrix contains zero-count cells:

```python
x = np.array([[0.0, 0.0], [4.0, 6.0], [8.0, 12.0], [12.0, 18.0]])

# dense -> [ 0.  0.  8. 12.  8. 12.  8. 12.]
# csr   -> [ 0.  0.  6.  9.  6.  9.  6.  9.]
```

With enough empty cells the sparse result is worse than merely different. For
row sums `[0, 0, 10]` the median is 0, so the sparse path divides by zero and
returns an all-zero matrix, silently destroying the data.

## Cause

`_normalize_total_helper` picks the target sum in two places. The numba/CSR
branch uses `np.median(counts_per_cell)`, while the dense branch uses
`_compute_nnz_median`, which drops zero counts first. CSC is converted to CSR,
so it inherits the sparse rule.

This looks like an oversight rather than a deliberate difference.
`_compute_nnz_median` was the single shared code path until #3571 split the
function to add the numba branch, and the new branch got a plain median. The
existing `test_normalize_matrix_types` does compare dense against other array
types for `target_sum=None`, but it uses `pbmc68k_reduced()`, which has no
empty cells, so the divergence never showed up.

## Fix

Use `_compute_nnz_median` in the sparse branch too, so both paths agree. One
line.

## Test plan

- New `test_normalize_total_target_sum_ignores_zero_count_cells` in
  `tests/test_normalization.py`, parametrised over `ARRAY_TYPES`. It asserts the
  non-empty rows sum to the non-zero median (20, not 15) and that the result
  matches the dense path. It fails on main for the sparse types.
- `tests/test_normalization.py`: 204 passed, 81 skipped (200 passed before this
  change, plus the new cases).
- `ruff check` and `ruff format --check` clean at the pinned v0.15.20.


flying-sheep:
thanks!

PR #4276, opened 2026-08-04
ci: run prek in GH actions

flying-sheep:


.vscode/settings.json
```diff
@@ -9,9 +9,6 @@
     "[python]": {
         "editor.defaultFormatter": "charliermarsh.ruff",
     },
-    "[toml]": {
```
Zethson, line 12: Why?

.pre-commit-config.yaml
```diff
@@ -1,40 +1,55 @@
...
-        exclude: tests/_data
-    -   id: check-added-large-files
-    -   id: check-case-conflict
-    -   id: check-toml
-    -   id: check-yaml
-    -   id: check-merge-conflict
-    -   id: detect-private-key
-    -   id: no-commit-to-branch
-        args: ["--branch=main"]
+- repo: https://github.com/flying-sheep/bibfmt
```
Zethson, line 2: Cool! Could we reduce the bus factor here if we ship this to so many people via our template, please?

.pre-commit-config.yaml
```diff
@@ -1,40 +1,55 @@
...
+- repo: https://github.com/tox-dev/pyproject-fmt
+  rev: v2.27.0
+  hooks:
+  - id: pyproject-fmt
+    groups: [format]
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.16.1
+  hooks:
+  - id: ruff-check
+    args: ["--fix"]
```
Zethson: BTW any opinion on `--exit-non-zero-on-fix`?

.github/workflows/autofix.ci.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Install dependencies
+        run: |
+          echo "::group::Install hatch"
+          uv tool install hatch
+          echo "::endgroup::"
+          echo "::group::Create environment"
+          hatch -v env create ${{ matrix.env.name }}
+          echo "::endgroup::"
+          hatch run ${{ matrix.env.name }}:session-info scanpy anndata
```
Zethson: Do we want to hardcode them specifically? Not all ecosystems need to use scanpy. Only anndata is very very likely

.github/workflows/autofix.ci.yml
```diff
@@ -0,0 +1,149 @@
...
+
+      - name: Run tests
```
Zethson: Why is this in the autofix ci? I think I'm missing something because the whole autofix CI workflow seems a bit bloated.

Zethson (commented):
Now I remember what you meant with autofixes. True, I had completely forgotten about this.

.github/workflows/autofix.ci.yml
Zethson: ok lol now you deleted the whole thing. This makes more sense.

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: autofix.ci
```
Zethson, line 1: Not sure whether I like the name tbh. I'm fine with just keeping it as is. It has a little bit of extra functionality

.vscode/settings.json
flying-sheep: unrelated, but `pyproject-fmt` has grown a few ways to format things that are impossible to configure “Even better TOML” to do, so they’ll fight.

.pre-commit-config.yaml
flying-sheep: Do we now? Sure, I can move it to scverse.

.github/workflows/ci.yml
flying-sheep: mandatory, autofix.ci requires this.

I thought at first the file name was supposed to be that but it’s this (that’s why I added the other file, that was a rename I tried to abort lol)

.pre-commit-config.yaml
Zethson: Oh boy, we're not. I somehow assumed you asked me for a review for a cookiecutter-scverse PR.

.github/workflows/autofix.ci.yml
Zethson: nevermind, I didn't notice that this is the scanpy repo

Zethson (approved):

.pre-commit-config.yaml
flying-sheep: I think I prefer it due to the way prek works: without this, prek (as pre-commit hook) will allow creating a commit, and then leave some new format changes uncommitted and unstaged.

Ideally, we could leave it out, and prek (when run as pre-commit hook) would automatically change the commit, then let it go through if only format changes need to happen.

.github/workflows/autofix.ci.yml
flying-sheep: no, but this is scanpy, so it’s there to be listed on top

.pre-commit-config.yaml
flying-sheep: that explains some other comments haha

flying-sheep:
@Zethson this has to wait until I’ve figured it out: https://github.com/autofix-ci/action/issues/40

PR #4288, opened 2026-08-11
feat: support acc shortcuts

flying-sheep:
`scanpy.get.aggregate` already handles them, but the plotting functions didn’t yet.

- [ ] Closes #


src/scanpy/plotting/_v2/_core.py
```diff
@@ -171,9 +173,9 @@ def _embedding(key: str, name: str, /) -> partial:
 
 def heatmap(
     adata: AnnData,
-    base: LayerAcc | GraphAcc = A.X,
+    base: str | LayerAcc | GraphAcc = A.X,
```
selmanozleyen, line 176: For example `Xenium_counts` here would cause returning `X` early instead of erroring since we only check `if spec == "X" or spec.startswith("X["):`

src/scanpy/plotting/_v2/_core.py
```diff
@@ -108,16 +109,17 @@ def scatter(
 
     """
     try:
-        i, j = kdims
+        i, j = _resolve_all(kdims)
```
selmanozleyen, line 112: Here `Xenium_counts[:,CD3]` silently yields `A.X[:, 'CD3']`

selmanozleyen (commented):
Hi,
this looks nice but if we want to ensure people use the accessors why do we allow strings then? Because I think there are some problems with the string format I'd like to mention first. I know it's not entirely in the scope of this PR but since we are exposing it here I find it relevant.

Also since the checks in the str format isn't robust and complete it might cause some confusion in scanpy. One more comment on `.` as the separator. I know that seurat v5 uses `.` uses it as a separator for example like `layers[counts.sample1]`

src/scanpy/plotting/_v2/_core.py
flying-sheep: handled in github.com/scverse/anndata/pull/2605

PR #4303, opened 2026-08-18
feat: move hashsolo to `sc.pp`

### LINKED ISSUE #4304: move `sc.external.pp.hashsolo` into `sc.pp`

Upstream `hashsolo` is now unmaintained, `solo` part of `scvi-tools`.

flying-sheep:
Basically just moving the code over and adding an `anndata.acc` based version of the API on top.

People had issues with our old API since it only supported `obs` columns:

- https://www.biostars.org/p/9561534/
- https://discourse.scverse.org/t/how-to-install-solo-sc-or-use-hashsolo-from-scvi/1702

The new API of course supports everything! Yay accessors!

- [x] Closes #4304


src/scanpy/external/pp/_hashsolo.py
```diff
@@ -1,296 +1,25 @@
...
-    )
-    most_likely_hypothesis = np.argmax(probs_hypotheses, axis=1)
-    return {
-        "most_likely_hypothesis": most_likely_hypothesis,
-        "probs_hypotheses": probs_hypotheses,
-        "log_likelihoods_for_each_hypothesis": log_likelihoods_for_each_hypothesis,
-    }
 
 
+@deprecated(Deprecation("1.13.0", "Use `scanpy.pp.hashsolo` instead."))
```
selmanozleyen: Wondering which one should be used because it seems inconsistent with
```python
scrublet = deprecated("Import from sc.pp instead")(_scrublet.scrublet)
scrublet_simulate_doublets = deprecated("Import from sc.pp instead")(
    _scrublet.scrublet_simulate_doublets
)

```

docs/conf.py
```diff
@@ -184,6 +184,7 @@
     "pp.filter_cells": (["np", "sp", "da"], []),
     "pp.filter_genes": (["np", "sp", "da"], []),
     "pp.harmony_integrate": (["np"], []),
+    "pp.hashsolo": (["np", "sp"], []),
```
selmanozleyen, line 187: This works silently for dask and will give a "# FutureWarning: The `numpy.column_stack` function is not implemented by Dask array."

src/scanpy/preprocessing/_hashsolo.py
```diff
@@ -0,0 +1,418 @@
...
+        prior for each hypothesis, in the order `[negative, singlet, doublet]`
+    number_of_noise_barcodes
+        number of barcodes to used to calculated noise distribution
+
+    Returns
+    -------
+    A 2d array of shape `(n_cells, 3)` with the probability of each hypothesis.
+
+    """
+    log_likelihoods, _, _ = _calculate_log_likelihoods(data, number_of_noise_barcodes)
```
selmanozleyen: the other returns of `_calculate_log_likelihoods` is never used elsewhere

src/scanpy/preprocessing/_hashsolo.py
```diff
@@ -0,0 +1,418 @@
...
+def _hashsolo(
+    data: NDArray,
+    *,
+    priors: ArrayLike,
+    clusters: NDArray | None,
+    number_of_noise_barcodes: int | None,
+) -> NDArray[np.float64]:
+    """Validate counts and run the bayes rule, optionally per cluster."""
+    if not check_nonnegative_integers(data):
+        msg = "Cell hashing counts must be non-negative"
```
selmanozleyen: we should also mention they should be an integer no? since people can give normalized log easily now

tests/test_hashsolo.py
```diff
@@ -0,0 +1,140 @@
...
+            from anndata.acc import A
+
+            df = request.param == "obsm-df"
+            adata.obsm["hto"] = counts.copy() if df else counts.to_numpy()
+            # a plain array has no column names, so they fall back to positions
+            names = HASHES if df else [str(i) for i in range(len(HASHES))]
+            return Hashed(adata, A.obsm["hto"], names)
+        case _:
+            pytest.fail(f"Unknown param {request.param!r}")
+
```
selmanozleyen, line 82: Maybe a test with 0-len, 1-len, 2-len hashes should be tried out to see if it errors gracefully. Because claude tells me this

"
 Silent all-NaN → every cell "Negative" when the derived noise-barcode count hits 0
Both return with no exception: adata.obsm["hashsolo"] is entirely NaN, adata.obs["hashsolo"] is {"Negative": n}, and the only signal is a RuntimeWarning: Mean of empty slice from np.mean(data_sort[:, :-n_barcodes]). One barcode gives IndexError: index -2 is out of bounds for axis 1 with size 1.
"

selmanozleyen: In general claude suggested these

```
test_too_few_noise_barcodes[1-hash] / [2-hashes-default] / [explicit-0]	RuntimeWarning: Mean of empty slice (warning-as-error) instead of the wanted ValueError	#1
test_pre_existing_clusters_missing_label	RuntimeWarning; without warnings-as-error, those rows are [0,0,0] → "Negative"	#2
test_pre_existing_clusters_singleton	RuntimeWarning: divide by zero; otherwise NaN probs	#2
test_pre_existing_clusters for the new ref path
```

```python

@pytest.mark.parametrize(
    ("n_hashes", "number_of_noise_barcodes"),
    [
        pytest.param(1, None, id="1-hash"),
        pytest.param(2, None, id="2-hashes-default"),
        pytest.param(len(HASHES), 0, id="explicit-0"),
    ],
)
def test_too_few_noise_barcodes(
    adata: AnnData, n_hashes: int, number_of_noise_barcodes: int | None
) -> None:
    """Without a noise barcode, the noise distribution is undefined."""
    with pytest.raises(ValueError, match=r"noise barcodes?"):
        sc.pp.hashsolo(
            adata,
            HASHES[:n_hashes],
            number_of_noise_barcodes=number_of_noise_barcodes,
        )


@pytest.mark.parametrize("kind", ["str", pytest.param("ref", marks=needs.anndata_acc)])
def test_pre_existing_clusters(adata: AnnData, kind: str) -> None:
    """Clustered demultiplexing equals demultiplexing each cluster on its own."""
    adata.obs["cl"] = np.where(np.arange(adata.n_obs) % 2, "a", "b")
    if kind == "ref":
        from anndata.acc import A

        ref = A.obs["cl"]
    else:
        ref = "cl"
    sc.pp.hashsolo(adata, HASHES, pre_existing_clusters=ref)

    np.testing.assert_allclose(adata.obsm["hashsolo"].to_numpy().sum(axis=1), 1)
    for cluster in ("a", "b"):
        sub = adata[adata.obs["cl"] == cluster].copy()
        sc.pp.hashsolo(sub, HASHES)
        np.testing.assert_allclose(
            sub.obsm["hashsolo"].to_numpy(),
            adata.obsm["hashsolo"].loc[sub.obs_names].to_numpy(),
        )


def test_pre_existing_clusters_missing_label(adata: AnnData) -> None:
    """Unlabeled cells must not silently turn into confident negatives."""
    adata.obs["cl"] = pd.Categorical(["a"] * (adata.n_obs - 10) + [None] * 10)
    sc.pp.hashsolo(adata, HASHES, pre_existing_clusters="cl")
    np.testing.assert_allclose(adata.obsm["hashsolo"].to_numpy().sum(axis=1), 1)


def test_pre_existing_clusters_singleton(adata: AnnData) -> None:
    """A one-cell cluster has zero variance; it must not yield NaN probabilities."""
    adata.obs["cl"] = ["a"] * (adata.n_obs - 1) + ["b"]
    sc.pp.hashsolo(adata, HASHES, pre_existing_clusters="cl")
    assert np.isfinite(adata.obsm["hashsolo"].to_numpy()).all()

```

selmanozleyen (approved):
Some edge cases inherited from main and one new behaviour with the pd.unique and loc usage instead of the old np.unique. Plus one nitpick on dead return variables. But the rest seems good

src/scanpy/external/pp/_hashsolo.py
flying-sheep: I think the difference is fine since the two `scrublet` versions are identical, whereas changing from the external hashsolo API to the sc.pp one needs to be called differently

docs/conf.py
flying-sheep: Yeah but that’s not really support isn’t it? It will immediately load all the data it uses into memory and not use dask to parallelize anything.

src/scanpy/preprocessing/_hashsolo.py
flying-sheep: `hashes` does mention that it’s supposed to be counts, and it crashes pretty fast, so I think it might be enough, but if you have a place in mind where to mention it, please say so.

selmanozleyen: I just meant clarifying 
```
msg = "Cell hashing counts must be non-negative integers"
```
instead of 
```
msg = "Cell hashing counts must be non-negative"
```
basically

selmanozleyen (approved):
nice, all looks very modern now also!

PR #4306, opened 2026-08-19
feat: add bbknn

### LINKED ISSUE #4302: Adopt `scanpy.pp.bbknn`
We already have bbknn in rapids-singlecell, porting it should be relatively easy.

@maltekuehl said we should make our neighbors implementation more flexible to support its defaults: https://github.com/scverse/scanpy/pull/3807

Intron7:
This PR add a scanpy native bbknn implementation because of the removal of external

fixes #4302

### DIFF
--- a/benchmarks/benchmarks/preprocessing_log.py
+++ b/benchmarks/benchmarks/preprocessing_log.py
@@ -67,6 +67,42 @@ class PreprocessingSuite:  # noqa: D101
         sc.pp.scale(self.adata, max_value=10)
 
 
+class NeighborsSuite:
+    """Benchmark neighbor graph construction.
+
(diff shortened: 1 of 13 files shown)


flying-sheep (commented):
Looks great!

I deduplicated the transformer creation method and adapted the heuristic a bit, and added a benchmark suite to check it. Take a look if that makes sense or if I missed something!

The only piece I think is missing is that we should get rid of legacy cruft, but I can take care of that:

- [x] no reason for `_accepts_legacy_random_state`
- [x] things like `batch_key` and so on should become something accepting `AdRef`s or strs in the `resolve` format (`"obs.obs_colname"` instead of `"obs_colname"`, and then `_key` of course makes no sense). 
- [ ] including `use_rep`: #4315

PR #4325, opened 2026-08-27
feat: add configurable columns to PAGA plots

### LINKED ISSUE #1203: sc.pl.paga: TypeError: paga() got an unexpected keyword argument 'ncols'
The current paga plot output a n_row =1 plot, which is not very user-friendly. It would be nice to have an option to control n_col as other plot functions in scanpy do. Thanks!

AtomicGlance:
## Summary

- Add an ncols parameter to sc.pl.paga for wrapping multiple color panels into a grid.
- Preserve the existing single-row layout when ncols=None.
- Keep colorbars aligned with grid panels and validate invalid column counts.
- Add regression coverage and a release-note fragment.

Closes #1203

## Tests

- pytest -q tests/plotting/legacy/test_paga.py -k paga_ncols -o filterwarnings=: 8 passed
- Ruff, Python byte-compilation, and git diff --check pass.
- The rest of test_paga.py passes 24 tests; one unrelated test needs to download the existing pbmc3k fixture, which is blocked in this environment.


AtomicGlance:
I synchronized this branch with current `scverse/scanpy:main` (5f2accbe). The upstream update includes the AnnData-version guard for `zarr_write_format`, which was causing the Python 3.14 fixture setup to fail before the PAGA tests ran. The PR diff remains limited to the PAGA implementation, tests, and release note. The local Hatch test environment is not installed here, so the focused and matrix validation should come from the new CI run.

PR #4333, opened 2026-09-01
feat: add `anndata.acc` support to `mask` params

### LINKED ISSUE #4331: `anndata.acc` based API for `mask`/`mask_{obs,var}`

Already done in `sc.get.aggregate`, TODO:

- `pp.sample(p=...)`
- `pp.scale(mask_obs=...)`
- `pp.pca(mask_var=...)`
- `tl.rank_genes_groups(mask_var=...)`
- `experimental.pp.normalize_pearson_residuals_pca(mask_var=...)`

flying-sheep:
TODO:

- [x] probably good to unify `_mask_arg` with `_check_mask`?
- [ ] think about if it makes sense to rename from `mask_{obs,var}` to `mask`:
  - not explicit from param name anymore which dim is masked
  - but explicit through argument (e.g. `mask="obs.xy"`)
  - also good fit for functions that can operate on either axis (sample, aggregate) or allow to mask either or both axes (e.g. `pca` could be fit on a subset or `scale` could be called on a subset) 

----

- [x] Closes #4331


src/scanpy/_settings/presets.py
```diff
@@ -93,7 +93,7 @@ class BasicEmbeddingPreset(NamedTuple):
 
 class RankGenesGroupsPreset(NamedTuple):
     method: DETest
-    mask_var: str | None
+    mask: str | None
```
ilan-gold, line 96: Why is this here if it is only ever `None`?

src/scanpy/_utils/__init__.py
```diff
@@ -988,14 +988,15 @@ def _resolve_axis(
     raise ValueError(msg)
 
 
-def obs_acc(obs_col: str) -> str | AdRef:
+def dim_acc(col: str, *, dim: Literal["obs", "var"] = "obs") -> str | AdRef:
```
ilan-gold, line 991:
```suggestion
def dim_acc(col: str, *, dim: Literal["obs", "var"]) -> str | AdRef:
```
right? why should `obs` be the default to the user?

src/scanpy/preprocessing/_pca/__init__.py
```diff
@@ -353,7 +352,7 @@ def pca(  # noqa: PLR0912, PLR0913, PLR0915
         adata.uns[keys.uns] = dict(
             params=dict(
                 zero_center=zero_center,
-                mask_var=mask_var_param,
+                mask_var=_ref_to_json(mask_param),
```
ilan-gold, line 355: So it seems like (correct me), previously we were just putting a raw accessor in the `uns`? And now we are putting JSON?

This _feels_ like it should be a separate PR or needs a bigger explanation or something (like a standalone release note).

src/scanpy/preprocessing/_highly_variable_genes.py
```diff
@@ -181,7 +181,7 @@ def _highly_variable_genes_seurat_v3(  # noqa: PLR0912, PLR0915
     )
     if batch_key is not None:
         aggregated_mean_var = aggregate(
-            adata_agg, by=obs_acc("__hvg_v3_batch_info__"), func=["mean", "var"]
+            adata_agg, by=dim_acc("__hvg_v3_batch_info__"), func=["mean", "var"]
```
ilan-gold, line 184:
```suggestion
            adata_agg, by=dim_acc("__hvg_v3_batch_info__", dim="obs"), func=["mean", "var"]
```

src/scanpy/preprocessing/_scale.py
```diff
@@ -142,20 +156,28 @@ def scale[A: _Array](
...
+# `stacklevel=2` skips `singledispatch`’s dispatcher frame
+@deprecated_arg("mask_obs", Deprecation("1.13.0", "Use `mask` instead."), stacklevel=2)
 def scale_array[A: _Array](
     x: A,
     *,
     zero_center: bool | Default = Default(preset=("scale", "zero_center")),
     max_value: float | None = None,
     copy: bool = False,
     return_mean_std: bool = False,
+    mask: NDArray[np.bool] | None = None,
```
ilan-gold, line 180: Is this type right?

src/scanpy/_settings/presets.py
flying-sheep: good point!

src/scanpy/preprocessing/_pca/__init__.py
flying-sheep: Currently I’m trying to put backwards-compatible stuff in there, but we might want to add the ability to store AdRefs/AdAccs in anndata and make scanpy 2.0 depend on that. What do you think?

src/scanpy/preprocessing/_scale.py
flying-sheep: You mean because it just allows an array and no reference? Then yes: if you call `scale(some_ndarray, mask=...)`, where would a reference pull anything from?

src/scanpy/preprocessing/_pca/__init__.py
ilan-gold: Oh I absolutely like that idea!

flying-sheep: Great! First step: https://github.com/scverse/anndata/pull/2642

src/scanpy/preprocessing/_scale.py
ilan-gold: That makes sense!

ilan-gold (approved):

Issue #2126, opened 2022-02-01
logreg with groups might not be working correctly all the time


Intron7:
- [X] I have checked that this issue has not already been reported.
- [X] I have confirmed this bug exists on the latest version of scanpy.
- [X] (optional) I have confirmed this bug exists on the master branch of scanpy.

---

If you us rank_genes_group with a subset of groups and change the order the reported output is messed up. In the example below the second result is like the fist result. However in the plot so cluster 2 has now the genes of cluster 1 and vice versa.

### Minimal code sample (that we can copy&paste without having any data)

```python
sc.tl.rank_genes_groups(adata, groupby="louvain",groups=["0","1","2"], method="logreg", use_raw= False)
sc.tl.rank_genes_groups(adata, groupby="louvain",groups=["0","2","1"], method="logreg", use_raw= False)
```

#### Versions

<details>
anndata==0.7.8
scanpy==1.8.2
sinfo==0.3.4
IPython==7.30.1
jupyter_client==7.1.0
jupyter_core==4.9.1
notebook==6.4.6

</details>

Issue #273, opened 2018-09-27
Reordering categories leads to issues with `sc.tl.rank_genes_groups` when using `method='logreg'`


fbnrst:
Hi,

Reordering the categories of groups in obs leads to shuffling of marker genes to the wrong groups when using `sc.tl.rank_genes_groups`. Interestingly, this only happens if I use `method='logreg`. It works fine with `method='t-test`. I am not sure how relevant this is. Maybe, one should not reorder categories. However, this was helpful to me when I wanted to plot the categories in a certain order. Here comes a minimal example:

```
import pandas as pd
import scanpy.api as sc

adata = sc.datasets.blobs(640, 3)
sc.tl.pca(adata)
sc.pp.neighbors(adata)
sc.tl.louvain(adata)

method='logreg'

sc.tl.rank_genes_groups(adata, 'louvain', method=method)
print(pd.DataFrame(adata.uns['rank_genes_groups']['names']).head(1))

adata.obs['louvain'].cat.reorder_categories(['2', '1', '0'], inplace = True)

sc.tl.rank_genes_groups(adata, 'louvain', method=method)
print(pd.DataFrame(adata.uns['rank_genes_groups']['names']).head(1))
```
The result looks like this:
```python
     0   1    2
0  570  63  126
     2   1    0
0  570  63  126
```
The expected result would be that the marker genes end up in the same groups as above (watch the change in the lowest row):
```python
     0   1    2
0  570  63  126
     2   1    0
0  126  63  570
```

fbnrst:
The same issue happens when renaming categories with the built-in rename categories function. I think, this definitely should not happen. Example:

```python
import pandas as pd
import scanpy.api as sc

adata = sc.datasets.blobs(640, 3)
sc.tl.pca(adata)
sc.pp.neighbors(adata)
sc.tl.louvain(adata)

method='logreg'

sc.tl.rank_genes_groups(adata, 'louvain', method=method)
print(pd.DataFrame(adata.uns['rank_genes_groups']['names']).head(1))

adata.rename_categories('louvain', ['Zero', 'One', 'Two'])

sc.tl.rank_genes_groups(adata, 'louvain', method=method)
print(pd.DataFrame(adata.uns['rank_genes_groups']['names']).head(1))
```
gives
```python
     0   1    2
0  570  63  126
  Zero  One  Two
0   63  126  570
```

falexwolf:
[...]

PR #4348, opened 2026-09-08
fix: correct binary logistic marker score direction

dnncha:
Binary `LogisticRegression.coef_[0]` points toward `clf.classes_[1]`. The current binary branch reports it under the first requested group without checking that group's encoded class. If the reported group has the lower category code, its marker ranking is reversed.

Orient the binary coefficient vector toward the reported group's class. Existing output shape and multiclass handling are preserved.

With a two-group synthetic matrix containing distinct A and B markers, the unmodified API ranks the B marker first under A when category order is A,B; reversing the categories similarly gives A's marker under B. An explicit target/reference comparison can also flip when only category order changes.

On Scanpy's released `pbmc3k_processed()` dataset, subset to 342 B cells and 1,144 CD4 T cells, run:

```python
sc.tl.rank_genes_groups(
    adata, "louvain", groups=["B cells"], reference="CD4 T cells",
    method="logreg", max_iter=1000,
)
```

Using the default raw expression matrix, category order `["B cells", "CD4 T cells"]` returns IL32, CD3D, S100A4, LDHB and CD3E at the top of the B-cell ranking. All top 20 have higher mean expression in CD4 T cells. Reversing category order returns HLA-DRA, CD74, HLA-DPB1, HLA-DRB1 and CD79A; all top 20 have higher mean expression in B cells. The first ordering's coefficients need to be negated. After this patch, both orders return the same B-cell ranking. The issue reproduces in releases 1.11.5 and 1.12.4 and current upstream base `ec374022343eb7ef80bbe3139264e37552cb79b4`.

Validation: 18 added public-API cases cover both category orders and target groups, default binary selection, nonconsecutive category codes, dense/sparse input, raw and layers. With the four existing logistic tests, unmodified source has 10 failures / 12 passes; patched source has **22 passes**. Both ranking test files: **60 passed, 35 skipped, 1 expected failure, 32 passing subtests**. These counts overlap. The full Scanpy suite was not run. Ruff lint and formatting checks pass.

Related older category-mapping discussions #273 and #2126 concern multiclass ordering; this patch addresses the binary coefficient orientation. No changed published conclusion or independent scientific review is claimed. Prepared with AI assistance. [The report and downloadable reproduction scripts, results and execution records are available here](https://cheerfulduck.com/research/audits/scanpy-binary-logreg-direction).


tests/test_rank_genes_groups_logreg.py
```diff
@@ -62,3 +62,66 @@ def test_rank_genes_groups_with_unsorted_groups():
         bdata.uns["rank_genes_groups"]["scores"]["Three"]
     ).to_numpy()
     np.testing.assert_equal(array_ad, array_bd)
+
+
+@pytest.mark.parametrize("categories", [["A", "unused", "B"], ["B", "unused", "A"]])
+@pytest.mark.parametrize("target", ["A", "B"])
+@pytest.mark.parametrize("representation", ["dense", "sparse", "raw", "layer"])
```
flying-sheep, line 69: please remove the `representation` parameter, the code paths are the same for dense, sparse, …

tests/test_rank_genes_groups_logreg.py
```diff
@@ -62,3 +62,66 @@ def test_rank_genes_groups_with_unsorted_groups():
...
+def test_binary_logreg_scores_point_toward_requested_group(
+    categories, target, representation
+):
+    from anndata import AnnData
```
flying-sheep, line 73: already imported in the module

src/scanpy/tools/_rank_genes_groups.py
```diff
@@ -597,11 +597,14 @@ def logreg(self, **kwds) -> Generator[_TestResult, None, None]:
         # not all codes necessarily appear in data
         existing_codes = np.unique(self.grouping.cat.codes)
         for igroup, cat in enumerate(self.groups_order):
+            cat_code: int = np.argmax(self.grouping.cat.categories == cat)
             if len(self.groups_order) <= 2:  # binary logistic regression
+                # Binary coefficients point toward classes_[1]; orient them
+                # toward the group under which the scores will be reported.
                 scores = scores_all[0]
+                if cat_code == clf.classes_[0]:
+                    scores = -scores
```
flying-sheep, line 606:
```suggestion
                scores = scores_all[0] if cat_code == clf.classes_[1] else -scores_all[0]
```

flying-sheep (commented):
Thanks! some small nitpicks, otherwise this is a good fix!

dnncha:
Thanks!

PR #4358, opened 2026-09-11
docs: correct tutorial cell-type mapping

fderop:
The tutorial’s cell-type mapping contradicts its own saved marker plot: it swaps the B-cell and erythroid labels and omits cluster 4. This corrects those entries and retains the annotation example.

| Cluster | Cells | Previous label | Corrected label |
| --- | ---: | --- | --- |
| 2 | 2,269 | Erythroid | B Cells |
| 3 | 1,759 | B Cells | Erythroid |
| 4 | 1,743 | Missing | Erythroid |

The panels below apply the original and corrected mappings to the same reproduced run. Cell membership, cluster numbers, and UMAP coordinates are identical. Gray cells lack a label in the original mapping.

Before and after correcting the tutorial cell-type mapping

The marker panels use the same cells, coordinates, and cluster numbers. MS4A1, PAX5, and CD79A support the B-cell label for cluster 2. HBA1, HBB, and GYPA support erythroid labels for clusters 3 and 4. Colors show log1p-normalized expression, with a separate full-range scale for each gene.

Expression of B-cell and erythroid markers on the same UMAP

Cluster 2 expresses MS4A1 in 81.8% of cells and PAX5 in 62.8%. Cluster 3 expresses HBA1 and HBB in every cell. [Marker measurements](https://github.com/fderop/scanpy/blob/651358bec69b68a2b358e9cdee63a5a84fa4895a/preview/marker-expression.csv) and the [figure script](https://github.com/fderop/scanpy/blob/651358bec69b68a2b358e9cdee63a5a84fa4895a/preview/create_figures.py) are available for review.

The contradiction is already present in the [March 2024 notebook](https://github.com/scverse/scanpy-tutorials/blob/aa99d1e2c755945698fe5e3df9850fbb90411a85/basic-scrna-tutorial.ipynb). Its saved marker plot is byte-for-byte identical to the current one. The error is therefore not a mismatch introduced by our rerun.

Validation: five fresh preprocessing and clustering runs from raw counts, using one, four, or eight neighbor-search workers, produced identical coarse assignments and numbers. Worker counts affected finer partitions. The corrected mapping labels all 17,041 cells. The reproduction used Scanpy `1.14.0.dev6` at `a6f1a2d2`; local formatting and pre-commit checks pass.


flying-sheep:
Good point! That style of assignment isn’t great anyway (reproducibility shouldn’t depend on randomness), and this mapping isn’t used afterwards, but you’re correct!

PR #4359, opened 2026-09-11
feat: switch to `use` and `out`

### LINKED ISSUE #2583: Remove inplace argument in scanpy 2.0

Other?

Another point that I'd like to throw into the scanpy 2.0 discussion: 

Right now, many functions have the `inplace` argument, that determines if a function should write back to `adata` or return the result instead. 

With this behavior it is hard to make correct type hints. While it is possible with `@overload`, it is cumbersome because it requires to type out the entire function signature twice. When I asked if it is possible to write these overloads in a more concise way [on stackoverflow](https://stackoverflow.com/questions/75757890/python-overload-single-argument?noredirect=1#comment133653817_75757890) several users argued that changing the return type based on an argument is an anti-pattern, and I think they convinced me. 

## Alternative approach
Have two API levels, e.g.

```py
def scanpy.tl.pca(adata: AnnData, **kwargs) -> None: ...
```
and
```py
def scanpy.lowlevel.tl.pca(data: np.ndarray | sp.spmatrix, n_pcs) -> np.ndarray: ...
```

Where the former is a wrapper for the latter. This allows to separate the implementation of the actual method using only numpy/scipy data types from the scverse-specific behavior.

### LINKED ISSUE #4334: `anndata.acc` replacement for functions taking `obsm`, `layer`, …

this probably also means coming up with an `in_`/`out` pair and maybe `copy`/`inplace`

flying-sheep:
Current design: 

- deprecate `copy` everywhere
- functions operating on a main matrix get `use` and `out` – default for `out` is to operate in place (`out=False`), while `out=True` means returning the matrix instead of writing it into AnnData.
- multi-in/out functions use `key_added`/`inplace` for the same functionality, like `Neighbors(dists=A.obsp["foo"], conns=A.obsp["bar"], meta=A.uns["foo"])` (where `A.uns` doesn’t even exist yet)
  - Variant: create `NamedTuple`s or so for the spots used. Probably what @ilan-gold would prefer, but we should meet to spec this out.

## Examples for `use`/`out` (`key_added` is unchanged):

```py
# in-place on X
sc.pp.log1p(adata)
# read counts, write X
sc.pp.log1p(adata, use=A.layers["counts"], out=A.X)
# return numpy array, equivalent to `sc.pp.log1p(adata.X)`
logged = sc.pp.log1p(adata, out=True)
```

`out` makes the singledispatch use some functions have (`sc.pp.log1p(adata.X)`) unnecessary, so we might deprecate that, what do you think?

Example for how this improves code:

```diff
-logged = sc.pp.log1p(adata_obs, copy=True)
-sc.pp.highly_variable_genes(logged)
-adata_obs = adata_obs[:, logged.var["highly_variable"]]
+sc.pp.log1p(adata_obs, out="layers.log1p")
+sc.pp.highly_variable_genes(adata_obs, use="layers.log1p")
+adata_obs = adata_obs[:, adata_obs.var["highly_variable"]]
```

## TODO:

- [x] `use` needs to be consistently documented
- [x] `use_raw` needs deprecating
- [ ] scale still writes `var["mean"]`/`["std"]` under `out=True`, should maybe gain a `return_stats` parameter or so
- [x] `regress_out` copies a backed array before rejecting it

----

- [x] Closes #4334, closes #2583


grst:
> functions operating on a main matrix get use and out – default for out is to operate in place, out=None means returning the matrix instead of writing it into AnnData.

If we want to discourage use of `inplace` we could also make the `out` param explicit:
 * not specified -> fails
 * out=False -> inplace
 * out=True -> return

flying-sheep:
Yeah, I think `bool` instead of `None|Default` is better. I thought that users probably don’t need to pass a dynamic value (since it completely changes how the function works), but `out=False` for inplace and `out=True` for return is just more intuitive.

grst:
Or sentinels, e.g. `A.INPLACE | A.RETURN`? But I guess the boolean is nicer to type.

flying-sheep:
Yeah, an enum would of course be clearest, but I think the bool is intuitive enough, don’t you think?

I currently left the default to be `False`, i.e. “overwrite what `use` points to”, but it doesn’t need to stay that, we could:

- force people to specify `out=...` as you suggest, or
- use function-dependent conventions like e.g. the defaults for `log1p` could be `use="layers.counts"` and `out="layers.logged"` or something.

grst:
> use function-dependent conventions like e.g. the defaults for log1p could be use="layers.counts" and out="layers.logged" or something.

I like that, this would establish some conventions

flying-sheep:
Yeah! I’ll meet with Ilan today about this, and we’ll probably hash something out. I intentionally left it out of this PR so far since the design is still changing, so I don’t want to muddle the discussion with bikeshedding about which conventional names to settle on.

flying-sheep:
Plan:
- remove `key_added`
- for multi-`out`, add global lookup and validation table for each function’s slots, sth. like `sc.OUTS["pp.neighbors"] == dict(distances=A.obsm["distances"], connectivities=A.obsm["connectivities"])`
- either:
  - add `prefix` parameter to all `use`/`out` functions, as mutually exclusive replacement to `out`
  - allow to use a class like `sc.Prefix("foo")` for `out`, e.g. `sc.pp.neighbors(..., use=sc.Prefix("foo"))`

flying-sheep:
OK, multi-out is blocked on not having `.uns` accessors then: https://github.com/scverse/anndata/issues/2657

Storing accessors for e.g. `neighbors` in anndata is https://github.com/scverse/anndata/issues/1979 (and the follow-up of storing accessors)

